PR #9163, opened 2016-12-15
Fixed , add setting info

zminhquanz:
Something wrong when i see texture coordspeedhack not disable when
enable software rendering
And i test vertex decoder JIT , i think it's not buggy ( avoid ) , it's
improve performance on heavy game ( GOW )

### DIFF
--- a/UI/GameSettingsScreen.cpp
+++ b/UI/GameSettingsScreen.cpp
@@ -263,15 +263,20 @@ void GameSettingsScreen::CreateViews() {
 	CheckBox *framebufferSlowEffects = graphicsSettings->Add(new CheckBox(&g_Config.bDisableSlowFramebufEffects, gr->T("Disable slower effects (speedup)")));
 	framebufferSlowEffects->SetDisabledPtr(&g_Config.bSoftwareRendering);
 
-	// Seems solid, so we hide the setting.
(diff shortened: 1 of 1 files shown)


LunaMoo:
I think you're confusing one thing - Vertex Decoder JIT isn't "disabled" now.
 The option for it was simply hidden from UI because it's enabled by default where possible and there shouldn't be any reason to turn it off.

zminhquanz:
@LunaMoo i think default it's disabled , i tried to hidden this option and performance is the same when i'm disabled this

zminhquanz:
Oh , i see , the default is enable , Sorry , i have to fix code again

moura464:
static const char *quality[] = { "Low", "Medium", "High", "Ultra"}; ???

ppsspp parou de funcionar :(

zminhquanz:
@moura464 there is no ultra setting for splines , bezier

hrydgard:
We consciously removed the Texture Coord Speedhack option and left it to always on by default, because we thought it was now free of problems. Is there a reason you want to reintroduce the ability to turn it off?

zminhquanz:
@hrydgard can you explain why Texture coord speedhack on recent build was removed

hrydgard:
Yes, because it's now the standard and works well so no need to be able to turn it off.

zminhquanz:
@hrydgard yes , i got it , you want to hide this code or remove this code ?

hrydgard:
I'll just remove that setting and merge. Thanks!

PR #9165, opened 2016-12-16
minor fixes

latot:
Hi, again other minor fixes, principally overflows.

### DIFF
--- a/Core/CoreTiming.cpp
+++ b/Core/CoreTiming.cpp
@@ -237,7 +237,7 @@ void Shutdown()
 
 u64 GetTicks()
 {
-	return (u64)globalTimer + slicelength - currentMIPS->downcount;
+	return (u64)globalTimer + (slicelength - currentMIPS->downcount);
 }
 
 u64 GetIdleTicks()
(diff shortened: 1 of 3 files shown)


ext/native/util/random/rng.h
```diff
@@ -15,7 +15,7 @@ class GMRng {
 		m_z = ~seed;
 		if (!m_z) m_z = 31337;
 	}
-	uint32_t R32() {
```
hrydgard: This intentionally clips the value to 32 bits.

A way that might avoid the warning might be:

    return (uint32_t)((m_z << 16) + m_w); 

perhaps. Or 

    return ((m_z << 16) + m_w) & 0xFFFFFFFF

Either way, it's hardly important and I would prefer to just revert the changes made to this file.

Core/CoreTiming.cpp
```diff
@@ -314,7 +314,8 @@ void ScheduleEvent(s64 cyclesIntoFuture, int event_type, u64 userdata)
 	Event *ne = GetNewEvent();
 	ne->userdata = userdata;
 	ne->type = event_type;
-	ne->time = GetTicks() + cyclesIntoFuture;
+        ne->time = GetTicks();
+        ne->time += cyclesIntoFuture;
```
hrydgard: I don't think this change is worth making. The old code is clear and doesn't actually have any problems.

hrydgard (changes requested):

Core/CoreTiming.cpp
latot: mm, hi, the problem here is for some reason values handled with this plus send an overflow, so it makes this undefined, that is the problem, but why works with this change and not with the old code i don't know.

latot:
Hi, i update some code following your advice, only as a note, ub(undefined behavior) don't are warning are bugs, and can be caused for a human error, or some hardware limits and similars, i think need to be fixed.

PR #9166, opened 2016-12-16
Fixed Inconsistent texture filtering in Disgaea 2 #4235

konchunas:
This issue is fixed for Disgaea 2, since this game manages to store some textures before address 0x08800000, but after 0x08400000.
I assume PPGE is placed even before 0x08400000, so this fix should be relatively safe


hrydgard:
Oh of course! I forgot that games can just grab the memory between 0x08400000 and 0x08800000 (there's a system call for it, sceKernelVolatileMemTryLock). Nice catch.

PR #9168, opened 2016-12-18
Cmake cleanup

Orphis:


hrydgard:
No complaints here, just have to make it pass the CI :)

PR #9169, opened 2016-12-18
FFmpeg and iOS updates

Orphis:
Update FFmpeg to be able to build iOS on new archs.


Fadi5555:
@Orphis Thank you so much for taking care of iOS port. We really appreciate a lot.

Is this will gonna fix stutter every 3-4 seconds as well?

Orphis:
No

PR #9176, opened 2016-12-20
Always prescale uv ("texcoord speedhack")

hrydgard:
This works fine and is definitely the way to go for a lot of reasons. We've had it on by default for a long time now so let's just delete the old code.


zminhquanz:
I think it's clearly code , tks @hrydgard

PR #9179, opened 2016-12-21
More gpu centralization

hrydgard:
Doing some groundwork in preparation for adding some new GPU backends. It's icky to work with heavily duplicated code.

In fact, I'm currently considering turning the Vulkan backend into "Modern", and have that call a few thin rendering interfaces, initially Vulkan only, then D3D11 and possibly Metal. Will likely not bother with DX12.

The current GLES and DX9 backends will remain as they are to match those older shader models. Some optimizations I'll do for the new ones are probably no good idea for the old ones.

### DIFF
--- a/GPU/Common/FramebufferCommon.h
+++ b/GPU/Common/FramebufferCommon.h
@@ -236,6 +236,8 @@ public:
 	void SetRenderSize(VirtualFramebuffer *vfb);
 	void SetSafeSize(u16 w, u16 h);
 
+	virtual void Resized() = 0;
+
 protected:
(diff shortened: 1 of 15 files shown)


hrydgard:
Removed the last couple of steps, might do them later.

Fadi5555:
@hrydgard wow that's nice good luck bro.
I'm really happy about gonna support metal as well. Of course will effect the performance.
Thanks for everything. That's really appreciate.
Keel going.

MarcoEstevez:
Great job

PR #9182, opened 2016-12-23
crosslink database & product id database

adenovan:
one step easier for user to play adhoc multiplayer with different game region (crosslink).


hrydgard:
Normally I'd say that long lists like this belong in a separate file instead of hardcoding them, but since lots of changes aren't likely since there are no more games being released, I'm OK with this.

Orphis:
That's an awful lot of generated code.
It would have been better with an array and iterating over it.

PR #9184, opened 2016-12-25
Some preparations for D3D11

hrydgard:
Not really much actual D3D11 code yet, just context creation and some simplifying refactorings in thin3d.


zminhquanz:
What difference between OpenGL and D3D11

hrydgard:
Can't use OpenGL in cross platform UWP apps. This will improve performance on all UWP platforms once it's done (this is very early prep work).

zminhquanz:
Oh . It will help on Windows 10 Mobile using D3D11 API

PR #9210, opened 2017-01-08
Hardware tessellation

xebra:
I implemented hardware tessellation, currently only on OpenGL.
This is using technique called "instanced tessellation". That is using basic method "instanced rendering".  And you can find a simple bezier sample implementation here (http://developer.download.nvidia.com/SDK/10.5/opengl/screenshots/samples/instanced_tessellation.html).
Of course, I know NVIDIA's sample code is licensed under the proprietary license, so I rewrited alsmost everything.
Anyway, this implementation doesn't use so modern API like "tessellation shader", so I think I can implement it to DX9 or other backend.
Well, I think this codes needs to fix compatibility with GLES or other glsl version because I only tested with Win7 64 and I don't implement any compatible code yet.

Here's the some performance test.
Tested with FF4CC in DEBUG mode
Thundaga
Software tessellation is FPS 6
ulus10560_00000

Hardware tessellation is FPS 60
ulus10560_00006

Mist dragon morphing
Software tessellation is FPS 21
ulus10560_00001

Hardware tessellation is FPS 60
ulus10560_00007

You can find the option here!
menu_00000


Orphis:
Sweet!
Yes, it needs some fixes for OpenGLES, doesn't seem to be too bad though.

zminhquanz:
@Orphis does OpenGLES 2.0 to support this ?

LunaMoo:
For GLES problems I think it will be enough to define all the texture 1D stuff as 2D(with 1 height) and GL_CLAMP defined as GL_CLAMP_TO_EDGE. Imo this will not be usable on mobiles through, this sure can be faster on desktops with powerful dedicated gpus, but that shader is actually extremely heavy and will tank fps down harder than software implementation in less optimal situations. Certain gpu's like those older mali's might just fail to compile it.

 ~

 Would be good to also gray out "Spline/Bezier curves quality" setting when Hardware Tesselation is activated since it doesn't affect it in any way. Either way, nice work:3.

zminhquanz:
That look good , and it's very nice to support other backend and older like DX 9 , OpenGLES 2.0

xebra:
@LunaMoo Thanks for advice and fix them.

xebra:
Note: Information about to fix compatibility with GLES 2.0 or older glsl version.

Using new APIs
+ Instanced rendering
glDrawElementsInstanced()
gl_InstanceID in glsl
If version 130, it needs code below for gl_InstanceID.
if (gl_extensions.EXT_gpu_shader4) {
	WRITE(p, "#extension GL_EXT_gpu_shader4 : enable\n");

+ Vertex texture fetch
glGetIntegerv(GL_MAX_VERTEX_TEXTURE_IMAGE_UNITS, &maxVertexTextureImageUnits);
if maxVertexTextureImageUnits > 0, enable

+ Floating point texture
if OES_texture_float is enabled??

hrydgard:
Really nice work!

I'll be able to review it in detail soon. Only concerns are if the additional shader bits would impose any performance hit somehow (probably very little), and obviously possible compatibility issues on mobile hardware, as commented.

zminhquanz:
I will going to review that

LunaMoo:
When saying about defining 1D to 2D on GLES I was thinking about something like [this](https://gist.github.com/LunaMoo/5b45907c6e4b0e55dd8f52f0445bad51).

 Didn't really tested that code on GLES device ~ hopefully works;p, but we definitely want to keep 1D where possible(aka OGL) - from a quick test Thundaga in FF4 the performance of the shader drops to 50% with 2D compared to 1D.:|

i30817:
This is only for spline curves right?

xebra:
@LunaMoo Oh, I see thanks again:)

@i30817 No, it supports both of spline and bezier. It is compatible code with spline and bezier.

zminhquanz:
@xebra tested on my PC , using GTX 950 , OpenGL Backend , that look good and compability well and didn't have isssues

xebra:
@zminhquanz Thank you for testing:)
I hope to effective in a situation like using powerless cpu and low fps gameplay with spline/bezier.
I think if you use powerful cpu, it doesn't any benefits unfortunately. And software tessellation is not so slow because of @unknownbrackets 's SIMD optimaization. However, our codes can more optimize for software tessellation.

DonelBueno:
Does any other game benefit from this change or just FF4?

zminhquanz:
@DonelBueno here is the list of game using spline / bezier
BEZIER:

Puzzle Bobble
SSX On Tour
Jeanne d'Arc

SPLINE:

Loco Roco 1 , 2
Idolmaster
Hexyz Force

BOTH:
Final Fantasy IV

GPU/Common/ShaderId.cpp
```diff
@@ -119,6 +131,12 @@ void ComputeVertexShaderID(ShaderID *id_out, u32 vertType, bool useHWTransform)
 
 		id.SetBit(VS_BIT_NORM_REVERSE, gstate.areNormalsReversed());
 		id.SetBit(VS_BIT_HAS_TEXCOORD, hasTexcoord);
+
+		id.SetBit(VS_BIT_BEZIER, doBezier);
```
hrydgard: I think I'd like to have an if (g_Config.bHardwareTesselation) around these, to avoid a little bit of work here if disabled.

GPU/Common/SplineCommon.cpp
```diff
@@ -223,6 +223,39 @@ static void spline_knot(int n, int type, float *knot) {
 	}
 }
 
+static void TessellateSplinePatchHardware(u8 *&dest, u16 *indices, int &count, const SplinePatchLocal &spatch) {
+	SimpleVertex *&vertices = (SimpleVertex*&)dest;
+
+	for (int tile_v = 0; tile_v < spatch.tess_v + 1; ++tile_v) {
```
hrydgard: I think this needs a comment to explain that it's generating simple input vertices for the spline-computing vertex shader.

GPU/Common/SplineCommon.cpp
```diff
@@ -223,6 +223,39 @@ static void spline_knot(int n, int type, float *knot) {
...
+			float u = ((float)tile_u / (float)spatch.tess_u);
+			float v = ((float)tile_v / (float)spatch.tess_v);
+
+			SimpleVertex &vert = vertices[tile_v * (spatch.tess_u + 1) + tile_u];
+
+			vert.pos.x = u;
+			vert.pos.y = v;
+
+			// For compute normal
+			vert.nrm.x = 1.0f / (float)spatch.tess_u;
```
hrydgard: These divisions can be moved outside the loop. The compiler might do that for us, but it's far from guaranteed and divisions can be a little expensive on some hardware. Might even be able to use a uniform, but not sure if that would make a difference.

GPU/Common/SplineCommon.cpp
```diff
@@ -767,6 +800,35 @@ static void _BezierPatchHighQuality(u8 *&dest, u16 *&indices, int &count, int te
 	dest += (tess_u + 1) * (tess_v + 1) * sizeof(SimpleVertex);
 }
 
+static void TesselateBezierPatchHardware(u8 *&dest, u16 *indices, int &count, int tess_u, int tess_v, GEPatchPrimType primType) {
+	SimpleVertex *&vertices = (SimpleVertex*&)dest;
+
+	for (int tile_v = 0; tile_v < tess_v + 1; ++tile_v) {
```
hrydgard: Same as above, needs a quick comment.

GPU/Common/SplineCommon.cpp
```diff
@@ -781,6 +843,31 @@ void TesselateBezierPatch(u8 *&dest, u16 *&indices, int &count, int tess_u, int
 	}
 }
 
+class IndexConverter {
```
hrydgard: Nice one.

GPU/GLES/ShaderManager.h
```diff
@@ -110,54 +110,110 @@ class LinkedShader {
...
 };
 
 enum {
-	DIRTY_PROJMATRIX = (1 << 0),
-	DIRTY_PROJTHROUGHMATRIX = (1 << 1),
-	DIRTY_FOGCOLOR = (1 << 2),
-	DIRTY_FOGCOEF = (1 << 3),
-	DIRTY_TEXENV = (1 << 4),
-	DIRTY_ALPHACOLORREF = (1 << 5),
+	_DIRTY_PROJMATRIX = 0,
```
hrydgard: I don't like these _DIRTY and the defines after. Please use the old format (though 1ULL etc of course), it's readable enough. Alternatively, use FLAG_BIT directly in the enum declaration.

hrydgard (changes requested):
This is great. And I guess it's finally time to move to a uint64 for the dirty flag... In fact, I've been meaning to dirty-flag things other than uniforms, like viewport state etc, which can now be added since there's space.

It's a bit unfortunate that only OpenGL is supported, but that can be fixed later (or maybe it's not even needed...)

Anyway, added a few minor comments.

xebra:
@hrydgard Thank you for your review in detail:) and sorry for the late to reply.
Now I'm trying to implement Hardware Tessellation to DX9 and Vulkan, and it's almost done but it's still buggy.
I'll fix them and reference to your advise, and commit them later.
Thanks!

xebra:
BTW, DX9 implementation probably supports only NVIDIA's gpu because VTF(vertex texture fetch) is probably only supported by NVIDIA.

xebra:
@hrydgard 

> I don't like these _DIRTY and the defines after. Please use the old format (though 1ULL etc of course), it's readable enough. Alternatively, use FLAG_BIT directly in the enum declaration.

I'm sure enum supports only 32 bits(64 bits are overflow), so I used tricky way to expand these to 64 bits.

Orphis:
Well, if Windows XP is dropped, there won't be any issue with DX9 anymore.

Also, you can use: `enum : uint64_t {}` to specify the storage class to be 64bit.

hrydgard:
Dropping XP does not mean dropping DX9 necessarily, although that might of course also happen, we'll see, but it doesn't seem like a pressing issue. 

As Orphis says, it's perfectly possible to have uint64_t enums, and I think you can do it even without "enum class" by specifying the values as, for example, 1ULL << 36.

xebra:
> Also, you can use: enum : uint64_t {} to specify the storage class to be 64bit.

@Orphis OH, really thank you for the useful information!

I think Hardware Tessellation is just an option so DX9 VTF issue does not mind any way. Or I think I can implement other way.

hrydgard:
@xebra  I fixed the DX9 bugs so please rebase on master, then you can see if it works better :)

xebra:
@hrydgard It worked, nice job! Thanks a lot:)
I can make progress to develop on dx9.

GPU/GLES/GPU_GLES.cpp
```diff
@@ -1020,8 +1020,19 @@ void GPU_GLES::Execute_Bezier(u32 op, u32 diff) {
 	int bz_vcount = (op >> 8) & 0xFF;
 	bool computeNormals = gstate.isLightingEnabled();
 	bool patchFacing = gstate.patchfacing & 1;
+
+	if (g_Config.bHardwareTessellation && g_Config.bHardwareTransform && !g_Config.bSoftwareRendering) {
+		gstate_c.bezier = true;
+		if (gstate_c.bezier_count_u != bz_ucount) {
+			shaderManager_->DirtyUniform(DIRTY_BEZIERCOUNTU);
```
xebra: Using shader manager(GLES) in Execute_Bezier()

xebra (commented):
@hrydgard Can I revert this changes ab9c1d4? (only spline/bezier exclude bbox)
This implementation needs to set dirty flags to each(gles/dx9/vulkan) shader manager in Execute_Bezier/Spline().

However, shader dirty flags look like almost same in all backends, so we may be able to unify that logic to common method like ShaderManagerCommon.

hrydgard:
@xebra  Ah yes, sorry about that. Yes, please revert - I'll think of a way to do it again later.

xebra:
@hrydgard OK thanks.

xebra:
I tried rebase and fix your request and push them.
Is this right?

I don't know how to use git well yet.

hrydgard:
Hm, looks like you rebased on an earlier commit than master because some of my stuff is in there that shouldn't be (like the >> to << change). 

So probably the master in your client that you rebased on wasn't updated to the upstream master.

Still, looks like it's free of conflict.

xebra:
I was wrong something...
Of course, I resolved conflicts, but many commits duplicated.
But if this commits(pull request) is no problem, I don't mind.
If it has critical issue, what do I need to do?
Anyway, I need to study git more.

zminhquanz:
@xebra you can go to the wiki to see how to keep pull request up to date
https://github.com/hrydgard/ppsspp/wiki/How-to-keep-a-pull-request-up-to-date

xebra:
@zminhquanz Thanks! Hmm, I'm using Tortoisegit on Windows but I think my way is almost same as the guide.

1. clone xebra/ppsspp as origin
2. add hrygard/ppsspp as upstream
3. fetch&rebase upstream
4. fix conflict if needed
5. pull upstream
6. pull origin
7. commit my changes
8. push origin

Maybe, I think my mistake is "pull origin"...
Usually, I don't do that way but some errors occured when I push to origin, so I pulled from origin and merged and push them.

hrydgard:
5 should be done before 3...

Or in cmdline git, on your branch:

    git fetch upstream
    git rebase upstream/master

hrydgard (approved):

hrydgard:
So apart from the rebase, what remains now? I'd like to merge this soon so I can do other changes and cleanup.

xebra:
Remains are other backends implementation, but it's not done and need more time.
However, OpenGL implementation is almost done. Also, GLES(for mobile?) compatibility is now researching.
Please merge these if you approved.
I'll try to commit the other backends implementation in another pull request.
Thanks.

hrydgard:
OK, let's do it that way.

hrydgard:
I cherry-picked your commits to a new branch, #9226 . I think I'm gonna merge that instead, once the buildbot is happy, for a cleaner history.

iOS4all:
I didn't find this option in iOS version of ppsspp any idea please?

hrydgard:
Take it easy, it has not even been merged yet. Soon, though can't guarantee it'll work on iOS right away, this is still work-in-progress.

xebra:
@hrydgard That sounds great! I appreciate your help.

iOS4all:
@hrydgard I edited my comment above before you reply. Yeah it seems not merged yet.
Anyway take your time and I hope it could work on iOS as well. We really need this option.
BTW all later versions of ppsspp on iOS are crashing when we select roms even the last one which is (1.3-299) and Last working version was (1.3-191) but beyond that are always crashes when we selected roms.
Thanks...

hrydgard:
I merged #9226, so please base your future work on the current master @xebra  :) Closing this.

xebra:
@hrydgard Thank you :)

@iOS4all 
>  I hope it could work on iOS as well.

I hope too!

iOS4all:
@xebra nice work. It's really appreciate a lot.
Waiting for your next work.
Cheers....

xebra:
@iOS4all Thanks :)

PR #9216, opened 2017-01-16
Fix high-DPI rendering/input on Windows

hrydgard:
Turns out this has been broken for a while (well, never working, but since we specified that we were DPI-aware while we really weren't on Windows, things have been wrong).

Also, I'd really like to use modern per-monitor DPI but we can't (without really dirty hacking) as long as we keep XP compatibility, because the XP-compatible Windows SDK only supports up to Windows 7 features, while per-monitor DPI is an 8.1 feature.

Before, on a 4k laptop:
bad

After:
proper

Note the razor sharp text and proper proportions.

### DIFF
--- a/Core/Core.cpp
+++ b/Core/Core.cpp
@@ -130,12 +130,21 @@ bool Core_GetPowerSaving() {
 	return powerSaving;
 }
 
+// TODO: Feels like this belongs elsewhere.
 bool UpdateScreenScale(int width, int height, bool smallWindow) {
-	g_dpi = 72;
+#ifdef _WIN32
+	// Use legacy DPI handling, because we still compile as XP compatible we don't get the new SDK, unless
+	// we do unholy tricks.
+	HDC screenDC = GetDC(nullptr);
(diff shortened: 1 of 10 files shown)


InusualZ:
Sorry, but WHO! use Windows XP? :)

hrydgard:
Agree, it's probably time to drop it. But it's a separate issue, most of this work will apply with a modern solution too.

LunaMoo:
@hrydgard why remove smallWindow condition?:3
 Not like I'm using it, but it was making x1 window size usable. If it affects other platforms android with hardware resolution set to x1 ~ which might be popular ~ will also not fit on screen.

 Also noticed when there's too many things on the list or the names are too long, things doesn't get wrapped correctly and fly all around window:
ui
 but that might be related to some earlier thin3D changes.

hrydgard:
Woops, that was not intended... I meant to fix smallWindow but forgot to check that it worked, and I obviously didn't get it right.

It will return soon!

PR #9223, opened 2017-01-19
ppsspp 1.3 stops when I start Metal Gear Solid Peace Walker

AXprime:
Hello.. I'm using ppsspp new . I finished Metal Gear Solid portable ops in ppsspp and it worked fine . But when I tried playing Metal Gear Solid Peace Walker in ppsspp 1.3 .. it stopped.. Do u have any idea how to get it work? Thanks in advance.. btw I'm using win7 64 bit. plz help me fast .. plz dont mind if i made any mistake using github .. i dont know how it works..

### DIFF
--- a/android/src/org/ppsspp/ppsspp/NativeActivity.java
+++ b/android/src/org/ppsspp/ppsspp/NativeActivity.java
@@ -958,7 +958,7 @@ public class NativeActivity extends Activity implements SurfaceHolder.Callback {
     		.setPositiveButton(defaultAction, new DialogInterface.OnClickListener(){
     			@Override
     			public void onClick(DialogInterface d, int which) {
-    	    		NativeApp.sendMessage("inputbox_completed", input.getText().toString());
+    	    		NativeApp.sendMessage("inputbox_completed", title + ":" + input.getText().toString());
     				d.dismiss();
     			}
     		})


AXprime:
i also tried joining the ppsspp forums.... but had no luck..

LunaMoo:
Maybe @hrydgard can help you with the forum problems, but github is not a forum, so stop treating it like one just because you can't register.

 As far as I can tell MGS PW works fine, finished it myself not long ago so most likely your problem is a broken copy and solution quite simple, depending if you bought the game on UMD or from the PSStore make a fresh proper dump or grab pbp. You could also revert ppsspp settings to default as there are lots of especially speedhacks which decreases compatibility resulting in black screen and/or crashes.

 Another thing - that's a second useless pull request you opened despite being told NOT to do that in the first one. Stop doing that, begging for attention and being annoying is always bad, but here will just get you ignored or worse. Closing.

AXprime:
Thanks Buddy for your help..

AXprime:
though ur suggestion didnt worked... i got far like installing data and then the story of the big boss and then the loading screen forever ... and i got that far by changing os win7 to xp...

hrydgard:
Hm, I'd try to NOT install data, but play anyway. I think we've had some issues with that with this game in the past.

I'll look into the forum issues in a while.

LunaMoo:
Yeah install data is completely unnecessary as the special features install data ocassionally unlock can be just unlocked via cwcheats(we have those in this game thread on the forums) and most games use it purely for performance when running from umd drive aka useless.

 That being said I tested the full install data(still have it ~ 886.19 MB as reported on game/info screen) when playing this game and it didn't really broke my game, just started new game to test, no problems here.

 Installing is another potential step for file corruption, but based on similar problems with other games(ie Tactics Ogre) and large amounts of potentially incorrectly dumped iso's in compatibility reports I would assume most install data problems at this point would be caused by warez scene and their habit of ripping games for better compression then users redownloading "fresh copy" of their favorite warez sites.:|

PR #9227, opened 2017-01-23
Centralize some more code, cleanup, hw tess uv bugfix

hrydgard:
also switch the other dirty-flags to 64-bit.

@xebra you'll want to wait for this one as well..


xebra:
Thanks! and I found another issue probably related these.
I'll commit later.

PR #9230, opened 2017-01-24
Fix texture animation with hw tess.

xebra:
@hrydgard I think this bug relate to rev 9d7983e.


hrydgard:
Ah yes, that is needed indeed. Deleted a bit too much..

PR #9235, opened 2017-01-25
The MemArena code has way too many ifdefs. Split it up by platform.

hrydgard:
This is, in a way, also preparation for adding a MemArenaDarwin for use on 64-bit iOS.


hrydgard:
Oops, not quite ready yet...

PR #9241, opened 2017-01-26
Follow the ARM64 ABI better (update the frame pointer).

hrydgard:
Also much simplified code to set up and tear down a stack frame, plus it updates the frame pointer correctly.

I was hoping this would help the iOS jit but it doesn't, there's something else going on. Possibly some alignment issue though I'm not sure yet.

Not sure if worth merging, either way before I merge I'm gonna optimize it some more.


iOS4all:
@hrydgard I want to give you a quick limited method to prevent that micro stutter.

If you have hdmi TV adapter for iOS devices just connect your device to TV By it and run ppsspp you'll notice that frame pointer is stable not changing by strong way and you'll notice there's no stutter at all. But when you disconnect it the the stutter will return back.
I just tell you that maybe it'll helps you for your investigation for fix this issue.

PR #9244, opened 2017-01-27
Script mesa3d enable for HW Tesselation working

mrcmunir:
By default mesa3d run Opengl 3.0 context and GLSL 1.30 so we needed force Opengl 3.3 context and run Ok HW tesselation
this script is workaround for that function work in mesa3d .


hrydgard:
Hm, we can't do this from code? Like requesting a higher GL version in the SDL init?

mrcmunir:
I tried by SDL in PCmain.cpp but I have error if I force by SDL 

SDL_GL_CreateContext failed: Could not create GL context: GLXBadFBConfig

With the variable by Mesa not crash I don't know why hmm..

hrydgard:
Okay, as as stopgap solution until that's fixed I guess this is fine, it's just a shell script anyway so doesn't affect anything else.

PR #9255, opened 2017-01-30
State optimizations (experimental)

hrydgard:
Avoids recomputing things like blendstate, vertexshaderID etc per draw call if none of the underlying information has changed. Should result in a 1-4% speedup overall, which seems to be the case though needs more testing.

Actually for D3D11 the speed increase is a bit bigger than that.

- [x] Need to change vertexFullAlpha logic
- [ ] Validate allowShaderBlend dirties
- [x] Brave Story missing octopus (related to DIRTY_BLEND_STATE somehow)
- [ ] MAX_REASONABLE_BLITS_PER_FRAME?
- [ ] SetDepthUpdated?


zminhquanz:
Do you see improvement in heavy games like GOW ?

hrydgard:
Minor. Needs a lot more testing, there's also a couple more optimizations I want to do before merging this (avoiding recomputing fragment shader IDs too).

daniel229:
Breaks brave story 
this branch


master

hrydgard:
Thanks!

hrydgard:
Just rebased on master. Issues probably still remain.

hrydgard:
Rebased again. Can't be merged as is because the checks need to be added to the other three backends as well, otherwise they might break (though probably not really).

hrydgard:
Now also optimizing out fragment shader ID calculation when possible.

With these changes, D3D11 goes from very fast to bonkers fast in games with many drawcalls.

Though there are still glitches to iron out so not ready for merge, this might have to wait for 1.5 because I'm intending to release 1.4 in about a week.

unknownbrackets:
Brave Story should be fixed now in GLES.  It's still wrong in D3D11, and I didn't check anything in D3D9/Vulkan.  Updated pull description with checkboxes for some other possible issues.

-[Unknown]

hrydgard:
I'm going to re-do this, by first reorganizing the state code without adding the dirty tracking, testing and merging that, and then applying the dirty tracking separately afterwards. 

That will make this more reviewable and less prone to bugs.

Issue #9189, opened 2016-12-26
Tactics orge on powervr is Pixelated/Blocky 2D in-battle sprites ( regression)


zminhquanz:
![screenshot_2016-12-26-20-26-52](https://cloud.githubusercontent.com/assets/11285640/21487664/82b99df2-cc01-11e6-8a46-dfab3a161c29.jpg)
![screenshot_2016-12-26-20-26-48](https://cloud.githubusercontent.com/assets/11285640/21487665/82c21b4e-cc01-11e6-8427-c8085a50b267.jpg)
![screenshot_2016-12-26-20-26-44](https://cloud.githubusercontent.com/assets/11285640/21487666/82d12c6a-cc01-11e6-95b0-eb3e5f305696.jpg)
i choose to render resolution 2x and upscale 2x, and the blurred not fix on powervr

hrydgard:
I think we've seen this before, some weird texture coordinate precision issue... seems to only affect PowerVR indeed.

zminhquanz:
And this game even on PC Windows , sometimes it's slow down

zminhquanz:
Regression from #7150 , bug is appear again on 1.3.0.1 @hrydgard

zminhquanz:
@hrydgard you can mention this issue

lokifrit:
was the pixelated issue for Tactics Ogre fixed? im using Asus zenfone 5, set to buffered rendering. When i switch to Read Frame Buffers to mem (GPU) the pixelated issue is gone, but it is slow

zminhquanz:
@lokifrit is doesn't fixed that , i think it's a regression

lokifrit:
what do you mean by regression? sorry for my ignorance

hrydgard:
regression means that something got worse since the last release, for example.

lokifrit:
oh okay. i tried versions 1.3 and 0.9. I noticed that final fantasy tactics has the same behavior.

hrydgard:
I believe the issue is that the game has a texture matrix enabled, which cause us to perform texture projection (including division), which unfortunately works horribly badly on older PowerVR due to a lack of precision in fragment shader calculations. I don't think the division is actually required, so one way might be to hack around it, another to detect in the texture matrix that no projection will be needed. I think I might try the second..

lokifrit:
is there any other phone model that doesn't have this problem?

hrydgard:
Yeah, nearly all of them.

[...]

PR #9258, opened 2017-01-30
Avoid unnecessary projection. Should fix "low-res look" in some games on older PowerVR

hrydgard:
#9189 
This is going to need some testing, shader generator changes like this are always a bit iffy...


hrydgard:
I think this one's fine, or at least mostly :) Let's merge and see how it goes.

PR #9264, opened 2017-01-31
Turn off Hardware Tessellation if device is unsupported.

xebra:
I think for GLES compatibillity fixes almost completely done.
TODO: To check and disable unsupported device name list(adreno? etc...)


GPU/GPUState.h
```diff
@@ -472,6 +472,9 @@ enum {
 	GPU_IS_MOBILE = FLAG_BIT(29),
 	GPU_PREFER_CPU_DOWNLOAD = FLAG_BIT(30),
 	GPU_PREFER_REVERSE_COLOR_ORDER = FLAG_BIT(31),
+	GPU_SUPPORTS_INSTANCE_RENDERING = FLAG_BIT(32),
```
hrydgard: There's actually space after FLAG_BIT(9) so you don't need to go to 64-bit here... yet :)

hrydgard (approved):

GPU/GPUState.h
xebra: Ah, right...

PR #9282, opened 2017-02-05
D3D9 code moving and cleanup

hrydgard:
This is simply some moving around of code to make the next step of my big FBO refactoring easier.

The goal is to share as much framebuffer management code as possible between the backends, to avoid further code duplication when D3D11 and Vulkan gain buffered rendering support later.


zackdreaver:
There are no builds for windows on orphis, either x32 or x64

http://buildbot.orphis.net/ppsspp/

hrydgard:
Thank you @zackdreaver  , I'll look into it.

xebra:
@hrydgard 
Need changes in dx_fbo.cpp
#include "d3d9_state.h"
to 
#include "gfx/d3d9_state.h"

xebra:
I was wrong and edited.

hrydgard:
Fixed!

Issue #5350, opened 2014-02-06
Mipmapping issue:Tactics Ogre Let Us Cling Together JPN/Chinese translate displaying wrong characters 


daniel229:
wrong
![01](https://f.cloud.github.com/assets/3481559/2097195/521138ae-8f14-11e3-999c-e5b83603a6f8.png)

correct 
The JPCSP builds display the correct characters once was that before this commit merged http://code.google.com/p/jpcsp/source/detail?r=2351
![02](https://f.cloud.github.com/assets/3481559/2097196/52154926-8f14-11e3-9f09-30d4f57fea36.png)

lockcole:
So is there no way to fix it?

I tried a recent build and I found that the problem still exists...

xebra:
This issue is not fixed yet. Now I'm trying but...

![uljm05753_00000](https://cloud.githubusercontent.com/assets/1607422/22403217/fda6c3e8-e653-11e6-8d68-3d516aafa83c.jpg)
![uljm05753_00002](https://cloud.githubusercontent.com/assets/1607422/22403220/fda9abd0-e653-11e6-8f05-40609cb97890.jpg)
![uljm05753_00003](https://cloud.githubusercontent.com/assets/1607422/22403219/fda97a8e-e653-11e6-88dd-061b195868a7.jpg)
![uljm05753_00004](https://cloud.githubusercontent.com/assets/1607422/22403218/fda928a4-e653-11e6-952e-6fd62d695844.jpg)

sum2012:
@hrydgard can we add this game in a compat system ?

hrydgard:
yes, we may have to do that... :/

sum2012:
So,let it fix in v1.5

xebra:
I confirmed the codes by ghost is still working in latest revision.
That's great!

![uljm05753_00000](https://cloud.githubusercontent.com/assets/1607422/22583927/f437af92-ea32-11e6-8a5f-33f061349028.jpg)
![uljm05753_00001](https://cloud.githubusercontent.com/assets/1607422/22583926/f435bb56-ea32-11e6-9d85-550e25f5b7ac.jpg)

hrydgard:
This is probably similar to that water blending issue we saw in another game, where a game set up multiple mipmap levels of the same size and uses mipmap bias to either blend or choose between them, in this case by explicitly setting the desired mipmap level for each character. ghost's code hacks around this by forcing a texture reload when the game chooses mipmap level. 

[...]

PR #9299, opened 2017-02-11
Fix fake mipmap issue related #5350

xebra:
I fixed fake mipmap issue about #5350, it is based ghost(@exhalatio?)'s codes.
This seems may be hacky, and using 3D texture is better solution.
However this issue needs to fix immediately. So I hope to fix this without any problem and I tested with some games. I think this codes working fine but it may need more testing.

Tactics Ogre - Graphics OK
Gran Turismo - Graphics & FPS60 OK
Outrun 2006 - Graphics OK

Also, GE_TEXLEVEL_MODE_CONST returns only integer values. I guess it's not "lod bias", it's "mipmap number". GE_TEXLEVEL_MODE_AUTO returns floating point number.
Researched in Gran Turismo

This is my guess.
GE_TEXLEVEL_MODE_AUTO
Autoselect mipmap from "lod bias" by gpu(graphics API).
GE_TEXLEVEL_MODE_CONST
Select mipmap directly from "mipmap number(lod bias)" by a hand.

So, we may need to fix this codes 

```
				switch (mode) {
				case GE_TEXLEVEL_MODE_AUTO:
					// TODO
					break;
				case GE_TEXLEVEL_MODE_CONST:
					// Sigh, LOD_BIAS is not even in ES 3.0..
#ifndef USING_GLES2
					glTexParameterf(GL_TEXTURE_2D, GL_TEXTURE_LOD_BIAS, lodBias);
#endif
					break;
				case GE_TEXLEVEL_MODE_SLOPE:
					// TODO
					break;
				}
```

to

```
				switch (mode) {
				case GE_TEXLEVEL_MODE_AUTO:
					// Sigh, LOD_BIAS is not even in ES 3.0..
#ifndef USING_GLES2
					glTexParameterf(GL_TEXTURE_2D, GL_TEXTURE_LOD_BIAS, lodBias);
#endif
					break;
				case GE_TEXLEVEL_MODE_CONST:

					break;
				case GE_TEXLEVEL_MODE_SLOPE:
					// TODO
					break;
				}
```

### DIFF
--- a/GPU/Directx9/TextureCacheDX9.cpp
+++ b/GPU/Directx9/TextureCacheDX9.cpp
@@ -605,7 +605,7 @@ void TextureCacheDX9::SetTexture(bool force) {
 	}
 
 	u8 level = 0;
-	if (gstate.getTexLevelMode() == GE_TEXLEVEL_MODE_CONST)
+	if (IsFakeMipmapChange())
 		level = (gstate.texlevel >> 20) & 0xF;
 	u32 texaddr = gstate.getTextureAddress(level);
 	if (!Memory::IsValidAddress(texaddr)) {
@@ -1014,7 +1014,7 @@ void TextureCacheDX9::BuildTexture(TexCacheEntry *const entry, bool replaceImage
 		maxLevel = 0;
 	}
 
-	if (gstate.getTexLevelMode() == GE_TEXLEVEL_MODE_CONST) {
+	if (IsFakeMipmapChange()) {
 		u8 level = (gstate.texlevel >> 20) & 0xF;
 		LoadTextureLevel(*entry, replaced, level, maxLevel, replaceImages, scaleFactor, dstFmt);
 	} else
@@ -1102,7 +1102,7 @@ void TextureCacheDX9::LoadTextureLevel(TexCacheEntry &entry, ReplacedTexture &re
 	int h = gstate.getTextureHeight(level);
 
 	LPDIRECT3DTEXTURE9 &texture = DxTex(&entry);
-	if ((level == 0 || gstate.getTexLevelMode() == GE_TEXLEVEL_MODE_CONST) && (!replaceImages || texture == nullptr)) {
+	if ((level == 0 || IsFakeMipmapChange()) && (!replaceImages || texture == nullptr)) {
 		// Create texture
 		D3DPOOL pool = D3DPOOL_MANAGED;
 		int usage = 0;
@@ -1123,7 +1123,7 @@ void TextureCacheDX9::LoadTextureLevel(TexCacheEntry &entry, ReplacedTexture &re
 			}
 		}
 		HRESULT hr;
-		if (gstate.getTexLevelMode() == GE_TEXLEVEL_MODE_CONST)
+		if (IsFakeMipmapChange())
 			hr = pD3Ddevice->CreateTexture(tw, th, 1, usage, tfmt, pool, &texture, NULL);
 		else
 			hr = pD3Ddevice->CreateTexture(tw, th, levels, usage, tfmt, pool, &texture, NULL);
@@ -1135,7 +1135,7 @@ void TextureCacheDX9::LoadTextureLevel(TexCacheEntry &entry, ReplacedTexture &re
 	}
 
 	D3DLOCKED_RECT rect;
-	if (gstate.getTexLevelMode() == GE_TEXLEVEL_MODE_CONST)
+	if (IsFakeMipmapChange())
 		texture->LockRect(0, &rect, NULL, 0);
 	else
 		texture->LockRect(level, &rect, NULL, 0);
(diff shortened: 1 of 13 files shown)


GPU/GLES/GPU_GLES.cpp
```diff
@@ -875,7 +875,8 @@ void GPU_GLES::FastRunLoop(DisplayList &list) {
 		const u8 cmdFlags = info.flags;      // If we stashed the cmdFlags in the top bits of the cmdmem, we could get away with one table lookup instead of two
 		const u32 diff = op ^ gstate.cmdmem[cmd];
 		// Inlined CheckFlushOp here to get rid of the dumpThisFrame_ check.
-		if ((cmdFlags & FLAG_FLUSHBEFORE) || (diff && (cmdFlags & FLAG_FLUSHBEFOREONCHANGE))) {
+		if ((cmdFlags & FLAG_FLUSHBEFORE) || (diff && (cmdFlags & FLAG_FLUSHBEFOREONCHANGE) 
+			&& (cmd != 0xc8 || (gstate.getTexLevelMode() == GE_TEXLEVEL_MODE_CONST && (0x00FF0000 & gstate.texlevel) != 0)))) { // Avoid always flushing when texlevel(0xc8).
```
hrydgard: Instead of adding this check here, remove the FLUSHBEFOREONCHANGE flag again and add EXECUTE instead, then perform this check inside a handler function and Flush() if necessary. This should be fast enough and cleaner than adding more checks to this already very hot loop.

GPU/Directx9/TextureCacheDX9.cpp
```diff
@@ -604,17 +604,20 @@ void TextureCacheDX9::SetTexture(bool force) {
 		lastBoundTexture = INVALID_TEX;
 	}
 
-	u32 texaddr = gstate.getTextureAddress(0);
+	u8 level = 0;
+	if (gstate.getTexLevelMode() == GE_TEXLEVEL_MODE_CONST)
+		level = (gstate.texlevel >> 20) & 0xF;
+	u32 texaddr = gstate.getTextureAddress(level);
```
hrydgard: I didn't consider this before, but picking the texaddr here will cause the game to create separate textures for each level. This means that the Tactics Ogre style of usage will actually probably perform well, while the choppy video issue from that other game will not be solved. 

Not sure if I like this or not...

hrydgard:
What's the reason this needs "to fix immediately" ?

I do see the reasoning that this is a sane approach when using a CONST miplevel, but it's still not pretty :)

LunaMoo:
Yeah, just to confirm this solution doesn't work for Misshitsu no Sacrifice nor Macross series which are affected by same "fake mipmap" issue.

GPU/GLES/GPU_GLES.cpp
xebra: I'll try to fix it.

xebra:
@hrydgard 
The reason is just simple, it's "to play games correctly".
If we know any hacky way, I think we should implement it a while instead of correct way.
Of course, if it doesn't cause any other problem.

@LunaMoo 
Thank you for the reaport about the two games.
Hmm, I didn't research Misshitsu and Macross. Well, it may seem not same issue. For now, I have no idea, I need research the games anyway.

Does this fix affected only Tactics Ogre JPN/CN?
Also, this pull request to use specialized patch for Tactics Ogre JPN/CN :p

GPU/Directx9/TextureCacheDX9.cpp
xebra: Anyway, I/we need more research about that.

hrydgard:
I try to avoid hacks as much as possible - if there's a nice solution then it's better to use that. Though in this case, this hack may be the only viable solution on OpenGL ES 2.0 devices so I might consider using it anyway.

As mentioned though, the hack will not work with games that want to use mipmap blending to transition between two images, as the hack makes the emulator create separate textures, instead of a blendable 3d texture.

xebra:
Ah, mipmap blending... Thank you for the explain what is the issue. 
I know this hack is to separate texture with each levels, so it effective with Tactics Ogre, but it's not effective to blending issues.
Yes, we also need to consider how to fix these issues on GLES 2.0 instead of using 3D texture.

xebra:
I added this hacks to compat system.

BTW, I probably understood about git how to "push" again to already existed pull requests with "Fetch&Rebase" :)
I was failing it in the past because it was rejected and I was forgetting to add a "force" option.

hrydgard:
Yeah, when you rebase you change the history, and that means you need the force flag.

hrydgard:
I like the code much better now. It's still on the edge of what I find to be acceptable hackiness-wise, but there's not likely to be a more reasonable solution for ES2.0 devices at least, and it really isn't very complex, so let's go with this for now.

xebra:
Thanks :)
and I found two issues, I'll fix it later.

PR #9305, opened 2017-02-14
Remember Last Remote ISO server

fireballwayne:
This change remembers the last remote iso server it was connected to and tries to connect to that first.  It also allows you to manually specify a server by editing the ini.
I have other changes in mind for remote ISOs.  But first a question.  Right now it uses reports.ppsspp.org to find the server which is down.  Is that a permanent change?  If so we need to find another server location method.
Other changes I have in mind:
A manual option in the GUI to specify server
Ability to read apache directory index output. 
Ability to read from subdirectory on webserver. 

I have it working with apache now, but I had to code some php to make the directory listing look like the ppsspp output.

### DIFF
--- a/UI/RemoteISOScreen.cpp
+++ b/UI/RemoteISOScreen.cpp
@@ -188,6 +188,17 @@ static bool FindServer(std::string &resultHost, int &resultPort) {
 	Buffer result;
 	int code = 500;
 
+	// Not sure why this is needed here, but windows fails without it.
+	net::Init();
+
+	// Try last server first, if it is set
+	if (g_Config.iLastRemoteISOPort && g_Config.sLastRemoteISOServer != "" && http.Resolve(g_Config.sLastRemoteISOServer.c_str(), g_Config.iLastRemoteISOPort) && http.Connect()) {
+		http.Disconnect();
+		resultHost = g_Config.sLastRemoteISOServer;
+		resultPort = g_Config.iLastRemoteISOPort;
+		return true;
+	}
+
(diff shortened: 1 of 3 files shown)


hrydgard:
Hm, no. I'll try to bring that server up again.

UI/RemoteISOScreen.cpp
```diff
@@ -189,6 +189,17 @@ static bool FindServer(std::string &resultHost, int &resultPort) {
 	Buffer result;
 	int code = 500;
 
+	// Not sure why this is needed here, but windows fails without it.
```
hrydgard: Well, you need to call net::Init() somewhere on Windows. Kind of odd to call it in here though... 

I'm going to clean our net init up before we merge this.

hrydgard:
Alright, cleaned that up. Please remove your call to net::Init.

fireballwayne:
Not sure I did that correctly.  Tried to rebase it  on your new code.  I'm new to git/github.

hrydgard:
Hm, no, something went wrong.

I assume you have added ppsspp as "upstream", and your clone as "origin".  (To add upstream, `git remote add upstream https://github.com/hrydgard/ppsspp.git` ).

Switch to your brand if you haven't already. Type:

    git fetch upstream
    git rebase upstream/master
    git push -f origin remoteiso

hrydgard:
@fireballwayne  wanna give it another try with those instructions?

fireballwayne:
Thanks that fixed it.

PR #9306, opened 2017-02-14
Import SPIRV-Cross. This will be used later to translate post-processing shaders...

hrydgard:
…to the various shader languages we use.

Eventually, this will make it possible to use post-processing with Vulkan and D3D11. Probably not DX9, though maybe.


hrydgard:
Can't get it to build on Android right now (silly issue with std::to_string), but merging anyway, as the main goal is D3D11. Will eventually need it on Android too to run post processing shaders on vulkan but that can wait.

mrcmunir:
@hrydgard  Broken linux build? 
CMake Error at ext/SPIRV-Cross/CMakeLists.txt:23 (message):
  Build out of tree to avoid overwriting Makefile

hrydgard:
Hm, travis should've caught that.. I'll look into it.

hrydgard:
After a `git submodule update --init --recursive`, things seem to build alright on my mac using CMake.

hrydgard:
Oh yeah, make sure to not just do "cmake .".  It's better to use ./b.sh.

mrcmunir:
@hrydgard  Ah I see working fine with ./b.sh  or mkdir build folder thanks

iOS4all:
@hrydgard nice job. You're an active person and always helps people who have issues.
Keep going man.

PR #9345, opened 2017-02-22
D3D11: Expand 16-bit textures to 32-bit if Win8+-only texture formats are not supported

hrydgard:
This should help compatibility of the D3D11 backend with Windows 7.


ufdada:
This seems to break texts in loco roco 2 on Windows 10

ppsspp-v1.3-705-gb94ad12-windows-amd64:
uces01059_00004

ppsspp-v1.3-709-g364b7e2-windows-amd64:
uces01059_00003

hrydgard:
Huh, I'll take a look.

hrydgard:
I have the game and it works perfectly fine here...

ufdada:
Weird, maybe it broke because i use the original psp fonts? Also i am using the European version (UCES01059).

It's also very interesting they sometimes seem to work properly... Like when they are generated a second time.

hrydgard:
Hm, interesting...

ufdada:
Ok while testing around i found the cause of the problem: it's the texture scaling which is broken. If i turn it off it works just fine

GPU/Directx9/GPU_DX9.cpp
```diff
@@ -473,11 +473,12 @@ void GPU_DX9::UpdateCmdInfo() {
 void GPU_DX9::CheckGPUFeatures() {
 	u32 features = 0;
 
-	features |= GPU_SUPPORTS_4BIT_FORMAT;
+	features |= GPU_SUPPORTS_16BIT_FORMATS;
 	features |= GPU_SUPPORTS_BLEND_MINMAX;
 	features |= GPU_SUPPORTS_TEXTURE_LOD_CONTROL;
 	features |= GPU_PREFER_CPU_DOWNLOAD;
 	features |= GPU_SUPPORTS_ACCURATE_DEPTH;
+	features |= GPU_SUPPORTS_16BIT_FORMATS;
```
ufdada: Is this duplication intended?

hrydgard: Nope, thanks.

hrydgard:
@ufdada Ah! That makes sense, and easy for me to test. I'll get it fixed soon.

ufdada:
Glad i could help you :)

PR #9347, opened 2017-02-22
Don't minimize PPSSPP when losing focus while fullscreen

### LINKED ISSUE #8729: Full screen deactivates when active work space is changed.
OS: `Slackware64-current`
`spectrwm-3.0.1-x86_64-1_SBo`
`ppsspp-2c84411_2016.05.07_master-x86_64-1_git`

When using ppsspp with the window manager spectrwm any time I toggle full screen on and then go to another work space (To complete other tasks) I find that ppsspp is no longer full screen when I return to the work space it is running in. The full screen should not reset every time the active workspace is changed. However full screen will not be reset if ppsspp itself is moved to another workspace.

asl97:
Fixes #8729.

edit: do tell if there is a better place to put the hint.

### DIFF
--- a/ext/native/base/PCMain.cpp
+++ b/ext/native/base/PCMain.cpp
@@ -420,6 +420,7 @@ int main(int argc, char *argv[]) {
 	bcm_host_init();
 #endif
 	putenv((char*)"SDL_VIDEO_CENTERED=1");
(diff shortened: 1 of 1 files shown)


hrydgard:
Should be fine there.

PR #9348, opened 2017-02-22
Recreate all view on resize

### LINKED ISSUE #9340: On-screen touch controls position isn't updated when using full screen keyboard toggle
The on screen controls isn't update to the new screen size when using the keyboard to toggle the full screen.

Note: It's the same with manually resizing the window. (not available on Qt?)

In order to get it to update the position of the controls, one have to open the menu (esc) and then close it (esc again).

Affect both SDL and Qt.

ps: title is kind of long, feel free to edit it

asl97:
Fixes #9340

Not quite sure about performance but the NativeResized shouldn't be call during normal play.

Should it be moved inside `if (uiContext) {`?

### DIFF
--- a/UI/NativeApp.cpp
+++ b/UI/NativeApp.cpp
@@ -992,6 +992,7 @@ void NativeMessageReceived(const char *message, const char *value) {
 
 void NativeResized() {
 	resized = true;
+	screenManager->RecreateAllViews();
 
 	if (uiContext) {
 		// Modifying the bounds here can be used to "inset" the whole image to gain borders for TV overscan etc.


hrydgard:
Nah, I think there's a better way, NativeApp shouldn't know about views (and screenmanager really shouldn't expose that function). Screen/ScreenManager already has a resized() method that is not properly hooked up, the class UIScreen should implement that and do the right thing.

I'll take a look tomorrow.

hrydgard:
I implemented an alternative fix: 900539e200ac6fc5fd9935d4ede984e34f6d9600

PR #9350, opened 2017-02-23
D3D11 postprocessing shaders

hrydgard:
As our post shaders are written in GLSL, we use glslang + SPIRV-Cross to translate them to SPIR-V and then to HLSL. This works, though there is some kind of filtering or precision issue with the 5X postshader. Oh well, at least it mostly works, worth merging.

This is the last missing feature in #9317, just needs some more debugging...

### DIFF
--- a/GPU/Common/ShaderTranslation.cpp
+++ b/GPU/Common/ShaderTranslation.cpp
@@ -19,6 +19,7 @@
 
 #include <memory>
 #include <vector>
+#include <sstream>
 
 // DbgNew is not compatible with Glslang
 #ifdef DBG_NEW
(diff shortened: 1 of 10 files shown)


zminhquanz:
I have a question , SPIR-V have it work on OpenGL ES 2.0 or 3.0

hrydgard:
SPIR-V is how you feed shaders to Vulkan, and it's also a generic "binary shader language" that's easy to transform. In this case we compile the OpenGL postprocessing shaders to SPIR-V, then decompile to HLSL.

LunaMoo:
Nice:), will see if I can find what's wrong with 5xBR, if I recall the most confusing part of at least manual translation of more complex shaders are inverted(transposed?) matrices between GLSL/HLSL, rows become columns etc.

hrydgard:
Editing posts here actually don't cause bumps :)

Yeah, I'll throw in fract->frac. That and the others really should be in spirv-cross, I reported it as an issue upstream.

PR #9353, opened 2017-02-24
Make sound recording sane and match video recording behaviour.

LunaMoo:
Current behaviour:
- records if the feature was enabled from the start of running the game, to the moment game was turned off assuming sound recording was still enabled(otherwise it never stops).

 Feels bad and confusing to users, well at least to one spamming on the forums, was confused myself when I tested it so with the proposed change:
 - records starting from when the feature was enabled, to the point it was disabled, just like video. 

Assuming most people want to record both at the same time that should make the feature much more useful.


hrydgard:
Thanks!

PR #9356, opened 2017-02-25
Hardware Tessellation on Vulkan

xebra:
This implementation being not tested enough yet, but at least working perfect with "bezier in pspgl", so probably it will work fine.

Also, DX11 implementaion has still some issues, so wait for a while.

### DIFF
--- a/GPU/Vulkan/DrawEngineVulkan.cpp
+++ b/GPU/Vulkan/DrawEngineVulkan.cpp
@@ -55,6 +55,9 @@ enum {
 	DRAW_BINDING_DYNUBO_BASE = 2,
 	DRAW_BINDING_DYNUBO_LIGHT = 3,
 	DRAW_BINDING_DYNUBO_BONE = 4,
+	DRAW_BINDING_TESS_POS_TEXTURE = 5,
+	DRAW_BINDING_TESS_TEX_TEXTURE = 6,
+	DRAW_BINDING_TESS_COL_TEXTURE = 7,
 };
 
 enum {
@@ -87,12 +90,12 @@ DrawEngineVulkan::DrawEngineVulkan(VulkanContext *vulkan)
 
 	InitDeviceObjects();
 
-	tessDataTransfer = new TessellationDataTransferVulkan();
+	tessDataTransfer = new TessellationDataTransferVulkan(vulkan);
 }
 
 void DrawEngineVulkan::InitDeviceObjects() {
 	// All resources we need for PSP drawing. Usually only bindings 0 and 2-4 are populated.
-	VkDescriptorSetLayoutBinding bindings[5];
+	VkDescriptorSetLayoutBinding bindings[8];
 	bindings[0].descriptorCount = 1;
 	bindings[0].pImmutableSamplers = nullptr;
 	bindings[0].descriptorType = VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER;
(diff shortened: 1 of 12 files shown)


iOS4all:
@xebra wow that's really amazing. Thanks for your hard work.
I just wondering if this can be port to iOS as well?

xebra:
@iOS4all Thanks! Well, I don't know anything about iOS :)

Nightbane112:
@iOS4all I really doubt that. Apple is going ahead with its own API, [Metal](https://developer.apple.com/metal/). Supported platforms for Vulkan is stated on the [Khronos website](https://www.khronos.org/vulkan/). There are some work going on to make Vulkan run on iOS & macOS via [MoltenVK](https://moltengl.com/moltenvk/) but you'll need to pay to use them.

hrydgard:
Nice work! I'm a little bit concerned about race conditions with the data transfer textures - if there are multiple splines drawn during a frame they could probably clash, unless we make sure to create a new texture for each transfer (and schedule them for destruction when the frame is fully done with context.Delete().QueueDeleteTexture()). Did you try this with Pursuit Force?

xebra:
@hrydgard Thanks! 

No, I can't try that because it can't work with/without HW tessellation.
Also FF4 has same reason.
However, At least, "bezier in pspgl" has two splines in one frame. Did you know it?

xebra:
And old image is queued for deleting by wipe() in unlock(). So I think it solve your concern.

hrydgard:
Oh, right, I forgot it worked that way :) Then yeah, this should be fine.

xebra:
Yeah, your backend framework is great!
Actually, I don't know about Vulkan well yet, but I will learn it more and I hope to contribute to backend more stable :)

ufdada:
Doesn't work for LocoRoco 2. Crashes directly on the main screen with HT, without it works fine.

xebra:
Oh, thanks. I'll look into it.

xebra:
Hmm, I can't see the problem. Seems working fine.
However, HW tess seems not faster than SW tess in DEBUG mode, so it needs more improved.

ufdada:
Weird... Well i have a nvidia card. I'll check for newer drivers, maybe that's the problem.

ufdada:
Okay when i tried to debug it, the following message came:

```
ERROR: [Driver] ImageView Code 0 : vkCreateImageView: cannot create an image view from this an image of this type
```
Here's the content of view_info at this time (VulkanImage.cpp line 332):
image

After continuing it shows this message

```bash
Assertion failed!

Program: D:\Programme\GitHub\ppsspp\PPSSPPDebug.exe
File: Vulkan\VulkanImage.cpp
Line: 273

Expression: res == VK_ERROR_OUT_OF_HOST_MEMORY || res == VK_ERROR_OUT_OF_DEVICE_MEMORY || res == VK_ERROR_TOO_MANY_OBJECTS
```
res actually has the value `VK_ERROR_VALIDATION_FAILED_EXT` at this time

Then i can just resume without crashing till this repeats again after a few seconds (the loco roco are shown correctly btw.). Also this is happening on the exact same spots in the main menu animation.

Video card is a nvidia 670 with 2 gb and the newest drivers

hrydgard:
Strange stuff. R32G32B32 is kind of an odd format with 12-byte pixels, might be worth trying R32G32B32A32 instead, though that behaviour is weird.

xebra:
Really strange, I have no idea for now.

GPU/Vulkan/DrawEngineVulkan.cpp
```diff
@@ -900,6 +947,40 @@ bool DrawEngineVulkan::IsCodePtrVertexDecoder(const u8 *ptr) const {
...
-{
+void DrawEngineVulkan::TessellationDataTransferVulkan::SendDataToShader(const float * pos, const float * tex, const float * col, int size, bool hasColor, bool hasTexCoords) {
+	int rowPitch;
+	u8 *data;
+
+	// Position
+	if (prevSize < size) {
+		prevSize = size;
+
+		data_tex[0]->CreateDirect(size, 1, 1, VK_FORMAT_R32G32B32_SFLOAT, VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL);
```
unknownbrackets: @xebra FWIW, if you look here:
https://vulkan.lunarg.com/doc/view/1.0.30.0/linux/vkspec.chunked/ch31s03.html#features-formats-mandatory-features-32bit

These are the formats which are mandatory - every device must have them.  `VK_FORMAT_R32G32B32_SFLOAT` is only mandatory for vertex buffers.  The safest thing, as @hrydgard suggested, is to use `VK_FORMAT_R32G32B32A32_SFLOAT` which is required always (even though it wastes a bit of space.)

We could also detect support and write the optimal texture based on what's available, though we'd need to write the data dynamically then.

-[Unknown]

unknownbrackets:
By the way, it would be ideal if `SendDataToShader` was changed into two functions, at least for some backends.  Modern GL would even support a buffer too, but I'm not sure about D3D11 (don't know it at all.)

Example:

```c++
bool useHW = g_Config.bHardwareTessellation && g_Config.bHardwareTransform && !g_Config.bSoftwareRendering;
float *pos = (float*)(decoded + 65536 * 18); // Size 3 float
float *tex = pos + count_u * count_v * 3; // Size 3 float
float *col = tex + count_u * count_v * 3; // Size 4 float

if (useHW) {
    tessDataTransfer->PrepareBuffers(pos, tex, col, count_u * count_v, hasColor, hasTexCoord);
}

// ... Populate ...

if (useHW) {
    tessDataTransfer->SendDataToShader(pos, tex, col, count_u * count_v, hasColor, hasTexCoords);
}

Vulkan::PrepareBuffers(float *&pos, float *&tex, float *&col, int size, bool hasColor, bool hasTexCoords) {
    auto posTexture = CreatePosTexture();
    pos = posTexture->Lock(0, &rowPitch);

    tex = ...
}
```

This would just avoid a copy, since it'd already be in the buffer and `SendDataToShader` could just `Unlock()` the textures.

Anyway, cool stuff.

-[Unknown]

hrydgard:
Same thing goes for D3D11, yeah.

xebra:
@unknownbrackets Yeah, your idea is cool! I was bothering the wasted copy. Thanks a lot :)

GPU/Vulkan/DrawEngineVulkan.cpp
xebra: Thanks for the information. It seems the @hrydgard's suggest is right. I will try to fix it.

PR #9364, opened 2017-02-27
Get rid of our own silly reimplementation of std::thread

hrydgard:


InusualZ:
Ship it!

hrydgard:
Just waiting for the build checks :)

Then I'm going to do the same with std::mutex, that one's a little bit more involved though.

Issue #9369, opened 2017-03-01
ppsspp direct3d 11 


moura464:
![aaaaaaa](https://cloud.githubusercontent.com/assets/22099440/23482938/a267f442-feaf-11e6-85e7-46640565a85f.png)

moura464:
Direct3d11 error  :/ ????

hrydgard:
I don't see anything wrong in the screenshot?

moura464:
![adasdadaeqeeee](https://cloud.githubusercontent.com/assets/22099440/23484469/dfe60f2e-feb5-11e6-9282-9317a1017aed.png)

mrcmunir:
@moura464 What it's you gpu and with driver version?

moura464:
@hrydgard http://www.rastertek.com/dx11s2tut03.html

tywald:
@moura464 can you name your graphics card?(placa gráfica).

hrydgard:
That might have gotten D3D11 working on some older GPUs.

moura464:
Build working ![1f603](https://cloud.githubusercontent.com/assets/22099440/23588419/1d0739d0-019c-11e7-82c4-a694bd94fdcb.png)

hrydgard:
great!

moura464:
@hrydgard Build travis ci ?????

PR #9387, opened 2017-03-05
Backport some code from my UWP branch to keep it small

hrydgard:
Might as well get this in and tested.

This might help #9369  .

### DIFF
--- a/Common/ArmCPUDetect.cpp
+++ b/Common/ArmCPUDetect.cpp
@@ -208,6 +208,13 @@ void CPUInfo::Detect()
 #endif
 	strcpy(brand_string, "Apple A");
 	num_cores = 2;
+#elif PPSSPP_PLATFORM(UWP)
+	strcpy(brand_string, "Unknown");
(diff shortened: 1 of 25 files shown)


hrydgard:
Actually this doesn't solve the whole problem for 9369, working on it.

xebra:
This can't compile HW tess.

```
error X4016: SV_InstanceID semantic cannot be used with 10Level9 targets
error X3548: in vs_4_0_level_9_1 uints can only be used with known-positive values, use int if possible
```
Also "featureLevel_" is not initialized.

moura464:
build working YES 1f609

xebra:
Ah, my issues was caused by uninitialized "featureLevel_" and "vs_4_0_level_9_1" .
I confirmed it can compile with vs_4_0.

hrydgard:
That's also what's causing trouble in #9361, working on a fix.

xebra:
Fixed 👍

PR #9388, opened 2017-03-05
Add PostShader option to force 60fps for animated effects

LunaMoo:
Option that will make animated post process effects consistent in games with variable fps or when it drops to 0 like during boot, loading screens or that PSPKVM homebrew etc., was requested on forums some time ago as well.

 Not sure if that's best way to do it, but works fine for post process animation without affecting fps counter.


hrydgard:
How does this interact with non-buffered rendering? Not that post effects work at all with that, just want to be sure it doesn't cause unexpected flicker or something ... If that works fine I'm happy to merge this.

LunaMoo:
I forgot about this interaction, thanks for reminder fixed now.

PR #9414, opened 2017-03-11
Update ppsspp_de.ts

Atalanttore:
Updated and almost completed German translation.


hrydgard:
Hm, we should really make the Qt stuff somehow reuse the other translation where possible.

Anyway, seems fine to me so let's merge.

PR #9424, opened 2017-03-13
Use $(DefaultPlatformToolset)_xp for easier VS builds with different toolsets.

### LINKED ISSUE #9423: Feature request - Build system - Visual Studio - Version independent solution
PCSX2 developers did this, it made things hassle free when multiple Visual Studio versions are supported like now. We have again 2 supported versions - Visual Studio 2015 and 2017.

LunaMoo:
Fixes #9423 

Seems like build was broken in https://github.com/hrydgard/ppsspp/commit/0079f745fe3cdc1946b583493a75e5aa39837138

error: no such remote ref b16f7e6819267e57c3c244808d1981f0ce34acbc
Fetched in submodule path 'ext/glslang', but it did not contain b16f7e6819267e57c3c244808d1981f0ce34acbc. Direct fetching of that commit failed.

Weirdly it exist:
https://github.com/hrydgard/glslang/commit/b16f7e6819267e57c3c244808d1981f0ce34acbc
@_@

### DIFF
--- a/Common/Common.vcxproj
+++ b/Common/Common.vcxproj
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<Project DefaultTargets="Build" ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <ItemGroup Label="ProjectConfigurations">
     <ProjectConfiguration Include="Debug|Win32">
       <Configuration>Debug</Configuration>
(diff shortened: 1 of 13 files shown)


Orphis:
I wonder if we shouldn't target the latest Windows SDK instead of 8.1.

LunaMoo:
WindowsTargetPlatformVersion was added automatically when changing PlatformToolset from VS, didn't really noticed.
 But I think that's the latest for _xp anyway. Can only set higher using v141 toolset(without _xp) and then it changes target platform from windows to windows 10

hrydgard:
Yeah, can't target 10 unless we drop xp support.

Orphis:
It's not about targeting 10, it's just using the latest SDK that has potentially tons of bugfixes rather than an old one. They can all target XP the same I believe.
Targeting XP by hand is actually simple, you just need to pass the subsystem option when linking and have some preprocessor macro defined. It's still the same compiler in the end!

hrydgard:
Wrong, the latest SDK cannot be used to target XP.

Orphis:
Hmmm, seems like you are right about the SDK. Wonder why I thought otherwise.
The rest still applies though if you ever want to target XP manually (eg from CMake).

PR #9435, opened 2017-03-16
Add Hyllian's xBR-lv2 Shader Accuracy (tweak by guest.r)

### LINKED ISSUE #9425: Low quality result of 5xBR postprocessing shader
I've noticed the 5xBR postprocessing shader in PPSSPP looks poorly comparing to other emulators.

5xbr_ppsspp
Here's a scalling result of PPSSPP (window size was set to 3xPSP)

3xbr
For comparision, here's the same frame captured then scale using 3xBR algorithm with ScalerTest.exe provided in [xBRZ project](https://sourceforge.net/projects/xbrz/) (I'm not sure which version of xBR it is though, the source for this tool doesn't seem to be available)

In the implement pull request @LunaMoo said it's "A simple and fast variant of 5xBR". I wonder if these artifacts shown above are the result of "simple and fast" or it's caused by a improper implement of the shader or something else in PPSSPP?

Also (not sure if it's related) if I enable 5xBR and set window size to 1xPSP, the result would be terrible. You can repreduce this behavior with any game so I'm not uploading any screenshots. Nothing like that would happen to other upscaling shaders like 4xHQ GLSL / Spline36 Upscaler.

(The game in screenshots above was NPJH90092, trial version of "Misshitsu no Sacrifice" which you can aquire from developer's site)

LunaMoo:
Not sure if we should include two variants of 5xBR, differences from the one we have includes:
 - blends the edges, resulting in smoother image ~ might help on lower/non standard resolution displays,
 - somewhat different color recognition ~ seems to deal with "Misshitsu no Sacrifice" transparent green layer being ocassionally blend together with background,
 - different corner handling keeping more squareish shapes,
 - performance(it's more than twice as demanding).

 And that last detail and ppsspp being popular on low end devices is primary reason why I'm adding it as new effect in this pr since while playing games and not comparing screenshots in many especially simpler games they'll produce pretty much same effect.

Fixes #9425 ~ at least it should handle this game better:
new:5xbrnew
old:5xbrold
might need to wait for @RadarNyan opinion since he was comparing higher res xBRZ with integer scale(from external tool) to low res 5xBR with potentially some weird scale and that might still not look good enough. It should at least be slightly better due to extra smoothing.

### DIFF
--- a/assets/shaders/5xBR-lv2.fsh
+++ b/assets/shaders/5xBR-lv2.fsh
@@ -84,6 +84,12 @@ float c_df(vec3 c1, vec3 c2) {
 
 void main() {
 
+ bool upscale = u_texelDelta.x > (1.6 * u_pixelDelta.x);
+ vec3 res = texture2D(sampler0, v_texcoord0.xy).xyz;
+
+ // Let's skip the whole scaling if output size smaller than 1.6x of input size
+ if (upscale) {
+
 	vec4 edri, edr, edr_l, edr_u, px; // px = pixel, edr = edge detection rule
 	vec4 irlv0, irlv1, irlv2l, irlv2u;
 	vec4 fx, fx_l, fx_u; // inequations of straight lines.
@@ -205,8 +211,8 @@ void main() {
 	res2 = mix(res2, mix(F, B, px.y), maximos.y);
 	res2 = mix(res2, mix(D, H, px.w), maximos.w);
 
-	vec3 res = mix(res1, res2, step(c_df(E, res1), c_df(E, res2)));
-
+	res = mix(res1, res2, step(c_df(E, res1), c_df(E, res2)));
+ }
 	gl_FragColor.xyz = res;
 	gl_FragColor.a = 1.0;
 }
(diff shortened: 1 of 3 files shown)


hrydgard:
I'm happy merging this, adding more shaders is a safe thing and not really annoying unless we have way too many. Should probably clean out some of the worst old shaders though, like the color space rotation :)

If you want to wait for RadarNyan's opinion that's fine, just let me know when you want this merged.

LunaMoo:
If you're fine with two similar shaders, then sure, you can merge.

 That issue had 3 problems:
 - those artifacts along green/transparent selection box,
 - lower resolution,
 - confusion with x1 window size that can't do any scaling.

 This new xBR lv2 variant definitely helps with the first while the latter is always limited by user display resolution / window size. And for the last one, to avoid glitches from rounding still being in effect without scaling, I just made a small change which will skip whole algorithm from both 5xBR effects when output res < (1.6x input res).

Fell:
As feedback from a user: "Hybrid" should be called "xBRZ + Linear" or something to make clear what it is. I was wondering for days!

hrydgard:
Yeah, that would make more sense..

PR #9459, opened 2017-03-19
Fix assets directory for non-mobile, non-Windows Qt builds

### LINKED ISSUE #9453: Qt builds do not recognize /usr/share/ppsspp/assets
I've noticed that if you do both an SDL and a Qt build, the SDL build will recognize /usr/share/ppsspp/assets, but the Qt build will not.

This behavior seems to be set here:

https://github.com/hrydgard/ppsspp/blob/fc648f5f32ae77f67acbf77c361527f1c720a0e9/UI/NativeApp.cpp#L317

I'd submit a pull request, but I'm not sure if this is intentional.

I'm referring to Linux builds, btw.

duganchen:
Qt builds on Linux, OSX, etc will now recognize assets in /usr/share/ppsspp/assets.

This fixes #9453

### DIFF
--- a/UI/NativeApp.cpp
+++ b/UI/NativeApp.cpp
@@ -320,7 +320,8 @@ void NativeInit(int argc, const char *argv[], const char *savegame_dir, const ch
 #elif defined(IOS)
 	// Packed assets are included in app
 	VFSRegister("", new DirectoryAssetReader(external_dir));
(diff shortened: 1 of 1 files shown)


hrydgard:
Though we should probably make sure it looks at local assets first - if you have one global install of an older ppsspp and a newer one in a directory, it might use the wrong assets which would be bad considering how the texture atlases work...

PR #9461, opened 2017-03-19
Fix linking against system GLEW and Snappy

### LINKED ISSUE #9443: Linux build broken (can not link snappy)
With latest git version I get the following error when trying to build with b.sh:

-- Found Snappy: /lib64/libsnappy.so  
CMake Error at CMakeLists.txt:389 (target_link_libraries):
  Target "Snappy::Snappy" of type UNKNOWN_LIBRARY may not be linked into
  another target.  One may link only to STATIC or SHARED libraries, or to
  executables with the ENABLE_EXPORTS property set.

Orphis:
Should fix #9443

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -386,7 +386,7 @@ add_subdirectory(ext)
 include_directories(Common)
 setup_target_project(Common Common)
 
-target_link_libraries(Common Snappy::Snappy)
+target_link_libraries(Common Ext::Snappy)
 
 if(WIN32)
 	include_directories(dx9sdk/Include)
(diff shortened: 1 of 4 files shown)


mrcmunir:
Target "armips" of type EXECUTABLE may not be linked into another target.
  One may link only to STATIC or SHARED libraries, or to executables with the
  ENABLE_EXPORTS property set.

armips needed too

Orphis:
`git submodule update --init --recursive`

mrcmunir:
Thanks fixed yet argh why the submodules are not automatically updated

PR #9481, opened 2017-03-22
Fix minor glitch with 5xBR effects.

LunaMoo:
Weirdly it only made a difference for one resolution setting of some snes emulator.


hrydgard:
Would be nice with some screenshots, but I'll take your word for it :)

LunaMoo:
slayers snes
~adding screenshot :3.

PR #9487, opened 2017-03-23
Revert "Fix minor glitch with 5xBR effects."

LunaMoo:
Sorry, it appears there's a difference / probably a bug / between D3d11 and OGL in what those uniforms passes, that change looked fine in d3d11, but made stuff ugly in ogl, affecting also native psp games:/.

### DIFF
--- a/assets/shaders/5xBR-lv2.fsh
+++ b/assets/shaders/5xBR-lv2.fsh
@@ -96,7 +96,7 @@ void main() {
 
 	vec2 pS  = 1.0 / u_texelDelta.xy;
 	vec2 fp  = fract(v_texcoord0.xy*pS.xy);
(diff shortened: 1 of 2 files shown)


hrydgard:
That's odd :/  Well I have noticed that there is something different with how 5xBR looks in D3D11 and it's not for the better. Wonder if the glslang+spirv-cross combo miscompiles something.

LunaMoo:
Maybe there's a difference in buffer width or height in d3d11?
I usually test a differences in post shaders by a code like that:
```
	float anim=(sin(u_time.x)+0.9)*0.55;
	if(v_texcoord0.x<anim) {
		//original code
	} else {
		//change I'm looking at
	}
```
and there was no difference at all in d3d11 aside from that snes case I looked at yesterday:|.

PR #9510, opened 2017-03-28
Fix hwtess gles

xebra:
Sorry fix again #9506.
The crash problem with mobile is not fixed yet unfortunately so I added a warning "experimental".


mrcmunir:
Broken with mesa3d driver now I can't selected

PR #9527, opened 2017-04-01
Cull out-of-bounds triangles that were not clipped

hrydgard:
Helps #5001, at least for Toca.

Looks like this may also help #2159, #8251, #5507 and #7026 .

This could also be ported to the other backends with a little bit of effort. Need a uniform to stash the NaN in.

Really not sure how big the guard band should be, and the right way to do this would be to use a geometry shader instead of killing triangles by setting w = NaN, which I'm not sure will do sane things on all hardware. Unfortunately geometry shaders are not available everywhere.


unknownbrackets:
I seem to remember it drops anything that even slightly goes outside the 4096x4096 grid - the one established by the viewport and offset parameters.  But this might be affected by the clip flag (I think that was only depth?)

-[Unknown]

GPU/Common/ShaderUniforms.h
```diff
@@ -43,6 +39,10 @@ struct UB_VS_FS_Base {
 	float blendFixB[4];
 	float texClamp[4];
 	float texClampOffset[4];
+	int spline_count_u;
+	int spline_count_v;
+	int spline_type_u;
+	int spline_type_v;
```
unknownbrackets: Well, I guess these are VS parameters, not FS, aren't they?

-[Unknown]

hrydgard: Right, I forgot we had that split. Didn't really mean to include that move anyway, I'll move them back.

hrydgard:
Yeah, I don't think the clip flag is involved. 4096x4096 sounds reasonable. Should probably transform that space into the clip space (using offset etc as you say) and pass the four "walls" in as uniforms...

unknownbrackets:
See here:
https://github.com/hrydgard/ppsspp/issues/2159#issuecomment-168463373
https://github.com/hrydgard/ppsspp/issues/8251#issuecomment-174250658

I think these two are affected by the same issue.

-[Unknown]

LunaMoo:
#5507 and #7026 could also be affected, they have some exploded vertices going wild far outside the screen.

hrydgard:
Yup, good memory, all of these are almost certainly the same thing. Now just got to figure out how to best implement it... This method works for me, but only because nVidia cards are smart enough to reject triangles with NaN coordinates instead of going insane - I believe it's undefined behaviour so it could also turn the moon into cheese, or whatever.

hrydgard:
This sounds promising though:

https://twitter.com/nthibieroz/status/718065039085846529

LunaMoo:
Sangokushi IX(Romance of the Three Kingdoms 9) still has black background with this, but at least that one has pretty easy workaround(disabling depth test via cwcheat).


 Tested Driver 76 as well which appears completely fixed:
ules00740_00001
 That I guess confirms this branch also works on AMD gpu's:].

 When it comes to this game, once it get's fixed it might generate new group complaining on weird slowdowns unaffected by rendering resolution which I belive is caused by ~ 15k drawcalls:3. Not sure if those drawcalls are even valid as a lot of them have 1 vertex:
vertices_weirdness_driver76
a lot of those also have really crazy coords:
```
X	Y	Z	U	V	Color
1120822.000000	-1120428.000000	205.275391	62.956650	888.494385	00000000
```

hrydgard:
OK, so I really don't understand this. I was under the impression that the PSP worked like this:

* Clip triangles to w=0, splitting as necessary.
* If any vertex is outside the 4096x4096 box, kill the whole triangle.

This means that we should not kill triangles with w < 0 in the vertex shader as they would have gotten clipped by the PSP and thus might end up inside the 4096x4096.

Anyway. I'm getting road triangle dropouts:

debugblack

I decided to create a zoomed-out debug mode which instead of dropping triangles, colored the out-ouf-bound vertices purple:

debugcolor

In wireframe:
debugwire

The game is performing manual clipping of the road into that big "road triangle" (clearly dropping triangles by moving a vertex way off to the right, causing the annoying stripes) but it happens that triangles that should be drawn have vertices way outside the allowed 4096x4096 square (not marked but should be obvious where it is) and reach all the way to the car, causing the visible triangle dropouts. I had expected those to be behind the camera but no, they aren't - they are behind the z=0 of the Z buffer though. So the near Z clip of the PSP doesn't work quite as I thought, otherwise we'd be having these dropouts on the real PSP as well.

Grrr.

hrydgard:
For now I left out the bottom border. If this fixes the rest of the games, then .. meh :) though not really happy about it.

Please test the other affected games again!

unknownbrackets:
Hmm, we can do more testing.  Thinking about it more, from the testing I previously did (was some time ago now though):

 * Depth:
    * CLAMPED when the "clip enabled" flag is ON, so -2.0 becomes 0.0.
    * CLIPPED when the "clip enabled" flag is OFF, so -2.0 is not drawn.  Note: this seems backwards from the name of the flag, and is what we always do.
    * May never have tested whether extreme depth values were culled.  Probably not?

 * Guard band:
    * Pretty sure drawing a box larger than the 4096x4096 size was culled.
    * Don't remember the behavior when only some vertices were inside, but I thought it did cull.
    * Could be the clip flag effects this in some way?
    * Definitely never tested vertical vs horizontal specifically.

With how it handles depth clipping (there's a "fuzz region" that is still clamped, iirc), I wouldn't put it past it to do something weird with the guard band, really...

-[Unknown]

GPU/Common/ShaderUniforms.cpp
```diff
@@ -23,6 +23,28 @@ static void ConvertProjMatrixToD3D11(Matrix4x4 &in) {
...
+		vpWidth = 480;
+	if (vpHeight == 0.0)
+		vpHeight = 272;
+
+	// We assume a symmetric guardband, even though it's not entirely correct to do so - but nearly everything does it
+	// this way and we have space for the NAN in the uniform.
+	// We also assume that everything behind the near clipping plane gets clipped and will thus not in reality
+	// exceed the guardband. This is a bit rough but should be ok.
+	float offsetX = gstate.getOffsetX();
+	float offsetY = gstate.getOffsetY();
```
unknownbrackets: I guess worst case they would just not cull things they should, right?  Maybe we remove these unused vars to reduce confusion?

-[Unknown]

GPU/D3D11/StateMappingD3D11.cpp
```diff
@@ -427,10 +427,11 @@ void DrawEngineD3D11::ApplyDrawState(int prim) {
 	if (rasterIter == rasterCache_.end()) {
 		D3D11_RASTERIZER_DESC desc{};
 		desc.CullMode = (D3D11_CULL_MODE)(keys_.raster.cullMode);
+		// desc.FillMode = gstate.isModeThrough() ? D3D11_FILL_SOLID : D3D11_FILL_WIREFRAME;
```
unknownbrackets: Maybe remove these debugging comments?

-[Unknown]

hrydgard: Done!

GPU/Common/ShaderUniforms.cpp
hrydgard: Well, in theory if the game has the viewport say at the far left of the 4096x4096 square, it could draw something very wide only to the right that we'd now cull but shouldn't. I don't think that's much of a concern though.

hrydgard:
This needs more work clearly, I'm getting some minor glitches in Wipeout on the left side.

hrydgard:
Did an experiment on an variant on D3D11/9 only  - so testers (like @LunaMoo :), please try both OpenGL and D3D11. In this case, we simply don't cull if Z is out of range.

unknownbrackets:
FWIW, neither option seems to fix Brave Story (although I still think it's from culling), but the pre-experiment version removed a chunk of the triangle, the post-experiment removes none of it.

-[Unknown]

hrydgard:
If it removed a chunk of the triangle, it has to be made out of multiple triangles...

Detailed hardware tests will be required to figure this out. Though as-is it does fix some games at least...

LunaMoo:
Driver 76 still works fine in d3d11(d3d9 looks about the same).
In OGL it still glitches a bit, much less than originally, there's like 1 diagonal(more vertical) line flashing for like 1 frame in the intro, can't catch it, but in the same place like 70% of the screen was covered with glitch pretty much constantly so one frame of one tiny line, really not that bad. There are also glitches coming out on top of some smaller buildings however that could be separate bug of OGL as their vertices seems to look ok:
vertices_weirdness_driver76b
the texture used for that glitch is also the texture used for all far/background buildings, so it looks like it draws their distance/simplified models at close range.


 Sangokushi IX now works in both OGL and d3d11 backends:)
uljm05842_00000
 ~ still broken in d3d9

hrydgard:
Thanks for your testing @LunaMoo . This will be a tricky one to get completely right...

sum2012:
@hrydgard Does not fix #2159 because DX9 still black screen
Please change your word "fix" to "help"

hrydgard:
Thanks @sum2012 , done. Though that can probably be fixed.

LunaMoo:
Sangokushi IX draws the black thing via:
```
X	Y	Z	U	V	Color
362.994385	139.997803	0.000000	185856.000000	35840.000000	9a9a9aff
468.992920	140.997803	0.000000	240128.000000	36096.000000	9a9a9aff
362.994385	155.997559	0.000000	185856.000000	39936.000000	9a9a9aff
468.992920	156.997559	0.000000	240128.000000	40192.000000	9a9a9aff
362.994385	170.997314	0.000000	185856.000000	43776.000000	9a9a9aff
468.992920	171.997314	0.000000	240128.000000	44032.000000	9a9a9aff
362.994385	185.997070	0.000000	185856.000000	47616.000000	9a9a9aff
468.992920	186.997070	0.000000	240128.000000	47872.000000	9a9a9aff
362.994385	200.996826	0.000000	185856.000000	51456.000000	9a9a9aff
468.992920	201.996826	0.000000	240128.000000	51712.000000	9a9a9aff
362.994385	215.996582	0.000000	185856.000000	55296.000000	9a9a9aff
468.992920	216.996582	0.000000	240128.000000	55552.000000	9a9a9aff
362.994385	231.996582	0.000000	185856.000000	59392.000000	9a9a9aff
468.992920	232.996338	0.000000	240128.000000	59648.000000	9a9a9aff
0.000000	0.000000	0.000000	0.000000	0.000000	000000ff
7999.877930	3999.938965	0.000000	4096000.000000	1024000.000000	000000ff

```
 I think only the last rectangle matters here which looks like it's way higher than what should be affected by this branch now and is indeed in other backends, so why fails in d3d9? Maybe it's related to it being rectangle prim and first vertex at (0,0) is within range ~ isn't rectangle divided to two triangles? Or maybe because the texture is disabled for this in opposite to other cases where this branch does work in d3d9?

Margen67:
Does this affect #7459?

thedax:
This actually does help #5507, but it causes issues with missing UI elements (the tachometer during races, parts of the initial loading screens to be black instead of white, and various graphics on the main menu to disappear).

hrydgard:
Hm. Until this is understood better, maybe it makes sense to commit it behind a compat flag which we specifically enable for TOCA and the few other games it helps. But I'm a bit disappointed it isn't working better than it is, I don't understand why.

unknownbrackets:
I think we need to do some hardware testing - it could be the PPS hardware does the culling in some strange way.  Maybe it doesn't even cull based on z at all, or something?

-[Unknown]

hrydgard:
Still a mystery, bumping to 1.6.0

hrydgard:
Rebased on master. Seems to work as well or not-well as before, depending on game.

Maarcis12:
Now NASCAR 07 works flawlessly, Codemasters games still has graphical issues: 
ppssppwindows64 2017-11-27 21-26-47-694
ppssppwindows64 2017-11-27 21-26-58-190

hrydgard:
That game (whichever it is?) looks like it has a completely different issue than the one discussed here. Feel like filing a new issue report?

Maarcis12:
Nuh, that's just how it looks.

unknownbrackets:
Is it possible the bugs here are related to the clipping flag?  Although, not sure, since the software renderer tries to do this using that flag, and has bugs itself...

-[Unknown]

hrydgard:
Could be that the behavior of the hardware z=0 clipper changes, leading to culling due to vertices out of bounds in different situations than we expect, who knows...

hrydgard:
Rebased on master, fixing conflicts again.

unknownbrackets:
Okay, so summarizing hrydgard/pspautotests#194 based on what would achieve those results:

1. Treat `gstate.isClippingEnabled()` as `gstate.isDepthClampEnabled()`.
    1.1. When disabled -> any vert outside [0, 1.0] (possible with some rounding padding) after projection is culled.
    1.2. When enabled -> verts are not culled based on depth, and depth is clamped to [0, 65535] (likely at fragment level - that's a separate problem to worry about for depth testing, etc.)
    1.3. In some backends, we use depth range to handle minz/maxz, which might have to be corrected for when culling.  Example: minz=16384, maxz=49152 -> means [-1.0, 1.0] is depth CLIP range, but outside [-2.0, 2.0] is culled.
    1.4. Using and populating `u_depthRange` for this may be the best bet?

2. Ignore `gstate.isClippingEnabled()` when it comes to X or Y.  Culling is performed regardless of the flag, which only seems to affect depth.

3. Use a uniform to specify the guardband, which is impacted by `gstate.getOffsetX()` and `gstate.getOffsetY()`.
    3.1 For example, `offset(X, Y)= (0, 0)` would cause culling of any vertices with X < 240 or Y < 136.  It is most commonly set to `(2048 - 480/2, 2048 - 272/2)`, the center of the 4096x4096 valid area.
    3.2. Ideally we'd reverse the transform in the uniform values, such that the min/max values could be compared with `gl_Position.xy / gl_Position.z`, I think?
    3.3. Unfortunately, this would probably require 4 floats (either min/max for each coord, or mul&add for each with hardcoded 2048 each way.)

4. We could potentially pass NAN as a dedicated uniform, or perhaps use `0.0` for w?

I think that covers it, but it's possible again that I've misunderstood how it works due to bias (see hrydgard/pspautotests#194).  What do you think @hrydgard?

-[Unknown]

hrydgard:
That all sounds good. but I expect that there might still be some remaining strange edge cases we can't foresee with the clipping or not and the resulting XY positions...

Passing on a couple more floats for the guardband and a NaN value shouldn't be much of an issue though, we still have some fluff to save in our uniform structs. I don't know how reliable passing 0.0 for W is but maybe that works. We can also enable all this on a per-game basis using compat.ini although would be good to avoid that.

unknownbrackets:
There's definitely more to it.  Persona 1 would draw hallways incorrectly from the above rules - see here:
https://github.com/hrydgard/ppsspp/issues/9622#issuecomment-410461523

-[Unknown]

hrydgard:
Oh. Yeah .. no idea about that one. It's a bit annoying how the PSP's rasterizer is mostly so normal, but has these bizarre edge cases....

I'm gonna rebase this again soon, for whatever it's worth.

hrydgard:
Rebased this on master again (squashed a couple of commits to make the rebase easier, there was some back-and-forth going on)

unknownbrackets:
I've confirmed the following so far:

 * That particular draw, with its model/view/projection values, should indeed draw and is definitely affected by the clipping flag.
 * If I tweak the coordinates, I can adjust to a position where the prim is skipped.
 * My original test used an ortho projection matrix, but if I switch to identity and change the viewport to scale 1:1 too, behavior is the same (implying that it's not before viewport.)

-[Unknown]

GPU/Vulkan/ShaderManagerVulkan.cpp
```diff
@@ -65,6 +65,10 @@ VulkanFragmentShader::VulkanFragmentShader(VulkanContext *vulkan, FShaderID id,
 		OutputDebugStringA(LineNumberString(code).c_str());
 		OutputDebugStringA("Messages:\n");
 		OutputDebugStringA(errorMessage.c_str());
+<<<<<<< HEAD
+=======
+		OutputDebugStringA(LineNumberString(code).c_str());
+>>>>>>> Experimental way to cull out-of-bounds triangles. Helps #5001, at least for Toca.
```
unknownbrackets: Oops.

-[Unknown]

unknownbrackets:
Ah... I tried taking the coordinates through M, V, P, and then viewport.  It still worked after MVP, which doubly confirmed the above - then I finally realized the obvious, d'oh.

The vertices that are outside the 4096x4096 space in this case ALSO have depth outside 65535.  If the depth is forced to a valid value, it kills the verts.  So basically, when the depth clamp flag is enabled, clamping bypasses killing the vertex.  But if it's not clamped, it's subject to the 4096x4096 cull.

Now it makes sense again.

-[Unknown]

hrydgard:
"bypasses killing the vertex" must then mean that the hardware performs some sort of clipping against the near plane, making the remaining verts in-bounds in X and Y, because clearly the 0-4096 thing is a hardware limitation.

And that would mean that the clamp flag performs clipping, and maybe the clip flag performs clamping? :)

unknownbrackets:
That makes sense.  In that way, I suppose it makes sense to call it a clipping flag.

But, it's important to note that depth clamp will cause an object entirely behind or in front of the frustum (i.e. before the near plane or after the far plane) to still draw, and not clip to those planes.  Meaning it only clips to the four planes on the sides.

It definitely does clamp depth, though.  If you consider the viewing frustum to be always 4096x4096 and depth 0-65535 (regardless of viewport of minz/maxz), then with depth clamp off, even a single vertex slightly outside the 0-65535 range will cause the triangle not to draw.  Same as outside 4096x4096.

But with it on - unless I'm confused - it will draw even the parts outside the frustum (on the near and far sides), just squished to the near and far sides like a deformed tin can.

-[Unknown]

hrydgard:
Oh I see. The phenomenon where clamping depth makes triangles stay makes sense for vertices that are closer to the camera than the near plane, but not beyond W=0. Those would, without clamping, end up with Z that are too large to interpolate and therefore the GPU throws them away. Maybe it's really that simple and games never really try to draw triangles that cross W=0....

unknownbrackets:
It's possible I'm not testing crossing w=0 and w=1 properly, though.  I'm not exactly sure how to cleanly test that, tbh.

-[Unknown]

rabbitboizaaa (approved):

Issue #9563, opened 2017-04-06
Windows Tablet UI problem (window position)


sum2012:
User report  from Baidu
His tablet is 1280 * 800
Start position problem 
V1.3 
https://postimg.org/image/hbv97h7l1/
V1.4
https://postimg.org/image/tdr38xoyz/

PR #9568, opened 2017-04-07
Windows: Make the window positioning code slightly saner and more straightforward.

hrydgard:
Hopefully this will help #9563 . @sum2012 , it's test time :)  Doing this as a pull request so you'll need to build your own.

### DIFF
--- a/Windows/MainWindow.cpp
+++ b/Windows/MainWindow.cpp
@@ -189,32 +189,27 @@ namespace MainWindow
 		}
 	}
 
-	static void GetWindowRectAtResolution(int xres, int yres, RECT &rcInner, RECT &rcOuter) {
-		rcInner.left = 0;
-		rcInner.top = 0;
-
-		rcInner.right = xres;
-		rcInner.bottom = yres;
-
-		rcOuter = rcInner;
-		AdjustWindowRect(&rcOuter, WS_OVERLAPPEDWINDOW, TRUE);
-		rcOuter.right += g_Config.iWindowX - rcOuter.left;
-		rcOuter.bottom += g_Config.iWindowY - rcOuter.top;
-		rcOuter.left = g_Config.iWindowX;
-		rcOuter.top = g_Config.iWindowY;
+	static void GetWindowSizeAtResolution(int xres, int yres, int *windowWidth, int *windowHeight) {
+		RECT rc{};
+		rc.right = xres;
+		rc.bottom = yres;
+		AdjustWindowRect(&rc, WS_OVERLAPPEDWINDOW, TRUE);
+		*windowWidth = rc.right - rc.left;
+		*windowHeight = rc.bottom - rc.top;
 	}
 
 	void SetWindowSize(int zoom) {
 		AssertCurrentThreadName("Main");
-		RECT rc, rcOuter;
-
 		// Actually, auto mode should be more granular...
+		int width, height;
 		if (g_Config.IsPortrait()) {
-			GetWindowRectAtResolution(272 * (int)zoom, 480 * (int)zoom, rc, rcOuter);
+			GetWindowSizeAtResolution(272 * (int)zoom, 480 * (int)zoom, &width, &height);
 		} else {
-			GetWindowRectAtResolution(480 * (int)zoom, 272 * (int)zoom, rc, rcOuter);
+			GetWindowSizeAtResolution(480 * (int)zoom, 272 * (int)zoom, &width, &height);
 		}
-		MoveWindow(hwndMain, rcOuter.left, rcOuter.top, rcOuter.right - rcOuter.left, rcOuter.bottom - rcOuter.top, TRUE);
+		g_Config.iWindowWidth = width;
+		g_Config.iWindowHeight = height;
+		MoveWindow(hwndMain, g_Config.iWindowX, g_Config.iWindowY, width, height, TRUE);
 	}
 
 	void SetInternalResolution(int res) {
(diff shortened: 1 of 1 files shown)


sum2012:
Ok,I will compile when back home

Windows/MainWindow.cpp
```diff
@@ -365,63 +360,58 @@ namespace MainWindow
...
-			// Center the window.
-			const int primaryScreenWidth = GetSystemMetrics(SM_CXSCREEN);
-			const int primaryScreenHeight = GetSystemMetrics(SM_CYSCREEN);
-			g_Config.iWindowX = (primaryScreenWidth - g_Config.iWindowWidth) / 2;
-			g_Config.iWindowY = (primaryScreenHeight - g_Config.iWindowHeight) / 2;
-			rc.left = g_Config.iWindowX;
-			rc.top = g_Config.iWindowY;
-			rc.right = rc.left + g_Config.iWindowWidth;
-			rc.bottom = rc.top + g_Config.iWindowHeight;
+		// Then center if necessary. One dimension at a time.
```
unknownbrackets: It's awfully annoying when apps assume that because you have dual displays, you want windows to open up centered half way on each.  Many people will have a larger virtual desktop with a TV, latop and monitor, or just two monitors.

Can this not use SM_CXVIRTUALSCREEN for centering?

-[Unknown]

unknownbrackets (changes requested):

Windows/MainWindow.cpp
hrydgard: This code is for when starting the first time - though I guess the visibility check can trigger it later too... Is it currently being annoying for you? I'm not sure the visibility check is a good idea anyway.

Should really switch to the newer fully multimonitor capable APIs (with DPI per monitor etc) instead, but then we have to do dynamic loading to avoid breaking XP...

unknownbrackets: > Though centering in the middle of a virtual screen is not always great either, you might end up straddling a screen edge.

More than "not always great", I think it is "always bad"?  PPSSPP correctly does _not_ straddle before this pull request.

I know plenty of people who have dual monitor setups.  Any of them could start PPSSPP for the first time.  I'm not sure we'd want the default startup position of the window to straddle to physical monitors for anyone.

It's even possible to configure (mostly seen with monitor + TV setups):

```
+----------+\\\\\\\\\\\
|          |\\\\\\\\\\\
|          |\\\\\\\\\\\
|          |\\\\\\\\\\\
|          |\\\\\\\\\\\
+----------+----------+
\\\\\\\\\\\|          |
\\\\\\\\\\\|          |
\\\\\\\\\\\|          |
\\\\\\\\\\\|          |
\\\\\\\\\\\+----------+
```

Which would (centering based on full virtual desktop size) result in PPSSPP defaulting like this:

```
+----------+\\\\\\\\\\\
|          |\\\\\\\\\\\
|          |\\\\\\\\\\\
|       ****\\\\\\\\\\\
|       *PPS\\\\\\\\\\\
+-------*PPSSPP*------+
\\\\\\\\\\\SSPP*      |
\\\\\\\\\\\*****      |
\\\\\\\\\\\|          |
\\\\\\\\\\\|          |
\\\\\\\\\\\+----------+
```

This is the reason why most applications and games just center based on the primary monitor (SM_CXSCREEN / SM_CYSCREEN) by default (it'd be e.g. in the top left monitor's center.)  Nothing stops the user from dragging it in the middle of two monitors if that's *really* what they want, afterward.

What's the motivation for using SM_CXVIRTUALSCREEN / SM_CYVIRTUALSCREEN for the default, as this pull request does?

I think the reason for the out of bounds detection was related to some issue with resizing resolution or changing monitors or something.  I don't remember well.  I suppose the issue is that PPSSPP could've been previously configured in a position that is no longer visible.

-[Unknown]

sum2012: Do I need to tell the reporter to test that which commmit\pull request break ?

hrydgard: So I completely missed that both CXSCREEN and CXVIRTUALSCREEN were in used before. I though it just fetched the same value twice for no reason.

I'll redo this.

sum2012:
The tester said that no fixed in https://github.com/hrydgard/ppsspp/pull/9568/commits/2d8910549e71a1a4891864055733c77f98d275aa

Windows/MainWindow.cpp
```diff
@@ -365,63 +360,60 @@ namespace MainWindow
...
+				((g_Config.iWindowX + g_Config.iWindowWidth) < (virtualScreenWidth + g_Config.iWindowWidth));
 
-			if (!visibleHorizontally)
-				g_Config.iWindowX = -1;
+			bool visibleVertically = ((g_Config.iWindowY + g_Config.iWindowHeight) >= virtualScreenY) &&
+				((g_Config.iWindowY + g_Config.iWindowHeight) < (virtualScreenHeight + g_Config.iWindowHeight));
 
+			if (visibleHorizontally)
+				resetPositionX = false;
 			if (!visibleVertically)
```
unknownbrackets: Should this be reversed - if it's visibleHorizontally reset X, visibileVertically, reset Y?  It seems like this resets the Y when it IS visible?

-[Unknown]

Windows/MainWindow.cpp
unknownbrackets: Thanks.

unknownbrackets (commented):
Even if it doesn't fix that issue, I think this is cleaner.

-[Unknown]

Windows/MainWindow.cpp
hrydgard: Oops.

hrydgard:
I'm thinking of also clamping the Y position to 0, or at least if it's less than a few pixels negative, so that the title bar never becomes invisible, making the window hard to move - what do you think?

unknownbrackets:
Probably makes sense, in practice it should be hard to move off screen that way (unless using the keyboard) anyway, and I'm sure I've seen apps/games that snap the title bar back onto screen.

-[Unknown]

hrydgard:
I was thinking as a workaround for whatever strange issue sum's friend is having, that should fix it..

hrydgard:
Two more possible fixes/workarounds added...

sum2012:
The reporter test on https://github.com/hrydgard/ppsspp/pull/9568/commits/0b7da0b6a40d0cc3aebeaa992af93a6d349fad8b
The size of windows a bit more than "x2" setting
if you have idea,you can fix it before merge.
Otherwise,the reporter also accept this fix

sum2012:
This is reporter 's screen of modify version

hrydgard:
Oh, so the window becomes larger than the screen. We should probably just maximize in that case, but that's for a separate pull later, maybe. I'm merging.

sum2012:
No,the reporter want x2 size
> We should probably just maximize in that case

PR #9588, opened 2017-04-10
Linux: Added support for Hidromancer gamepad device driver

ghost:
Hidromancer DS3/Multi-device driver support.  Tested and working.

It is crazy, however, that ppsspp will not recognize any device that is not on this list, meaning that all devices/device drivers must register on this whitelist in order to be used with ppsspp.

Please merge and I'll post an issue about this in a bit.


hrydgard:
I agree that it's not great, and we should go with some default mappings for devices not in the list.

PR #9629, opened 2017-04-22
Add Simple highlight in Savedata Manager

zminhquanz:
I implement about this , but i don't know how to implement pic1 in save data direction to show on background
Like this on real PSP
snap000


xiushudongfang:
Maybe #9501 is a good reference.

zminhquanz:
@xiushudongfang i didn't mention about that , i want to implement savedata manager to show in background like picture on above , not choose another picture custom background

xiushudongfang:
I mean that it is a reference.

zminhquanz:
I closed this because i need to more time to read this code

PR #9640, opened 2017-04-24
Fix postprocess shader submenu.

### LINKED ISSUE #9639: Postprocess shaders list no longer accessible from windows menu
Seems caused by #9558, if it's too slow I guess it wouldn't be that bad to just remove that list, it still works fine in the nicer ui.

This https://github.com/hrydgard/ppsspp/blob/1a9e93d4974a35e99088f96c6441d7c1dc9fea0e/Windows/MainWindowMenu.cpp#L177 - seems to be taken initially disabling the creation of submenu which then can't be accessed and ends up completely disabled.

LunaMoo:
Not sure what was the meaning of the original code, I saw that CreateShadersSubmenu was called twice, but since the first time menuShaderInfo was 0, the submenu was never created and menuShaderInfo  never set to shader info and it just generally failed.

 I think that check was ment to skip re-creating when it's same, so I changed it that way.

 Fixes #9639

### DIFF
--- a/Windows/MainWindowMenu.cpp
+++ b/Windows/MainWindowMenu.cpp
@@ -174,7 +174,7 @@ namespace MainWindow {
 		}
 		std::vector<ShaderInfo> info = GetAllPostShaderInfo();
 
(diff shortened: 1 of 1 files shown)


hrydgard:
Looks like a typo, yes.

Issue #8196, opened 2015-11-09
Mouse buttons not read for input


GcHS:
Mouse buttons 1(left click),3(middle click),4(Xbutton1),and 5(Xbutton2) are not read for input by PPSSPP. In addition, though the input configurator reads scroll up and scroll down, neither seems to actually produce the desired input whilst in-game.

ppmeis:
As they decribed on forum: http://forums.ppsspp.org/showthread.php?tid=6626

Mouse buttons could not be used on controls because emulator use them for navigating under menus.

benderscruffy:
can this be closed ???

hrydgard:
I think there might be possible improvements here, though there's little demand for playing PSP games by mouse anyway... I'll leave open.

RADIOACTIVEshrek:
I feel like this feature is a must, especially for games like Coded Arms, which are now getting mouse support via a mouse injector.

PR #9644, opened 2017-04-26
Mouse Control(Windows only for now)

LunaMoo:
No idea how to do that for other desktop platforms(RawInput code?), but seems to work decently on windows.

 Uses workaround for mapping mouse(axis as well as left button) which is in form of checkbox that get's added when mouse control is activated, after selecting it, user can map whatever with a mouse once, then has to activate it again, not pretty but works, maybe this can be done better, I just had no ideas how to solve that nicely.:|

 Outside of mapping workaround, mouse is only enabled in-game, and using it also disables double click fullscreen switch, cursor is hidden in-game as well. Tested mostly on MGS with my Right Analog Aim patch and it worked quite well with default settings, but also this add's two sliders for sensitivity and another for smoothing.

 PS for that first line of UI/GameSettingsScreen.cpp this removes BOM, not add's it;p.


hrydgard:
Seems nice. My only concern is how to make users understand the limitations, and not start complaining about how mouse won't work in X game which has a pointer controllable with the stick, or in Y FPS shooter...

MarcoEstevez:
Maybe special bindings section under settings, actually keyboard and joystick works together, why not a third one activatable by a flag with a nice warning: "Might not work properly on every game"

LunaMoo:
Mouse can be mapped to any psp keys, so left stick, right stick, d-pad or face keys all would work without any game patches. But yeah when it does work it does not work exactly like a native mouse support in pc games and less understanding people might complain, customizable per-game patches adding better analog support help, but aren't required.:]

 With recent forum threads about it([here](http://forums.ppsspp.org/showthread.php?tid=22881) and [here](http://forums.ppsspp.org/showthread.php?tid=22857) seems that people who actually want to use mouse might not care as much how well it resembles true mouse control as long as it works ~ kind of a reason why I'm trying to push that feature from the closet.


 I see one potentially bothersome problem and not sure how to deal with it ~ in windowed mode mouse cursor can go outside of the window - assuming most people will just play games in fullscreen it might be desired to leave window with mouse at will, however I don't have a multi-monitor setup to test and I think this could be a problem on those.

 BTW is the missing break; in [here](https://github.com/hrydgard/ppsspp/blob/e17772b58d647ed73fc7c177cdc4feaecb938e76/Windows/RawInput.cpp#L121) intentional(to avoid mapping syskey used for windows menu)? Asking because there was some thread with someone complaining about being unable to map ALT key;p.

LunaMoo:
I guess I'm happy how it works now, seems user friendly enough and good enough to make mouse usable even in window mode.

 Partially(windows only) covers #4650 and #8196, hopefully didn't made too much mess in the code. (*hides* >.>)

hrydgard:
Huh, pretty sure that missing "break;" is a mistake :P  Feel free to correct it.

I think the code looks alright, hopefully it won't break anything else :)

PR #9651, opened 2017-04-29
Minor OCD fix;].

LunaMoo:
Just a bunch of things I noticed after my last contribution which started bugging me, like inconsistent upper/lower case ~ sorry;p. If anyone has a request for more minor OCD cleaning that fits here I will gladly add.

 Also changed a bit one CWC comment which was confusing to people who can't read what the code there does(to be fair that code type even with documentation would be very confusing ~ since it's only documented on some japanese site and googre tlansrate from japanese sucks:]).

### DIFF
--- a/UI/GameSettingsScreen.cpp
+++ b/UI/GameSettingsScreen.cpp
@@ -576,7 +576,7 @@ void GameSettingsScreen::CreateViews() {
 		return UI::EVENT_CONTINUE;
 	});
 #if defined(USING_WIN_UI)
-	controlsSettings->Add(new ItemHeader(co->T("Mouse", "Mouse Control Settings")));
+	controlsSettings->Add(new ItemHeader(co->T("Mouse", "Mouse settings")));
 	CheckBox *mouseControl = controlsSettings->Add(new CheckBox(&g_Config.bMouseControl, co->T("Use Mouse Control")));
 	mouseControl->OnClick.Add([=](EventParams &e) {
 		if(g_Config.bMouseControl)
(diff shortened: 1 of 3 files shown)


UI/GameSettingsScreen.cpp
```diff
@@ -576,8 +576,8 @@ void GameSettingsScreen::CreateViews() {
 		return UI::EVENT_CONTINUE;
 	});
 #if defined(USING_WIN_UI)
-	controlsSettings->Add(new ItemHeader(co->T("Mouse", "Mouse Control Settings")));
-	CheckBox *mouseControl = controlsSettings->Add(new CheckBox(&g_Config.bMouseControl, co->T("Use Mouse Control")));
+	controlsSettings->Add(new ItemHeader(co->T("Mouse", "Mouse settings")));
+	CheckBox *mouseControl = controlsSettings->Add(new CheckBox(&g_Config.bMouseControl, co->T("Use mouse control")));
```
hrydgard: This changes the translation key so you'll have to update all the inis. Instead you can make the same tweak in en_US.ini.  Yeah, this is annoying..

hrydgard (changes requested):

UI/GameSettingsScreen.cpp
LunaMoo: Were there any translations added? This early it seems fine, but I can do it via ini I guess.

hrydgard: Oh right, this was the new stuff, sorry :)

PR #9654, opened 2017-04-29
Fix "copy type" pointer cwc

LunaMoo:
Blah I honestly have no idea if that's how this should work because the only documentation of those pointer codes seems to be in:
http://wikiwiki.jp/cwcheat/?CODE%A4%CE%B8%FA%B2%CC
;]
had hard time to even find which of those was broken with people who use those not being very informative, ended up just making the code work with some existing cheats.
 At worst it's now broken differently I guess, worked for a bunch of stuff through.>.>


hrydgard:
I'll trust your judgment on the cheat code stuff, I really have no idea :)

PR #9659, opened 2017-05-01
Fix: in some extreme case on-screen touch controls get stuck on iOS

xsm1997:
In some extreme case, such as backing to home screen when pressing touch controls, touching screen with whole palm, the on-screen touch controls on iOS may get stuck and become fully unresponsive.
In this case actually, iOS send a "touchesCancelled" message to views, which indicates that the touches had been abnormally cancelled.
Fixed absence of processing the message. After some simple tests, it will not get stuck again.

### DIFF
--- a/ios/ViewController.mm
+++ b/ios/ViewController.mm
@@ -344,6 +344,17 @@ static GraphicsContext *graphicsContext;
 	}
 }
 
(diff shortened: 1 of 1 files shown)


hrydgard:
Nice catch!

PR #9664, opened 2017-05-04
Handle alt+tab and menu better when mouse is confined in-game.

LunaMoo:
~~A bit workaroundish, don't know how to detect leaving drop down menu any better and I wanted to enable cursor when it's opened in-game, so set it to be trapped again on touch/mouse click in-game.~~ Found a better way with WM_EXITMENULOOP.

 Also handles alt+tab when the game is running and mouse is confined, it would still confine it, since somehow I forgot about such possibility>.>,  it also covers opening other PPSSPP windows like disassembly etc. or any other loss of focus ie. by other app.


hrydgard:
Looks fine.

PR #9697, opened 2017-05-17
Making ppsspp bootable from ARC Browser

kobol123:
Making ppsspp capable to boot a game directly from ARC Browser Catalog

### DIFF
--- a/Core/Config.cpp
+++ b/Core/Config.cpp
@@ -377,9 +377,11 @@ static ConfigSetting generalSettings[] = {
 	ConfigSetting("LastRemoteISOPort", &g_Config.iLastRemoteISOPort, 0),
 	ConfigSetting("RemoteISOManualConfig", &g_Config.bRemoteISOManual, false),
 	ConfigSetting("RemoteISOSubdir", &g_Config.sRemoteISOSubdir, "/"),
+	
 
 #ifdef __ANDROID__
 	ConfigSetting("ScreenRotation", &g_Config.iScreenRotation, 1),
+	ConfigSetting("ArcPath", &g_Config.arcpath, ""),
 #endif
 	ConfigSetting("InternalScreenRotation", &g_Config.iInternalScreenRotation, 1),
 
--- a/Core/Config.h
+++ b/Core/Config.h
@@ -232,7 +232,13 @@ public:
 	// UI
 	bool bShowDebuggerOnLoad;
 	int iShowFPSCounter;
-
+	
+	// Boot Arc Path
+	std::string arcpath;
+	
+	
+	
+	
 	// TODO: Maybe move to a separate theme system.
 	uint32_t uItemStyleFg;
 	uint32_t uItemStyleBg;
--- a/UI/MainScreen.cpp
+++ b/UI/MainScreen.cpp
@@ -917,6 +917,13 @@ UI::EventReturn MainScreen::OnDownloadUpgrade(UI::EventParams &e) {
 	} else {
 		LaunchBrowser("market://details?id=org.ppsspp.ppsspp");
 	}
+	
+	// Boot From Arc then clean path
+	if (g_Config.arcpath != ""){
+	screenManager()->switchScreen(new EmuScreen((g_Config.arcpath).c_str()));
+	g_Config.arcpath = "";
+	
+	}
 #else
 	// Go directly to ppsspp.org and let the user sort it out
 	LaunchBrowser("http://www.ppsspp.org/downloads.html");
--- a/android/jni/app-android.cpp
+++ b/android/jni/app-android.cpp
@@ -617,13 +617,13 @@ extern "C" void Java_org_ppsspp_ppsspp_NativeApp_resume(JNIEnv *, jclass) {
 
 extern "C" void Java_org_ppsspp_ppsspp_NativeApp_pause(JNIEnv *, jclass) {
 	ILOG("NativeApp.pause() - pausing audio");
-	AndroidAudio_Resume();
+	AndroidAudio_Pause();
 }
 
 ///Boot From ARC path
 
 extern "C" void Java_org_ppsspp_ppsspp_NativeApp_bootArc(JNIEnv env*, jclass, jstring path) {
-	ILOG("NativeApp.resume() - Booting from Arc");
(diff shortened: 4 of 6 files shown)


unknownbrackets:
Can you go into more detail about why it's necessary to do more here?  Does it send multiple intents?

Using g_Config isn't the right way, and I don't see the need for a new bootArc function.  That's why I'm trying to understand the problem you had that caused you make these changes.

-[Unknown]

kobol123:
ARC Browser is an app that is used as a Catalog of games that will boot directly from an activity inside ARC Browser the emulators so is why is there a new special intent for ARC to boot ppsspp games.

hrydgard:
But we already support starting games through intents. You're not adding a listener for some kind of special event here.

kobol123:
Only send the path to boot the games. If i send data from ARC activity with path to PpssppActivity it will boot the games directly from that intent?

kobol123:
In init i can see the shortcutparam that is the path but i can not see a boot command from that shortcutparam...

kobol123:
If you have a listener to the content of path i suggest boot from that command instead in MainScreen Class because is when everything is instantiated.

hrydgard:
Should all already be handled by this, assuming that we pass the parameter in from Java correctly:

https://github.com/hrydgard/ppsspp/blob/master/android/jni/app-android.cpp#L568

kobol123:
Ok but is not handling that well. It need to Instantiate everything first. Don't you think that? Because may not handle all objects in the app if for instance is called the boot before an object used by the app is created...

hrydgard:
So you still haven't explained what problem you are seeing. When you start a game in PPSSPP from ARC today, what happens? does it just start up to the normal start screen instead of booting a game?

hrydgard:
So I just installed ARC File Manager, clicked a CSO file, it asked me which app to open in, I chose PPSSPP and it booted straight into the game with no issues. Though funnily, it does so within the context of ARC so when you try to switch app it looks like it's ARC running the game.

Anyway, I don't see what's being improved here.

kobol123:
Comment from my facebook i leave you with that from the developer of ARC. I feel like a monkey here. "PPSSPP (free version) does work according to some users, but it seems a bit unstable. It hasn't always worked when I tried it." . Sorry if is all ok i will not participate on this sorry Henrik and we already talked about it if you remember.

hrydgard:
Well, to merge this I would need some kind of evidence or explanation for how it improves things, and I haven't gotten that yet. So we'll leave this alone until more information is available.

If the developer of ARC knows something we're doing wrong, point them here.

kobol123:
You only want things instead of give. That is your philosophy. Is proved that is not working well and i have proves that not everything is working well, but i will not reveal... i know what you want and what you do. When you pointing to others things that don't exist and the others have obligations to prove. That is not my philosophy. You also didn't prove me that you are 100% right, yet, so i will keep go on...

kobol123:
What prove you want? That is not the best way to call the boot.Period.

hrydgard:
I'm sorry but I still don't understand what you mean. We support starting games through intents already, and it works - what is there to fix? How have you proved that the current method is not working well? As far as I can tell, it all works.

kobol123:
What I'm trying to say is that the way is booting now is not the best way. Because you are sending params to arg without you know when it will execute the boot. And if that command is executed before all app be load with all objects, maybe something will not went well. And ,(do you can specify when that boot happens in sequence?), sorry about trying to help...

kobol123:
If you show that command after UI been load, i will forget all and ask excuse, sorry...

kobol123:
In other hand screenmanager() is dependent of other classes... That shouldn't be load first? And not talking of others that depend of that. But if MainScreen class is load completely all adjacent will be guaranteed. Im i wrong?

kobol123:
When you tested in ARC did you test ppsspp FREE or the GOLD version?

hrydgard:
Both. There is an issue if you try to start it through ARC before giving the app permission to read from disk first, it hangs and that should be fixed, but your pull request won't do the trick.

I really, really can't see what's wrong with our current approach. See, it works like this:

In .PpssppActivity, we register an Intent filter, like this:

```             <intent-filter>
				<action android:name="android.intent.action.VIEW" />
				
				<category android:name="android.intent.category.DEFAULT" />
				<category android:name="android.intent.category.BROWSABLE" />
				
				<data
					android:host="*"
					android:mimeType="*/*"
					android:pathPattern=".*\\.iso"
					android:scheme="file" />
...

```

This will *start* a PpssppActivity when the intent happens. This derives from NativeActivity which is the whole app, really. It saves the parameter here:

```
		Intent intent = getIntent();
		String action = intent.getAction();
		if (Intent.ACTION_VIEW.equals(action)) {
			String path = intent.getData().getPath();
			super.setShortcutParam(path);
```
which does this:
```	public void setShortcutParam(String shortcutParam) {
		this.shortcutParam = ((shortcutParam == null) ? "" : shortcutParam);
	}
````

this.shortcutParam is used when we run the C++ code NativeInit:

```		NativeApp.init(model, deviceType, languageRegion, apkFilePath, dataDir, externalStorageDir, libraryDir, cacheDir, shortcutParam, Build.VERSION.SDK_INT, Build.BOARD);
```
Which in turn will look at that parameter and start the game.

Explain to me please why this is not enough. This seems to be exactly the right way to do it.

And that's why I'm pushing back, because your explanations don't make sense.

kobol123:
Yes i know that i did saw exactly that lines and functions in the code the only thing i couldn't see was the boot process in the app when is initiated i know when is executed and from where it comes. And sorry unknown and you are right when saying about the two intents it will execute two times from same intent. But what I'm tried to say is that when it executes the code from that function i included it will guarantee that nothing will be lose through the boot and with the function being initiated somewhere in the middle of process is not so secure thats why i asked to know exactly where is initiated the boot because i know where is executed that is exactly in same class MainScreen but you can have the class but not all adjacent subclasses load when you send "boot" param to be initiated.

kobol123:
You don't need to use ARC to test that action use instead : adb shell am start -n org.ppsspp.ppsspp/org.ppsspp.ppsspp.PpssppActivity -a android.intent.action.VIEW -d "file:///storage/path/to.iso"

PR #9709, opened 2017-05-21
Don't use partial-restart unless debugger is present.

hrydgard:
Works around #9666 for most practical purposes.


unknownbrackets (approved):
Still think the g_Config thing is a shame, but yeah.

-[Unknown]

PR #9726, opened 2017-05-26
Remove "Always Depth Write" setting. One step forward for #8171

hrydgard:
This option was only for working around a problem in Saint Seiya and Jeanne D'Arc on ES 2.0 devices where we can't blit depth buffers. Don't really think it's worth keeping, ES 2.0 devices are a diminishing part of the market and they are probably mostly not fast enough to run these games very well anyway. 

#8171


zminhquanz:
I think is not necessary at present , because on GLES 2.0 device , not powerful enough to emulate well , so i vote to remove it

unknownbrackets (approved):
It's a terrible workaround for Jeanne d'Arc anyway.

-[Unknown]

Issue #6779, opened 2014-08-25
Musou orochi 2 special title missing background - UNIMPL sceJpegDecodeMJpegSuccessively


daniel229:
JPCSP works fine
![01](https://cloud.githubusercontent.com/assets/3481559/4025949/5275da74-2c09-11e4-9d9e-530821999e59.png)
![02](https://cloud.githubusercontent.com/assets/3481559/4025950/52789a2a-2c09-11e4-8f2e-ce63186c52a1.png)

JPCSP
![03](https://cloud.githubusercontent.com/assets/3481559/4025955/7f68dd60-2c09-11e4-9df6-b87e87bca85d.png)

LunaMoo:
Would be nice if someone with the game checks it with the [recent PPSSPP](http://buildbot.orphis.net/ppsspp/), jpeg decoding should work fine now.

sum2012:
![1](https://cloud.githubusercontent.com/assets/2177532/26758331/b076df8c-490d-11e7-9011-0a3273a7b73d.jpg)
Thanks @LunaMoo It work now.
Closing this issue

Issue #1504, opened 2013-04-24
No image displayed on Official PlayView


rikimtasu:
Might be similar to other playview PSN iso+dlc. Logs here:
https://gist.github.com/anonymous/e86e9ae3610bab69c96b

PR #9750, opened 2017-05-31
Implement sceJpegDecodeMJpeg & sceJpegDecodeMJpegSuccessively

LunaMoo:
#1504 and #6779 might be fixed with this, but I have neither to test.

 I implemented this based on a weeb clock app "Bijin Tokei Portable" which is pretty much a collection of jpg's for every single minute in day:], works fine with this:
bijintokeiportable
Previously black background as mentioned [in the forums](http://forums.ppsspp.org/showthread.php?tid=18966).

Also changed some formatting and fixed the debug jpeg dump there.

Full list of games/apps from the report logs it should help:
```
TIGER & BUNNY オンエアジャック！
メルルのアトリエPlus~アーランドの錬金術士3~OfficialPlayView™
Best of Solitaire
美人時計ポータブル
無双OROCHI２ Special
```

### DIFF
--- a/Core/HLE/sceJpeg.cpp
+++ b/Core/HLE/sceJpeg.cpp
@@ -50,6 +50,10 @@ void __JpegDoState(PointerWrap &p) {
 	p.Do(mjpegHeight);
 }
 
+static int getWidthHeight(int width, int height) {
+	return (width << 16) | height;
+}
+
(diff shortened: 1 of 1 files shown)


hrydgard:
What a strange app :)   Code looks fine, and should be safe since there's no state.

PR #9769, opened 2017-06-02
Implement software skinning for the Vulkan backend. 

### LINKED ISSUE #9753: Software skinning not working properly in Vulkan
capture _2017-05-31-19-04-29

capture _2017-05-31-19-07-20

I'm using an LG G6, and on the latest build (1.4.2-81) with Vulkan Project Diva Extend and Project Diva 2nd display these weird models. This wasn't a problem when I was using 1.4.2-46. I tried turning off a 60fps cheat, and it didn't change anything.

The actual performance is not that much different from previous versions, and Vulkan seems to work a bit better than OpenGL, although on both backends my device seems to slow down drastically after a few minutes.

hrydgard:
Fixes #9753.

Also starts preparing for vertex caching.

This also opens the door to unify a bunch of code in DrawEngine between the backends.

### DIFF
--- a/GPU/Common/DrawEngineCommon.h
+++ b/GPU/Common/DrawEngineCommon.h
@@ -88,7 +88,7 @@ protected:
 	// Cached vertex decoders
 	u32 lastVType_ = -1;
 	std::unordered_map<u32, VertexDecoder *> decoderMap_;
-	VertexDecoder *dec_;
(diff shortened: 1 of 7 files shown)


iOS4all:
@hrydgard can you add metal as backend for iOS?
Now we have: OpenGL, DirectXD, Vulkan, and hopefully metal. To complete all difference types of backends.

PR #9781, opened 2017-06-05
Patch unusual way of checking for firmware version.

### LINKED ISSUE #8266: Anti-piracy/CFW measure prevents iDOLM@STER SP's DLC and Office Mode from being accessed.
uljs00169_00008
Like the title says, trying to access Office Mode in any of the iM@S SP games (Perfect Sun, Missing Moon and Wondering Star) that have the "Download Plus" update gives the error screen above, which translates to "Couldn't detect the system software. This game/software may run on system software that is not supported." (Translation was provided by a friend of mine).
After doing a bit of digging I discovered that the Download Plus update implements an anti-custom firmware mechanism that blocks CFW users from using Office Mode and DLC released after the update. (More info here: http://www.project-imas.com/wiki/THE_iDOLM@STER_SP_DLC)

EDIT: JPCSP is able to access the mode just fine, but like I say in the last comment, recent versions of the emulator can't load the DLC in both the base game and the updated version for some reason.

LunaMoo:
As left in the comment a few idolmaster games use this pauth function as a tricky way of checking for firmware version. Don't think it even matters what they're trying to decrypt as it's trying to decrypt tiny file on boot and decides whenever user has a good psp firmware or not:].

I think this change is safe, this would be 3rd know game to use pauth files, the other two would freeze/end in endless loop anyway even when we admit failing.

Fixes #8266

### DIFF
--- a/Core/HLE/scePauth.cpp
+++ b/Core/HLE/scePauth.cpp
@@ -71,7 +71,9 @@ static int scePauth_F7AA47F6(u32 srcPtr, int srcLength, u32 destLengthPtr, u32 w
 	fwrite(key, 1, 16, fp);
 	fclose(fp);
 
-	return -1;
+	// We failed decrypting and dumped encrypted files, some games like Idolmaster
+	// use this to check for firmware version, so let's still return no problem.
+	return 0;
 }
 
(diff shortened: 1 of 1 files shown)


hrydgard:
Ugly, but seems reasonable - shouldn't fail under normal circumstances.

PR #9828, opened 2017-07-03
WIP: Work towards NDK 15 support

hrydgard:
Trying to build with Android Studio 3.0 canary 5 together with the latest NDK, 15.1.

Having some really strange issues. The new headers seem very stingy about allowing stat64 so I removed it when building with support for old Android versions. Then I was getting crashes in the zip library so decided to skip it and implement support for Android's native asset API, which we should have done ages ago. That worked, but then I ran into mmap failing all over the place and I'm having more issues. This is when testing on a Galaxy S3. 

That's where I am right now. Will try on some different phones later but this isn't good...


hrydgard:
OK, the issue is that _FILE_OFFSET_BITS really isn't working. It also breaks with the older SDK but differently.

I'm going to split and rework this stuff.

PR #9857, opened 2017-07-21
Fix crash on FreeBSD x86_64 when starting a game

jbeich:
`MAP_32BIT` code appears to have regressed by #6694 but the fallback doesn't work on FreeBSD with defaults limits. So, I'm trying to get rid of [a downstream hack](https://github.com/freebsd/freebsd-ports/blob/064f9430ec/emulators/ppsspp/files/patch-Common_MemoryUtil.cpp).

https://lists.freebsd.org/pipermail/freebsd-stable/2011-December/065032.html

### DIFF
--- a/Common/MemoryUtil.cpp
+++ b/Common/MemoryUtil.cpp
@@ -157,7 +157,7 @@ void *AllocateExecutableMemory(size_t size) {
 	}
 #else
 	static char *map_hint = 0;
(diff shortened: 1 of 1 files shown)


hrydgard:
Hm, that's alright, I suppose. But I don't see how anything guarantees that our code pages will actually also stay below the 32-bit limit and thus reachable.. hm.

PR #9876, opened 2017-08-01
Android audio refactor

hrydgard:
Prepare for the future addition of AAudio support. In addition, this makes sure we no longer accidentally rely on auto-initialization of globals in any way. Might help #9771.


hrydgard:
Looks like the travis failures are just build server timeouts.

PR #9899, opened 2017-08-14
Vulkan state optimizations

hrydgard:
Now that we have dirty-flags for most state, make better use of them in the Vulkan backend to avoid submitting a lot of redundant commands.

This improves performance quite noticeably on Adreno at least.


DonelBueno:
A little off-topic, but I own a Xiaomi Mi 5, which sports an Adreno 530 (Snapdragon 820). 

It should support Vulkan, right? But the option for Vulkan on PPSSPP is grayed out.

Do I need to take an extra step to make it available?

hrydgard:
Grayed out? It should either be visible, or not. If it's not visible, it's not available (the GPU driver on the devices does not have Vulkan support).

DonelBueno:
Sorry, you're right, it's not visible.

But shouldn't it? Adreno 530 supports Vulkan AFAIK.

hrydgard:
Only if there's a Vulkan driver on the device - there's nothing we can do if there isn't one.

Issue #9825, opened 2017-07-02
Controller not working under Linux with SDL2 (and an incomplete and nasty hack to fix it)


marty1885:
Hello, I found that PPSSPP with SDL under Liunx don't response to my DS4 controller no matter what. Have anyone encounter this?
After digging into the source. I found that the events in `SDLJoystick.cpp` never happens and the controller seems not to be initialized in SDL. And I have come up with an incomplete and nasty hack to fix this. ([link](https://github.com/marty1885/ppsspp))(Good enough for playing Project DIVA) Which requires a controller to be connected to the PC when PPSSPP is launched.
Is my patch going at the right direction? What should I improve so I could contribute this for PPSSPP?

OS: Arch Linux
Kernel: 4.11 AMD64
Compiler: GCC 7.1.1
Controller: DualShock4 Gen2
Interface: USB

unknownbrackets:
This might be related the the "game controller db" not having your controller in it.  I think there's another issue for this, and specifically there was some discussion that things not found in the db should fall back to some sane handling - which apparently is not currently happening.

-[Unknown]

marty1885:
I'll have a look into that.
I have also found that the DPad on DS4 are handles as `hats` under SDL. Which PPSSPP currently does not support. How much discussion and code will be required to add a new input type? (I haven't look into PPSSPP's core yet. I have only working at the SDL part.)
Also, the problem of DS4 not working seems to start after I upgraded my kernel from 4.9 to 4.10 a while ago. Maybe the kernel changes how it handles DS4?

hrydgard:
You could just handle hat directions as digital buttons, and inject them the same way, I think. Will just have to define a few new keycodes for them. They should probably not be analog axes.

logos88:
@marty1885 kernel 4.10 included support for DS4 v.2 which didn't have until then (only v1)
https://wiki.gentoo.org/wiki/Sony_DualShock#cite_note-ds4v2-support-3

marty1885:
@logos88 I know. But It just works. maybe some patches are added in my kernel. Or it might be some non-mainline driver I have been testing and I accidentally installed them.

wodencafe:
Hi guys, I have this same issue.

Using an official PS3 controller, it works. But does not work when using a "Generic" version of the PS3 controller from Amazon. This "Generic" PS3-Like controller works for almost all of the other emulators and applications I've tried, such as dolphin-emu and desmume.

[...]

PR #9902, opened 2017-08-16
Add controller entries from gabomdq/SDL_GameControllerDB

hexacera:
Before using gabomdq/SDL_GameControllerDB as submodule, there's a few entries that exists in our current file but not in that repository. I have created a [pull request](https://github.com/gabomdq/SDL_GameControllerDB/pull/149) for that repo, and before that get merged, this could be a direct temporary fix to #9872 and #9825 and many more potential controller support issues.


hrydgard:
Thanks!

PR #9905, opened 2017-08-17
Implement vertex caching for Vulkan.

hrydgard:
Since vertex uploads are so efficient on Vulkan, the benefit is smaller than seen on some of the other backends (often less than 5%). So it's probably just barely worth adding this code...

Abuses a VulkanPushBuffer as a linear allocator, kept between frames, then completely wiped when full.


iOS4all:
@hrydgard please if you don't mind. Can you implement metal Api for iOS after finishing vulkan please?

PR #9927, opened 2017-08-23
Integrate sceUsbGps and sceUsbCam with Android

Florin9doi:


Core/HLE/sceUsbGps.cpp
```diff
@@ -15,24 +15,57 @@
...
 
-enum UsbStatus {
+enum GpsStatus {
 	GPS_STATE_OFF = 0,
 	GPS_STATE_ACTIVATING1 = 1,
 	GPS_STATE_ACTIVATING2 = 2,
 	GPS_STATE_ON = 3,
 };
 
+GpsStatus gpsStatus = GPS_STATE_OFF;
```
hrydgard: This is state, so if you want save states to work correctly this needs to be saved. Implement a DoState function like most other HLE modules do.

android/AndroidManifest.xml
```diff
@@ -14,13 +14,16 @@
...
+    <uses-feature android:name="android.hardware.gamepad" android:required="false" />
+    <uses-feature android:name="android.hardware.location.gps" android:required="false" />
+    <uses-feature android:name="android.hardware.location.network" android:required="false" />
 
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
     <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />
-    <uses-permission android:name="archos.permission.FULLSCREEN.FULL" />
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
     <uses-permission android:name="android.permission.INTERNET" />
+    <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
```
hrydgard: Requiring new permissions adds friction to the update process as people will be asked for it. Extremely few people are going to need this so I'm a bit dubious that it's worth supporting this on older devices where we can't do the runtime permission question...

hrydgard (commented):
Most of the code looks quite good, you've done a good job matching the style and so on. I only have some doubt about requiring a new Android permission for an extremely niche feature like this...

Florin9doi:
Here is a test program I've used; http://lukasz.dk/mirror/forums.ps2dev.org/viewtopica9f1.html?t=7222#49724

Core/HLE/sceUsbGps.cpp
Florin9doi: done

Florin9doi:
Updated with some Camera features: 
device-2017-08-27-125025

The camera images are correctly processed by **Invizimals** games, but they have other unrelated graphical problems.
**Eye Pet** don't starts at all.

Core/HLE/sceKernel.cpp
```diff
@@ -261,6 +263,7 @@ void __KernelDoState(PointerWrap &p)
 		__UmdDoState(p);
 		__UtilityDoState(p);
 		__UsbDoState(p);
+		__UsbGpsDoState(p);
```
hrydgard: New DoState functions should be added at the end of the list, for compatibility with old states. Actually come to think of it, that might not be an issue after all, but that's how we've always done it so please do so anyway in case I'm missing something...

Florin9doi: done

android/AndroidManifest.xml
```diff
@@ -14,13 +14,20 @@
...
+    <uses-feature android:name="android.hardware.camera" android:required="false" />
+    <uses-feature android:name="android.hardware.camera.autofocus" android:required="false" />
 
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
     <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />
-    <uses-permission android:name="archos.permission.FULLSCREEN.FULL" />
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
     <uses-permission android:name="android.permission.INTERNET" />
+    <uses-permission android:name="archos.permission.FULLSCREEN.FULL" />
+    <uses-permission-sdk-23 android:name="android.permission.ACCESS_COARSE_LOCATION" />
```
hrydgard: Oh! I didn't notice this little trick, had no idea you could do this. This alleviates all my concerns about adding permissions.

(https://developer.android.com/guide/topics/manifest/uses-permission-sdk-23-element.html for those living in ignorance like me)

hrydgard (approved):
Finally got around to reading the rest of the code. You've done a really good job!

Endarko:
I have not found the camera mode feature in android 6.0, ppsspp build a new one

PR #9938, opened 2017-08-29
Patches to make it possible to compile with mingw, pt. 1

inactive123:
This is the first of many commits where we try to make it possible to compile PPSSPP with Mingw in addition to the MSVC support that already exists.

### DIFF
--- a/Common/Common.h
+++ b/Common/Common.h
@@ -62,13 +62,6 @@ private:
 
 #elif defined(_WIN32)
 
-// Check MSC ver
-	#if !defined _MSC_VER || _MSC_VER <= 1000
-#ifndef __MINGW32__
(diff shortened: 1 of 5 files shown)


Common/FileUtil.h
```diff
@@ -26,12 +26,14 @@
 #include "Common.h"
 
 #ifdef _WIN32
```
hrydgard: This should probably simply be changed to #ifdef _MSC_VER to identify MS' compiler and libs.

Common/Common.h
```diff
@@ -64,7 +64,9 @@ class NonCopyable
 
 // Check MSC ver
```
hrydgard: Please just delete this entire check, it's quite silly.

PR #9950, opened 2017-08-31
Mingw patches (pt. 3)

inactive123:
After these changes, it should be possible to compile and link the libretro core with mingw. However, it crashes somewhere during initialization, so I still need to figure out where to go from here.


hrydgard:
Seems like a spurious buildbot error. Let's merge.

PR #9955, opened 2017-09-04
Add support for iPad Pro Smart Keyboard (or physical keyboards)

imxieyi:
It works fine without any latency.
Demo Video: https://youtu.be/F8D0Zu1xrJ0
However, the method of parsing iCade Controller can also read several keys on Smart Keyboard. But it can only read key press and then gets stuck. So I disabled the support for iCade.
Further work should be made to recover support for iCade.
**UPDATE:**
It seems that iCade and other physical keyboards can also be supported in this way. Since I don't have those accessories, I can't test them by myself.


hrydgard:
Seems fine, and useful.

PR #9966, opened 2017-09-13
Add uniform to make post process effects aware of videos.

LunaMoo:
And an example effect ~ subtle AA in-game/heavy smoothing for videos.

  I got the idea for doing this some time ago while replying to some thread [in here](https://forums.ppsspp.org/showthread.php?tid=23393&pid=128142#pid128142) about low res videos.
 My post there includes a bunch of example screenshots of the video part of the shader included here althrough the version here is free of any coloring.

 I think new uniform for that is nice to have since making an effect to look good in both, low res videos and high res 3D games is kind of impossible, with this people might get more creative, through with how blurry some videos tend to be, the results might vary.:]


hrydgard:
Yeah, this is a good idea.

PR #9985, opened 2017-09-28
Remove unnecessary ifdef that broke cursor hiding on SDL.

hrydgard:
Thanks Bracket-. See #9548 comments.


Bracket-:
Forgot the test at PCMain.cpp#L861

hrydgard:
Indeed, fixed.

Bracket-:
Thanks! 👍

PR #9991, opened 2017-10-01
gamecontrollerdb.txt: Add 8Bitdo NES30 Pro mappings for Linux

ghost:
Adds 8Bitdo NES30 Pro USB and bluetooth mappings for Linux

Minor things to note in case anyone with the controllers come across this:  
* Yes, '8Bitdo NES30 Pro   8Bitdo NES30 Pro' is what's reported for USB
* The power/home button is not available in USB mode so there's no guide button mapped
* Triggers support both axis and button (righttrigger b9, lefttrigger b8)
* Yes, trigger axis are swapped in USB and bluetooth modes. I didn't check button input  

I noticed PPSSPP doesn't map the triggers correctly whether I use axis or button input in gamecontrollerdb.txt, PPSSPP sets them to Pad1.bN which don't appear to exist. I can manually configure the triggers through PPSSPP and it sets them to Pad1.Touch Major+ and Pad1.Touch Minor+.


hrydgard:
Hm, odd about the defaults, should probably look into that at some point...

PR #9992, opened 2017-10-01
Reorder SDL init a bit, allowing us to read the fullscreen option early.

### LINKED ISSUE #9086: ppsspp sdl linux fails to enter full screen on startup
Though fullscreen option is enabled in config, ppsspp starts in windowed mode (starting the game does not change that). If after that I go to graphics options, I see "Full screen" checkbox checked. If I press that checkbox, ppsspp goes to full screen and checkbox stays checked.

hrydgard:
Fixes #9086


Bracket-:
Thanks! I can confirm it fixes issue for me.

Issue #9987, opened 2017-09-29
Regression: ppsspp does not recognize gamepad.


orbea:
## Description

With the SDL2 or Qt5 frontends PPSSPP does not recognize my gamepad (Horipad One using the xpad driver). It however does work with the PPSSPP libretro core and RetroArch.

### Bisect Results

I found this regressed in two places.

First bad commit for gamepad buttons.
```
db2cc49e0dd1b04d8b20ca4ccc52648e72f6b738 is the first bad commit
commit db2cc49e0dd1b04d8b20ca4ccc52648e72f6b738
Author: Neil Munday <neilmunday@users.noreply.github.com>
Date:   Sun Sep 18 23:23:36 2016 +0100

    Modified SDL2 code to make full use of SDL2's Game Controller API.

:040000 040000 927add97bc9c6db07039a7c7d0062553223fdc83 4eaae6da805457195278ae60be251a2ac8a660fe M	SDL
:040000 040000 164419420df51754566f5650953970dacadd21bb c8c7643f65832a16d4b23620e2ef78ff841e2652 M	ext
```
db2cc49e0dd1b04d8b20ca4ccc52648e72f6b738

First bad commit for analog sticks.
```
commit 67425d7d477d1e95dfa3347df607806dc214c873
Author: Neil Munday <neilmunday@users.noreply.github.com>
Date:   Sun Sep 18 00:33:28 2016 +0100

    Initial attempt at adding SDL's game controller database to PPSSPP.

:040000 040000 cc7ac605476a55810de8959b6b3abe0566509d80 927add97bc9c6db07039a7c7d0062553223fdc83 M	SDL
:040000 040000 d5819be803066a9199987095feb8623c1c8e31e1 18f17488d4d0ac57789eb40d0c324e3571607f62 M	assets
```
67425d7d477d1e95dfa3347df607806dc214c873

### Version/Commit
You can find this information under Information/System Information

`ppsspp-2017.09.29_1b6c5c4_master-x86_64-1_git`

### Environment information

- OS: `Slackware64-current`

orbea:
@neilmunday Can you take a look at this regression?

orbea:
I submitted a PR to add my gamepad config here. https://github.com/hrydgard/ppsspp/pull/9988

However if I understand this issue correctly if a gamepad is not included in `assets/gamecontrollerdb.txt` ppsspp will not allow the user to manually configure their gamepad as it used to before the above commits which still seems like a bug.

[...]

PR #9997, opened 2017-10-03
Create default mapping for unknown control pads in SDL's controller database

neilmunday:
Updated SDL/SDLJoystick.cpp in response to #9987 

Summary:

If a control pad is not found in SDL's game controller database then a default mapping (based on the Sony PlayStation 3 Dual Shock controller) will be created and added to the in-memory database. Note: this new mapping is temporary and is not written to the control pad database.

### DIFF
--- a/SDL/SDLJoystick.cpp
+++ b/SDL/SDLJoystick.cpp
@@ -50,22 +50,36 @@ void SDLJoystick::setUpControllers() {
 }
 
 void SDLJoystick::setUpController(int deviceIndex) {
-	if (SDL_IsGameController(deviceIndex)) {
-		SDL_GameController *controller = SDL_GameControllerOpen(deviceIndex);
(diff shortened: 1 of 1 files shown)


hrydgard:
Nice. The occasional wrong button mapping is better than it working at all for unknown joysticks. Only drawback I can see is that motivation for adding new known-good mappings might be smaller.. oh well.

carlsmith:
Button mappings are easy to fix. The analogue sticks are the problem. There are third party apps that will map control pad inputs to keyboard keys reliably, and that works perfectly with PPSSPP, but they can only map the analogue sticks to binary keys or buttons, like D-pads.

At the moment, I can only use PPSSPP with games that work with a D-pad, like Wipeout, even though I'm using a pad with analogue sticks.

PR #10001, opened 2017-10-04
Add ini-only "Savedata Upgrade" setting.

### LINKED ISSUE #10000: Errorcode 80110306 when trying to load a savegame in this game...
Hi.

## Environment Specifications:
PPSSPP version: 1.4.2.0 ( [this](https://ppsspp.org/files/1_4_2/ppsspp_win.zip) )
S.O: Windows 10 x64

## Affected game:
Title: **Gunhound EX**
Disc ID: **NPJH-50723**
ISO Download -cut-
(THE DOWNLOAD IS JUST FOR EDUCATIVE PURPOSES IF NEEDED BY THE DEVELOPERS OR HELPER USERS TO TEST OUT THE PROBLEM, if this violates the rules I'm sorry and then please delete the url)

## Affected savegame:
Savegame download: [here](https://www.gamefaqs.com/psp/681782-kisou-ryouhei-gunhound-ex/saves)
Savegame folder name: **NPJH5072300GHSAVE00**

## Problem Description:
The emulator displays this message when initializing the game due it can't load the savegame for some reason:

errorcode

After that, the game works as normally... but it can't load savegames.

## Steps to reproduce the problem:
1. Download the specified savegame linked above.
2. Run your copy of **Gunhound EX** in PPSSPP.
2. Wait for the errorcode to appear.

## Research:
I'm 100% sure this is not a problem of user rights neither a corrupted save data, it is an issue with the emulation. Also be aware that I have a private collection of many, many many games that runs perfectly in PPSSPP and the savegames too. Please avoid typical things like "_you have admin privileges?_" and that..

I had another similar problem with another japanese game ( **Strike Witches - Hakugin no Tsubasa** ) which It was unable to load a savegame and it gaves me an error code 80110306 or maybe 80110307, I'm not totally sure, but the thing is that I solved it by following this strange guide:

solution
Source: http://forums.afterdawn.com/threads/help-saved-game-give-error-code-80110306-and-also-80110307.580858/

Yes, I don't know how and why, but doing those weird things (which are not much good explained at all) I solved the errorcode in that other game and my savegame start loading correctly. Unfortunately that thing didn't worked for **Gunhound EX**.

Also I noticed that the **Gunhound EX** is not listed in the official PPSSPP compatibility list: 
http://forums.ppsspp.org/showthread.php?tid=1473
( However **Strike Witches - Hakugin no Tsubasa** is not listed there too, but it is perfectly playable game. )

It was unable for me to find any info about this strange error. I read that maybe it could be due to some games requires a higher firmware and then... it breaks when trying to load a savegame or something like that?. Or maybe just what happens is the game is not compatible with PPSSPP? maybe it never get tested by the developers...due it does not appear in the compatibility list?.

## Other:
I'm glad to contribute to PPSSPP's project being my issue the magic number #10000. Is there any reward for this, like... you know, a month of free Domino's Pizza or some coupon to buy a PS4?. ;)

Thanks for read!

LunaMoo:
Fixes #10000 

The option does similar thing as the workaround for old/buggy PPSSPP savedata, except deals with outdated PSP firmware(only cfw?) problem.
 Made it ini-only, non default, but not per game since if someone would be affected by it chances are he might have a lot of such saves for many games.

### DIFF
--- a/Core/Config.cpp
+++ b/Core/Config.cpp
@@ -754,6 +754,7 @@ static ConfigSetting systemParamSettings[] = {
 #endif
 	ConfigSetting("WlanPowerSave", &g_Config.bWlanPowerSave, (bool) PSP_SYSTEMPARAM_WLAN_POWERSAVE_OFF, true, true),
 	ReportedConfigSetting("EncryptSave", &g_Config.bEncryptSave, true, true, true),
+	ConfigSetting("SavedataUpgrade", &g_Config.bSavedataUpgrade, false, true, false),
 
 	ConfigSetting(false),
 };
--- a/Core/Config.h
+++ b/Core/Config.h
@@ -420,6 +420,7 @@ public:
 	int iButtonPreference;
 	int iLockParentalLevel;
 	bool bEncryptSave;
+	bool bSavedataUpgrade;
 
 	// Networking
 	bool bEnableWlan;
--- a/Core/Dialog/SavedataParam.cpp
+++ b/Core/Dialog/SavedataParam.cpp
@@ -647,6 +647,12 @@ void SavedataParam::LoadCryptedSave(SceUtilitySavedataParam *param, u8 *data, u8
 			} else {
 				WARN_LOG_REPORT(SCEUTILITY, "Savedata loading with detected hashmode %d instead of file's %d", decryptMode, prevCryptMode);
 			}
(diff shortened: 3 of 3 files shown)


hrydgard:
Yeah, I guess that's alright as ini-only.

PR #10022, opened 2017-10-18
Better path fix x86_64 android

mrcmunir:
Not sure why not needed before when updated my toolchain to NDK15 now  no found -landroid  and error: cannot open crtbegin_dynamic.o: No such file or directory 

So we pull it safe in case not detect sysroot properly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1617,7 +1617,7 @@ add_library(${CoreLibName} ${CoreLinkType}
 if(ANDROID)
   set(CoreExtraLibs ${CoreExtraLibs} android)
   if(X86_64)
-    set(CMAKE_SHARED_LINKER_FLAGS "-Wl,-Bsymbolic")
+    set(CoreExtraLibs ${CoreExtraLibs} android -Wl,-Bsymbolic")
   endif()
 endif()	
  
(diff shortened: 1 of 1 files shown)


mrcmunir:
Well after some days investigate  

With
```set(CoreExtraLibs -Wl,-Bsymbolic)```

```Error:(148) undefined reference to 'ANativeWindow_release'
Error:(1045) undefined reference to 'ANativeWindo´´´

So fix renference needed add android link too.

I.e 
```set(CoreExtraLibs android -Wl,-Bsymbolic)```

The ${CoreExtraLibs} is not mandatory necesary but ok. 

If you have this problem you should apply this pull request to linker correctly under x86_64 android.. sigh

hrydgard:
Hm. It's strange that this would work but not setting CMAKE_SHARED_LINKER_FLAGS . Not a big fan of passing in flags in a variable that's supposed to hold library names...

mrcmunir:
That's what the compiler complains about not detected some flags from android and Corenativelelibs.
It is rare that compile did not ask before Passing from NDk13 to NDK15.
I have also upgraded to NDk16 in case of any regression but the problem continues

I'll see what alternative there is without declaring these variables.

mrcmunir:
@hrydgard Done Sorry Apparently needed to separate -Bsymbolic to detect some previous flags.

hrydgard:
Oh, of course. Good one.

PR #10076, opened 2017-11-09
Upgrade glslang to a fresh version (used to compile Vulkan shaders).

hrydgard:
Might help #10065 ?


hrydgard:
Stupid buildbot not building for mac/ios... Not a code problem.

Issue #10077, opened 2017-11-09
Vulkan: Crash when switching apps on Android


hrydgard:
This happens because we tear down Vulkan and delete the context while keeping the DrawEngine around, which keeps a vulkan_ pointer around, which is now invalid...

So either we need to change it to not actually delete the vulkan_ object, or we need to update all the cached vulkan_ pointers everywhere. Or make a better way to get to the vulkan_ object.

Found multiple other bugs too of course.

PR #10079, opened 2017-11-10
Vulkan lost-device fixes

hrydgard:
With this, task switching on Android with Vulkan becomes considerably more stable.

It might be possible to keep around a lot of Vulkan objects through task switches, but I choose to tear everything down and rebuild it like we do with GL. Task switches are slightly sluggish but I feel this is the right way since it reduces memory consumption while we're in the background, and thus the risk to be kicked by the OS if other apps are being used.

Helps #10077


oli0371:
I am using s7 edge exynos with mali cpu and the latest version vulkan is not stable as previous version, when i enter setting when scrolling the ui glitch

hrydgard:
@oli0371 I've seen this too on S8, seems to be a platform/driver bug. It seems to help to change Display Resolution to 2x PSP in Graphics Settings.

PR #10092, opened 2017-11-11
Vulkan hardware tesselation fixes

hrydgard:
Should also be faster.

Should fix the crash reported in #10089 .


hrydgard:
This turns out not to be enough to fix hw tess on Vulkan on Mali at least, looking into it.

PR #10126, opened 2017-11-14
Fix AVIdump Memory Leak

### LINKED ISSUE #10125: Awful Memory Leak while dumping AVI.
Existing before new changes, it seems to leak every frame quickly filling ram, so something somewhere has to be cleared and it isn't, not sure if I can find it myself, so opening this issue.:]

Edit:
I noticed it's calling av_frame_free when closing file... shouldn't this happen for every frame?

LunaMoo:
Fixes #10125

:)


hrydgard:
Some problems are easy :)

PR #10143, opened 2017-11-15
Update .gitignore for libretro builds.

orbea:
This would be helpful for the libretro port. I am submitting it here because I rather change upstream files in the libretro repo as little as possible even if its just a `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,7 +62,6 @@ build*/
 # Temp file used by jenkins windows build (TODO: remove)
 desc.txt
 
-
 Logs
 Memstick
 memstick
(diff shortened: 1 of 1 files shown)


hrydgard:
Do you need them all in the root or could you put this .gitignore in some subdirectory?

orbea:
Most of them can be put in the `libretro` directory in the libretro repo, but that ends up missing many `.o` files in `Common`, `Core`, `GPU` and `ext`. I think the cleanest solution is putting them all in the root directory, but the choice is yours. :)

orbea:
If you prefer I could move most of them to the libretro repo except for `*.o`?

orbea:
After thinking about that, I think that might be preferable so I made that change.

PR #10154, opened 2017-11-18
fix missing NDEBUG macro ,change all /W3 to /W4 and /MT to /MD

peaceshi:
fix missing NDEBUG macro for release. because of it , change all /W3 to /W4 and /MT to /MD.


hrydgard:
Do not change /MT to /MD please.

hrydgard:
Got nothing against fixing NDEBUG though.

PR #10166, opened 2017-11-20
Update armips to latest. See #10161

hrydgard:


unknownbrackets (approved):
LGTM, jit tests and assembler feature still works.

-[Unknown]

PR #10168, opened 2017-11-21
Cleanup GPU vendor detection a bit, use it to avoid the issue mentioned in #9545

hrydgard:


LunaMoo:
Seems like you reverted that recent armips update by accident.:)

hrydgard:
Oops, the perils of using multiple computers. Fixed.

hrydgard:
@LunaMoo Please check if this does the right thing on you GPU - it should, though. 

Build errors seems to be the usual, apple stuff build server down..

LunaMoo:
It's good, compat hack doesn't activate.:)

PR #10219, opened 2017-12-01
Fix cwcheat import

LunaMoo:
Didn't notice I messed it up when I replaced some substr;c. Sorry.


hrydgard:
BTW when it's 2 chars or more, I don't mind the substrs as much. But whatever :)

LunaMoo:
I saw [Unknown] using it for 2 chars earlier, so thought it's still better:P.
Now I think in those two lines could as well just skip the check for first character completely, cheat.db files made for psp doesn't have any other lines and we only import from them. But yeah it's whatever as long as it works:].

PR #10225, opened 2017-12-01
Add hotkey to toggle audio and video dumping together in sync.

LunaMoo:
I think that's how most people would want to use such feature, much more comfortable than going through few settings screens.


hrydgard:
Nice, looks safe enough to squeeze into 1.5.

PR #10242, opened 2017-12-03
Use a VulkanDeviceAllocator for thin3d textures. Saves on allocations.

hrydgard:
~With this merged, just click around the UI for a few secs and you hit 	_assert_msg_(G3D, found, "Failed to find allocation to free! Double-freed?");~

~This can't be good. Wonder if the bug is in the VulkanDeviceAllocator or elsewhere...~

Never mind, was just a silly bug.


hrydgard:
Oh, turns out this is because we don't use the allocator properly when creating textures the old way...

hrydgard:
OK, that's it, now it works. Oops :)

hrydgard:
Last one is an interesting one though.. Please double-check but I think it's right.

hrydgard:
Actually nevermind, misunderstood.

unknownbrackets (approved):

PR #10253, opened 2017-12-04
RIFF reader cleanup and sanity checks

hrydgard:
We only use this class for reading the menu music from .AT3 files embedded in PBPs, so remove a bunch of irrelevant stuff and clean it up.  Then add a few checks. Hopefully will fix a crash seen in #10202


unknownbrackets (approved):
Looks reasonable to me, glad to see it more readable.

-[Unknown]

PR #10393, opened 2017-12-11
Vulkan: Texture upload through compute, experimental texture scaling too

hrydgard:
This performs plain RGBA8 texture uploads using a compute shader directly.

Mostly intended as a proof of concept for now, but we could easily replace the compute shader with something that does smart upscaling, or even do the unswizzling and format conversion in it, avoiding lots of CPU work. 

This one takes pushbuffer memory and writes to a storage image. This unfortunately requires that the destination format supports being a storage image, which on most devices limits us to RGBA8888.

An alternative approach would be to have the compute shader read from the pushbuffer and write to a device-local buffer, and then issue a vkCopyBufferToImage from the latter to the image, which should be cheap and would allow us to use this to upload 16-bit format textures as well.

EDIT: Added code for alternative approachs as well - though for upscaling, we probably always want to output RGBA8888.

### DIFF
--- a/Common/Vulkan/VulkanImage.cpp
+++ b/Common/Vulkan/VulkanImage.cpp
@@ -224,7 +224,6 @@ void VulkanTexture::Touch() {
 }
 
 VkImageView VulkanTexture::CreateViewForMip(int mip) {
-	// Create the view while we're at it.
 	VkImageViewCreateInfo view_info = { VK_STRUCTURE_TYPE_IMAGE_VIEW_CREATE_INFO };
 	view_info.image = image_;
 	view_info.viewType = VK_IMAGE_VIEW_TYPE_2D;
--- a/Common/Vulkan/VulkanImage.cpp
+++ b/Common/Vulkan/VulkanImage.cpp
@@ -121,9 +121,10 @@ bool VulkanTexture::CreateDirect(VkCommandBuffer cmd, VulkanDeviceAllocator *all
 	if (initialLayout != VK_IMAGE_LAYOUT_UNDEFINED && initialLayout != VK_IMAGE_LAYOUT_PREINITIALIZED) {
 		switch (initialLayout) {
 		case VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL:
+		case VK_IMAGE_LAYOUT_GENERAL:
 			TransitionImageLayout2(cmd, image_, 0, numMips, VK_IMAGE_ASPECT_COLOR_BIT,
-				VK_IMAGE_LAYOUT_UNDEFINED, VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL,
-				VK_PIPELINE_STAGE_TOP_OF_PIPE_BIT, VK_PIPELINE_STAGE_TRANSFER_BIT,
+				VK_IMAGE_LAYOUT_UNDEFINED, initialLayout,
+				VK_PIPELINE_STAGE_ALL_GRAPHICS_BIT, VK_PIPELINE_STAGE_TRANSFER_BIT,
 				0, VK_ACCESS_TRANSFER_WRITE_BIT);
 			break;
 		default:
@@ -208,10 +209,10 @@ void VulkanTexture::GenerateMip(VkCommandBuffer cmd, int mip) {
 		VK_ACCESS_TRANSFER_READ_BIT, VK_ACCESS_TRANSFER_WRITE_BIT);
 }
 
-void VulkanTexture::EndCreate(VkCommandBuffer cmd, bool vertexTexture) {
+void VulkanTexture::EndCreate(VkCommandBuffer cmd, bool vertexTexture, VkImageLayout layout) {
 	TransitionImageLayout2(cmd, image_, 0, numMips_,
 		VK_IMAGE_ASPECT_COLOR_BIT,
-		VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL, VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL,
+		layout, VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL,
 		VK_PIPELINE_STAGE_TRANSFER_BIT, vertexTexture ? VK_PIPELINE_STAGE_VERTEX_SHADER_BIT : VK_PIPELINE_STAGE_FRAGMENT_SHADER_BIT,
 		VK_ACCESS_TRANSFER_WRITE_BIT, VK_ACCESS_SHADER_READ_BIT);
 }
@@ -220,6 +221,27 @@ void VulkanTexture::Touch() {
 	if (allocator_ && mem_ != VK_NULL_HANDLE) {
 		allocator_->Touch(mem_, offset_);
 	}
+}
+
+VkImageView VulkanTexture::CreateViewForMip(int mip) {
+	// Create the view while we're at it.
+	VkImageViewCreateInfo view_info = { VK_STRUCTURE_TYPE_IMAGE_VIEW_CREATE_INFO };
+	view_info.image = image_;
+	view_info.viewType = VK_IMAGE_VIEW_TYPE_2D;
+	view_info.format = format_;
+	view_info.components.r = VK_COMPONENT_SWIZZLE_R;
+	view_info.components.g = VK_COMPONENT_SWIZZLE_G;
+	view_info.components.b = VK_COMPONENT_SWIZZLE_B;
+	view_info.components.a = VK_COMPONENT_SWIZZLE_A;
+	view_info.subresourceRange.aspectMask = VK_IMAGE_ASPECT_COLOR_BIT;
+	view_info.subresourceRange.baseMipLevel = mip;
+	view_info.subresourceRange.levelCount = 1;
+	view_info.subresourceRange.baseArrayLayer = 0;
+	view_info.subresourceRange.layerCount = 1;
+	VkImageView view;
+	VkResult res = vkCreateImageView(vulkan_->GetDevice(), &view_info, NULL, &view);
+	assert(res == VK_SUCCESS);
+	return view;
 }
 
 void VulkanTexture::Destroy() {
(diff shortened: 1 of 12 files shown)


hrydgard:
There are a bunch of other variants that might be interesting for texture upscaling. For example, could first copy the small texture into an image, then have a compute shader sample that image as a texture (in order to get free conversion to float and interpolation if desired) and generate an upscaled image as data in a buffer, then copy that into the final image. Or use stores for the latter. But I think pretty much anything will be way faster than doing it on the CPU.

unknownbrackets:
Cool, will check it out later.  I guess this means switching from xbrz though on Vulkan, right?  (I think the shader variant we have is xBR, not sure how important the difference is...)

-[Unknown]

hrydgard:
Well, there's no reason you couldn't port XBRz to run in a compute shader. All that template usage might need to be collapsed down to functions but other than that should be fine, though tedious work heh. Remember that in compute shaders you can write multiple pixels for every invocation if you like so wouldn't need to change the algorithm pretty much at all.

unknownbrackets:
Hmm, fair enough.  Now where's a GLSL++ compiler when you need one?

-[Unknown]

hrydgard:
Actually... http://alphanew.net/index.php?section=alphanew&site=overview&lang=eng&newsID=111

though development of that stuff seems to have stalled?

hrydgard:
And yeah, it would be by far easier to just use XBR. Does XBRz have a noticeable quality advantage?

I think the XBR shader we have could be optimized quite a bit as a compute shader by using shared memory, there's a lot of redundant access to neighbour pixels - but should still be wicked fast as-is.

LunaMoo:
I think primary differences would be:
 - different handling of some details, because xBR algorithms have many variants for those, no idea which xBRZ is using:P, 5xBR uses variant A I think, and 5xBRv2 has a few different ones included as options,
 - post process doesn't have access to 1:1 graphics so often ends up looking bad,
 - post process scales from native res to as big as the output res is when optimally it should be x5, hence it might look rough in limited resolution like 720p.

 I think none of that will matter that much when it will work per texture.:]

hrydgard:
Actually we do give the postprocess access to 1:1 graphics, don't we?

LunaMoo:
1:1 yes, but to whatever game outputs and many 2D games stretch textures sometimes even for UI, not to mention effects that include zooming in like some 2D explosions etc., so it's not as great as texture scaling.

hrydgard:
Yeah, that's true of course.

GPU/Vulkan/TextureCacheVulkan.cpp
```diff
@@ -595,7 +648,21 @@ void TextureCacheVulkan::BuildTexture(TexCacheEntry *const entry, bool replaceIm
 			break;
 		}
 
-		bool allocSuccess = image->CreateDirect(cmdInit, w * scaleFactor, h * scaleFactor, maxLevel + 1, actualFmt, VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL, VK_IMAGE_USAGE_TRANSFER_DST_BIT | VK_IMAGE_USAGE_SAMPLED_BIT, mapping);
+		VkImageLayout imageLayout = VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL;
+		VkImageUsageFlags usage = VK_IMAGE_USAGE_TRANSFER_DST_BIT | VK_IMAGE_USAGE_SAMPLED_BIT;
+		// If we want to use the GE debugger, we should add VK_IMAGE_USAGE_TRANSFER_SRC_BIT too...
```
unknownbrackets: We can call `host->GPUDebuggingActive()` to determine if it's open.

-[Unknown]

GPU/Vulkan/TextureCacheVulkan.cpp
```diff
@@ -657,19 +724,32 @@ void TextureCacheVulkan::BuildTexture(TexCacheEntry *const entry, bool replaceIm
...
+						vkCmdBindDescriptorSets(cmdInit, VK_PIPELINE_BIND_POINT_COMPUTE, upload_.GetPipelineLayout(), 0, 1, &descSet, 0, nullptr);
+						vkCmdPushConstants(cmdInit, upload_.GetPipelineLayout(), VK_SHADER_STAGE_COMPUTE_BIT, 0, sizeof(params), &params);
+						vkCmdDispatch(cmdInit, (mipWidth + 15) / 16, (mipHeight + 15) / 16, 1);
+						vulkan_->Delete().QueueDeleteImageView(view);
+					} else {
+						entry->vkTex->texture_->UploadMip(cmdInit, i, mipWidth, mipHeight, texBuf, bufferOffset, stride / bpp);
+					}
 				}
 				if (replacer_.Enabled()) {
 					replacer_.NotifyTextureDecoded(replacedInfo, data, stride, i, mipWidth, mipHeight);
```
unknownbrackets: This path possibly won't work with saving new textures, right?

-[Unknown]

hrydgard: Hm, no, especially not if we change it to do deswizzle and depal on the GPU .. though, we could of course read back the results of that and write that to disk.

GPU/Vulkan/TextureCacheVulkan.cpp
hrydgard: Though textures could be created before it was opened..

Common/Vulkan/VulkanImage.cpp
```diff
@@ -388,14 +389,35 @@ void VulkanTexture::GenerateMip(VkCommandBuffer cmd, int mip) {
...
 	TransitionImageLayout2(cmd, image, 0, 1,
 		VK_IMAGE_ASPECT_COLOR_BIT,
-		VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL, VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL,
+		layout, VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL,
 		VK_PIPELINE_STAGE_TRANSFER_BIT, vertexTexture ? VK_PIPELINE_STAGE_VERTEX_SHADER_BIT : VK_PIPELINE_STAGE_FRAGMENT_SHADER_BIT,
 		VK_ACCESS_TRANSFER_WRITE_BIT, VK_ACCESS_SHADER_READ_BIT);
 }
 
+VkImageView VulkanTexture::CreateViewForMip(int mip) {
+	// Create the view while we're at it.
```
hrydgard: Note to self: remove irrelevant comment

GPU/Vulkan/TextureCacheVulkan.cpp
unknownbrackets: Maybe we flush the texture cache when it opens?  Not ideal but it isn't completely terrible.

-[Unknown]

hrydgard: Sure. Or we just always set the flag here - I can't imagine there's much of a cost to it really.

hrydgard:
Rebased on master.

QWEmct:
Hi. Could you make it possible to use Doom 3 using PPSSPP ?
Calculating the correct shadow position in space with all interactions of any number of objects and light sources is unique in Doom 3.
He now works on the Android using the Voodoo Engine with big problems.
Also, you once said that you will be able to make vibration in the games for PPSSPP, will it be implemented ?
Thank you for your work.

hrydgard:
@QWEmct  Please stop posting nonsense. Doom 3 is not a PSP game, and the PSP has no vibration functionality so there's nothing there to emulate.

QWEmct:
I understand that your time is limited, but could you build a small project on this ?

screenshot_2018-03-25-22-33-25-918_com n0n3m4 diii4a

GPU/Vulkan/TextureCacheVulkan.cpp
```diff
@@ -68,6 +68,84 @@ static const VkComponentMapping VULKAN_1555_SWIZZLE = { VK_COMPONENT_SWIZZLE_B,
...
+layout(push_constant) uniform Params {
+	int width;
+	int height;
+} params;
+
+void main() {
+	uint x = gl_GlobalInvocationID.x;
+	uint y = gl_GlobalInvocationID.y;
+	// Kill off any out-of-image threads to avoid stray writes.
+	// Should only happen on the tiniest mipmaps as PSP textures are power-of-2,
```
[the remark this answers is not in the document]
hrydgard: It should be completely negligible in all cases.

hrydgard:
Rebased it on master.

unknownbrackets:
As a test, I created:
https://github.com/unknownbrackets/ppsspp/tree/compute-upload

It uses a shader from #10818.  However, that shader incorrectly handles alpha: it changes any pixel with 0 alpha to black, which won't work for many PSP game usages.

-[Unknown]

hrydgard:
Yup, that's the direction I was planning to take this :)  Would probably also not be too hard to modify the shader to do something half-sensible for alpha.

GPU/Vulkan/TextureCacheVulkan.cpp
unknownbrackets: At least for now, it seems like the debugger still works.

-[Unknown]

hrydgard: It does work but with undefined behavior - so results will vary depending on GPU. In practice most layout transitions are no-ops on NVIDIA, some are on AMD and Intel too.

unknownbrackets:
Okay, force pushed then.  Added a check so if the ini setting is disabled, we don't even try to compile the shaders.

I'd propose we merge this (it's disabled by default) once we start merging 1.10.x things.  I'm thinking the next step is to probably make it able to read some ini for the texture shader, like with post shaders.  Then people can try out different texture upscaling options more easily.

When we get it to a good place and make sure we detect if unsupported or anything, we could make it a more visible option.

-[Unknown]

hrydgard:
Yeah, agreed.

Only issue I see is that people will start complaining that they need this for the other backends too. But I think we're going to have to start to go in the direction of leaving some backends behind.

unknownbrackets:
Reminder about this one (potentially could remove "do not merge" from the title.)

-[Unknown]

PR #10394, opened 2017-12-11
Small display layout editor improvements

LunaMoo:
This:
- corrects touch controlled zoom to max out at 10 to match it's slider,
- changes bounds to make it easier to use high zoom into corners(useful when testing subtle shader effects),
- stick display to edges,
- makes "center" respect that stickiness(saw something alike requested on forums some time ago).

### DIFF
--- a/UI/DisplayLayoutScreen.cpp
+++ b/UI/DisplayLayoutScreen.cpp
@@ -100,10 +100,11 @@ bool DisplayLayoutScreen::touch(const TouchInput &touch) {
 			int windowUpperEdge = local_dp_yres / 4;
 			int windowLowerEdge = windowUpperEdge * 3;
 			// And stick display when close to any edge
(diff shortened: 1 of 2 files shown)


hrydgard:
Not gonna review the math in detail, I trust you got it right :)

LunaMoo:
Yeah, this can be pretty confusing:], but I tested it, including rotation, so should be fine. Also limited smallest zoom to 1 as it really shouldn't go under and that's how slider was limited as well:].

LunaMoo:
Appveyor likes to randomly fail it seems:P.

I found the culprit of breaking both this and touch control layout screens on resize as https://github.com/hrydgard/ppsspp/commit/900539e200ac6fc5fd9935d4ede984e34f6d9600

Seems whatever fixed linux broke windows eh;p.

hrydgard:
Doesn't look like a real breakage to me, sigh. Sometimes buildbots are more annoying than useful :P

LunaMoo:
Yeah the breaking stuff I ment about resizing problem, going to open a new issue for that.

PR #10526, opened 2018-01-12
Update Inno installer : resolve #10367, resolve #10368, resolve #10370 

### LINKED ISSUE #10367: Windows installer: different 32/64-bit behaviour
What happens?

On Windows 64-bit, installer installs 32-bit version by default (folder **C://Program Files (x86)/PPSSPP/** instead of C://Program Files/PPSSPP/ and creates the shortcut to **PPSSPPWindows.exe** instead of PPSSPPWindows64.exe), so 32-bit version is using by default.

What *should* happen?

Could you make auto-check Windows version and different behaviour of the installer? Or even make two separate 32- and 64-bit installers?

Windows 10 64-bit, PPSSPP 1.5.4 (installer).

### LINKED ISSUE #10368: Windows installer: localization
What happens?

Although PPSSPP is translated to many languages, but its Windows installer is not, there's only English.

What *should* happen?

Is it possible to localize Windows installer?

Windows 10 64-bit, PPSSPP 1.5.4 (installer).

### LINKED ISSUE #10370: Windows installer: cosmetic fix
What happens?

CCleaner shows the name of the program as "PPSSPP version 1.5.4", but there is already version number in it's separate right column.

What *should* happen?

You can remove "version 1.5.4" from the name and set "PPSSPP" only.

Windows 10 64-bit, PPSSPP 1.5.4 (installer).

Florin9doi:
#10367, #10368, #10370

### DIFF
--- a/ppsspp.iss
+++ b/ppsspp.iss
@@ -1,21 +1,73 @@
+#define ApplicationName 'PPSSPP'
+#define ApplicationVersion GetFileVersion('PPSSPPWindows.exe')
+
+[Languages]
+Name: "en"; MessagesFile: "compiler:Default.isl"
+Name: "ar"; MessagesFile: "compiler:Languages\Armenian.islu"
+Name: "br"; MessagesFile: "compiler:Languages\BrazilianPortuguese.isl"
+Name: "ca"; MessagesFile: "compiler:Languages\Catalan.isl"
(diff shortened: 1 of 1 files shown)


hrydgard:
Nice, thanks!

In the future, use "fixes" instead of "resolve" and put it in the pull description, that will make github auto-close them :)

PR #10533, opened 2018-01-15
Allow up to 10x window size on Windows.

AkiraMiyakoda:
4x can be too small for today's high resolution monitors.


hrydgard:
Yeah, fair enough.

PR #10537, opened 2018-01-17
Minor refactoring: Make a place for shared Vulkan objects like simple linear samplers

hrydgard:
VulkanStockObjects, like our D3D11 stock objects.


hrydgard:
Actually, skipping this for now, not worth it. Thought I needed a sampler in yet another place but changed the approach.

PR #10636, opened 2018-02-22
iOS: fix hang on exit

KentuckyCompass:
This prevents PPSSPP from hanging after the user taps Exit, due to a bug I introduced in the previous pull request adding support for GLRenderManager.


hatemalhaneen (commented):
لتحميل

PR #10672, opened 2018-03-02
Android: Fix fallback from Vulkan to GL.

### LINKED ISSUE #10663: My PPSSPP Crash and close everytime I open the app.
My ppsspp suddenly won't open. It crashes everytime i open the app and going back to previous version the app only makes black screen. Work on 1.4.2. Haven't had this issue before until i updated it to the new version. 

My phone is oppo a71. And the version 1.5.4-605 gives me no lag and every game works perfectly on the 1.54-605 Until i updated to new version the app wont let me open. It doesn't go to the menu. On 1.4.2 it let me open but it gave me huge performance drop.

hrydgard:
Should fix #10663


hrydgard:
No longer kills off all the EGL code, let's keep it around for later.

PR #10718, opened 2018-03-13
Fix a typo extra " in some header

mrcmunir:
Fix a typo introduced in https://github.com/hrydgard/ppsspp/pull/10715


hrydgard:
Thanks :)

PR #10742, opened 2018-03-18
Map SDL controller's `back` button to Sony DualShock 4 V2's share button

crocket:
It turns out that `back` button is actually `select` button on PSP.

2018-03-18_sun_10 33 25

1920px-psp-1000

I confirmed this by clicking touchpad in a game.

`select` button maps to `share` button on Sony DualShock 4 V2.

Update 1. During internet search, I discovered that SDL game controller was modeled after xbox 360 controller.

xbox-360-controller-2

As you can see, BACK button is at the same position as Sony DualShock 4 V2's share button. Guide button is glowing in green.

Compare the preceding picture with Sony DualShock 4 V2.

dualshock 4


hrydgard:
Yeah, I guess that makes more sense. Some people with existing set ups might be confused by the change though...

PR #10757, opened 2018-03-20
Enable FlushInstructionCache on UWP, it's been allowed finally.

hrydgard:
Required for JIT on ARM. Works on my Lumia, though it isn't exactly fast.

However, things are still very shaky and I don't think I'm gonna spend a lot more time on this due to the very limited usage of this platform....


Toff-kun:
@hrydgard Thank you so much for this fix! JIT finally works on my 950 XL and it's actually much faster! I can reach 100% in most of my games now, so I'd say PPSSPP is finally worthwhile on W10M. 
The crackling audio is the biggest party pooper now, but I'm happy this worked out the way it did. :) I know my words can't change your plans, but I hope you don't give up on this platform just yet. 🤞 Thank you for all the work you've done and keep doing for this amazing emulator, Henrik!

hrydgard:
@Toff-kun Glad it works for you, on my Lumia it's still very slow but hey, nice :)

Issue #10766, opened 2018-03-22
[Feature Request] Less blurry Natural Colors PP shader


leopard20:
The Natural Colors Postprocessing shader is awesome, but it's a bit too blurry. Do you think you could tweak it a little to be better? The FXAA level of blurriness is just fine IMO.

hrydgard:
pull requests accepted :)

leopard20:
Would've done that if I knew how!

leopard20:
How can I compile the .apk from source code without Android Studio (like an online buildbot or something)?

PR #10774, opened 2018-03-23
Removed blur effect from Natural Color PP shader

Leopard20:
Fixes https://github.com/hrydgard/ppsspp/issues/10766.
Here's a couple of photos for comparison:

**Original Natural Color PP Shader**:
screenshot_2018-03-24-00-39-05-223_org ppsspp ppsspp

**Modified Version**:
screenshot_2018-03-24-00-38-52-281_org ppsspp ppsspp

P.S: Should I remove the copyright comment at the start of the code? I'm not familiar with licencing issues and legal stuff!


hrydgard:
Copyright is fine, but the diff is weird, seems like you changed line endings of the file or something, so it thinks you changed every single line?

leopard20:
Almost everything's changed actually!

hrydgard:
Yeah but not everything, the diff should have kept things like the ifdef etc. Something is wrong...

leopard20:
Dunno. Beats me. Maybe because I uploaded both files at the same time? Or a github bug?
Did you check the builds? Do they work okay?

P.S: The vertex shader file is not shared with any other shader right? Because that could cause problems.

leopard20:
I tried changing the structure and copying the text manually but still the same...

hrydgard:
Huh, the diff looks right on your next-to-last commit: https://github.com/hrydgard/ppsspp/pull/10774/commits/9a1beaee41dec731042fefef4f3b5532e96a646e

Oh by the way don't worry about the build breaking, this file is separate and not part of the build - if it breaks it's just being unstable..

leopard20:
Well in that commit I just made some changes to the **new** natural.fsh file to see if it was a github bug. I really didn't change anything! So it doesn't count! 😉

leopard20:
Yep you're probably right.

hrydgard:
Just split it into two shaders, one with and one without?

leopard20:
Yes. Seems like this way everyone is gonna be happy! Let me see if I can do it by myself. If not please feel free to do it yourself.

PR #10780, opened 2018-03-24
Libretro port.

aliaspider:


hrydgard:
clang says:

    /home/travis/build/hrydgard/ppsspp/libretro/libretro.cpp:429:3: error: multiple conversions from switch condition type 'std::atomic<EmuThreadState>' to an integral or enumeration type
  switch (emuThreadState)`

Unfortunately I think you're gonna have to make that one an std::atomic<uint32_t> and cast.

aliaspider:
I still don't exactly understand that error, but a cast to the enum type does seem to work. let's wait for travis.

libretro/LibretroGLContext.cpp
```diff
@@ -0,0 +1,51 @@
...
+	g_Config.iGPUBackend = (int)GPUBackend::OPENGL;
+	return true;
+}
+
+void LibretroGLContext::Shutdown()
+{
+	LibretroGraphicsContext::Shutdown();
+	libretro_get_proc_address = nullptr;
+#if 0
+   NativeShutdownGraphics();
```
hrydgard: Nit: Indentation miss.

libretro/LibretroGraphicsContext.h
```diff
@@ -0,0 +1,139 @@
...
+};
+
+extern "C" retro_hw_get_proc_address_t libretro_get_proc_address;
+
+#ifdef _WIN32
+class LibretroD3D9Context : public LibretroHWRenderContext {
+	public:
+	LibretroD3D9Context() : LibretroHWRenderContext(RETRO_HW_CONTEXT_DIRECT3D, 9) {}
+	bool Init() override { return false; }
+#if 0
```
hrydgard: Can we get rid of all these ```#if 0``` ?  Also they're mis-indented.

libretro/LibretroVulkanContext.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+	vk->WaitUntilQueueIdle();
+
+	vk->DestroyObjects();
+	vk->DestroyDevice();
+	vk->DestroyInstance();
+	delete vk;
+	vk = nullptr;
+
+#if 0
```
hrydgard: Get rid of this.

libretro/LibretroVulkanContext.h
```diff
@@ -0,0 +1,24 @@
+#pragma once
+
+#include "Common/Vulkan/VulkanLoader.h"
+#include "libretro/LibretroGraphicsContext.h"
+
+class LibretroVulkanContext : public LibretroHWRenderContext {
+	public:
+	LibretroVulkanContext() : LibretroHWRenderContext(RETRO_HW_CONTEXT_VULKAN, VK_MAKE_VERSION(1, 0, 18))
+	{
+#if 0
```
hrydgard: Again, why ```#if 0``` ?

hrydgard (changes requested):
Some nits.

libretro/LibretroGraphicsContext.h
aliaspider: they are just visual reminders about things that need to be added/fixed. I can get rid of them if that's preferred.
I'll fix the indent, I relied too much on clang-format heh.

libretro/LibretroGraphicsContext.h
```diff
@@ -71,13 +71,13 @@ class LibretroD3D9Context : public LibretroHWRenderContext {
...
 	LibretroD3D9Context() : LibretroHWRenderContext(RETRO_HW_CONTEXT_DIRECT3D, 9) {}
 	bool Init() override { return false; }
-#if 0
-   void InitDrawContext() override
-   {
-      draw_ = Draw::T3DCreateDX9Context();
-      draw_->CreatePresets();
-   }
-#endif
+
```
hrydgard: Here it was #if 0,  but you enabled it. Why wasn't it needed before or why is it needed now?

aliaspider: that part is just placeholder/deadcode at the moment since there is no D3D support in libretro yet. the main reason it was initially disabled was because the build was missing the thin3d files, but that it isn't the case anymore.

hrydgard (approved):

hrydgard:
And it's upstream!

PR #10782, opened 2018-03-24
Added new Natural Colors PP shader (removed blur) 

Leopard20:
Fixes https://github.com/hrydgard/ppsspp/issues/10766.
Here's a couple of photos for comparison:

**Original Natural Color PP Shader**:
screenshot_2018-03-24-00-39-05-223_org ppsspp ppsspp

**Modified Version (called Natural Colors (No Blur) )**:
screenshot_2018-03-24-00-38-52-281_org ppsspp ppsspp


leopard20:
Please let me know if I failed to alter some files in the process.

LunaMoo:
Preferably you should not change existing natural colors at all to avoid messing people's configs and confusing them, including it's name. The blur was part of the original effect, so you should rather mention the new variant is blurless than changing old one to nc+blur :P

leopard20:
Right! Let me change em again! 😩

leopard20:
@hrydgard 
If you don't mind me asking, why do you use both Travis CI and App Veyor? Don't they both do the same thing?!

LunaMoo:
Appveyor is for visual studio / windows builds while travis for everything else, it's common to break windows while developing for example on linux and the opposite, so it helps.
Neither appveyor or travis does anything for pr with no changes affecting build, so if you finished making changes I could merge this... BUT if that's not too much to ask would be nice if you squashed all your commits into a single one(if you're new with git, that's a good thing to learn anyway:P), since all of those commits are unnecessary clutter which just add more to git history for no reason.

leopard20:
If things look good then sure you can merge it.

> would be nice if you squashed all your commits into a single one

Sorry I don't know how to do this!

TheComputerGuy96:
@leopard20 https://stackoverflow.com/a/5201642

leopard20:
Unfortunately I don't use Git. Any other solution?

hrydgard:
Oh, are you doing this using the github text editor? :)  You better learn real git...  but I can squash it for you when I merge.

leopard20:
Yeah. Thanks.

marosis:
Thank you leopard20!@Hrydgard this is it what i wanted!:D

marosis:
Now it is looks like Real PSP screen

leopard20:
@marosis Glad you like it!

@LunaMoo @hrydgard @unknownbrackets 
It looks like the name is not correct though. It says NaturalA but it should be Natural Colors (No Blur). I won't make another pull request just one of you guys fix it. Thanks.

vnctdj:
I really like it too, thanks @leopard20 :)

leopard20:
@vnctdj 
You're welcom!

PR #10789, opened 2018-03-25
Fixed new shader name not displaying correctly

Leopard20:
Sorry. I don't know why the name wasn't correct. This should fix it.


hrydgard:
That's weird? We should be displaying the name...

leopard20:
I know. But just displayed the tag [ ], NaturalA!

hrydgard:
Got it, it's a bug. I'm fixing it.

hrydgard:
Fixed in the upcoming build, this isn't needed.

Issue #10793, opened 2018-03-25
The new shader(naturalA) 


abdellahking:
So this is not an issue but just ask : the new shader(naturalA) suppose to do this to games

![screenshot_20180325-141738](https://user-images.githubusercontent.com/36864704/37875729-30a42268-303b-11e8-969d-adcbcba29b43.png)
![screenshot_20180325-143121](https://user-images.githubusercontent.com/36864704/37875751-64e09408-303b-11e8-9f17-d944da8a85b1.png)

But the Call of duty :rtv
Didn't do like this ahe work perfectly

What hardware and operating system are you running PPSSPP on? GPU might matter if it's a graphical issue.
Mali t720 quad core
Android 6.0
OPENGL

leopard20:
That's weird. I tested most games with this shader and they work ok (GoW, GTA, Assassin's Creed, Flatout, Burnout, NFS, Project Diva Extend, Def Jam, PoP)

marosis:
Yes i have same problem and it's more blury  than when i don't have eyeglasses https://photos.app.goo.gl/QMpiZlSuMmmVR2cq2

abdellahking:
Hmmm. Thats wierd maybe the problem on my device?

marosis:
Leopard20 yes?! And did you tried latest version?https://photos.app.goo.gl/iI17zOd8JgXlCe7g1

marosis:
Flatout have same problem

leopard20:
@marosis I guess you're using the Natural Colors shader!

@abdellahking No, like I said it happens with openGl.

BTW guys, this new shader is called Natural Colors (No Blur), if you still see it as NaturalA you can update your PPSSPP to fix it.

leopard20:
More screenshots:
On Vulkan:
![screenshot_2018-03-25-18-51-37-543_org ppsspp ppsspp](https://user-images.githubusercontent.com/16115820/37876176-14d79e84-305e-11e8-824a-4af4f9ae799c.png)

![screenshot_2018-03-25-18-53-42-074_org ppsspp ppsspp](https://user-images.githubusercontent.com/16115820/37876185-293ace6e-305e-11e8-8bc4-4b52f4a8105a.png)

OpenGL:
![screenshot_2018-03-25-18-52-34-496_org ppsspp ppsspp](https://user-images.githubusercontent.com/16115820/37876194-3e8a69dc-305e-11e8-8633-1152bdebb483.png)

[...]

PR #10799, opened 2018-03-25
Fixed the new shader error on OpenGL ES

Leopard20:
Fixes https://github.com/hrydgard/ppsspp/issues/10793.

~~This reverses commit https://github.com/hrydgard/ppsspp/commit/296481703fb41742b73f1bdd48276967fd70d529 by @LunaMoo.~~
Added set Alpha to 1.0.


assets/shaders/naturalA.fsh
```diff
@@ -63,16 +63,12 @@ void main()
     float d2=dot(abs(c20-c02),dt)+0.0001;
     float hl=dot(abs(c01-c21),dt)+0.0001;
     float vl=dot(abs(c10-c12),dt)+0.0001;
-
-    float md = d1+d2;   float mc = hl+vl;
-    hl*=  md;vl*= md;   d1*=  mc;d2*= mc;
```
hrydgard: So you just remove these multiplications, won't that change the look? No idea what's going in this shader anyway though :)

I'm guessing what happened before sometimes is that md and mc both became 0, ww became 0, and then the long line below divides by (3.0*ww) which could end up 0.0. And dividing by 0.0 is bad... Maybe better to just check for that? Not sure.

leopard20: Nope. It looks the same (99% +)

That was the only solution that came to my mind. Don't merge for now to see if I can find a better fix.

leopard20:
I found something strange. Even though it looks fine in some games, Assassin's Creed still look like this (openGl):
screenshot_2018-03-25-22-11-02-335_org ppsspp ppsspp

The original one on Vulkan (to prove they look the same):
screenshot_2018-03-25-22-11-34-928_org ppsspp ppsspp

assets/shaders/naturalA.fsh
leopard20: @hrydgard 
Quick question: If this was because of a miscalculation, then wouldn't the issue occur on all backends? But this only happens on OpenGl ES (not even OpenGl).

LunaMoo:
For your reversing of addition of alpha see https://github.com/hrydgard/ppsspp/commit/7eaec209d3feabfecfa338b9cee213d6f1295cd2, I don't remember why it was added exactly, but some mobiles might had problems otherwise.

 Also since this fix doesn't really fix anything, just hides most of the problem, the actual problem is most likely not even related. 
 More likely actual bug is within one of the functions used, for example pow, which I remember does have a problem on some drivers in certain situations, for example pow(x,2). For a solution try checking for pow(x,y) where y = 2 and replace it with simple x*x. Overall there are other cases where results of some functions might not be defined and drivers will show a difference there, hence vulkan will differ from ogl etc...

marosis:
don't forget to set different languages

leopard20:
@marosis 
Why?

leopard20:
@LunaMoo 
Thanks for the suggestion. I'm working on it.

marosis:
I toight it could be usefull😥

marosis:
Tought*

PR #10818, opened 2018-03-27
POC(do not merge): D3D realtime texture scaling filters.

aliaspider:
this adds an option to bybass cpu based scalers and do the scaling on GPU at sampling time instead.

the GPU cost is obviously very high, but this comes with the benefit of completely negating the small delays that generally occur with texture loading when the scaling filters are enabled.

probably the more sane approach would have been to scale the textures once with the gpu, with compute shaders ideally, but this was a lot simpler to make, and worth doing just to check out how good/bad it will perform.

I also added two other scaling filters for testing, they don't have a cpu implementation yet and will just fallback to bicubic if realtime scaling is disabled.

only xbr and the two new filters have a gpu implementation, I will see if I can add the rest later.


hrydgard:
Cool try, and nice to get the scalers in shader format - but yeah, this is not the way I want to go - this scales (no pun intended) very badly with rendering resolution due to the high pixel shader cost.

aliaspider:
- fixed requiring a restart when toggling the option from the internal UI, it only worked from the win32 ui before.

- it doesn't look like the effect is being applied twice (I tested this by making it output flipped colors, which would have ended up being displayed normally if that were the case). I think it is just being applied to all textures including the ones that are bypassing the texture cache, like the initial warning screen, which is a plain software rendered framebuffer iirc. the final draw to the swapchain is more likely using a different path, since it can set a post-processing shader.

- added bicubic and some other test filters that are a bit faster and better suited for realtime scaling than something heavyweight like xbr.

- I can't seem to find a working fix for the edge problem with convolution scalers, it looks like it is trying to sample outside the texture, which is plain black. the problem also happens with the cpu based bicubic filter. edge detection scalers like xbr will just ignore the bad samples, which is why they end up looking correctly.

LunaMoo:
xBRZ can be applied to itself as much as we want so it would not be an issue of being applied twice, the issue I mentioned was about situation where game renders the result of it's draw to a texture which by this will be taken as native resolution texture and upscaled, in result it doesn't really differ from a post process shader upscaling native res which in games that mix 2D and 3D graphics looks poorly on the 3D elements.

 This happens for example in Harvest Moon Boy & Girl that also happens to work poorly with cpu scaling and at the same time is pretty bothersome case for texture replacement as well:
ulus10142_00000

aliaspider:
oh yeah it works pretty badly for that game heh, and flipping color has no effect so that also confirms that the effect is applied twice.
I'll see if I can find a workaround for this, without messing with the other cases where applying the filter to the framebuffer turned out to be an improvement.

aliaspider:
I found a solution for the render to target textures which will now be correctly skipped. they are prescaled with the internal resolution factor anyway.
the software framebuffers are still being processed by this. (it turns out the reason they are skipped by cpu scalers is the fact that they have the `STATUS_CHANGE_FREQUENT` flag set).

Harvest Moon Boy & Girl now looks pretty good with xbr, except for a small issue with text which I'll hopefully figure out soon.

glad to see at least one game where this approach is actually beneficial, since the game is constantly mapping new textures, which makes pre-scaling less practical. and the fact that the game engine is pretty basic and running at only 30fps makes the GPU performance hit even less noticeable.

zminhquanz:
It will solve problem about texture scaling when object movement . For example : Ys Seven map , when you move , it's very blurred and broken

aliaspider:
in the case of "Harvest Moon Boy & Girl", looking at the texture dump, it seems that at least the main sprite is being updated in place in VRAM when it changes direction, so it is constantly being bypassed by the cpu scaler for several frames. 
given that this games mixes 2D and 3D, it can benefit from increased internal resolution, where a post shader would unable to filter the sprites correctly.

hrydgard:
Closing, while cool we're not going to go with this approach, instead we upscale textures with a compute shader as they are uploaded to the GPU (#10393). Unfortunately this is currently only implemented for Vulkan, but everybody who can should use the Vulkan backend anyway now as that's where future work will be focused the most.

PR #10827, opened 2018-03-28
Fix libretro compile for gles platforms

webgeek1234:
Specifically fixes android builds for me.


hrydgard:
Yeah, this looks right.

PR #10828, opened 2018-03-28
More libretro fixes

webgeek1234:
First commit fixes running on android. Core needs to be statically linked since libretro cores are shipped as single shared libraries.

Second commit fixes said shared library being enormous. I don't know of a better way to do this. Open to suggestions if there are.

The core is crashing on exit for some reason. Will investigate that later as I have time. This should make it usable at least, though. Tested on Tegra X1, Lineage 15.1 (Oreo).


aliaspider:
you could try garbage collection: build with `-ffunction-sections` `-fdata-sections` and then link with `-Wl,--gc-sections`.

webgeek1234:
I could submit that for the buildbot build scripts, but it'd be the only path that would need it. And it wouldn't help local builds either. I know I can do it manually, but that gets tedious every time.

orbea:
The libretro buildbot already strips binaries, but only or `unix` and `windows`. I'm not sure why this is not done for android? Maybe it helps debugging to not strip on android?

orbea:
The libretro buildbot already strips binaries, but only on `unix` and `windows`. I'm not sure why this is not done for android? Maybe it helps debugging to not strip on android?

webgeek1234:
Well, would you look at that. I didn't look far enough. Okay, fair enough. I'll expand that in a PR that direction. Afaik, there's no reason to. Most of the Android cores are stripped, probably in the mass of Makefiles somewhere. I'll revise this later tonight and pull that commit.

orbea:
I'm not really an android user and I suspect that script would be stripping android cores already unless there was a reason to not. I would like to know what that reason is before I start making changes for platforms I don't use. :)

webgeek1234:
The thought that came to mind right after I replied was that to strip, you have to have a bin targeting the platform. So, that would be more of a pain to get. I'll see if I can't come up with an easy way to build the path to it in the ndk. It'd be easier to just use this since it's already done. And I've tested that core does run. Debugging lines go into logcat, so there's no trouble there unless someone want's to do full tracing. Normal users aren't that, though.

webgeek1234:
@aliaspider: That didn't help. Tried the vulkan renderer and it does not have the crash on exit. So, it appears to be GL related.

aliaspider:
what ?
the garbage collection suggestion was for reducing the binary size.

webgeek1234:
Oh, misunderstood. Um. That didn't change significantly either. Other than stripping, I've only been able to drop it from 115 MB to 103 MB. Stripping takes it to 12 MB.

Any clue why the GL path would hang on exit, causing a force close when input stops responding? That doesn't happen on the Vulkan path.

webgeek1234:
Okay, I've dropped the strip commit. Working on the libretro side to find the best way for the recipe script to handle stripping. The other commit is still needed to make a usable android core, though. I'll investigate the crash and submit a separate PR if I come up with anything.

Issue #10823, opened 2018-03-28
Ultimate Ghost 'n Goblins Crashes


Panderner:
The game has crashes trying to start the game

hrydgard:
Please follow these instructions:
https://github.com/hrydgard/ppsspp/wiki/How-to-report-an-issue-properly

adennen:
I have this issue too, I'm running the latest source version on a Mac, the crash is a EXC_BREAKPOINT (SIGTRAP) in TextureCacheCommon::ApplyTexture().I think it's being caused by line 1520: _assert_(!entry->texturePtr); commenting this line out fixes the crash for me

hrydgard:
Huh, interesting.

hrydgard:
Same game used to leak a lot of textures in a different way before: https://github.com/hrydgard/ppsspp/issues/9089

hrydgard:
Try now.

adennen:
It's working for me

PR #10833, opened 2018-03-29
When binding a framebuffer to a texcache entry, delete its texture. Should help #10823

hrydgard:


hrydgard:
spurious build error, apparently.

unknownbrackets (approved):

PR #10845, opened 2018-04-01
Add SSAA flag for post process shaders

### LINKED ISSUE #10535: Natural Postprocessing Shader is awesome but Blurry
So natural looks very blurry unless you upscale the rendering resolution. This is bad when you want to run full screen as you must have `rendering resolution: auto` on to get good detail. It would be nice for there to be an upscale switch on post processing so that it looks better. Natural looks so good then.

So a hack would be to guesstimate what upscale you need by some calculations done by taking your screen resolution into account and setting the appropriate rendering resolution. Then setting it up a level or two.

LunaMoo:
It simply enforces ~~2x~~ chosen multiplier of display resolution, ignoring user render res, best used with blurry effects.

fixes #10535, althrough doesn't change natural colors shader which is in that's issue title as I think modifying existing effects with enforced double auto resolution is bad, instead add's an example SSAA effect using simplified gauss filter.

 When it's used and resolution info is printed on the screen it also add's "(supersampling)" text.

### DIFF
--- a/GPU/Common/FramebufferCommon.cpp
+++ b/GPU/Common/FramebufferCommon.cpp
@@ -1752,17 +1752,20 @@ void FramebufferManagerCommon::Resized() {
 	}
 
 	postShaderIsUpscalingFilter_ = shaderInfo ? shaderInfo->isUpscalingFilter : false;
+	postShaderIsSSAAFilter_ = shaderInfo ? shaderInfo->isSSAAFilter : false;
 
 	// Actually, auto mode should be more granular...
 	// Round up to a zoom factor for the render size.
 	int zoom = g_Config.iInternalResolution;
-	if (zoom == 0) {
+	if (zoom == 0 || postShaderIsSSAAFilter_) {
 		// auto mode, use the longest dimension
 		if (!g_Config.IsPortrait()) {
 			zoom = (PSP_CoreParameter().pixelWidth + 479) / 480;
 		} else {
 			zoom = (PSP_CoreParameter().pixelHeight + 479) / 480;
 		}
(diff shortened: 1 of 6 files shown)


GPU/Common/FramebufferCommon.h
```diff
@@ -379,6 +379,7 @@ class FramebufferManagerCommon {
 	bool usePostShader_ = false;
 	bool postShaderAtOutputResolution_ = false;
 	bool postShaderIsUpscalingFilter_ = false;
+	bool postShaderIsSSAAFilter_ = false;
```
hrydgard: Why not make this an integer, so you can make an even more extreme 3x3 SSAA filter in the future? :)

LunaMoo: :3 done.

GPU/Common/FramebufferCommon.cpp
```diff
@@ -1752,17 +1752,20 @@ void FramebufferManagerCommon::Resized() {
 	}
 
 	postShaderIsUpscalingFilter_ = shaderInfo ? shaderInfo->isUpscalingFilter : false;
+	postShaderIsSSAAFilter_ = shaderInfo ? shaderInfo->isSSAAFilter : 0;
```
hrydgard: Sorry for nagging, but I of course also meant to rename it :)   postShaderSSAAFilterLevel ?

LunaMoo: Sorry for forgetting about that;p, a holiday-of-too-much-food made me lazy.

hrydgard:
Oh, sorry for the delay, it's fine.

PR #10887, opened 2018-04-08
Less intrusive volume change indicator on iOS

konistehrad:
A quality-of-life issue, but iOS used to cover a large portion of the screen with the volume indicator when the system volume was changed:

img_0013

 I've taken an [Objective-C port](https://github.com/iMokhles/SubtleVolume/blob/master/subtleVolumeObjC/subtleVolumeObjC/SubtleVolume.mm) of [SubtleVolume](https://github.com/andreamazz/SubtleVolume), spruced it up with some tighter animation code and padding support, and integrated it into PPSSPP. Now, when you change the volume, you're presented with a bar at the top of the screen:

img_0009

Tested on iOS 11.3 on iPhone 7, 10.3.3 on iPhone 6. Layout dimensions similar to those found in Instagram.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -722,10 +722,12 @@ elseif(IOS)
 		ios/PPSSPPUIApplication.mm
 		ios/SmartKeyboardMap.cpp
 		ios/SmartKeyboardMap.hpp
+		ios/SubtleVolume.h
+		ios/SubtleVolume.mm
 		ios/iCade/iCadeReaderView.h
 		ios/iCade/iCadeReaderView.m
 		ios/iCade/iCadeState.h)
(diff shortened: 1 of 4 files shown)


LunaMoo:
We indent with tabs, even if you have different personal preference, you should always match the project you're contributing to, so please correct all of that.

hrydgard:
Yeah, apart from the indentation issue, I'm fine with this.

konistehrad:
Ah, my apologies must be the settings Xcode defaulted to. I've changed all `.h` and `.mm` files I've updated to use tabs, and rebased on upstream master. Thanks again!

konistehrad:
Weird, I use the Exit button pretty judiciously, and absolutely tested it as part of this patch and in the builds I use basically on the daily. (See https://github.com/hrydgard/ppsspp/pull/10887/commits/3f90c97f39d2efd7a55037e938d9acb8c4f36836 ) ; that said, if there's a way I can help triage let me know.

hrydgard:
Maybe it only happens if you exit without changing the volume? That way you could have missed it in your testing :)

PR #10921, opened 2018-04-15
Add support for File app (file browser) on iOS 11

imxieyi:
This option enables File app support on iOS 11+, making it possible to download games and manage data without any computer:
img_0520


unknownbrackets (approved):

konistehrad:
That's awesome thank you!! 👍

PR #10949, opened 2018-04-21
Fix build with ffmpeg 4.0

valpackett:
Compatibility with the new ffmpeg version 4.0. Like https://github.com/RPCS3/rpcs3/pull/4471

### DIFF
--- a/Core/HLE/sceMpeg.cpp
+++ b/Core/HLE/sceMpeg.cpp
@@ -884,13 +884,16 @@ public:
 		}
 	};
 #ifndef USE_FFMPEG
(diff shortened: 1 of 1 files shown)


LunaMoo:
Looks fine, but feels pointless, PPSSPP doesn't work properly with newer versions of FFMPEG so people who build PPSSPP with it, will get a broken experience ~ see #9026.

myfreeweb:
Yeah, videos look broken, but they're usually not the most important part of games. PPSSPP is built with system ffmpeg on FreeBSD because the bundled version won't build without patching it.

PR #11010, opened 2018-05-06
Shader depal (Vulkan): Apply a half-texel offset when bilinear filtering, fixing #11008

hrydgard:
Also skip three texture samples if all indices are equal. Haven't actually benchmarked that though, might ditch it.

### DIFF
--- a/GPU/Vulkan/FragmentShaderGeneratorVulkan.cpp
+++ b/GPU/Vulkan/FragmentShaderGeneratorVulkan.cpp
@@ -175,29 +175,37 @@ bool GenerateVulkanGLSLFragmentShader(const FShaderID &id, char *buffer) {
 				doTextureProjection = false;
 			}
 
-			if (doTextureProjection) {
-				WRITE(p, "  vec4 t = textureProj(tex, %s);\n", texcoord);
-				if (shaderDepal) {
-					WRITE(p, "  vec4 t1 = textureProjOffset(tex, %s, ivec2(1, 0));\n", texcoord);
-					WRITE(p, "  vec4 t2 = textureProjOffset(tex, %s, ivec2(0, 1));\n", texcoord);
-					WRITE(p, "  vec4 t3 = textureProjOffset(tex, %s, ivec2(1, 1));\n", texcoord);
+			if (!shaderDepal) {
+				if (doTextureProjection) {
+					WRITE(p, "  vec4 t = textureProj(tex, %s);\n", texcoord);
+				} else {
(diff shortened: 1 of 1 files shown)


GPU/Vulkan/FragmentShaderGeneratorVulkan.cpp
```diff
@@ -175,29 +175,37 @@ bool GenerateVulkanGLSLFragmentShader(const FShaderID &id, char *buffer) {
...
-			}
-
-			if (shaderDepal) {
+				WRITE(p, "  vec2 tsize = textureSize(tex, 0);\n");
+				WRITE(p, "  vec2 fraction;\n");
+				WRITE(p, "  bool bilinear = (base.depal_mask_shift_off_fmt >> 31) != 0;\n");
+				WRITE(p, "  if (bilinear) {\n");
+				WRITE(p, "    uv_round = uv * tsize - vec2(0.5, 0.5);\n");
+				WRITE(p, "    fraction = fract(uv_round);\n");
+				WRITE(p, "    uv_round = (floor(uv_round) + vec2(0.5, 0.5)) / tsize;\n");  // We want to take our four point samples at pixel centers.
```
aliaspider: shouldn't this be `uv_round = (uv_round - fraction) / tsize;` instead ?

hrydgard: ```uv_round - fraction```, sure (changing to that), but since we use a point sampler I want to be well away from the borders between pixels - not having the 0.5,0.5 offset for the actual lookups makes it noisy.

aliaspider: `uv_round - fraction` will already give you the center, since `fraction` is the distance to the left center. adding 0.5 to that would land you at the border.

hrydgard: something else is off somewhere in that case, because without the offset, the toon effect is broken in DBZ: Tenkaichi. Need to dig more into this .. after 1.6 :)

aliaspider: oh interesting.
well if it works correctly then the theory was just wrong then xD
now I have to test this too..

aliaspider: ah I see, I missed the fact that `uv_round` still had that `vec2(0.5, 0.5)` in it, and you needed to add that back. 
I was seeing it this way heh :
```fraction = fract(uv * tsize - vec2(0.5, 0.5));
uv_round = uv - fraction / tsize;```

PR #11028, opened 2018-05-09
Show PPSSPP version only

vnctdj:
As you already know, that's a common option for Linux programs.
I think this is a basic feature which is currently missing :)

Tested for SDL, but not for Qt as I don't use it, but it should be the same.
I'm not sure if this would be useful for other platforms?


unknownbrackets (approved):

hrydgard:
Yeah, why not.

PR #11032, opened 2018-05-10
Create vital directories on Android storage permission grant. 

hrydgard:
Should help #11020


Leopard20 (changes requested):
Perhaps you should add another line to your function `CreateDirectoriesAndroid`:

```File::CreateDir((g_Config.memStickDirectory + "PSP/SYSTEM").c_str());```

leopard20:
Also, creating directories after permission is granted doesn't clear the whole directory, does it?

hrydgard:
Oops, yeah definitely. And no, CreateDir only creates a directory if it doesn't exist, it does nothing beyond that.

leopard20:
Ok good! :)

hrydgard:
Tested it myself by renaming the PSP directory on my device, and it looks like it works fine.

PR #11072, opened 2018-05-25
Android immersive mode: Fix display size stretch and touch screen inaccuracy

hrydgard:
Happened when toggling immersive mode. When in immersive, we need to call getRealMetrics, which is only available from API 17 unfortunately. The Android display size APIs are not super well conceived...


leopard20:
Almost all phones that had on-screen navigation buttons support Android 4.2 onwards IIRC, so it shouldn't pose a problem!

PR #11107, opened 2018-05-30
Unbreak -march=native build on x86

jbeich:
See [error log](https://ptpb.pw/SneQ). Affects both Clang and GCC.


unknownbrackets (approved):
Just noting: this is because it uses the 64-bit version of the crc intrinsic, which isn't supported in 32-bit mode.

This really seems like an upstream bug:
https://github.com/google/cityhash/blob/master/src/city.cc#L553-L555

But we already had the ANDROID thing for some reason, so this deviation from upstream shouldn't be a big deal.

-[Unknown]

hrydgard:
Yeah this is fine.

PR #11111, opened 2018-05-31
depal shader use high int precision for opengl es

weihuoya:
fx Prince of Persia Rival Swords - Yellow Screen #10522


unknownbrackets (approved):
Nice catch.  I wonder if there are other places / shaders we need to increase the precision for...

-[Unknown]

hrydgard:
Good find! I'm gonna sneak this into 1.6.3.

Issue #11113, opened 2018-05-31
PSP2 is crashing while using Vulkan and changing the resolution


vsub:
That includes using Auto and do any kind of window sizing.
When it crashes,first it shows a dialog with Yes and No and as info

Critical
dstrect offset y + extent > height
Try to continue?
Yes   No

Choosing No,kills the emulator and choosing Yes
Critical
Lost the Vulkan device!
Try to continue?
Yes   No

Choosing No,kills the emulator and choosing Yes shows the same dialog and you can hear the music for a second

It doesn't crash only on 1.4.2 and on the latest beta,there are some weird blinking things at the title screen while using vulkan on 1.6

This is the log
56:25:898 user_main    E[SCEKERNEL]: HLE\sceKernel.cpp:675 UNIMPL sceKernelReferGlobalProfiler(deadbeef)
56:41:539 user_main    E[ME]: HLE\sceMpeg.cpp:1971 UNIMPL sceMpegAvcInitYCbCr(09092700, 1, 480, 272, 086091c0)
56:41:539 user_main    E[ME]: HLE\sceMpeg.cpp:1971 UNIMPL sceMpegAvcInitYCbCr(09092700, 1, 480, 272, 08638f80)
56:41:539 user_main    E[ME]: HLE\sceMpeg.cpp:1971 UNIMPL sceMpegAvcInitYCbCr(09092700, 1, 480, 272, 08668d40)
56:41:539 user_main    E[ME]: HLE\sceMpeg.cpp:1971 UNIMPL sceMpegAvcInitYCbCr(09092700, 1, 480, 272, 08698b00)
56:41:840 user_main    E[ME]: HLE\sceMpeg.cpp:1335 UNIMPL sceMpegAvcDecodeFlush(09092700)
56:42:098 SOUND_THREAD E[ME]: HLE\sceAtrac.cpp:1695 80630016=sceAtracResetPlayPosition(0, 0, 0, 0): first byte count not in valid range
56:48:365 SOUND_THREAD E[ME]: HLE\sceAtrac.cpp:1695 80630016=sceAtracResetPlayPosition(0, 0, 0, 0): first byte count not in valid range
57:03:834 SOUND_THREAD E[ME]: HLE\sceAtrac.cpp:1695 80630016=sceAtracResetPlayPosition(0, 0, 0, 0): first byte count not in valid range
[...]

PR #11118, opened 2018-06-01
When creating temp framebuffers for download, size them using bufferWidth/Height instead of width/height.

hrydgard:
Or maybe we should make sure to only download within the width/height by
adding more clamps at the beginning of ReadFramebufferToMemory but seems
more dangerous, at least if we want to sneak this one into 1.6.3..

Plus some minor things.

Should help #11113


unknownbrackets (approved):
This makes sense - maybe I'm remembering wrong that something would download from or transfer from this area, then, because it couldn't have been working?  Or maybe this explains black textures somewhere.

-[Unknown]

Issue #11116, opened 2018-06-01
1.6.2 Play Store detected crashes


hrydgard:
```
 #00  pc 000000000048340e  /mnt/asec/org.ppsspp.ppsspp-2/lib/arm/libppsspp_jni.so (_ZN14ManagedTexture10GetTextureEv+5)
  #01  pc 00000000004bb857  /mnt/asec/org.ppsspp.ppsspp-2/lib/arm/libppsspp_jni.so (_ZN9UIContext5BeginEv+46)
  #02  pc 000000000042c04f  /mnt/asec/org.ppsspp.ppsspp-2/lib/arm/libppsspp_jni.so (_ZN9EmuScreen8renderUIEv+42)
  #03  pc 000000000042be23  /mnt/asec/org.ppsspp.ppsspp-2/lib/arm/libppsspp_jni.so (_ZN9EmuScreen6renderEv+130)
  #04  pc 00000000004bb065  /mnt/asec/org.ppsspp.ppsspp-2/lib/arm/libppsspp_jni.so (_ZN13ScreenManager6renderEv+84)
  #05  pc 00000000004251cb  /mnt/asec/org.ppsspp.ppsspp-2/lib/arm/libppsspp_jni.so (_Z12NativeRenderP15GraphicsContext+794)
  #06  pc 000000000042050f  /mnt/asec/org.ppsspp.ppsspp-2/lib/arm/libppsspp_jni.so (_Z20UpdateRunLoopAndroidP7_JNIEnv+22)
  #07  pc 0000000000421855  /mnt/asec/org.ppsspp.ppsspp-2/lib/arm/libppsspp_jni.so
  #08  pc 00000000002b72a5  /mnt/asec/org.ppsspp.ppsspp-2/lib/arm/libppsspp_jni.so (_ZNSt6__ndk114__thread_proxyINS_5tupleIJNS_10unique_ptrINS_15__thread_structENS_14default_deleteIS3_EEEEPFvvEEEEEEPvSA_+24)
  #09  pc 0000000000016f23  /system/lib/libc.so (_ZL15__pthread_startPv+30)
  #10  pc 0000000000014f43  /system/lib/libc.so (__start_thread+6
```
Thought we had fixed this one, apparently not.

```
#00  pc 000000000007f392  /system/vendor/lib/hw/vulkan.msm8937.so (_ZN16A5xDescriptorSet24HwUpdateWriteDescriptorsEPK22QglDescriptorWriteData+273)
[...]

PR #11119, opened 2018-06-01
Never bind a NULL image view in Vulkan, not even through thin3d

hrydgard:
Some mobile drivers really don't like it, should fix some crashes seen in #11116. This simply moves the nullTexture concept from the DrawEngineVulkan to thin3d vulkan.


unknownbrackets (approved):
Seems to be working okay for me.

-[Unknown]

PR #11137, opened 2018-06-04
Opengl save cache before clear

weihuoya:
maybe use `GL_OES_get_program_binary` or `GL_ARB_get_program_binary` extension to cache linked program, too.


hrydgard:
Hah. Ooops. Nice catch.

PR #11221, opened 2018-06-26
Modify D3D11 hwtess texture buffer to structured shader buffer.

xebra:
Like Vulkan.

@hrydgard I have an issue.
The following commit causes drawing bugs with huge spline/bezier like "FF4 mist dragon morphing" or "Spline Surface demo in jpcsp demos folder".
So I think we need to keep 32bits.
Edit: 648b4eac(Shrink the vertexCount field in DeferredDrawCall to the 16 bits required.)

BTW, the brand new GL rendering manager is awesome!
Really nice performance and really nice work :)

### DIFF
--- a/GPU/D3D11/DrawEngineD3D11.cpp
+++ b/GPU/D3D11/DrawEngineD3D11.cpp
@@ -695,71 +695,38 @@ rotateVBO:
 #endif
 }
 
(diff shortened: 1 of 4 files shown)


hrydgard:
Hm, yeah, you're probably right about the vertex count. It normally doesn't go very high but in these tesselation cases it can...

hrydgard:
I committed in a quick fix for the vertexCount.

xebra:
Thanks!

PR #11254, opened 2018-07-15
Reduce the estimated transform cost of an additional light from 10 to 7 cycles

hrydgard:
See #5530.


unknownbrackets (approved):
As you said before, probably best to just find out what breaks (if anything.)

-[Unknown]

PR #11281, opened 2018-08-01
add libretro hooks for memory pointers to allow usage of new retroarc…

RetroSven:
…h cheat code searching and rumble-on-memory change features


RetroSven:
Thanks!

PR #11331, opened 2018-08-26
Allow choosing to use the system snappy or not.

orbea:
I think it can be useful to configure to use the system snappy or the included static snappy. Opinions?


hrydgard:
Why not.

PR #11332, opened 2018-08-28
Add: Get user-preferred locale from OS (linux only)

AreaScout:
This will set the region locale on linux to support different language on games they do support it

### DIFF
--- a/SDL/SDLMain.cpp
+++ b/SDL/SDLMain.cpp
@@ -19,6 +19,7 @@ SDLJoystick *joystick = NULL;
 #include <cassert>
 #include <cmath>
 #include <thread>
+#include <locale>
 
 #include "base/display.h"
 #include "base/logging.h"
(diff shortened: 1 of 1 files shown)


SDL/SDLMain.cpp
```diff
@@ -176,6 +177,13 @@ void LaunchEmail(const char *email_address) {
 }
 
 std::string System_GetProperty(SystemProperty prop) {
+#ifdef __linux__
```
hrydgard: This check should be done in the case SYSPROP_LANGREGION, not out here, and preferably only once during execution

AreaScout:
proof: GoW playing on ODROID in German Video :)

[IMAGE ALT TEXT HERE](https://www.youtube.com/watch?v=QegJlwflkZk)

SDL/SDLMain.cpp
hrydgard: note that you can declare variables in a switch case, you just have to wrap the entire case in { }.

AreaScout:
Ok I have investigated further on this, this PR works also on macOS if ``defined(__APPLE)`` is added, I have tested it on my mac mini (early 2009) freshly updated to 10.12.5 🤣and it works.

The funny thing is, I have made this PR in the hope it will solve a problem with some games not having e.g german language on game play, but I realized that it's working also without this PR now !!?? I swear it was not working 3 weeks before when having region language different from menu language settings.

howsoever maybe this was fixed somewhere in between, is this PR still useful then ? On Windows the region is set according to the OS language setting.

I will change it to affect also macOS, let me know if I should close this PR or not

AreaScout:
as it looks like macOS and Linux are the only OS's that are using this, i have removed the ``#if`` condition

PR #11354, opened 2018-09-03
Add: ban only Mali-4x GPU from 3.x context, they do not support GLES 3.x

AreaScout:
Here are various GPU Render Strings:

Mali-400
Mali-450
Mali-T600
Mali-T620
Mali-T720
Mali-T760
Mali-T820
Mali-T830
Mali-T880

As of Mali-Tx they support OpenGL ES 3.1 
https://en.wikipedia.org/wiki/Mali_(GPU)#Variants


hrydgard:
If you want to make changes to a pull request, you don't need to open a new one, you can just force-push over it.

AreaScout:
I was not about to change this, I did close this because PR because it was wrong, there are some checks before so it never reaches that string compare on 3.x contexts

PR #11390, opened 2018-09-16
Add: allow touch screen events on fullscreen

AreaScout:
## Description
Disable full screen check for touch screen events

## Motivation and Context
It also works in window mode, moving the window itself on the desktop is done with xinput touch events, every touch event that is done inside the window is handled by SDL2 -> very cool !
No special handling is needed for it

## How Has This Been Tested?
Laptop with touch screen on Ubuntu 18.04
Didn't test it on macOS but i guess that should work the same way

### DIFF
--- a/SDL/SDLMain.cpp
+++ b/SDL/SDLMain.cpp
@@ -684,7 +684,6 @@ int main(int argc, char *argv[]) {
 				}
 			case SDL_FINGERMOTION:
 				{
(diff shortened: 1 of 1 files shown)


hrydgard:
Alright, I guess this should be fine. I can't recall why these checks were added?

AreaScout:
I just couldn't test it in desktop environment, that's why I placed those checks in the first place

Issue #11154, opened 2018-06-08
GPD XD (original): Display resolution (HW scaler) causes crashes


radiocaravan:
When I select a Display resolution other than the default (2x PSP), the app crashes instantly.

GPD XD (original), LegacyROM, PPSSPP Gold 1.6.3

ChurroBoi:
[gpd_xd_res-crash.txt](https://github.com/hrydgard/ppsspp/files/2089224/gpd_xd_res-crash.txt)
Here, I made a crash log.

ChurroBoi:
ALSO, if you aren't patient enough to wait for a fix you can go to the ppsspp.ini in the SYSTEM folder and set AndroidHwScale = 0 to make full resolution.

radiocaravan:
I made this change in the INI file... it worked great.  AND it also magically fixed the Manual Scaling issue.

hrydgard:
Good find. Yeah we should probably not default that to On, it causes too many issues....

hrydgard:
I'm gonna close this as a duplicate of #11063 because it's almost certainly has the same causes.

Issue #11151, opened 2018-06-07
Display distorted on Nexus Player Android TV 8.0


RoylandRW:
This is all that I can see when opening the app on my Nexus Player Android TV 8.0, tried deleting settings backup, tried uninstalling and reinstalling, tried most things I can do without actually being able to access the apps settings menu, cause it is not viewable or changeable in this state. https://photos.app.goo.gl/ZJtodgT7eS9Y19Jh6

hrydgard:
Probably a GPU driver bug, quite an unfortunate one though :(

hrydgard:
@RoylandRW  can you please test the latest builds? It might be fixed now...

hrydgard:
@RoylandRW  Sorry the play store version won't be enough, you'll need to get the latest from the buildbot and install manually via usb (or one of the other ways to sideload): https://buildbot.orphis.net/ppsspp/

RoylandRW:
This one? v1.6.3-383-g22a536fb0

RoylandRW:
I wish I'd seen your response before I'd updated from the play store. Should I delete the play store version I have before installing this one?

hrydgard:
Sorry for the late reply @RoylandRW , thank you very much for reporting, a bit surprising that it didn't help, maybe it needs even bigger alignment. Probably better to force scaling off, can you try to edit /PSP/SYSTEM/ppsspp.ini on your Nexus (I guess just copy the file to your computer and back) and change AndroidHwScale under [Graphics] to 0?

hrydgard:
@RoylandRW could you please try again with a fresh build, deleting ppsspp.ini first? It should no longer even try to use hardware scaling so hopefully this glitch will be gone (unless you enable it, and we should probably detect TV devices and get rid of it...)

hrydgard:
@RoylandRW haven't got a sec to try it? Would be great to get confirmation before releasing 1.7.0.

hrydgard:
No confirmation, moving to 1.8.0 for now.

hrydgard:
Very likely to be fixed now, closing.

PR #11392, opened 2018-09-16
Android: Don't use "hw scaling" on modern devices by default.

hrydgard:
It seems to cause more problems than it's worth, see #11151 and multiple other issues with rotation etc.

Should we force this by renaming the config option? And what should the cutoff be? Really old devices usually benefit quite noticeably.

See also #11154 and #11198

### DIFF
--- a/Core/Config.cpp
+++ b/Core/Config.cpp
@@ -510,6 +510,12 @@ static bool DefaultTimerHack() {
 
 static int DefaultAndroidHwScale() {
 #ifdef __ANDROID__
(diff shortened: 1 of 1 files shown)


unknownbrackets (approved):
Hm.  We could try resolution, but then there's old tablets and small screen modern devices as edge cases.

We could simply have a list of GPU string matches (Mali-4, etc.) that we continue the default on, maybe.  Anyone with a more obscure GPU would default off.

-[Unknown]

PR #11415, opened 2018-09-23
Update ffmpeg submodule

hrydgard:


hrydgard:
OK so this breaks badly on Linux, that can't be good...

hrydgard:
Thank you! That one's from June? I completely forgot about it, sorry.

It's good if we can build on Trusty yeah...

PR #11421, opened 2018-09-25
Fix: prevent linkage error if EGL and GLES is not in one big blob

AreaScout:
## Description
Building for GLES2 on my desktop PC 32bit VM Linux Ubuntu machine will give a linkage error:

``lib/libnative.a(gpu_features.cpp.o): In function `CheckGLExtensions()':``
``gpu_features.cpp:(.text+0x141f): undefined reference to `eglQueryString'``
``gpu_features.cpp:(.text+0x144a): undefined reference to `eglGetCurrentDisplay'``
``gpu_features.cpp:(.text+0x145a): undefined reference to `eglQueryString'``
``lib/libnative.a(gl3stub.c.o): In function `gl3stubInit':``
``gl3stub.c:(.text+0xb): undefined reference to `eglGetProcAddress'``
``gl3stub.c:(.text+0x1c): undefined reference to `eglGetProcAddress'``
``gl3stub.c:(.text+0x2d): undefined reference to `eglGetProcAddress'``
``gl3stub.c:(.text+0x3e): undefined reference to `eglGetProcAddress'``
``gl3stub.c:(.text+0x4f): undefined reference to `eglGetProcAddress'``
``lib/libnative.a(gl3stub.c.o):gl3stub.c:(.text+0x60): more undefined references to `eglGetProcAddress'`` ``follow``
``collect2: error: ld returned 1 exit status``
``make[2]: *** [PPSSPPSDL] Error 1``

## Motivation and Context
If i.e libEGL.so and libGLESv2.so don't symbolic link to one driver binary (for MALI it is libmali.so), then
we get those linkage errors.
i.e Mesa has EGL and GLESv2 separated 

## How Has This Been Tested?
Ubuntu 14.04 inside VirtualBox with GalliumVMware backend driver to Mesa

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -179,7 +179,7 @@ endif()
 include_directories(ext/native)
 
 if(NOT OPENGL_LIBRARIES AND USING_GLES2)
-	set(OPENGL_LIBRARIES GLESv2)
+	set(OPENGL_LIBRARIES GLESv2 EGL)
 endif()
 
 if(USING_EGL)


AreaScout:
Binary was build with:

``cmake -DUSING_EGL=OFF -DUSING_GLES2=ON -DUSE_FFMPEG=YES -DUSE_SYSTEM_FFMPEG=NO .``

hrydgard:
Seems okay, and buildbot is happy, so let's try it.

AreaScout:
@mrfixit2001 

That's wrong if you compile PPSSPP with -DUSING_EGL=OFF like plain SDL2 compile should do, EGL is done inside SDL2, the only reason we have to link to libEGL separately with this configuration is because of the use of the `eglQueryString` and `eglGetCurrentDisplay` function outside of SDL2, we are doing this to get EGL Info which was empty before this commit 19dd285 

Linking could get right without linking to libEGL if you have GLES and EGL functions inside one big library which is the case for e.g. libMali.so but it will fail if they are separated.

Regards,
Daniel

mrfixit2001:
@AreaScout thanks for the speedy reply!

I can confirm that PPSSPP runs fine with EGL=OFF and GLES2=ON, but ofc it depends on the environment and use-case. 
In my case, I'm using the mali blob, and as you said they are all one big library.

In any event - including EGL in the OPENGL_LIBRARIES definition is done later in CMakeLists when EGL is enabled. So including it along with the GLESv3/v2 value isn't necessary and thus limits the flexibility of the config options.

But... it's technically not breaking anything at this point, which is why I didn't submit  PR, just wanted to provide my insight.

AreaScout:
Well, PPSSPP runs on a lot of devices, I would just leave it as it is, if you want and SDL2 build it is wrong to turn on USING_EGL if you don't have troubles to obtain an EGL surface, initially that switch was for different render backends IIRC (was it RPi?), the flexibility is that you can choose to use EGL done inside SDL2 or from within PPSSPP with respect of those three functions

``` 
eglGetCurrentDisplay
eglQueryString
eglGetProcAddress
```

There is already a SDL2 function for `eglGetProcAddress` wich could be taken instead and the other two will be implemented in SDL3

https://wiki.libsdl.org/SDL2/SDL_EGL_GetCurrentEGLDisplay

So, in the future building PPSSPP with SDL3, EGL can be skipped entirely

RG,
Daniel

PR #11427, opened 2018-09-27
Android: Reduce the number of build variants

Florin9doi:


hrydgard:
huh! Had no idea you could do that. This makes sense.

PR #11449, opened 2018-10-06
Remove duplicate library; use the existing one

Florin9doi:


unknownbrackets (approved):
Yay, less pollution of the main directory.

-[Unknown]

PR #11450, opened 2018-10-06
Use androidgitversion version 0.4.5

Florin9doi:


hrydgard:
Any particular improvements? I don't mind upgrading though.

Florin9doi:
Nothing worth mentioning

PR #11488, opened 2018-10-23
Windows: Option to show offset scale in memory view

dl471:
I've been messing around in the memory view a lot recently so I ended up trying to make a few things to make things a bit more in line with what I'm personally used to where I couldn't find any equivalents. One of the modifications was adding the option to have offsets be drawn at the top of the hex view. This is what it looks like.

withoffsets_cropped

I made space for it by changing what lines are drawn on the fly since it seemed the most straightforward compared to messing around with multiple DC/bitmaps. I also added a checkbox to toggle it on and off to the memory view window, though said option does not affect the memory view embedded in the diassembler window. It seemed like it might have been too cramped to have it in the disassembler and having it be controlled by a checkbox in a different window seemed like it would be a bit strange.  (That said, the screenshot provided includes it turned on in both as a reference). I tried following along with the existing code style, though I'm not sure if I got it entirely right.

Finally, the memory view doesn't seem to have changed in a while. Would there be interest in merging things relating to it at all?


Windows/ppsspp.rc
```diff
@@ -304,6 +304,7 @@ BEGIN
     CONTROL         "Normal",IDC_MODENORMAL,"Button",BS_AUTORADIOBUTTON | WS_GROUP,248,9,40,9
     CONTROL         "Symbols",IDC_MODESYMBOLS,"Button",BS_AUTORADIOBUTTON,291,9,43,8
     GROUPBOX        "Mode",IDC_STATIC,241,0,104,22
+	AUTOCHECKBOX  "Show Offsets",IDC_SHOWOFFSETS,365,9,55,8
```
hrydgard: Wrong type of tabs

hrydgard:
Looks good to me, will merge after 1.7.0 is out, which should be fairly soon.

Is there even a reason to have the checkbox? Those offsets don't take up a lot of space.

Windows/ppsspp.rc
dl471: Oops, I missed that. I'll change it to four spaces like the rest.

dl471:
Thanks, that's good to hear.

The checkbox was primarily a user choice thing just in case people didn't like it and wanted to turn it off. I can remove it if you feel it's pointless.

unknownbrackets (approved):
I think it makes sense to always show, but I don't really mind the checkbox either.  LGTM too.

FWIW, memory view hasn't been implemented in the cross platform / remote UI yet, unknownbrackets/ppsspp-debugger#1.

-[Unknown]

PR #11504, opened 2018-10-28
Make more lines translatable in credits

vnctdj:
Fixes https://github.com/hrydgard/ppsspp/issues/11499

I'm not sure my code is pretty though, I'm a novice in C++ ;)


hrydgard:
Yeah, why not.

PR #11505, opened 2018-10-28
Add a second FrameSkip option - one for setting a specific # of frames to skip and one for setting a % of frames to skip

mrfixit2001:
I've only been able to test this on my linux build for ARM 32-bit, but it works well. Might need some tweaking for other environments.

### DIFF
--- a/Core/Config.cpp
+++ b/Core/Config.cpp
@@ -621,6 +621,7 @@ static ConfigSetting graphicsSettings[] = {
 	ReportedConfigSetting("AndroidHwScale", &g_Config.iAndroidHwScale, &DefaultAndroidHwScale),
 	ReportedConfigSetting("HighQualityDepth", &g_Config.bHighQualityDepth, true, true, true),
 	ReportedConfigSetting("FrameSkip", &g_Config.iFrameSkip, 0, true, true),
+	ReportedConfigSetting("FrameSkip2", &g_Config.iFrameSkip2, 0, true, true),
 	ReportedConfigSetting("AutoFrameSkip", &g_Config.bAutoFrameSkip, false, true, true),
 	ConfigSetting("FrameRate", &g_Config.iFpsLimit1, 0, true, true),
 	ConfigSetting("FrameRate2", &g_Config.iFpsLimit2, -1, true, true),
--- a/Core/Config.h
+++ b/Core/Config.h
@@ -143,6 +143,7 @@ public:
 	bool bSustainedPerformanceMode;  // Android: Slows clocks down to avoid overheating/speed fluctuations.
 	bool bVSync;
 	int iFrameSkip;
+	int iFrameSkip2;
 	bool bAutoFrameSkip;
 	bool bFrameSkipUnthrottle;
 
--- a/Core/HLE/sceDisplay.cpp
+++ b/Core/HLE/sceDisplay.cpp
@@ -554,7 +554,7 @@ static void DoFrameTiming(bool &throttle, bool &skipFrame, float timestep) {
 	bool useAutoFrameskip = g_Config.bAutoFrameSkip && g_Config.iRenderingMode != FB_NON_BUFFERED_MODE;
 	bool forceFrameskip = (fpsLimiter == FPSLimit::CUSTOM1 && g_Config.iFpsLimit1 > 60) || (fpsLimiter == FPSLimit::CUSTOM2 && g_Config.iFpsLimit2 > 60);
 	int frameSkipNum;
-	if (flips < 50) { frameSkipNum = g_Config.iFrameSkip; } else { frameSkipNum = g_Config.iFrameSkip2; }
+	if (g_Config.iFrameSkip2 > 0) { frameSkipNum = ceil( flips * (static_cast<double>(g_Config.iFrameSkip2) / 100.00) ); } else { frameSkipNum = g_Config.iFrameSkip; }
 	if (g_Config.bAutoFrameSkip || forceFrameskip) {
 		// autoframeskip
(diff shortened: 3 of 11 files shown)


owengque:
working on android??

owengque:
please build for android to.

mrfixit2001:
I think I touched every instance of frame skip code throughout the repo, so I'd say test it on android and let me know if it doesn't work :)

owengque:
I have added .diff at the end of the url but only the text appears.
How do I make it into an APK?
please teach me. I beg for help

owengque:
how do i patch to APK?

mrfixit2001:
You will need to compile it yourswlf

LunaMoo:
Lacks description, but overall I don't like it, if all this is for is doubling the frameskip automatically when fps is higher than usual it could just be a stupid checkbox to do that through in practice, auto-frameskip should already be doing same thing. There's no need to create two frameskip options and renaming standard frameskip to xxfps frameskip has no sense.

owengque:
I want to try it but I don't know. please teach me how to patch to apk or I ask you to send me your patch modification APK.
I really want to try it.
please.

mrfixit2001:
It's not to "double the frame skip", its to allow customization of a specific frame skip to apply depending of the game's current fps. Doesn't have to be double.

mrfixit2001:
@owengque - I'm on a Linux platform, you'll need to google how to compile APK from source.

mrfixit2001:
:) read the code, it would still work because it checks < 50fps for the first setting, not hard set at 30. If the preference is to update the setting descriptions to that then that's an easy change.

mrfixit2001:
Should technically update dynamically for the current fps as well every time the function is called

mrfixit2001:
But you do raise a good point. I wonder if the option to identify the fps limit would be wise. Make that 50 a variable the user can set... Or maybe I should scrap the whole idea and offer an alternative "frame skip %" setting instead. Something to think about for sure.

hrydgard:
Why have you separated it into individual commits for each file?

mrfixit2001:
It's just how GitHub handled it. I committed from the web interface because I didn't have access to my build environment at the time and couldn't do a proper push.

mrfixit2001:
Ok, the PR has been updated to implement a new setting which calculates how many frames to skip based off the set percentage of the current FPS in the game.

A few things... 
- Again, I've only tested this on Linux. Windows, QT, and libretro changes are my best guesses at making it work.
- One thing I don't know the best way to accomplish is how to tell the menu to auto-set frame skip number to zero if frame skip percent is set, and vice versa. Works fine without this, but it would be a good idea to implement this to avoid any confusion to the user. Currently the code will prioritize a frame skip % setting over the number, so if they set both then the percent is used.
- Since this PR started as one request and turned into another, let me know if you need me to close and resubmit as a new PR.

hrydgard:
Setting a % of frames to skip is an interesting idea, even though it never will yields a smooth framerate it might still look better to some people to have the framerate maybe at 40 but inconsistent.  At least gives you more steps to choose from than just 60, 30, 20, 15 etc.

Calling it Frameskip2 isn't great though..  FrameskipPercentage? Not sure

mrfixit2001:
Good point on the variable naming. I was being lazy :) Let me know if
there's any other tweaks you'd like me to make before adjusting so I can
get them all done at one time. Can someone test it on other platforms as
well?

I can confirm that, on my ARM Linux build anyway, a setting of 3% results
in almost all my ROMs playing at between 85% and 99% fps (when I enable
option 3 in the showfps to monitor it), regardless of the game's native fps.

unknownbrackets (changes requested):
Agreed on the naming.  I'm slightly worried that some users may be confused by just the `#` and `%` difference.  If we're changing them anyway, perhaps:

 * Number of frames to skip
 * Percent of frames to skip

Not sure if that's too long, though.  I've thought before actually about changing the label to "Max ..." when auto frameskip is checked, just so people understand what it does.

It's probably also not clear that when the % skip is set, the # skip is ignored.  I think we need to make this clear in the options somehow.  Perhaps # becomes disabled if % is set, or the 0 option should have a different label like "Use Percent" and automatically clear % if changed?

Or maybe it makes sense to combine the auto option with the mode, e.g.:

Frame skipping mode (`g_Config.iFrameSkipMode`)
  * Fixed count
  * Max count
  * Fixed percent
  * Max percent

-[Unknown]

unknownbrackets:
Oh, and I'm normally opposed to squashing PRs but this one probably needs it, heh.

-[Unknown]

mrfixit2001:
Agreed on the changes. I share your concern that users may be confused on how it works. How about this as an alternative, the new setting becomes "frame skip type" with the options of "number of frames" and "percent of current fps". The number is still set in the first setting, and the second determines how that number is used. Let me know what you think before I make any changes. When we all agree on the best approach I'll close this PR and resubmit ("squash" it haha)


On Oct 30, 2018, at 10:12 PM, Unknown W. Brackets <notifications@github.com> wrote:

Oh, and I'm normally opposed to squashing PRs but this one probably needs it, heh.

-[Unknown]

leopard20:
I'm a bit confused. Does it do what I requested in this issue https://github.com/hrydgard/ppsspp/issues/11372 or is it something else that it's intended for?

mrfixit2001:
@leopard20 - no, your request is to only enable frameskip if the fps % drops below 80%, this setting allows for a dynamic frameskip setting that's calculated from the current fps. For example... if set to 3% then a game running at 30fps would skip 1 frame, and a game running at 60fps would skip 2 frames.

PR #11537, opened 2018-11-06
CMake: Allow disabling Wayland support with USE_WAYLAND_WSI

### LINKED ISSUE #11536: CMake: USE_WAYLAND_WSI doesn't allow disabling Wayland support
### What happens?

Building PPSSPPSDL on Linux with Wayland installed but `USE_WAYLAND_WSI=OFF` still links against Wayland (if installed on the system). The option seems to be only used to force check that wayland support will be enabled here: https://github.com/hrydgard/ppsspp/blob/f81dd83aa053316f1be7b12b627cf528b4cdc801/CMakeLists.txt#L149-L151

### What *should* happen?

I'd want the option to define whether Wayland should be used or not, and not simply whether its absence is a fatal error.

As such I'd propose something like the following diff. Opening an issue for feedback instead of a PR for now, as the current behaviour might be intended - I'm just suggesting to change it.
```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 4c15cba0d..9129a6659 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -134,7 +134,7 @@ option(USE_FFMPEG "Build with FFMPEG support" ${USE_FFMPEG})
 option(USE_SYSTEM_SNAPPY "Dynamically link against system snappy" ${USE_SYSTEM_SNAPPY})
 option(USE_SYSTEM_FFMPEG "Dynamically link against system FFMPEG" ${USE_SYSTEM_FFMPEG})
 option(USE_SYSTEM_LIBZIP "Dynamically link against system libzip" ${USE_SYSTEM_LIBZIP})
-option(USE_WAYLAND_WSI "Set to ON to require Wayland support for Vulkan" ${USE_WAYLAND_WSI})
+option(USE_WAYLAND_WSI "Set to ON to require Wayland support for Vulkan" ON)
 option(USE_ADDRESS_SANITIZER "Use Clang memory sanitizer" ${USE_ADDRESS_SANITIZER})
 
 if(UNIX AND NOT (APPLE OR ANDROID) AND VULKAN)
@@ -145,13 +145,14 @@ if(UNIX AND NOT (APPLE OR ANDROID) AND VULKAN)
 		message("NOT using X11 for Vulkan")
 	endif()
 	# add_definitions(-DVK_USE_PLATFORM_XCB_KHR)
-	find_package(Wayland)
-	if(USE_WAYLAND_WSI AND NOT WAYLAND_FOUND)
-		message(FATAL_ERROR "Could not find libwayland, but USE_WAYLAND_WSI was enabled. Failing.")
-	endif()
-	if(WAYLAND_FOUND)
-		include_directories(${WAYLAND_INCLUDE_DIR})
-		add_definitions(-DVK_USE_PLATFORM_WAYLAND_KHR)
+	if(USE_WAYLAND_WSI)
+		find_package(Wayland)
+		if(NOT WAYLAND_FOUND)
+			message(FATAL_ERROR "Could not find libwayland, but USE_WAYLAND_WSI was enabled. Failing.")
+		else()
+			include_directories(${WAYLAND_INCLUDE_DIR})
+			add_definitions(-DVK_USE_PLATFORM_WAYLAND_KHR)
+		endif()
 	endif()
 endif()
 
 ```

Mageia 7 x86_64 (Linux), PPSSPP 1.7.2/master.

### PR DESCRIPTION
**Pull request by akien-mga**
Closes #11536.

### CODE REVIEW

File: CMakeLists.txt [hunk 1]
```diff
@@ -134,7 +134,7 @@ option(USE_FFMPEG "Build with FFMPEG support" ${USE_FFMPEG})
 option(USE_SYSTEM_SNAPPY "Dynamically link against system snappy" ${USE_SYSTEM_SNAPPY})
 option(USE_SYSTEM_FFMPEG "Dynamically link against system FFMPEG" ${USE_SYSTEM_FFMPEG})
 option(USE_SYSTEM_LIBZIP "Dynamically link against system libzip" ${USE_SYSTEM_LIBZIP})
-option(USE_WAYLAND_WSI "Set to ON to require Wayland support for Vulkan" ${USE_WAYLAND_WSI})
+option(USE_WAYLAND_WSI "Set to ON to require Wayland support for Vulkan" ON)
```
hrydgard: Was this change intended?

akien-mga: I wish CMake had a "on/off/auto" switch like autotools (and it's not often that I wish for autotools-like things :P).

hrydgard:
Alright, I think this should be fine.

Let's wait for the buildbot to go green before merge.

akien-mga:
I'll just tweak the hint string for `USE_WAYLAND_WSI` then to be more explicit.

akien-mga:
Yeah CI fails, and any build against older SDL2 or without Wayland libraries would fail too, that's probably not a good change as is.

Alternatively I could drop the fatal error, so that the scenarios are:
- `USE_WAYLAND_WSI=ON` (default): Test for Wayland, if found, ok (like before)
- `USE_WAYLAND_WSI=ON` (default): Test for Wayland, if not found, continue silently without Wayland support (like before without defining `USE_WAYLAND_WSI=ON`)
- `USE_WAYLAND_WSI=OFF`: Don't test for Wayland (new)

akien-mga:
BTW should I move the option up near `USING_X11_VULKAN` in the `# :: Environments` section?

akien-mga:
I've pushed an updated version which does what I describe in https://github.com/hrydgard/ppsspp/pull/11537#issuecomment-436207027.

hrydgard:
Seems ok.

PR #11545, opened 2018-11-07
F-droid lite version - create stubs for Moga Controller

Florin9doi:


hrydgard:
I'm honestly thinking about just ripping out the Moga support. It's not clear that they're very popular and their support library I believe is still stuck on Android 16 or something and requires ugly workarounds.

PR #11590, opened 2018-11-25
Attempt to fix glsl depal error reported in #11588 (ATI/AMD)

hrydgard:


unknownbrackets (approved):
Seems okay, but worried about what the deeper problem is here.

-[Unknown]

PR #11632, opened 2018-12-03
SDL: Allow toggling fullscreen for GLES2 on desktops.

orbea:
Fixes https://github.com/hrydgard/ppsspp/issues/11627

Note: This works for linux, but I am not sure what other platforms rely on the old behavior other than android and IOS which should be defined under `MOBILE_DEVICE`.


hrydgard:
Android doesn't use this code, only I can think of is blackberry, which is deprecated.

Issue #11629, opened 2018-12-03
(cmake) Configure warnings


orbea:
### What happens?

Configuring ppsspp with `cmake-3.13.1` prints several warnings.
```
CMake Warning (dev) at CMakeLists.txt:117 (option):
  Policy CMP0077 is not set: option() honors normal variables.  Run "cmake
  --help-policy CMP0077" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.
 
  For compatibility with older versions of CMake, option is clearing the
  normal variable 'X86'.
This warning is for project developers.  Use -Wno-dev to suppress it.
 
CMake Warning (dev) at CMakeLists.txt:118 (option):
  Policy CMP0077 is not set: option() honors normal variables.  Run "cmake
  --help-policy CMP0077" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.
 
  For compatibility with older versions of CMake, option is clearing the
  normal variable 'X86_64'.
This warning is for project developers.  Use -Wno-dev to suppress it.
 
CMake Warning (dev) at CMakeLists.txt:129 (option):
  Policy CMP0077 is not set: option() honors normal variables.  Run "cmake
  --help-policy CMP0077" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.
 
  For compatibility with older versions of CMake, option is clearing the
  normal variable 'HEADLESS'.
This warning is for project developers.  Use -Wno-dev to suppress it.
 
CMake Warning (dev) at CMakeLists.txt:134 (option):
  Policy CMP0077 is not set: option() honors normal variables.  Run "cmake
  --help-policy CMP0077" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.
 
  For compatibility with older versions of CMake, option is clearing the
  normal variable 'USE_FFMPEG'.
This warning is for project developers.  Use -Wno-dev to suppress it.
```
Full cmake log - https://pastebin.com/19Bqt15M

### What *should* happen?

Configuring ppsspp with cmake should not print warnings.

### What hardware, operating system, and PPSSPP version?  On desktop, GPU matters for graphical issues.

OS: `Slackware64-current`
Linux: `4.19.5`
ppsspp: https://github.com/hrydgard/ppsspp/commit/95ffa15a842730e8ddc467d28d50d7995e258c13
cmake: `3.13.1`

orbea:
[...]

PR #11633, opened 2018-12-04
cmake: Fix CMP0077 policy warnings.

orbea:
Fixes https://github.com/hrydgard/ppsspp/issues/11629

This does a few things.

1. It uses `set()` instead of `option()` to avoid the new warnings with `option()`. It seems the way to avoid the warnings would be to change the variable names which would break build script compatibility, set the policy which may break in the future or use `set()` instead.
2. This required ensuring that undefined variables are set so that contents of `CMakeCache.txt` do not differ before and after.
3. Some of the previous checks and defaults no longer make sense and can just be handled with `set()` so they were removed.
4. Fixes some misleading indentation.

The before and after behavior should be the same with this, but please review this!

There should also be no difference between the old and new `CMakeCache.txt` files.


orbea:
I added a second commit which removes the `RASPBIAN` define, as best as I can tell its unused.

I also force pushed a few times to try to fix issues found by travis...

orbea:
I think I am misunderstanding the issue, I will close this and open an alternative later. Sorry for the noise.

PR #11650, opened 2018-12-09
Attempt to improve Discord Rich Presence

Abbanon:


hrydgard:
Oh right, forgot about this one :)

Not sure what the best text is, but this will do for now.

PR #11652, opened 2018-12-10
Update UWP port

pent0:
Update some missing dependencies and changes to the UWP port.
Tested on my Lumia 525 running Windows 10 Mobile 10.15063

image

image


Core/MIPS/ARM/ArmJit.cpp
```diff
@@ -46,7 +46,7 @@
 
 using namespace ArmJitConstants;
 
-static void DisassembleArm(const u8 *data, int size) {
+void DisassembleArm(const u8 *data, int size) {
```
pent0: MSVC keeps telling me this is an unresolved external symbol although It's not, so I remove the static and it works.
Not sure why

hrydgard:
Cool, will look at this soon.

hrydgard:
No objections here, this all seems fine. Thanks!

PR #11682, opened 2018-12-19
GPU driver test, first part: Discard+stencil and Discard+depth

hrydgard:
This test verifies that "fragment discard" affects Z and Stencil operations as well as Color - which some Adreno generations seems to have problems with.

It can be accessed from Settings/Tools/Developer Tools/GPU Driver Test.

The output should look like this:

gpudrivertest

If it fails, the text will probably be invisible covered in green or red.

@unknownbrackets  can you see if this fails as expected on your Adreno hardware?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -850,6 +850,7 @@ list(APPEND NativeAppSource
 	UI/PauseScreen.cpp
 	UI/GameScreen.cpp
 	UI/GameSettingsScreen.cpp
+	UI/GPUDriverTestScreen.cpp
 	UI/TiltAnalogSettingsScreen.cpp
 	UI/TiltEventProcessor.cpp
 	UI/TouchControlLayoutScreen.cpp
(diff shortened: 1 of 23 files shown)


hrydgard:
I'm gonna go ahead and merge this so people can try it, can always add more tests later.

unknownbrackets:
Currently, the tests as-is pass.

When depth write is disabled (`depthWriteEnabled = false`;), "TEST OK" is just two boxes (obviously wrong) on the left two, while they continue to work on NVIDIA.  Obviously the third test does not work with depth write off.

The same occurs with depth test disabled on depthStencilWrite (but depth write enabled on all.)  That probably most directly represents the original bug.

I'm pretty sure it was happening even with depth testing on, but I might have not tried enough scenarios.  Would need to dig into it further to reproduce, but short on time until the weekend.

-[Unknown]

hrydgard:
If the depth test is disabled, depth write has no effect. but yeah, we still write to the stencil buffer, and that doesn't work then for Adreno. I see. 

Alright I think I have what I need to create a repro-ing test now. You don't need to dig further, only try my next tests which I'll add later today :)

Emulatorer:
Something wrong in UI of GPU Driver Test depth and back are close each other
### Screenshot
screenshot_2018-12-20-21-09-36

PR #11684, opened 2018-12-20
GPU driver test: This should finally repro one of the Adreno bugs

hrydgard:
... hopefully. See comment in https://github.com/hrydgard/ppsspp/pull/11682 .

### DIFF
--- a/UI/GPUDriverTestScreen.cpp
+++ b/UI/GPUDriverTestScreen.cpp
@@ -47,6 +47,10 @@ GPUDriverTestScreen::GPUDriverTestScreen() {
 GPUDriverTestScreen::~GPUDriverTestScreen() {
 	if (discardWriteDepthStencil_)
 		discardWriteDepthStencil_->Release();
+	if (discardWriteDepth_)
+		discardWriteDepth_->Release();
+	if (discardWriteStencil_)
+		discardWriteStencil_->Release();
 	if (drawTestStencilEqualDepthAlways_)
 		drawTestStencilEqualDepthAlways_->Release();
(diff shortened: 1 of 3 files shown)


hrydgard:
The issue: If using discard in the fragment shader of a pipeline that writes to stencil, the stencil result will be written to all pixels, not just the non-discarded ones.

How to test:

Get a recent build of PPSSPP, for example from https://buildbot.orphis.net/ppsspp/ , switch to the Vulkan backend.

Go to Settings/Tools/Developer Tools/GPU Driver Test.

Buggy driver output (Adreno 500, 600 series), note the bright green boxes instead of TEST OK:

https://media.discordapp.net/attachments/478294170662862852/525357246155325450/vulkan-adreno-bug.png?width=1072&height=603

Non buggy driver (Adreno 300, 400 series) looks the same except there's TEST OK on all the green rectangles.

mrcmunir:
screenshot_2018-12-21-01-12-24-614_org ppsspp ppsspp
For me returned test ok on all test

hrydgard:
Interesting, maybe only 500 series is affected?

hrydgard:
Thanks. Xiaomi Mi A1 has Adreno 506, which is another data point showing that this is broken on Adreno 500 series.

Nucleoprotein:
I got update to Android Pie yesterday, but nothing changed :(

Emulatorer:
Oppo a3s android 8.0
ppsspp git build v1.7.5-336
screenshot_2018-12-30-10-24-12-68

PR #11717, opened 2019-01-05
Added a check to the pointer for memcpy

z2442:
This check enables the newer builds of daedalus to run.

### DIFF
--- a/Core/HLE/sceKernelInterrupt.cpp
+++ b/Core/HLE/sceKernelInterrupt.cpp
@@ -684,7 +684,9 @@ const HLEFunction Kernel_Library[] =
 
 static u32 sysclib_memcpy(u32 dst, u32 src, u32 size) {
 	ERROR_LOG(SCEKERNEL, "Untested sysclib_memcpy(dest=%08x, src=%08x, size=%i)", dst, src, size);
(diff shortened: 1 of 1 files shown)


hrydgard:
Rather strange that it would pass in invalid parameters to memcpy, no? But I guess if it runs on the real PSP...

hrydgard:
Actually if it's a copy to ME space or something, that could be alright. Didn't think that was mapped in user space though but maybe the target applications runs in kernel mode?

mygizli04:
Is this windows only or something? I still get an insta-crash. Here's the crash report: https://pastebin.com/ecxrAQr1

z2442:
You have to remove the media engine.prx from Daedalus and disable fast memory then it is able to boot into the ui. Also yes Daedalus does run in kernel mode I’m pretty sure since it needs to unload and load a bunch of modules.

mygizli04:
Now i only get a black screen.

z2442:
What is your psp model set to should be 2000/3000. Also are you using Daedalus from my repo?

mygizli04:
It's 2000-3000 and yes, it's from your repo ( i don't know how to build it so i downloaded the [latest](https://github.com/z2442/daedalus/releases) prerelease )

PR #11718, opened 2019-01-06
[iOS Port] Making the Launch Screen not look like crap on newer devices

mygizli04:
Hello there, it's me again. Well the last one i made looked like crap on newer devices and took up some space ( well KB's of space is still space ). I couldn't find ANYTHING about fonts or logo images on the project that was being generated, so i simply reimported them which has no downsides anyway.


CMakeLists.txt
```diff
@@ -1927,7 +1927,9 @@ set(NativeAssets
 	assets/rargray.png
 	assets/unknown.png
 	assets/zip.png
-	assets/logo.png
+	assets/Roboto-Condensed.ttf
```
hrydgard: Roboto-Condensed.ttf is already listed here on line 1921.

hrydgard:
Apart from the above minor comment, this seems fine.

PR #11751, opened 2019-01-31
Update Vulkan headers

hrydgard:
Also disables our current bad use of VK_KHR_dedicated_allocation. Will fix that properly later.

Should not really need review.


hrydgard:
Seems to only be random timeout failures in the CI check. Merging.

PR #11773, opened 2019-02-06
D3D9: Disable range culling on really old NVIDIA cards

hrydgard:
See #11638. Should fix it.


unknownbrackets (approved):
I guess this makes sense, bit of a shame about the deviceID detection...

-[Unknown]

PR #11777, opened 2019-02-08
Gate fewer effects behind "Disable slow framebuffer effects".

hrydgard:
Similar to #11776 but for Disable Slow Framebuffer Effects.

Really, these two settings should probably be combined, and only affect effects that need readbacks.


hrydgard:
Makes a little bit of progress on #11610 .  

The idea of doing this step by step is to get reports about bad performance regressions so we can fix them.

unknownbrackets (approved):

PR #11826, opened 2019-02-23
ARM/ARM64: Ignore invalid immediate addresses in delay slots, which may be conditional.

hrydgard:
Should do something in x86's safe mem too, but leaving for later.

Replaces #11824


vit9696:
Nice catch. I can confirm that this change also fixes the issue.

I also confirm that _dbg_assert_msg_ is disabled in Release builds on iOS, so it looks good to me.

PR #11831, opened 2019-02-23
Bundle libSDL inside app on macOS, fixes #11830

### LINKED ISSUE #11830: Bundling SDL within PPSSPP on macOS
@hrydgard, last time I grumbled about it, @angelXwind said it was your desire not to bundle libSDL within PPSSPP builds on macOS. I think there was some misunderstanding from the sides here, and request it to be changed.

For instance, I do not use HomeBrew, and constantly having to fix provided binaries to locate the right lib (/usr/local/lib → /opt/local/lib) makes me unhappy. Putting aside having to install something like libSDL at all. I believe PPSSPP should launch on stock macOS without this mess.

Something as simple as the script below puts libSDL binary inside app Frameworks and makes the resulting build portable. I admit that Qt framework, which you also use, does exactly the same thing via `macdeployqt`, which is slightly more advanced.

```bash
SDL=$(otool -L PPSSPPSDL.app/Contents/MacOS/PPSSPPSDL | grep -Eo /.+libSDL.+dylib)
if [ "$SDL" != "" ]; then
  if [ -f "$SDL" ]; then
    SDLNAME=$(basename "$SDL")
    mkdir -p PPSSPPSDL.app/Contents/Frameworks || exit 1
    cp -r "$SDL" PPSSPPSDL.app/Contents/Frameworks || exit 1
    install_name_tool -change "$SDL" "@executable_path/../Frameworks/$SDLNAME" PPSSPPSDL.app/Contents/MacOS/PPSSPPSDL || exit 1
  fi
fi
```

I think we should include its invocation in cmake, and given that you use Travis, upload macOS builds to https://buildbot.orphis.net/ppsspp/index.php and/or some other place. I kind of hope to see release builds instead of snapshots too.

vit9696:


unknownbrackets (approved):
Makes sense to me, we bundle redists on Windows too.  libzip isn't a problem, right (page says to install libzip too)?

-[Unknown]

vit9696:
Static one is used in my case, as I did not install it and did not provide USE_SYSTEM_LIBZIP=1.

PR #11879, opened 2019-03-12
Fix apparent bug in #11094, fixes #11878 and likely #11798

### LINKED ISSUE #11878: Sonic Rivals controls semi-broken
Holding down digital buttons doesn't work. Can still play somewhat using analog stick. Seems similar to #11798.

hrydgard:
@unknownbrackets  please take a look :)


unknownbrackets (approved):
This must be right, but it's strange that it still made the test pass... that said, now I'm not seeing a test fail even with the change reverted, and I didn't note which test it was.

-[Unknown]

PR #11885, opened 2019-03-14
Correctly handle memory types in VulkanPushBuffer. Should help #11882.

hrydgard:
Can't cache memorytype, need to ask it anew for every allocation....

See #11882


hrydgard:
This one I want in any 1.8 point updates, so I'm merging.

PR #11938, opened 2019-03-26
Add texture dump/replace hotkeys.

### LINKED ISSUE #11734: Easier access for Replace/Save new textures
In regards to future texture modders, I think it would be a good idea to give them easier access to these two buttons, as they will be visited, a lot, specially if it's a huge game, or one with duplicate textures popping each second.

There is the DevMenu, yes. But while I tried to use it, I found out that even after checking the boxes, nothing is affected until you pause and go to settings, so it's not viable. A suggestion for a perfect place for those would be...

image

A secret shortcut for that would also be appreciated too.

LunaMoo:
Should fix #11734
althrough did no menu changes, only added hotkeys that aren't set by default with OSM and clearing gpu cache.


hrydgard:
Yet more strings, not very critical ones though...

PR #11942, opened 2019-03-27
Fixes iOS exit freezing, Selective PTRACE syscall

jeeeyul:
See details on #11905


unknownbrackets (approved):
Seems reasonable to me.  We prefer cases in switches aligned to the switch, but I'm not too concerned with applying that convention to mm files.

-[Unknown]

PR #11961, opened 2019-04-05
SaveState: Reduce savestate size using zstd.

xebra:
Zstandard(zstd) is a better fast-compression library than Snappy in compression ratio.
https://github.com/facebook/zstd

## Benchmarks

| Compressor name         | Ratio | Compression| Decompress.|
| ---------------         | ------| -----------| ---------- |
| **zstd 1.3.4 -1**       | 2.877 |   470 MB/s |  1380 MB/s |
| zlib 1.2.11 -1          | 2.743 |   110 MB/s |   400 MB/s |
| brotli 1.0.2 -0         | 2.701 |   410 MB/s |   430 MB/s |
| quicklz 1.5.0 -1        | 2.238 |   550 MB/s |   710 MB/s |
| lzo1x 2.09 -1           | 2.108 |   650 MB/s |   830 MB/s |
| lz4 1.8.1               | 2.101 |   750 MB/s |  3700 MB/s |
| snappy 1.1.4            | 2.091 |   530 MB/s |  1800 MB/s |
| lzf 3.6 -1              | 2.077 |   400 MB/s |   860 MB/s |

## Author
Yann Collet (lz4, xxhash, etc.)

## Version
1.3.8

## License
Zstandard is dual-licensed under BSD and GPLv2.


hrydgard:
Maybe generally, but could you present some numbers (compression ratio, speed I trust) on PSP savestates?

LunaMoo:
Aren't rewind snapshots compressed? How does that affect them? Afaik even on powerful desktop it causes a slight frame drop/stutter when state is created making the whole feature really annoying to use, if it's affected by compression, from that list something like lz4 1.8.1 with just slightly higher compression, but much better compression and decompression speed would be better assuming that data would even be real for psp games savestates.

xebra:
Just with my feeling, the speed is almost same with Zstandard and Snappy, and compression ratio reduces average 40%(20-120%). I will present actual comparison data later.
Also, I will try to reduce the code size.


> Aren't rewind snapshots compressed? How does that affect them?

Sorry, what do you mean? What is rewind snapshot?

unknownbrackets:
Could we use zstd as a submodule rather than copying it in?  Also, would need to be added to cmake/etc.

Rewind snapshots aren't exactly compressed - it keeps only different blocks but it doesn't use snappy or zstd (snappy was slower than just keeping the changed blocks.)  Rewind basically creates a save state every X frames.

-[Unknown]

xebra:
I will learn how to add zstd as submodule, and how to write cmake script.
Also, I understood what is the 'rewind snapshots', I didn't know this feature.

xebra:
Results of the test code.
名称未定-1

Some comparison test results.
zstd compression level.
1: Fastest
3: Default
9: Highest compression ratio
名称未定-1_
名称未定-2_
名称未定-3_
名称未定-4_
名称未定-5_
名称未定-6_
名称未定-7_
名称未定-8_

unknownbrackets:
Any luck with the submodule?

-[Unknown]

PR #12012, opened 2019-05-02
Enable PNG screenshots for all platforms

vnctdj:
Following https://github.com/hrydgard/ppsspp/pull/12009 :)

As a side note, what does the following line mean in my build log?
`-- Could NOT find PNG (missing: PNG_LIBRARY PNG_PNG_INCLUDE_DIR) 
`
I thought it could prevent me from using PNG in PPSSPP, but that's definitely not the case!


hrydgard:
That's looking for a system-installed png library, and if it's not there, we'll use the one shipped with PPSSPP.

vnctdj:
OK, thanks!

PR #12028, opened 2019-05-11
Add CPU string in Windows ARM

driver1998:
Reads processor name on Windows ARM, supports both UWP and Desktop.


Gamemulatorer:
Just OT 😅 PPSSPP doesn't detect my android cpu name on device info > system information > cpu information
IMG_20210701_030827
Using cpu-z my android cpu name
IMG_20210701_025459

PR #12032, opened 2019-05-12
POC: D3DCompiler on D3D9

driver1998:
This enable D3DCompiler_47 on D3D9, side-by-side with D3DX.

Mainly for ARM, but it also means DirectX runtime updates is no longer needed for x86/64 (we ship D3DCompiler_47.dll with PPSSPP anyway)

define PPSSPP_API_D3D9_D3DCOMPILER = 1 for D3DCompiler, and PPSSPP_API_D3DX9 for D3DX9.

### DIFF
--- a/UI/DevScreens.cpp
+++ b/UI/DevScreens.cpp
@@ -47,7 +47,11 @@
 #ifdef _WIN32
 #include "Common/CommonWindows.h"
 // Want to avoid including the full header here as it includes d3dx.h
+#if PPSSPP_API(D3DX9)
 int GetD3DXVersion();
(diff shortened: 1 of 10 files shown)


hrydgard:
I'd love to get rid of D3D9X altogether. Are there any drawbacks to this, like only working on newer versions of Windows or something? 

We are already dropping XP support in 1.9.0.

driver1998:
I haven't notice any, although I only have two machines to test on and maybe 5 games to test. ;)

hrydgard:
Let's do this. If we don't get any issues reported, I'll kill off the D3D9X path.

PR #12052, opened 2019-05-19
VS2017: Add hack to automatically select the newest installed Windows 10 SDK

hrydgard:
VS2019 already does this by default if you specify 10.0.

Should hopefully fix the buildbot.

Had to move the property sheet imports upwards in the project files because otherwise it wouldn't find headers, for some reason. I hope this won't have any other bad effects...


hrydgard:
We don't have any CI that checks Windows at the moment except for the broken buildbot that only builds master, so I'll just merge this.

LunaMoo:
Seems the buildbot is still broken, maybe it's something else like the config order causing it to try and build win arm/arm64 versions since it got broken starting from the arm windows merge which was days before the SDK change reverted back to 10.0(so it was broken while still using 8.1).

hrydgard:
Darn, I really thought this would take care of it. But getting this bizarre error:

    LINK error LNK1158: cannot run 'rc.exe'

teamcity


@Orphis , could you take a look? Maybe simply reinstall the latest Windows 10 SDK?

LunaMoo:
Maybe it's the new version set in
https://github.com/hrydgard/ppsspp/commit/b22b54b636712f914697541b700de3b75c53c10b#diff-a1fc988beeb37b5a72a3b99f9537d408L2 I saw in the history you changed that once to older for CI and it was updated with the arm win merge.

unknownbrackets:
I think it's working now.

-[Unknown]

PR #12070, opened 2019-05-30
Improvements to WordWrapper

CookiePLMonster:
This PR started as a change to word wrapping (basically 1.), but when rebasing and cleaning up for submission I realized more can be done. Therefore, this PR introduces several styling changes and introduces a functionality change to `WordWrapper` class:

- When wrapping before word, left trim whitespaces to prevent newlines from starting indented
- Changes `lastLineStart_` to `size_t` to cut down on nasty typecasts
- Replaces `operator +=` with explicit call to `append` to make intent more clear and reduce memory allocations

That's how the change looks in action:

Before:
image

After:
image

Notice that a wrapped string is now left trimmed of whitespaces, so it doesn't appear indented.

When analyzing the safety of this change @unknownbrackets mentioned I could test this by feeding it a very long word which would have wrapped, and here's the result - it appears to look identical before and after the change.

Before:
image

After:
image

### DIFF
--- a/ext/native/util/text/wrap_text.cpp
+++ b/ext/native/util/text/wrap_text.cpp
@@ -74,7 +74,7 @@ std::string WordWrapper::Wrapped() {
 	return out_;
 }
 
-void WordWrapper::WrapBeforeWord() {
+bool WordWrapper::WrapBeforeWord() {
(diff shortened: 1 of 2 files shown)


hrydgard:
Thanks, this looks right.

Issue #11226, opened 2018-06-27
Analog axis keeps tiny values even when there's no input(mouse/gamepad analog)


LunaMoo:
Following the discussion in the [forum](https://forums.ppsspp.org/showthread.php?tid=24439) about mouse control not working as easily as third party one(glovePIE) only when mapping axis to PSP buttons I made https://github.com/hrydgard/ppsspp/compare/master...LunaMoo:AxisToButtonThreshold however I noticed a separate problem - if too small threshold is allowed, PPSSPP get's input even if there's none.

 It's like something doesn't clear input properly and a tiny value not always, but very commonly "sticks" to emulated analog axis:], I noticed it as well in the past while making right analog patches but I thought back then that my gamepad just is old and missbehaves so I added high deadzone by default to all my patches, but I guess it could just be some issue in PPSSPP itself.

 Steps to reproduce problem, either use right analog patch and reduce it;s deadzone to very small value or any game with the branch linked above,, threshold set to low value (ie 0.5) and mouse axis mapped to anything which is easily observable like camera movement. Possibly could also be the case with standard gamepad when setting deadzone to very small value, but I moved and not sure where my gamepads are so I currently can't test that;p. 

 In case it's platform specific I'm using windows 10 x64.

AkiraJkr:
I noticed that too, everytime I began a game, it was normal, however after moving a bit to, say, down, it began to walk on it's own to the right.

Windows 7 btw.

PR #12110, opened 2019-06-16
Fix mouse scroll for key mapping

### LINKED ISSUE #11043: Mouse scroll doesn't work when mapped via mouse control
Mouse scroll doesn't work when mapped via mouse control on windows as reported in [here](https://forums.ppsspp.org/showthread.php?tid=24311), would take a look myself, but affraid I'm not going to have much time this week and was invited to a party in the weekend;p, might forget later without issue.

LunaMoo:
Fixes #11043 (which I totally forgot about;p ~ sorries)

### DIFF
--- a/Windows/MainWindow.cpp
+++ b/Windows/MainWindow.cpp
@@ -103,8 +103,10 @@ extern ScreenManager *screenManager;
 
 #define TIMER_CURSORUPDATE 1
 #define TIMER_CURSORMOVEUPDATE 2
+#define TIMER_WHEELRELEASE 3
 #define CURSORUPDATE_INTERVAL_MS 1000
 #define CURSORUPDATE_MOVE_TIMESPAN_MS 500
+#define WHEELRELEASE_DELAY_MS 16
 
 namespace MainWindow
 {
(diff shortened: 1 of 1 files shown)


LunaMoo:
Please don't merge yet, it causes some weird wheel interaction in UI;c, I guess it needs that fix has to be limited to in-game similar to how mouse movement is.

LunaMoo:
Heh, I think it's still bad overall, going to have to think about it more as it can get stuck.
 Does WM_MOUSEWHEEL is only called when a wheel event is passed? Or maybe something related to #11226

LunaMoo:
As far as I tested this works fine now althrough may be a bit hacky. I set a variable in WM_MOUSEWHEEL event when wheel is "key_down" and check for that variable under default case to send "key_up".

 Moved WM_MOUSEWHEEL to DisplayProc as well, dunno what "some reason" was forcing it in WndProc, but works fine nowadays in DisplayProc.

 If this solution is acceptable I can squash the commits later.

LunaMoo:
Heh I squashed, but it still sucks I need to release it elsewhere, default case in DisplayProc still requires some event to trigger I guess.

LunaMoo:
Ended up using a timer using 16ms delay to ensure wheel is released on next frame even in 60 fps games. I think it's ready now.:]

hrydgard:
Yeah, I suppose that'll be alright...

PR #12116, opened 2019-06-21
Vulkan: Implement pretransform (performance optimization)

### LINKED ISSUE #12099: Possible VK mobile optimization: Pre-rotate the image.
ARM says that it can be beneficial to render the image pre-rotated. We already do this on Windows Mobile so it shouldn't be too hard to also implement for VK/Android.

Instructions are here:
https://arm-software.github.io/vulkan_best_practice_for_mobile_developers/samples/surface_rotation/surface_rotation_tutorial.html

Would expect maybe a few percent boost from something like this at most, though.

hrydgard:
Apply the pretransform that the operating system wants when drawing to the backbuffer.

This should save a lot of memory bandwidth on mobile devices that can't rotate images natively in the display engine. Fixes #12099.

Hopefully this will close the gap in the few instances where GL is currently faster (since GL drivers do this automatically).

Needs a bit of testing before merge.

Explanation for why this helps is here:
https://arm-software.github.io/vulkan_best_practice_for_mobile_developers/samples/surface_rotation/surface_rotation_tutorial.html

### DIFF
--- a/GPU/Common/FramebufferCommon.cpp
+++ b/GPU/Common/FramebufferCommon.cpp
@@ -734,7 +734,7 @@ void FramebufferManagerCommon::DrawPixels(VirtualFramebuffer *vfb, int dstX, int
 		// We are drawing to the back buffer so need to flip.
 		if (needBackBufferYSwap_)
 			std::swap(v0, v1);
-		flags = (DrawTextureFlags)(flags | DRAWTEX_TO_BACKBUFFER);
+		flags = flags | DRAWTEX_TO_BACKBUFFER;
 		float x, y, w, h;
 		CenterDisplayOutputRect(&x, &y, &w, &h, 480.0f, 272.0f, (float)pixelWidth_, (float)pixelHeight_, ROTATION_LOCKED_HORIZONTAL);
 		SetViewport2D(x, y, w, h);
@@ -809,7 +809,7 @@ void FramebufferManagerCommon::DrawFramebufferToOutput(const u8 *srcPixels, GEBu
 		std::swap(v0, v1);
 
 	DrawTextureFlags flags = g_Config.iBufFilter == SCALE_LINEAR ? DRAWTEX_LINEAR : DRAWTEX_NEAREST;
-	flags = (DrawTextureFlags)(flags | DRAWTEX_TO_BACKBUFFER);
+	flags = flags | DRAWTEX_TO_BACKBUFFER;
 	if (cardboardSettings.enabled) {
 		// Left Eye Image
 		SetViewport2D(cardboardSettings.leftEyeXPosition, cardboardSettings.screenYPosition, cardboardSettings.screenWidth, cardboardSettings.screenHeight);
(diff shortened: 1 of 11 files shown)


ext/native/thin3d/thin3d.cpp
```diff
@@ -352,6 +353,35 @@ DrawContext::~DrawContext() {
...
+		rect.x = rth - rect.h - origY;
+		rect.y = origX;
+		std::swap(rect.w, rect.h);
+		break;
+	}
+	case DisplayRotation::ROTATE_270: {
+		float origX = rect.x;
+		float origY = rect.y;
+		// TODO
+		std::swap(rect.w, rect.h);
```
hrydgard: Hm, I copy-pasted this from the D3D11 code and didn't notice this case wasn't fully implemented.

Will do that before merge.

GPU/Common/FramebufferCommon.cpp
```diff
@@ -733,6 +734,7 @@ void FramebufferManagerCommon::DrawPixels(VirtualFramebuffer *vfb, int dstX, int
 		// We are drawing to the back buffer so need to flip.
 		if (needBackBufferYSwap_)
 			std::swap(v0, v1);
+		flags = (DrawTextureFlags)(flags | DRAWTEX_TO_BACKBUFFER);
```
unknownbrackets: Maybe an `operator |` makes sense...

-[Unknown]

ext/native/math/lin/matrix4x4.h
```diff
@@ -111,6 +111,34 @@ class Matrix4x4 {
 		zz = 1.0f; 
 		ww = 1.0f;
 	}
+	// Exact angles to avoid any artifacts.
+	void setRotationZ90() {
+		empty();
+		float c = 0.0f;
+		float s = 1.0f;
+		xx = c;		xy = s;
```
unknownbrackets: nit: Prefer against tabs after first non-tab?

-[Unknown]

unknownbrackets (approved):
Cool.  Build failures still seem to be from the SPIRV cross update.

-[Unknown]

ext/native/math/lin/matrix4x4.h
hrydgard: Agreed, I'll fix that and delete some unused code.

GPU/Common/FramebufferCommon.cpp
hrydgard: Done

hrydgard:
Yeah, not quite sure how to fix that .. we shouldn't need python at all though.

unknownbrackets:
Someone reported seeing this:

broken pre-transform in Vulkan

-[Unknown]

hrydgard:
Wacky, is that new?

unknownbrackets:
This range was given: https://github.com/hrydgard/ppsspp/compare/v1.8.0-303-g12ab276a7...v1.8.0-316-g068ec3059

-[Unknown]

PR #12134, opened 2019-06-25
UWP: Fix overscan on Xbox One

CookiePLMonster:
This PR corrects overscan values when using the UWP version of PPSSPP on Xbox One. Previously PPSSPP would display with on around 80% of the screen and aligned to top left (so with black bars on right and bottom). With this fix, it displays fullscreen properly.

Windows is unaffected, and reportedly no issues on Windows Phone either.

That's how PPSSPP looks on X1 **without** this fix:
image


hrydgard:
Hm, ok. I don't remember why that code was added, but let's try without it.

driver1998:
I hope this is not the code I added to handle window resizing....  I will do some test later.

driver1998:
Well, it is not, so it should still work properly.

PR #12142, opened 2019-06-29
Misc code sanitize improvements for Windows and UWP

CookiePLMonster:
This PR brings several misc Windows-specific improvements to the codebase. Those are, per commit:

- Fixed an exit crash in `CGEDebugger::~CGEDebugger()` when attempting to reference a freed pointer. Not only it was freed, but also the operation performed there (`DestroyWindow`) is technically invalid because `CGEDebugger` never owned this window to begin with.
- Cleaned up `getFilesInDir` a little and modified it to properly detect hidden files/directories on Windows. Detecting if filename starts with dot doesn't apply there, so it's now only done for non-Windows builds.
- `xinput9_1_0.dll` will now be used if both `xinput1_4.dll` and `xinput1_3.dll` are absent - this can happen on a Windows XP/Vista/7 system without DirectX End User Runtimes. Additionally, `XInputGetState` will be used if `XInputGetStateEx` is not found - this applies to `xinput9_1_0.dll` and some XInput wrappers which do not implement this export.
- Removed an empty `KeyboardDevice` and migrated the corresponding input code to use `std::unique_ptr` instead of `std::shared_ptr`. There was no need for a `shared_ptr` with their costly ownership tracking, since ownership was never duplicated or passed around.
- Threading in `InputDevice.cpp` has been simplified to join on a thread instead of using a condition variable to signal thread termination. Thread termination has also been changed to use an `atomic_flag`, as technically writing and reading to the same memory at the same time is undefined behaviour (even with a `volatile` keyword).


CookiePLMonster:
PR is now ready for review.

hrydgard:
Yeah, I'm fine with all of this. Merge time.

PR #12173, opened 2019-07-14
Ignore mapped mouse input for UI

LunaMoo:
This is a workaround for a problem which will happen if user maps mouse actions like for example mouse wheel into psp buttons that does something else in the UI, especially cross or circle.

 Might help #12156, but I can't reproduce that issue in any way on my system.

### DIFF
--- a/Common/KeyMap.cpp
+++ b/Common/KeyMap.cpp
@@ -281,14 +281,15 @@ void UpdateNativeMenuKeys() {
 	int confirmKey = g_Config.iButtonPreference == PSP_SYSTEMPARAM_BUTTON_CROSS ? CTRL_CROSS : CTRL_CIRCLE;
 	int cancelKey = g_Config.iButtonPreference == PSP_SYSTEMPARAM_BUTTON_CROSS ? CTRL_CIRCLE : CTRL_CROSS;
 
-	KeyFromPspButton(confirmKey, &confirmKeys);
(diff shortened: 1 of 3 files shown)


LunaMoo:
Decided to change it a bit and just ignore mouse input in the UI, doesn't matter if mouse control is active or not since most mouse input can still be mapped without mouse control and could be mapped even after it get's disabled.

unknownbrackets (approved):

i30817:
Does this also work for touchpads used as mouse? I stopped using standalone PPSSPP after i found the UI was automatically putting my touchpad as one of the analog joysticks.

Which was a problem on the UI because moving the mouse 'left' cause the menus to go 'back'. I was completely unusable.

LunaMoo:
If the touchpad is detected as DEVICE_ID_MOUSE then it's mapped controls will be ignored in the UI, but I don't use linux, nor I have a touchpad, so I have no clue how it is detected, nor what's your actual problem since analog axis doesn't really do anything in the UI on Windows.

hrydgard:
I like this better than the original solution, will merge.

As for the linux touchpad problem, it should be reported in a separate issue with more details.

i30817:
@hrydgard  I had already did and was it was left open in https://github.com/hrydgard/ppsspp/issues/11139, but i can bump it with a new downloaded ppsspp and transcribe the ini files on ~/.config/ppsspp for the current version, which i should have done to confirm it's that the problem.

Issue #9859, opened 2017-07-24
Allow remapping touch controls(mainly right thumbstick)


andreyorst:
First of all thanks for great PSP emulator. I've played Monster Hunter, and other games where camera control is important. And quite often we control `Camera` with `D-Pad`. When I play ppsspp on my PC, I usually map joystick's second stick to `D-pad`, to control camera with my right thumb. But when I play on my Android phone, there is no other option but to use `combo keys` to create second `D-Pad` on the right side of the screen. Another optional `Analog stick` to control `D-Pad` would be wery useful.

This is my configuration for MHFU:
![image](https://user-images.githubusercontent.com/19470159/28529373-0e0a8b8c-7099-11e7-85a7-6295e9af2c82.png)
Buttons `1` and `2` mapped to `right` and `left` on `D-pad`  and it is working workaround, but if I'll add two more buttons to control up and down behaviour it will become less convenient, and diagonal control will be still imposible.

hrydgard:
Yeah, that's a good idea. Noted as a feature request and for 1.6, maybe I'll get it into 1.5.

andreyorst:
Ah, I didn't knew about it at all. Definitely will check those "PS3 remasters".

As I was thinking about this second stick, I came to this: this stick should have saparate submenu as `combo keys` do. And it should not just control `D-Pad` but also should feature custom mappings because there are some games like Call Of Duty, which used `cross, triangle, square and circle` to control camera, which was really painful. And I believe that games that use `R` and `L` buttons to rotate camere would also get some benefits from this custom mapped stick.

Maybe it is too much, but it will change the way of playing with touch input.

unknownbrackets:
Just noting that the Vita also allows this (though with its physical right stick):

http://www.guidingtech.com/assets/postimages/2013/07/Vita-Right-Stick.png

Arbitrary mapping would be better (I know of games that use L/R and ▲/X), but a menu of options might be faster to configure and simpler.

Some might map both analogs to the left analog, which could get a little tricky too... aside from the dpad mapping problems.

-[Unknown]

jajabinx35:
Hi all,

Firstly apologies for bumping this thread up again...I'm Also a big fan of ppsspp and after reading this thread, I thought that this is actually a really good idea. 

For example, on my phone (oneplus 5), I map the volume up and down as the L and R buttons (with the awesome options ppsspp offers, you can rotate the screen to landscape reverse so that the volume buttons are on the top)

[...]

PR #12182, opened 2019-07-18
Add right analog for touch controls.

LunaMoo:
Not to be confused with #9859 which this doesn't fulfill despite that issue name as it's a request for mapping other controls for the analog which this doesn't.

 It only add's a right analog to be used as a proper right analog. This is useful for HD remasters and even more for games patched(ie through CWCheat) to get real right analog controls for example Metal Gear Solid Peace Walker or Ghost in the Shell Stand Alone Complex.

 If set to visible, resetting touch location to defaults will move face keys up and place right analog under reflecting dpad and left analog stick.

Screenshot_2019-07-18-04-05-42-596_org ppsspp ppssppluna

### DIFF
--- a/Core/Config.cpp
+++ b/Core/Config.cpp
@@ -859,6 +859,7 @@ static ConfigSetting controlSettings[] = {
 	ConfigSetting("LKeyX", "LKeyY", "LKeyScale", "ShowTouchLTrigger", &g_Config.touchLKey, defaultTouchPosShow, true, true),
 	ConfigSetting("RKeyX", "RKeyY", "RKeyScale", "ShowTouchRTrigger", &g_Config.touchRKey, defaultTouchPosShow, true, true),
 	ConfigSetting("AnalogStickX", "AnalogStickY", "AnalogStickScale", "ShowAnalogStick", &g_Config.touchAnalogStick, defaultTouchPosShow, true, true),
+	ConfigSetting("RightAnalogStickX", "RightAnalogStickY", "RightAnalogStickScale", "ShowRightAnalogStick", &g_Config.touchRightAnalogStick, defaultTouchPosHide, true, true),
 
 	ConfigSetting("fcombo0X", "fcombo0Y", "comboKeyScale0", "ShowComboKey0", &g_Config.touchCombo0, defaultTouchPosHide, true, true),
 	ConfigSetting("fcombo1X", "fcombo1Y", "comboKeyScale1", "ShowComboKey1", &g_Config.touchCombo1, defaultTouchPosHide, true, true),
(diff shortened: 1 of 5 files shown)


unknownbrackets:
I think this is a good thing, but I'm slightly concerned that it'll lead to confused users when it doesn't work in 99% of games unless there's some way to tell them it needs cheats or there's mapping or something...

-[Unknown]

LunaMoo:
Made it clear on the visibility screen. Right analog touch is disabled by default, so anyone who will want to activate it will see that it doesn't do anything in PSP games.

hrydgard:
Still a little concerned by the wording, as it implies that PPSSPP plays things other than PSP games (which strictly is true, but for most normal purposes isn't). Maybe it should only appear at all when playing remasters and patched game?

marosis:
So what just use right analog stick as, triangle= Y Up, cross= Y down, square= X left and circle= X right. This is really good for FPS games

LunaMoo:
Ok, changed to "(not used by most games)".
 Can't really make it to "only patched games" since patches can be applied by different ways some not detectable in any easy way and maintaining some kind of list would be a chore. At most I could hide touch right analog completely through ini-only setting since remasters only use it as d-pad alternative and patches are even less popular than fps patches and I don't see people even willing to learn how to make them.

 As for requests about mapping digital controls to touch analog as stated on the start this PR has nothing to do with that. I'm also not intrested in doing it at all since for me the benefit of touch analog stick exists only when it's used at much larger size than my thumb which is really useful to gain better control over sensitivity with true analog control, but that's actually something that most PSP games doesn't really support even for left stick. There might be people for which this is positive improvement, but to me personally it would at most be the opposite and would rather replace "digital analog stick" into 4 digital touch buttons since analog used digitally typically has to use some high threeshold and requires bigger and more tiresome movement for exactly same digital actions.

Emulatorer:
Right Analog doesn't work in Monster Hunter 3rd HD 🤔
Screenshot_2019-07-25-06-36-47
Screenshot_2019-07-25-06-36-56
Im using ppsspp luna edition 😅
@LunaMoo Do I need to use cwcheat for MH3rd HD?

Saramagrean:
It work for me without cheat. :)

Saramagrean:
It work for me without cheat. :)

In the village can not rotate the camera angle, will only be used when playing in quests.

Emulatorer:
@Saramagrean thanks for feedback now I can really enjoy playing Monster Hunter 3rd HD via PPSSPP :3

hrydgard:
Decided I can accept the new wording. Merging.

PR #12226, opened 2019-08-08
Vulkan: Add some hacks to make Killzone run at playable speeds on mobile

hrydgard:
Really gross though, we're disabling post-processing effects and there's a small artifact on the right side of the screen.

See #6207.

### DIFF
--- a/assets/compat.ini
+++ b/assets/compat.ini
@@ -460,6 +460,13 @@ ULUS10323 = true  # SR2
 ULES00940 = true  # SR2
 ULET00958 = true
 
+[KillzoneHack]
+UCUS98646 = true
+SCES52004 = true
+SCUS97402 = true
+SCKA20048 = true
+SLPM66151 = true
+
 [BlockTransferAllowCreateFB]
 NPJH50686 = true  # Digimon Adventures (JP, and English patches...)
 ULJS00078 = true  # MotoGP
--- a/Core/Compatibility.cpp
+++ b/Core/Compatibility.cpp
@@ -59,6 +59,7 @@ void Compatibility::CheckSettings(IniFile &iniFile, const std::string &gameID) {
 	CheckSetting(iniFile, gameID, "DisableAccurateDepth", &flags_.DisableAccurateDepth);
 	CheckSetting(iniFile, gameID, "MGS2AcidHack", &flags_.MGS2AcidHack);
 	CheckSetting(iniFile, gameID, "SonicRivalsHack", &flags_.SonicRivalsHack);
+	CheckSetting(iniFile, gameID, "KillzoneHack", &flags_.KillzoneHack);
 	CheckSetting(iniFile, gameID, "BlockTransferAllowCreateFB", &flags_.BlockTransferAllowCreateFB);
 	CheckSetting(iniFile, gameID, "YugiohSaveFix", &flags_.YugiohSaveFix);
 	CheckSetting(iniFile, gameID, "ForceUMDDelay", &flags_.ForceUMDDelay);
--- a/Core/Compatibility.h
+++ b/Core/Compatibility.h
@@ -59,6 +59,7 @@ struct CompatFlags {
 	bool DisableAccurateDepth;
 	bool MGS2AcidHack;
 	bool SonicRivalsHack;
+	bool KillzoneHack;
 	bool BlockTransferAllowCreateFB;
 	bool YugiohSaveFix;
 	bool ForceUMDDelay;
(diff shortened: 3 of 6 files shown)


unknownbrackets:
I know you know this, but I'm very against non-optional changes that make emulation knowingly wrong.  I realize what the game is doing here is terrible... but I'd hate to just disable the effect and require people to know about and hunt down an obscure hidden ini setting to be able to see full effects (even on a powerful desktop.)

-[Unknown]

hrydgard:
Yes, hence it's a work-in-progress thing and I'm very reluctant to do this. I'm thinking of just leaving this around as kind of a guide if someone wants to try something smarter....

hrydgard:
One way to possibly make it run faster would be to somehow forcibly split the main framebuffer so that the strip of padding data which it uses for temporary storage sits in its own framebuffer. Then it would at least just run slow instead of insanely slow (still won't be able to reorder the draws to avoid render targets switches because the strip is reused many times).

Dotty:
I absolutely love hacks which remove hideous, excessive, and/or costly post processing.
You could could do like PCSX2 does, and have it set up as an optional "aggressive" game specific hack. Maybe by ID rather than CRC.

i30817:
> Maybe by ID rather than CRC

They do

unknownbrackets:
> You could could do like PCSX2 does

*shudder*.  Perhaps we should create a small minigame you have to play before each emulation session, and your score will determine how good the emulation is.  I think it'd result in an overall better and more consistent experience.

> so that the strip of padding data which it uses for temporary storage sits in its own framebuffer

Yeah, this seems like the best way.  In theory, if it's doing entire draws in a detectable way, we might even be able to not reuse the same FB for each "framebuffer margin" temporary.

What vertex type does it use for this draw?  If it uses through mode, it should be easier...

-[Unknown]

LunaMoo:
No, for making PPSSPP into PCSX2-like, it's known for it's hacks and the fact that most of activity in the project goes around those hacks, but that's really awful way to go which only turned many people off from their project.

hrydgard:
I think we need to define an ethos for GPU emulation. I think it goes something like this:

* It's OK to render stuff at higher resolution, provided that rendering at 1x produces the correct result and rendering at higher resolution is a very close match visually (things like mip levels can differ, not much else).

* It's OK to reorder computations in time, provided we know that the same output is achieved in the end. This is done in the Sonic and Metal Gear Acid 2 speed hacks, where doing things the original way produces a render pass count that is just not doable on modern mobile drivers and barely by desktop drivers.

* It's not OK to drop computations that affect the final result. Not even if the effect is small.

A few more points may be needed, but I think we can all agree on these. By these rules, the Sonic and Metal Gear Acid 2 speed hacks are A-OK while this Killzone speed hack is right out as-is.

unknownbrackets:
Yeah.  To be fair, a big concern of mine is that we often focus on one slow area, but you never know if a small effect might also make rain stop working in some area of the game, or something.

The only thing I'd personally change is "It's not OK to drop computations that affect the final result *without an option*."  For example, vertex cache.

-[Unknown]

Dotty:
Sorry, I just meant the option. Having it in a clearly labelled switch (Aggressive, not normal, removes graphics) that's not set on by default. Not the entire concept of using nothing but workarounds and removing graphics to just deal with problems; that would be terrible.

sum2012:
Note that a chinese people in baidu use v1.6.3 as a solution

assets/compat.ini
```diff
@@ -460,6 +460,13 @@ ULUS10323 = true  # SR2
 ULES00940 = true  # SR2
 ULET00958 = true
 
+[KillzoneHack]
+UCUS98646 = true
+SCES52004 = true
```
Panderner, line 465: Umm?

Panderner:
In compat.ini commit these are game IDs next to UCUS98646 are Killzone for PS2. Why are you supposed to add ps2 game IDs to PPSSPP's compat.ini?

hrydgard:
Oh, I must have googled badly for the game IDs.

Panderner:
And the UCES00279 and UCKS-45041 are missing in compat.ini

unknownbrackets:
I recommend looking at related game versions here for the IDs (including two demos and three EU releases):
https://report.ppsspp.org/game/UCUS98646_0.30

-[Unknown]

hrydgard:
Yeah, i usually check that, gamefaqs release data, and a generic googling. Wasn't so diligent here apparently.

Anyway, i think I can do better than this for this game later, i have a plan (split off the offscreen part of framebuffers). Closing.

Gamemulatorer:
This game is the most lag psp games on ppsspp android?
Screenshot_2021-05-10-18-15-17-647_org ppsspp ppsspp
Screenshot_2021-05-10-18-16-25-424_org ppsspp ppsspp

hrydgard:
This was replaced with #15934 .

PR #12234, opened 2019-08-12
Audio: Remove little-used resampler option.

hrydgard:
Practically no modern devices run at 44khz by default anyway, so it's nearly useless.


hrydgard:
Hm, turns out libretro uses this path currently. I'll hold off on this for now.

PR #12236, opened 2019-08-12
A few simple Vulkan optimizations

hrydgard:
Just some very minor Vulkan CPU optimizations and cleanups.


unknownbrackets (approved):

Emulatorer:
How about OGL minor optimization?

hrydgard:
@Emulatorer  None of these particular changes apply to GL at all (except for the stencil optimization, which we already do on the GL side).

PR #12237, opened 2019-08-13
Vulkan: Simple blacklist of device names that are not allowed to check for Vulkan

hrydgard:
Will help #12093.

Sad stuff..

### DIFF
--- a/Common/Vulkan/VulkanLoader.cpp
+++ b/Common/Vulkan/VulkanLoader.cpp
@@ -248,8 +248,8 @@ bool VulkanMayBeAvailable() {
 		return g_vulkanMayBeAvailable;
 
 	std::string name = System_GetProperty(SYSPROP_NAME);
-	for (size_t i = 0; i < ARRAY_SIZE(device_name_blacklist); i++) {
-		if (name == device_name_blacklist[i]) {
+	for (const char *blacklisted_name : device_name_blacklist) {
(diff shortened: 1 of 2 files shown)


CookiePLMonster:
I feel gross even suggesting this, but maybe catching the exception in the offending function would be "less dirty"?

Common/Vulkan/VulkanLoader.cpp
```diff
@@ -239,6 +246,16 @@ void VulkanSetAvailable(bool available) {
 bool VulkanMayBeAvailable() {
 	if (g_vulkanAvailabilityChecked)
 		return g_vulkanMayBeAvailable;
+
+	std::string name = System_GetProperty(SYSPROP_NAME);
+	for (size_t i = 0; i < ARRAY_SIZE(device_name_blacklist); i++) {
```
CookiePLMonster: Range based for loop would be a bit cleaner here, I think. Yay for not using macros.

hrydgard:
I'm not sure how we could reliably recover from a crash caused by the detection.

Switched to range based for loop, I always forget those work on plain arrays.

unknownbrackets (approved):
I'm concerned that there may eventually be a new SHIELD Tablet with exactly the same name but without this problem, but not sure what to do about that, really...

-[Unknown]

PR #12241, opened 2019-08-13
Update PowerSaveModeReceiver

Florin9doi:
On emulator, power saving updates are delivered immediately using POWER_SAVE_MODE_CHANGED.
On Samsung Note9 with Android9, POWER_SAVE_MODE_CHANGED is never received, so isPowerSaveMode() need to be checked every time the application is resumed.


unknownbrackets (approved):
Cool, thanks.

-[Unknown]

PR #12250, opened 2019-08-17
Android: Get our base pointer in a saner way than YOLO, if possible.

hrydgard:
Should help #12249.

Tried this on a SHIELD TV 2017 and a Galaxy S8+ with no issues.


Emulatorer:
YOLO means?

hrydgard:
"You only live once" - just a little joke, we previously got the pointer by jumping out in the unknown pretty much.

Emulatorer:
okay 😅

PR #12258, opened 2019-08-20
Fixes delete button for downloaded games from homebrew store in UWP

Antx64:
This PR fixes the delete function for homebrew store downloads. Also, had to add condition to void Register_sceUsbMic() function. It was causing errors on build for UWP.

### DIFF
--- a/Core/HLE/sceUsbMic.h
+++ b/Core/HLE/sceUsbMic.h
@@ -16,5 +16,7 @@
 // https://github.com/hrydgard/ppsspp and http://www.ppsspp.org/.
 
 #pragma once
-
+//UWP doesn't like this for some reason
+#if !PPSSPP_PLATFORM(UWP)
 void Register_sceUsbMic();
+#endif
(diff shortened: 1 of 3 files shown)


Core/HLE/sceUsbMic.h
```diff
@@ -16,5 +16,7 @@
 // https://github.com/hrydgard/ppsspp and http://www.ppsspp.org/.
 
 #pragma once
-
+//UWP doesn't like this for some reason
```
Florin9doi: sceUsbMic.c/h should be added to CoreUWP/CoreUWP.vcxproj and CoreUWP/CoreUWP.vcxproj.filters

hrydgard: I just did that on master, so preferably just rebase this pull request on master.

hrydgard:
Good catch. Please rebase on master and remove the sceUsbMic checks, and I'll merge.

Antx64:
Just did the rebase and git pushed. Thanks

PR #12260, opened 2019-08-20
Update PowerSaveModeReceiver (again)

Florin9doi:
The previous patch had an issue: PowerSaveModeReceiver was registered in NativeActivity::Initialize() which is called only once, but when the activity is recreated (eg: caused by device rotation), the receiver wasn't registered against the new activity context. This is the reason why the Intent was not received  sometimes.
In this patch, PowerSaveModeReceiver is now registered in onCreate, so it will be called again when the activity is recreated; the explicit call from onResume is removed; and all 3 intents are handled my the same receiver. 
The changes are required because POWER_SAVE_MODE_CHANGED is not delivered to receivers declared in manifest: https://stackoverflow.com/a/38859740

### DIFF
--- a/android/AndroidManifest.xml
+++ b/android/AndroidManifest.xml
@@ -102,14 +102,6 @@
             </intent-filter>
         </activity>
 
-        <receiver android:name=".PowerSaveModeReceiver">
-            <intent-filter>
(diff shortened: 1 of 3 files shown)


hrydgard:
Cool, thanks!

PR #12264, opened 2019-08-21
Support for notched phones

Florin9doi:
Needs testing, some settings may became inaccessible

### DIFF
--- a/android/res/values-v28/styles.xml
+++ b/android/res/values-v28/styles.xml
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+	<color name="black">#FF000000</color>
+	<style name="ppsspp_style">
+		<item name="android:windowFullscreen">true</item>
+		<item name="android:windowNoTitle">true</item>
+		<item name="android:navigationBarColor">@color/black</item>
+		<item name="android:windowLayoutInDisplayCutoutMode">shortEdges</item>
+	</style>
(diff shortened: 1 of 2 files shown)


hrydgard:
Thanks! I've tried similar stuff but had issues on my notched phone, I'll test it later before I merge.

hrydgard:
Yeah, this is gonna need additional work. With fullscreen off, I get that white edge I could never get rid of:
fullscreen_off

And with fullscreen on, the UI is clipped and needs fixing: 
fullscreen_on

Right now I don't see how it's worth it - PSP games at approximately 16:9 just aren't wide enough for this to matter at all.

unknownbrackets:
It seems like the ideal thing would be to make the notch area black.

-[Unknown]

hrydgard:
Yeah, which is what currently happens without this pull request..

unknownbrackets:
I guess I might want to put buttons near the notch, at most... then again, I'm planning to avoid notch phones for as long as I can.

-[Unknown]

Florin9doi:
>Right now I don't see how it's worth it - PSP games at approximately 16:9 just aren't wide enough for this to matter at all.

You're right

hrydgard:
Well, I do want to extend the UI out into the notched area later - but I'm gonna postpone it to 1.10.

Halo-Michael:
And Please add FullScreen Button for iOS.

PR #12280, opened 2019-08-26
(Libretro) Buildfix - Update Makefile.common

inactive123:
This is needed for a couple of targets to compile.


hrydgard:
Alright, but why close?

twinaphex:
My bad, this was not a necessary PR. I have another one coming up which is actually necessary. My bad for wasting your time with this one.

LunaMoo:
You still should learn how to use git properly, first of all you should use branches and not trunk/master for any work, it'll make things much simpler for your git workflow. Then when working with branches, you don't have to remove the branch to commit completely different code(mistake or change of approach doesn't matter), just force push your new local changes to same branch and PR will automatically get the new code.

PR #12297, opened 2019-09-01
Vulkan: Prepare for adding a second init path for VulkanContext.

hrydgard:
Will eventually be used for libretro. Just wanted to try this part on CI.


hrydgard:
Hm, I just noticed how the libretro init actually works. It overwrites some Vulkan function pointers with its own no-ops from libretro_vulkan.cpp, and those don't support the extensions we were trying to enable, hence the issues.

I think it would be better to rip all that out, the code would be a lot more understandable. But we have to check what extensions the actual libretro Vulkan initialization enables.

PR #12323, opened 2019-09-13
WIP: Nintendo Switch (libnx) Port - Reference PR

m4xw:
After talking with @hrydgard, we decided I should open the PR already (also with 1.9.0 around the corner).
There are still quite a few Issues with the code that I didn't clean up yet.
I am aware of almost all issues.

- GL Loader macro hackery (can be cleaned up and made properly for all platforms, but since we HAVE to resolve ALL functions dynamically on switch, I went for the lazy way)
- CMake Issues with static platforms (will need help here to solve that, also needs a change in the glslang CMake file to not link to OS dependent, better solutions?)
- Some Issues with SDL GL context creation, we need 0,0 to get full GL as ver, otherwise we will only get GLES, I just whipped that together without considerations for other platforms currently.
- Standalone works with GLES2, GLES3, GL. Libretro with GL.
- File API macro shenanigans
- Lots of macro declares could be unified in a header (open for consideration)
- JIT Alloc stuff needs more sanity checks (can sometimes randomly fail, but rarely, likely due to aslr placement, not sure yet).
- No more masking needed (still needs more checks as said above)
- I also probably broke other platforms here and there (also am aware of most places, but I guess you all know the drill).
- Toolchain missing strnlen, realpath and std::thread::detach (made it a minefield..)
- So some stuff had to actually be joined (Local file loader etc), otherwise crashes..
- Consistent indentation and line-endings
- Need to fix some whacky Path management that I have right now (because of how hbmenu/hbloader works)
- I probably forgot quite a bit, but just don't hesitate to ask.

This is still in a WIP state since I kinda got sidetracked with the standalone port when I started the cleanup.
A minimal libretro common is added, it can be cut even further, in the end I just needed a few headers but not the full statemachine (glsm). Ignore that for now.

Also the gl_common stuff was generated with some notepad++ macros.
I made a branch with the squashed history of this PR and libretro-common ripped out for easy review here: https://github.com/m4xw/ppsspp/commit/7ff8d94c03ff2f099e26ab1f5ec2d9e5d045c96e

The current rebased changes (today) seem to run fine but further testing is needed from my part.

Total relevant changes ~800loc.
**Don't merge it anytime soon**, also I might swap the branch later to "switch", which currently still has non-squashed changes)


m4xw:
Pushed my rebased branch, still in testing but so far no regressions, will push it to my other branches later on.
@hrydgard  @unknownbrackets

LunaMoo:
Just a nitpick/annoyance, I saw comments that some of that code was copied from elsewhere, but could you fix the indent? We use tabs in this project. Also place curly braces on the end of the line instead of a new line please. It's pretty ugly to read stuff when parts of the same code are using different style.

m4xw:
@LunaMoo 
grafik

UI/NativeApp.cpp
```diff
@@ -498,10 +498,10 @@ void NativeInit(int argc, const char *argv[], const char *savegame_dir, const ch
 	else if (getenv("HOME") != NULL)
 		config = getenv("HOME") + std::string("/.config");
 	else // Just in case
-		config = "./config";
+		config = "config";
 
-	g_Config.memStickDirectory = config + "/ppsspp/";
-	g_Config.flash0Directory = File::GetExeDirectory() + "/assets/flash0/";
+	g_Config.memStickDirectory = g_Config.internalDataDirectory + config + "/ppsspp/";
+	g_Config.flash0Directory = g_Config.internalDataDirectory + "assets/flash0/";
```
unknownbrackets: These changes shouldn't be needed (or the above), since it should be using the `#elif PPSSPP_PLATFORM(SWITCH)` path on switch.

-[Unknown]

ext/native/thread/executor.cpp
```diff
@@ -10,7 +10,10 @@ void SameThreadExecutor::Run(std::function<void()> func) {
 }
 
 void NewThreadExecutor::Run(std::function<void()> func) {
-	std::thread(func).detach();
+	std::thread thr = std::thread(func);
+#ifndef HAVE_LIBNX
+    thr.detach();
```
unknownbrackets: Ah, I missed this one.  Seems like we could just make it a simple property of `NewThreadExecutor`?

-[Unknown]

Common/CommonFuncs.h
```diff
@@ -31,6 +31,8 @@
 
 #if defined(_M_IX86) || defined(_M_X86)
 #define Crash() {asm ("int $3");}
+#elif defined(HAVE_LIBNX)
+#define Crash() { while(1); };
```
unknownbrackets: Hm, this seems decidedly not a crash... can we just use like `*(int *)0 = 0x1337;`?

-[Unknown]

unknownbrackets (commented):
Seems like there are multiple cores updating the libretro headers.  Maybe it'd be best to update them once in a separate pull?

-[Unknown]

Common/CommonFuncs.h
m4xw: We can also install a full exception handler if needed + do a context restore, got it all ready in another project.
But yes, we can even throw a fatal, but that would cause a full reboot.

ext/native/thread/executor.cpp
m4xw: Either way works for me ;)

UI/NativeApp.cpp
m4xw: Ok I can double check this.

m4xw:
> 
> 
> Seems like there are multiple cores updating the libretro headers. Maybe it'd be best to update them once in a separate pull?
> 
> -[Unknown]

Yea it caused me some hassle already, I still have some stupid changes in there I think that are just redundant, need to take a look on that.

Not sure which of these are even required anymore.

m4xw:
@unknownbrackets just fyi pushed my rebase, I left extra commits in for now so we can see some relevant changes, will squash it the next time, ignore the gitlab file for now.
You might want to import the following toolchain related changes related to the launch browser stuff already again (SDL/SDLMain.cpp)
image

Core/HLE/sceKernelSemaphore.h
```diff
@@ -17,6 +17,11 @@
 
 #pragma once
 
+#ifdef HAVE_LIBNX
```
unknownbrackets, line 20: You can remove this check now, we renamed it.

-[Unknown]

Core/HLE/sceKernelMutex.h
```diff
@@ -17,6 +17,11 @@
 
 #pragma once
 
+#ifdef HAVE_LIBNX
```
unknownbrackets, line 20: You can remove this check now, we renamed it.

-[Unknown]

Common/GL/GLInterface/GLInterface.cpp
```diff
@@ -7,7 +7,7 @@
 
 #ifdef __ANDROID__
 #include "Common/GL/GLInterface/EGLAndroid.h"
-#elif PPSSPP_PLATFORM(SWITCH)
+#elif defined(HAVE_LIBNX)
```
unknownbrackets, line 10: Is this intentional?  Any specific reason not to use `PPSSPP_PLATFORM(SWITCH)`?

-[Unknown]

ext/native/file/file_util.cpp
```diff
@@ -17,6 +17,8 @@
 #include <string>
 #include <set>
 #include <algorithm>
+#include <strings.h>
```
unknownbrackets, line 20: The strings.h include should be there now (higher up, and not on Windows to avoid conflicts.)

-[Unknown]

ext/native/base/stringutil.h
```diff
@@ -4,6 +4,7 @@
 #include <cstdio>
 #include <cstring>
 #include <string>
+#include <strings.h>
```
unknownbrackets, line 7: This one is lower down, again now on Windows for cross plat reasons.

-[Unknown]

UI/GameSettingsScreen.cpp
```diff
@@ -1164,8 +1165,10 @@ void GameSettingsScreen::onFinish(DialogResult result) {
 	KeyMap::UpdateNativeMenuKeys();
 
 	// Wipe some caches after potentially changing settings.
+#ifndef HAVE_LIBNX
```
unknownbrackets, line 1168: Was there a problem this caused?

This will prevent some features from working, like post-processing shaders and certain settings changes.

-[Unknown]

Core/Util/GameManager.cpp
```diff
@@ -112,6 +112,14 @@ bool GameManager::Uninstall(std::string name) {
 }
 
 void GameManager::Update() {
+#ifdef HAVE_LIBNX
```
unknownbrackets, line 115: There's every chance this may all be working now without these hacks - it should now join automatically after the install completes.  I think none of the changes in this file are needed anymore, hopefully.

-[Unknown]

Core/MIPS/MIPSDebugInterface.cpp
```diff
@@ -33,6 +33,8 @@
 #include "Core/MIPS/MIPS.h"
 #include "Core/System.h"
 
+#include <strings.h>
```
unknownbrackets, line 36: This is added above, and not on Windows.

-[Unknown]

Core/HLE/sceKernelThread.h
```diff
@@ -26,6 +26,12 @@
 // There's a good description of the thread scheduling rules in:
 // http://code.google.com/p/jpcsp/source/browse/trunk/src/jpcsp/HLE/modules150/ThreadManForUser.java
 
+#ifdef HAVE_LIBNX
+// Conflicting types
+#define Thread _Thread
+#define ThreadContext _ThreadContext
```
unknownbrackets, line 32: We've renamed both of these conflicts - you can remove these lines.

-[Unknown]

unknownbrackets (commented):
Looked for a reason for vertexjit crashes, didn't see one.

Are ALL_CALLEE_SAVED / ALL_CALLEE_SAVED_FP correct on Switch?

-[Unknown]

UI/GameSettingsScreen.cpp
m4xw: IIrc some stuff triggered a context reset(?) and that caused a crash

SDL/SDLMain.cpp
```diff
@@ -555,7 +564,7 @@ int main(int argc, char *argv[]) {
 	if (mode & SDL_WINDOW_FULLSCREEN_DESKTOP) {
 		pixel_xres = g_DesktopWidth;
 		pixel_yres = g_DesktopHeight;
-		g_Config.bFullScreen = true;
+		g_Config.bFullScreen = false;
```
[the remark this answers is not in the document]
m4xw: cant remember, mightve been smth with docked maybe

Common/StringUtils.cpp
```diff
@@ -27,6 +27,9 @@
 
 #if PPSSPP_PLATFORM(SWITCH)
 #define _GNU_SOURCE
+#undef __GNU_VISIBLE
+#define __GNU_VISIBLE 1
+#include <stdio.h>
 #include <cstdio>
```
unknownbrackets, line 33: So using `cstudio` doesn't work?  I guess we should just only do `stdio.h` then.

-[Unknown]

Common/CodeBlock.h
```diff
@@ -10,6 +10,14 @@
 #include "Common/Log.h"
 #include "Common/MemoryUtil.h"
 
+#if PPSSPP_PLATFORM(SWITCH)
+#include <cstdio>
+#include <switch.h>
+// This is a bit of a whacky solution, however this is the simplest implementation for it
+// given the restrictions in Horizon
+extern Jit* activeJitController;
```
unknownbrackets, line 18: This isn't needed anymore right?  I haven't looked at the rest much, just want to make sure this is an artifact from rebasing still.  Doesn't look like it's actually used anymore, aside from setting it.

-[Unknown]

m4xw: Sorry for late answer, yea i think this is a leftover from when writable/exec memory werent properly split in the emitter. Will clean that up ext time

m4xw: Removed in next rebase

m4xw:
Review commit: https://github.com/m4xw/ppsspp/commit/871d5de20f07576195d3db8e467fad7bdada2b45

PR #12333, opened 2019-09-17
Android OpenSL initial queue fix

hrydgard:
Discovered (thanks to @jimmyw ) that I was queueing up the wrong amount of buffers and wrong size data when setting up OpenSL.

Fixing both issues and cleaning the code a little.

Android doesn't really seem to mind though, this doesn't appear to change anything about playback.


jimmyw:
Thanks for the credit! :)

Issue #12381, opened 2019-10-02
Touch screen offset issues again


hrydgard:
Getting multiple reports of this, and another from @weihuoya in https://github.com/hrydgard/ppsspp/pull/12033 who suggests adding back

``` NativeApp.setDisplayParameters(pixelWidth, pixelHeight, (int) densityDpi, refreshRate);```

to NativeActivity::surfaceCreated.

I'm going to try that and also create a touch test screen for users to check the touch accuracy.

hrydgard:
Reopening, still a little off on Pocophone F1, which is quite annoying... Trying to figure it out.

PR #12382, opened 2019-10-02
Touchscreen: Apply weihouya's fix, add a touchscreen test screen. 

hrydgard:
See #12381.

Also adds a new screen in Developer Tools that can be used to check for touchscreen misalignment. It includes a fullscreen checkbox to be able to directly toggle.


unknownbrackets (approved):
Without multi-window mode, I'm pretty sure this essentially reverses #12033.  But doesn't sound like it's fixed it for everyone after all...

Note #12033 was new in 1.9.x.

-[Unknown]

PR #12421, opened 2019-10-15
Update iOS JIT hack

sbingner:
The current hack being used to get read/write/execute memory on iOS leaves the process in a very bad state where it can never exit.  I fixed the hack for another project and somebody pointed out that I should let you guys know about it.  The attached hack will get the same RWX memory state but without leaving the process in a bad state that does not allow it to be killed or exit.

### DIFF
--- a/ios/main.mm
+++ b/ios/main.mm
@@ -28,7 +28,7 @@ bool get_debugged() {
 	if (rv==0 && flags&CS_DEBUGGED) return true;
 
 	pid_t pid = fork();
-	if (pid) {
+	if (pid > 0) {
 		int st,rv,i=0;
 		do {
 			usleep(500);
--- a/ios/main.mm
+++ b/ios/main.mm
@@ -15,9 +15,53 @@
 #include "base/NativeApp.h"
 #include "profiler/profiler.h"
 
-#define CS_OPS_STATUS	   0   /* return status */
-#define CS_DEBUGGED		 0x10000000  /* process is currently or has previously been debugged and allowed to run with invalid pages */
+#define CS_OPS_STATUS	0		/* return status */
+#define CS_DEBUGGED	0x10000000	/* process is currently or has previously been debugged and allowed to run with invalid pages */
+#define PT_ATTACHEXC	14		/* attach to running process with signal exception */
+#define PT_DETACH	11		/* stop tracing a process */
 int csops(pid_t pid, unsigned int  ops, void * useraddr, size_t usersize);
+#define ptrace(a, b, c, d) syscall(SYS_ptrace, a, b, c, d)
+
+bool get_debugged() {
+	int flags;
+	int rv = csops(getpid(), CS_OPS_STATUS, &flags, sizeof(flags));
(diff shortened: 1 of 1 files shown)


rweichler:
:fire::fire::fire::fire::fire::fire::fire::fire::fire::fire::fire::fire::fire::fire::fire::fire::fire::fire::fire::fire::fire::fire::fire::fire::fire::fire:

ios/main.mm
```diff
@@ -15,9 +15,53 @@
...
 int csops(pid_t pid, unsigned int  ops, void * useraddr, size_t usersize);
+#define ptrace(a, b, c, d) syscall(SYS_ptrace, a, b, c, d)
+
+bool get_debugged() {
+	int flags;
+	int rv = csops(getpid(), CS_OPS_STATUS, &flags, sizeof(flags));
+	if (rv==0 && flags&CS_DEBUGGED) return true;
+
+	pid_t pid = fork();
+	if (pid) {
```
rweichler: This `if` chain is a bit wonky. `if(pid)` basically means `if(pid != 0)`. The `else if(pid == 0)` captures all possible other values for `pid` so the `else` block below contains dead code.

Also, according to [the `fork(2)` man page](https://www.freebsd.org/cgi/man.cgi?fork(2)#RETURN_VALUES) errors return -1. 

So I think this chain should be something like:

```c
if(pid == -1) {
    // fail
} else if(pid == 0) {
    // child process
} else {
    // parent process
}
```

sbingner: Yeah that was supposed to be `pid>0` 😒

hrydgard:
Can't say I completely understand what's going on here, but if it fixes it, it fixes it...

sbingner:
> Can't say I completely understand what's going on here, but if it fixes it, it fixes it...
For posterity and edification:

The problem with the original hack was that it says it expects to be debugged but thing ever debugs it.  I fixed this by having something actually debug it.

1. It spawns a child, and waits for that child to exit
2. That child connects to its parent and debugs it.
3. Once the child is attached to the parent, it just detaches and exits.  This permanently marks the parent as having been debugged.
4. The parent can then continue normal execution and pages that are not CS valid or are later created not valid do not cause problems because the kernel knows that a debugger may have modified code in the parent.

hrydgard:
Ah, that explains it! Very cool!

Halo-Michael:
Thanks!

ios/main.mm
rweichler: ah, makes sense lol

jet082:
This is directed towards @sbingner (I cannot seem to find another way to contact you):

Does this hack work on non-JB devices on iOS11+?

I have recently been working on getting more cores working on retroarch. I got flycast to compile and work with its interpreter core. However, I would like to know if it is worth the time to try grafting this onto flycast so as to get some JIT optimization and reach fullspeed.

https://github.com/jet082/flycast

I apologize for the bother - since I do not own a JB device, nor one on anything previous to iOS 13.2, I would just like to know whether this is even possible before diving in.

PR #12431, opened 2019-10-19
Change Bloom Hack aggressive detection to include GOW games bloom.

LunaMoo:
Aggressive was kind of useless with balanced detecting all known cases already, so it should be fine to just dump GOW case into aggressive.

The result is on the ugly side, through some might still preffer that than the double image, maybe in the future effects like that could be additionally filtered somehow to keep them blurry, but without the aliasing or use the detection to remove them and just use post process for similar, but higher quality effects.

 Works in both GOW games as a side effect might improve performance slightly when not using x1 render res, but probably not noticeably.

Effects sampled at high res:
NPEG00023_00000

Effects at x1:
NPEG00023_00001


hrydgard:
The check is kind of odd, but yeah, let's go for it.

GOW's bloom is quite strangely implemented.

PR #12457, opened 2019-10-30
Fix vcrs disassembly

kotcrab:
`Dis_Vcrs` is using `_VS` instead of  `_VD`, this doesn't seem to be intended.


unknownbrackets (approved):
Thanks, I'm sure this was a mistake too.

-[Unknown]

PR #12462, opened 2019-11-03
Vulkan: Enable renderpass merging for all games

hrydgard:
Was introduced in 1.9 for GoW games, but some other games can also benefit (for example Star Wars - The Force Unleashed).

This should have no visible side effects if it's correct (which I believe it is), so why not just enable it globally. If we do find problems with this, they should be fixed instead of disabling it again.


unknownbrackets (approved):
Hm.  Do we need to add `dst` to `touchedFramebuffers` in the `VKRStepType::COPY` and `VKRStepType::BLIT` cases?  Hopefully a rare case.  Can't remember if there was a reason this wasn't needed...

It'd be interesting to expose a debug statistic of renderpasses per frame this reduces.

-[Unknown]

hrydgard:
If you enable the GPU profiler, you can see by looking at the list how many renderpasses were reduced away (though they'll look the same as renderpasses that we already removed through clear elimination).

So yeah, an actual number could be nice too indeed.

hrydgard:
Alright, I think this is ready now. Adding the copy/blit destinations to touchedFramebuffers does not hurt the optimization in the God of War games.

PR #12471, opened 2019-11-07
Make style more consistent

vnctdj:
While working on my tool to automate updating `ppsspp-lang`'s files, I've found these few lines of code which annoyed my parser.


vnctdj:
Thanks!

Did you all see that cool looking abbreviated commit hash? :-D

PR #12477, opened 2019-11-10
Support multiple instance of localhost Multiplayer on 1 PC

anr2me:
I only briefly tested this on Win10 with Metal Slug XX, because the game i have on this laptop is only games with AdhocMatching so i can't test it on regular Adhoc, ~~and apparently AdhocMatching is broken in the latest version of PPSSPP (not sure since when it's broken), when i tried to join a host it gets frozen (deadlock/race-condition issue during AdhocMatching callbacks maybe?)~~

### DIFF
--- a/Core/HLE/proAdhoc.cpp
+++ b/Core/HLE/proAdhoc.cpp
@@ -22,9 +22,7 @@
 // All credit goes to him!
 
 #if !defined(_WIN32)
-#include <netdb.h>
 #include <unistd.h>
-#include <netinet/in.h>
 #endif
 
 #include <cstring>
--- a/Core/HLE/proAdhoc.cpp
+++ b/Core/HLE/proAdhoc.cpp
@@ -21,6 +21,12 @@
 // This is a direct port of Coldbird's code from http://code.google.com/p/aemu/
 // All credit goes to him!
 
+#if !defined(_WIN32)
+#include <netdb.h>
+#include <unistd.h>
+#include <netinet/in.h>
+#endif
+
 #include <cstring>
 #include "util/text/parsers.h"
 #include "Core/Core.h"
@@ -55,6 +61,10 @@ std::recursive_mutex peerlock;
 SceNetAdhocPdpStat * pdp[255];
 SceNetAdhocPtpStat * ptp[255];
 uint32_t localip;
+
+bool isLocalServer = false;
+uint8_t PPSSPP_ID = 0;
+sockaddr localIP; // This might serves the same purpose with existing "localip" above, but since this is copied from my old code so here it is (too lazy to rewrite the code)
 
 int isLocalMAC(const SceNetEtherAddr * addr) {
 	SceNetEtherAddr saddr;
@@ -1286,7 +1296,7 @@ int getActivePeerCount(void) {
 int getLocalIp(sockaddr_in * SocketAddress){
 #if defined(_WIN32)
 	// Get local host name
-	char szHostName[128] = "";
+	char szHostName[256] = "";
 
 	if(::gethostname(szHostName, sizeof(szHostName))) {
 		// Error handling 
@@ -1296,11 +1306,26 @@ int getLocalIp(sockaddr_in * SocketAddress){
 	pHost = ::gethostbyname(szHostName);
 	if(pHost) {
 		memcpy(&SocketAddress->sin_addr, pHost->h_addr_list[0], pHost->h_length);
+		if (isLocalServer) {
+			SocketAddress->sin_addr = ((sockaddr_in*)&localIP)->sin_addr;
+		}
 		return 0;
 	}
 	return -1;
 #else
 	memcpy(&SocketAddress->sin_addr, &localip, sizeof(uint32_t));
+	char szHostName[256] = "";
+	gethostname(szHostName, sizeof(szHostName));
+	struct hostent* pHost = 0;
+	pHost = gethostbyname(szHostName);
+	if (pHost) {
+		memcpy(&SocketAddress->sin_addr, pHost->h_addr_list[0], pHost->h_length);
+		if (isLocalServer) {
+			SocketAddress->sin_addr = ((sockaddr_in*)&localIP)->sin_addr;
+		}
+		return 0;
+	}
+	//return -1;
 	return 0;
 #endif
 }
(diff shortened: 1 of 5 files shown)


LunaMoo:
Hi, could you look at:
https://github.com/ANR2ME/ppsspp/compare/adhoc_fix...LunaMoo:AdhocStuff_from_THBranch

 That's a comparison between this branch of yours and mine which I separated from my hackish branch that had some of your old work merged and supposedly even not including hacks has a better compatibility for multiplier. Outside of hacks I only made some small changes like replacing all: 
````
real_time_now()*1000000.0
````
with 
````
CoreTiming::GetGlobalTimeUsScaled()
````

But maybe the rest of the differences is fixing the adhocMatching you mentioned that's broken in PPSSPP's current master. Feel free to salvage that code, it's yours anyway.

 You could also take my changes for README.md where I removed duplicate credits we already had in it's proper place and  UI/MiscScreens.cpp where I added you to the credits and aemu to the 3rd party tools used to keep it consistent with everything else and finally place your name there at the same time.

**Comment by ANR2ME** [comment]
> 
> 
> Hi, could you look at:
> [ANR2ME/ppsspp@adhoc_fix...LunaMoo:AdhocStuff_from_THBranch](https://github.com/ANR2ME/ppsspp/compare/adhoc_fix...LunaMoo:AdhocStuff_from_THBranch)
> 
> That's a comparison between this branch of yours and mine which I separated from my hackish branch that had some of your old work merged and supposedly even not including hacks has a better compatibility for multiplier. Outside of hacks I only made some small changes like replacing all:
> 
> ```
> real_time_now()*1000000.0
> ```
> 
> with
> 
> ```
> CoreTiming::GetGlobalTimeUsScaled()
> ```
> 
> But maybe the rest of the differences is fixing the adhocMatching you mentioned that's broken in PPSSPP's current master. Feel free to salvage that code, it's yours anyway.
> 
> You could also take my changes for README.md where I removed duplicate credits we already had in it's proper place and UI/MiscScreens.cpp where I added you to the credits and aemu to the 3rd party tools used to keep it consistent with everything else and finally place your name there at the same time.

Although CoreTiming::GetGlobalTimeUsScaled() vs real_time_now()*1000000.0 may affect stability (ie. disconnection issue) but it shouldn't cause the game to freeze for an indefinitely time, that's why i suspects there is a deadlock/race-condition, but i haven't debug the code yet, because i need to make sure multiple instance feature to work properly first to make testing adhoc easier on a single laptop.

Btw, how do i exclude Android from multiple instance feature? since android doesn't have **shm_open** and may need to replace it with **ashmem** which is a hassle since i'm not familiar with it (i don't even know what to imports)

**Comment by ANR2ME** [comment]
Btw, currently on every Reset/changing game the PPSSPP_ID keeps increasing thus having different IP, so the adhoc server need to be re-pointed to the right IP on every reset/changing game which can be a hassle. 

It would be nice if the PPSSPP_ID can be made persistent but i forgot how :) i think the PPSSPP_ID Initialization & Cleanup need to be moved somewhere else

File: Core/HLE/proAdhoc.cpp [hunk 1]
```diff
@@ -1296,11 +1304,26 @@ int getLocalIp(sockaddr_in * SocketAddress){
...
+	struct hostent* pHost = 0;
+	pHost = gethostbyname(szHostName);
+	if (pHost) {
+		memcpy(&SocketAddress->sin_addr, pHost->h_addr_list[0], pHost->h_length);
+		if (isLocalServer) {
+			SocketAddress->sin_addr = ((sockaddr_in*)&localIP)->sin_addr;
+		}
+		return 0;
+	}
+	//return -1;
```
ANR2ME: I prefer to return -1 (failed) if the address didn't passed the validity test tho, but since i'm not trying to be in the way of the person who wrote the previous code so i just commented my code :D just in case that person have plans for his code in the future

LunaMoo:
Please avoid looking at https://github.com/LunaMoo/ppsspp/commit/4ec7f4f26b3ece66b64d8b2830a4b16718f55682 and instead look at the branch I compared to which is https://github.com/LunaMoo/ppsspp/commits/AdhocStuff_from_THBranch, all the adhoc stuff before adding hacks is in this commit https://github.com/LunaMoo/ppsspp/commit/c9de850c7d597dc32c5c064b3afedf3a0942672b 

They're different as I only started testing adhoc on my TemporaryHackishBranch later on and fixed some stuff I missed on the way that were missing in the initial merge while this separate branch was made specially in case someone wanted to merge multiply instances functionality as well as somewhat improved compatibility and just included all currently used adhoc code.

ANR2ME:
> 
> 
> Please avoid looking at [LunaMoo@4ec7f4f](https://github.com/LunaMoo/ppsspp/commit/4ec7f4f26b3ece66b64d8b2830a4b16718f55682) and instead look at the branch I compared to which is https://github.com/LunaMoo/ppsspp/commits/AdhocStuff_from_THBranch, all the adhoc stuff before adding hacks is in this commit [LunaMoo@c9de850](https://github.com/LunaMoo/ppsspp/commit/c9de850c7d597dc32c5c064b3afedf3a0942672b)
> 
> They're different as I only started testing adhoc on my TemporaryHackishBranch later on and fixed some stuff I missed on the way that were missing in the initial merge while this separate branch was made specially in case someone wanted to merge multiply instances functionality as well as somewhat improved compatibility and just included all currently used adhoc code.

Currently i'm trying to separate pull request of multiple-instance feature from adhoc bug fix.
I'll create a different pull request for things related to adhoc compatibility/stability when i found a solution, as i remembered the last time i worked with AdhocMatching the callbacks is corrupting some registers when returned and some even failed to return from callbacks (within 250ms) because the callback runs on any random PSP Thread (current thread). 
I did some experiment last night, testing to run AdhocMatching callbacks on idleThread(by calling __KernelSwitchOffThread before __KernelDirectMipsCall) seems to be the one failed to return within 250ms.
And currently i'm experimenting to run the mipscall on AdhocThread but having difficulty converting SceUID to Thread to be used on __KernelSwitchContext before calling __KernelDirectMipsCall

This is the code i'm having issue and don't know how to solve it: Error C2440	'static_cast': cannot convert from 'KernelObject *' to 'T *'
`Thread *t = kernelObjects.GetFast<Thread>(threadAdhocID);`

PS: On JPCSP when AdhocMatching Started it will create a new Thread, even though this thread is just a dummy/loop that calls HLE function (the actual code is running on JPCSP side) at least mipscall will runs on it's own thread thus won't corrupt registers on any random thread.

ANR2ME:
i probably should rebase this in to a single commit, and the branch name was suppose for fixing bugs related to adhoc instead of feature addition >.< i should change the branch name also

PR #12482, opened 2019-11-12
Better Remote Disc Streaming

ghost:
Hi.I am here to request better Remote Disc Streaming.Like seriously.Everytime i want  to stream my games to phone i need to reset settings 10 times on PC/Phone.It is lacking all we actually need.We need better UI for that function and more pieces like streaming save states(optional).Having problems with port and all that is ridiculous and i think emulator needs advanced locating and hosting options like LOBBY for that instead of having to change manual IP adresses if willing to host 2 devices for game.


hrydgard:
This should just be an issue, not a pull request. Closing, try again.

Yes, Remote disc streaming is lacking some features and polish for sure, and does need more work.

PR #12486, opened 2019-11-13
Rebindable touch right analog

iota97:
Allow Right analog stick to act as D-Pad or face button pad.

sd
Screenshot_2020-01-17_16-34-44

### DIFF
--- a/UI/GamepadEmu.cpp
+++ b/UI/GamepadEmu.cpp
@@ -29,6 +29,7 @@
 #include "gfx/texture_atlas.h"
 #include "math/math_util.h"
 #include "ui/ui_context.h"
+#include "Core/Util/AudioFormat.h"  // for clamp_u8
 
 static u32 GetButtonColor() {
 	return g_Config.iTouchButtonStyle != 0 ? 0xFFFFFF : 0xc0b080;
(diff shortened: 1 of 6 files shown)


marosis:
Finaly Somebody did it!

HiImBrala:
God Finally

hrydgard:
Kind of an oddball feature but I definitely see the point in this game and a few others!

I do think it should default to off though.

LunaMoo:
Don't like the idea of how it's implemented. Another analog stick for dpad, then people will remind about wanting face buttons as analog as well and we get yet another one? Then yet another for combo buttons because someone probably will request that as well. This should be more generic  / remappable.

iota97:
The default config is "defaultTouchPosHide" as the right analog, so it should be hidden/off by default i think (unless i am missing something).

About the implementation i see the point of having it customizable, but i think is better to make one for DPad (and another for face button if need), its only 1/2 stick at the end and it will cover 99% of the situation (If i recall correctly most game either use DPad, face button, or move the camera in a way is not really stick friendly).
Making it custom with every possibility will require a quite complex UI that will just be confusing for the end user IMHO.
Have DPad and face as a option is better than nothing anyway.

LunaMoo:
Some games require holding some button to enter aim mode to move camera. On physical controls we can map multiply PSP keys to stick to simplify the controls. With just this, it'll still not be possible on touch controls and if someone ever will want to make mappable touch controls to solve that, this code will become obsolete.

hrydgard:
Well, to get this perfect we should really implement some kind of scripting language support so that perfect game-specific controls can be created that does things like holding down a button while a stick is moving, etc. Not sure if we should let perfect be the enemy of the good here.

iota97:
I removed the extra analog and made the already present RightAnalog act as either RightAnalog, DPad or Face button (config in the control tab).

Right analog mode allow to switch these 3 mode.

Right analog button on press allow to set a key that will be held when the finger is on the analog (it wont require it to be moved as some game might have in/out animation between camera mode making the moving left to right toggling them).

Any suggestion is more than welcome on both code and naming of setting.

PilipinoAko:
Its only in gamepad? Or need to map first

LunaMoo:
I don't like some details still.

- "Right analog settings" are on "Game Settings Screen" which some users can definitely miss when just trying to activate right analog touch controls which is on completely different screen - I think those settings should be moved into the screen where touch buttons are enabled, it would be enough to add buttons above "Toggle all" in visibility screen for those.

- Defaulting analog to d-pad is purely prefference of some people, I don't like that for example Monster Hunter fans dictate default settings for anything, if it was made by someone playing shooters then what, it would default for face buttons? Then someone enables it without knowing what it is, assumes analog working as analog and then sees his d-pad get's pressed when moving right analog and game reacting on his analog movement as on button movement. Maybe just replace the checkbox option of the right analog enabling in visibility screen with a selection box and default it to off/invisible.

- if I read it correctly Right analog is still moved even if it's not used as right analog, not sure if will cause problems in any games, but there are remasters which while don't have a proper right analog support, do map it for other buttons natively, potentially creating confussion/double input.



 Aside from those, why changing lang submodule revision? Accident?

iota97:
Moved he setting on visibility screen as suggested. (See above screenshot for reference)

Default to right analog again.

About double input, i'm not sure as i don't know any game to test it, i tried to set the stick ID to 2 if other mode are used, i think the game read from it as the only difference from the left and right stick was ID 0 or 1.

About the lang submodules, i can't seem to get it sorted to match master, any suggestion here is very welcome.

LunaMoo:
Oh I forgot it starts from 0 and didn't notice it's still another stick, but that's probably worse.
 We can freely add a second analog because the function originally supported it, however when you're sending more analog sticks that never existed I think PPSSPP would write more data than expected which might be harmless 90% of the time, but could as well break stuff.
 Better just use right analog(stick 1), but use a check if(g_Config.iRightAnalogMode!=0) instead of check for (stick==2) for custom then else for sending analog stick(__CtrlSetAnalogX and __CtrlSetAnalogY does that) ~ to avoid sending analog data when it's not used as right analog.

Also:
"Right analog button on press" ~ seems confusing when you don't know what it's for. It should be kept grayed out unless right analog mode is not set to custom ones and a more descriptive name would be nice as well, maybe even literally "Keep this button pressed when custom right analog is used".


 As for accidental changes and reverting them, both the initial change and revert goes into git history, you can "squash" commits together to get rid of that.

iota97:
I'm not sure how to gray out the button if right analog is set as a right analog, the code pass a pointer to an option to change it dynamically when other option are changed, ATM:

"rightAnalog->SetEnabledPtr(&g_Config.touchRightAnalogStick.show);"

I'm not sure about the best way to handle 2 disabling condition.

hrydgard:
You can make a class member variable, update it at appropriate places and send it into SetEnabledPtr. Or just use SetEnabled?

I'll look at this more soon.

iota97:
Added a static bool and updating the condition when a checkbox, a checkbox choice or the other option is changed, not very pretty, there probably is a better way.

PilipinoAko:
Waiting

LunaMoo:
As it stands it's the last thing here, so maybe a simpler and better way, would be simply to not disable that option, but making it work with standard right analog as well, renaming it accordingly.

UI/GamepadEmu.cpp
```diff
@@ -431,9 +434,43 @@ void PSPStick::ProcessTouch(float x, float y, bool down) {
...
+				else
+				__CtrlButtonUp(analogButton[rightAnalogMode][1]);
+			if (dy < -0.5f)
+				__CtrlButtonDown(analogButton[rightAnalogMode][3]);
+			else
+				__CtrlButtonUp(analogButton[rightAnalogMode][3]);
+		}
+
+		if ((stick_ == 1 || stick_ == 2 ) && g_Config.iRightAnalogPress != 0) // Right analog button press
+				__CtrlButtonDown(button[g_Config.iRightAnalogPress-1]);
```
LunaMoo: Nitpick - one tab too many.

UI/GamepadEmu.cpp
```diff
@@ -431,9 +434,43 @@ void PSPStick::ProcessTouch(float x, float y, bool down) {
...
+		if(stick_ == 2) { // Custom right analog
+			int rightAnalogMode = g_Config.iRightAnalogMode - 1; // 0 is the right analog mode
+			__CtrlButtonUp(analogButton[rightAnalogMode][0]);
+			__CtrlButtonUp(analogButton[rightAnalogMode][1]);
+			__CtrlButtonUp(analogButton[rightAnalogMode][2]);
+			__CtrlButtonUp(analogButton[rightAnalogMode][3]);
+		}
+
+		if ((stick_ == 1 || stick_ == 2 ) && g_Config.iRightAnalogPress != 0) // Right analog button press
+				__CtrlButtonUp(button[g_Config.iRightAnalogPress-1]);
```
LunaMoo: Here as well.

LunaMoo (approved):

iota97:
There is still some problem, look like resetting analog position when the stick get added to the UI race condition with the memset to 0 from "sceCtrl.cpp".
Also ATM it write in the "unused" part of the control data struct.
The commit fix the race, but i wonder if is better to rename 2 byte of the struct or add some comment about that.

iota97:
Made if way more customizable. @LunaMoo any suggestion?

iota97:
Separated the custom analog logic as another class, i think it's cleaner than using stick ID branch, also this doesn't write to SceCtrl in weird way that should avoid bug with emulation.
The UI allow to freely remap it, so it should be versatile enough this time.
Any suggestion is more than welcome as always.

Schwipsy:
Hey! I was looking for this since I was playing Phantasy Star Portable and the camera moves with the DPAD.
I'm quite a newbie with github and stuff, how do I get this to work on my PPSSPP on android?. I also want to use the combo toggle keys. (which is #12623 i think)
Thanks!

iota97:
You would need to clone the repo, merge the PR locally and then build the APK (and install it as any other APK).

I have have an APK ready, it contains my others PR as well (I would advise to not enable any new setting in the graphic hack section of the OpenGL rendering).

If you don't wanna to deal with building it yourself you can just download it [here](https://drive.google.com/open?id=1FR4uN-vk6WdbLZynw6pK5PHBntkldlvI).

Schwipsy:
dude, you're a hero, i'll download the APK and test it out. thank you SO much.

PilipinoAko:
Thank to for apk now ican play call of duty or other fps game

hrydgard:
Can't quite decide what to do with this - it's good functionality, but somewhat confusing given the lack of a second stick on the real PSP, and the UI is different from the other binding functionality.

Maybe it's worth it though...

LunaMoo:
Personally I was against this PR in it's initial form because I really thought a full control mapping should be used for touch as well and this initially was more about very specific game usage than anything else, but now it at least allows customization matching many different games and as you said yourself "Not sure if we should let perfect be the enemy of the good here.".


 Also I don't think we can jump over people's lack of imagination and/or knowledge, for example standard mapping for physical buttons is very plain and easy to understand, but some users still don't have idea they can map single physical button to multiply PSP keys and other way around and looking at some support threads on forums or reddit, apparently there are people which doesn't even know PPSSPP allows remapping keys to begin with.

 To solve this, we'd have to have a detailed manual included with usage examples like old software did, but then we hit another problem that most people doesn't read and nowadays might even lack enough concentration to be able to read. This problem could be solved by video tutorials, but then if some guy who shouts to the camera and fakes emotions or half naked girl records their own, completely wrong guides they will end up more popular anyway. So ultimate solution would be to record proper tutorials by some hot, barely wearing any clothes girl faking emotions and shouting to the camera. ;o JK, but sometimes I really think the world works that way nowadays.

unknownbrackets:
Makes sense to me too.  With a right analog touch control I feel like I'd be swimming in touch controls, but that's not at all a problem with this change.

Just to note: I'm usually very against game-specific hacks, since there are easily over 1000 PSP games even if you don't count region/version duplicates.  But, right analog is something that doesn't exist on the PSP and homebrew/romhacks can add support for.  I'd actually be in favor of having a compat.ini or similar flag that indicates the right analog layout for games (maybe it'd be a different ini.)

-[Unknown]

iota97:
I think a hidden setting in a compat.ini might do more harm than good:

1) Some game have multiple control option (COD allow face button but also a button toggle to make left analog aim, some other use face or dpad).

2) If is not clear to the user what is happening when he will move the right analog in a menu and press random button will be kinda annoyed.

3) A lot of game indeed.

About control being crowed it's also a matter of using a phone or a tablet.

hrydgard:
Something we definitely need is a better way to download, organize and use CWcheats, and maybe we could even have recommended cheats for some games that enable the right stick for example in the future, similar to Unknown's idea. But that's quite separate from this one.

I think I'll go ahead and merge it.

Issue #12309, opened 2019-09-05
[Feature Request] More available options for the FPS/Speed Counters


Abbanon:
It would be nice to have additional options for PPSSPP's current FPS and Speed counters, such as:
 - Text/Font Size: to scale it smaller or larger
 - Location: Upper Left, Lower Left, Upper Right (default), Lower Right
 - Type: Have it display in the game screen (like default), or if you're on a computer have it display in the window border (alongside the app name) if you're playing in windowed mode

This really isn't anything needed, but would be a very useful option for those who would want it.

Emulatorer:
Similar #11524?

PR #12518, opened 2019-12-20
Customizable FPS counter

iota97:
Add some additional setting for the FPS/Speed counter:

- Position: Any screen corner
- Color: Green, White, Red, Black, Blue
- Size: from 0.5 to 3.0 (was 0.7)
- Toggle Shadow
- Rectangle background: None, White, Black, Transparent white, Transparent black

Previews (second image is really small bottom right in white)

asd


zminhquanz:
Hmm , i think you need to add the graph like RPCS3 emulator

iota97:
Fixes #12309 beside view type on the window border.

Abbanon:
How does this look when showing framerate by itself, and speed by itself?

iota97:
Pretty much the same, the background is based on the size of the text and the text is aligned to border in the same way.

s

Emulatorer:
This is very nice to be added to v1.10.0 milestone :)

Emulatorer:
@iota97  can you add color yellow in the FPS counter color choices because I prefer color yellow because it's look good  imo :)
Screenshot_20200415-092023
Color yellow FPS counter from GLtools

LunaMoo:
Why not just add a custom color to the list that would use ini-only setting to make color configurable, it's one of those things that you set once for your personal prefference anyway.

iota97:
Not a big fan on ini-only TBH as it's not obvious for many user (especially a RGB flipped for endianess one).

Wanted to try a better color settings, but I got a problem with scroll view + slider as it steal the focus or something and I can't drag it when the scroll appear (with a smaller screen).
Also using a slider popup in a popup make the background all black as you can't stack them.

Screenshot_2020-04-15_08-44-56

Emulatorer:
@iota97 do you have apk soI can test :)
sorry for asking here 😅

iota97:
It have my other changes as well, still should do the job:

https://drive.google.com/open?id=1AgLYKAFd878No_zyxWcdpE5hA3rUxt2c

Emulatorer:
@iota97 thanks and keep up the good work 👍

LunaMoo:
Don't get me wrong I am lazy and like change things from UI, but just look at your screenshot, 1 line setting turned into half of the screen of new settings which most users will only change once and leave it like that and many others will not use the feature at all. In my personal builds I have an option that hides most settings which I almost never change when just wanting to play due to clutter in PPSSPP UI, official version does not have anything like that yet and with each tiny feature that has to come with dozen of settings it literally begs for it, otherwise PPSSPP UI will turn into a prettier, but just as cluttered and confusing thing as that retrostuff people spend more time at setting things than actually playing games(which is an AWFUL user experience).

 Not againt those settings, but I think we'll have to decide on some ways to de-clutter the UI to allow quicker access to things that actually matter more than once on the start.

Abbanon:
That's true, it might be a good idea to compress all of the FPS layout options into its own pop-up (like the colour customizer does) to get rid of Settings menu bloat.

So there could be three settings:
- FPS Counter (for Disable, FPS, Speed, Both)
- FPS Counter Layout Editor (pop-up containing the Position, Size, Shadow, and Background options)
- Color Customization (like you currently have it)

unknownbrackets:
Hm, I think a separate screen would work better for the array of options than a large scrollable pop up.

On one hand, I think this is a lot of options for a little fps counter.  On the other hand, I think it's likely the main overlay some people may display, so customization makes a lot of sense.  I'm torn.

I think if we're showing a color picker, we at least need a preview swatch that updates as you drag the sliders.  Not sure if maybe we should use HSV or even a grid... seems like a lot for a few numbers on the screen.  Still torn, sorry.

-[Unknown]

iota97:
A separate screen seem better (popup will not work anyway as it will require popup stacking for other option).

Screenshot_2020-04-17_08-45-42
Screenshot_2020-04-17_08-45-54

About the color previous I'm not really sure how to make it tho', with a preview HSV make sense indeed. 

For the customization while color is totally a personal preference position is way more useful and game based as well for not covering the HUD, not sure what's the best way here.

Emulatorer:
@iota97 Can you make FPS counter colors selection like this? it's possible?
Screenshot_20200423-184229
That's from Xposed app called Flat Style Bars Indicators.

fahadfoyjur:
Iota your horrible hack worked but not any performance improvment plz any solution

Gamemulatorer:
@hrydgard can you add this to v1.10.0 milestone ?

hrydgard:
Still can't decide if this one is worth all the extra translation strings. I'm thinking about just making the default look better instead (though I do realize people have many different preferences....)

fahadrashid7705:
I want to play some psp games in 30fps but vsync not worked . What do i do ? Sorry for my bad english .

hrydgard:
@fahadrashid7705 That's entirely irrelevant to this issue, please make a new issue and describe clearly what's not working.

iota97:
Thinking on this it went kinda way too far...

What you think about just having the position in the UI (as it can move the counter out of game hud) and leave color and size (?) as a ini only setting (I guess this can be set once if a user want it anyway)?

EX-e-LD:
@iota97 
Hi.Can we expect it in 1.11 update?This would be really neat

itsXPCT:
@iota97 Hi,me again.Is this launching this update?

TIA

LunaMoo:
>  I don't know how to edit zim/atlas. 😂

 You don't have to edit atlas files, just run buildatlas.sh to re-generate it from new atlas script / assets that should be already present if the PR is adding some changes to it.

Saramagrean:
> 
> You don't have to edit atlas files, just run buildatlas.sh to re-generate it from new atlas script / assets that should be already present if the PR is adding some changes to it.
> 
> Thank you for advice! It work!! 👍

XiaomiRedmi9:
Any news updates about this??

XiaomiRedmi9:
@iota97 can you possibly add a battery percentage just like this please 🥺
IMG_20210721_104935

iota97:
I feel like LunaMoo is right on this, closing. (I won't delete the branch on my fork in case someone want this anyway).

unknownbrackets:
A compromise could be keeping the ini options but not adding the UI?

I don't think the rendering adds much complexity, but I know we want to avoid complicating the options to avoid burying settings.

-[Unknown]

iota97:
I'm not sure how much ini option are accessible to average user.

I think it's probably simpler to use a 3rd party overlay for fps at some point rather than dealing with app specific ini config (at least on desktop, not sure on Android if there are any convenient app for that and neither if ini editing got worse with scoped storage).

NABN00B:
There's already a layout editor built into PPSSPP. If it would be fixed up and generalized for not-only-touch-controls usage, then the FPS counter, the chat button, the devmenu button and everything else could all be made customizable there.

hrydgard:
Doing that right, including good UI design that works for all input devices, is not a small task though, and not something I personally plan to prioritize soon at least.

PR #12559, opened 2020-01-11
Expand PRX decryption capabilities

DaveeFTW:
These patches extend ppsspp's ability to decrypt PRXs. Currently ppsspp is limited to decryption of type 0, 1, 2 and 6 executables of certain keys.

This patch:
* Adds proper decryption of type 0, 1, 2, 5, and 6 PRX.
* Removes decryption key limitations.
* Adds PRX GZIP decompression - used in some homebrew.
* Adds Pauth decryption support.

I've tested:
* Star Wars Battlefront: Elite Squadron - Tag: 0xC0CB167C, type 1
* Ridge Racer - Tag: 0x08000000, type 0
* Monster Hunter Portable 3rd - Tag: 0xD91681F0, type 6

Homebrew:
* Worm Warrior - Tag: 0xADF305F0, type 2
* Block Out II - Tag: 0xADF305F0, type 2, GZIP compressed

Currently the decryption routine attempts decryption of all types until it is successful. This is not optimal, and ppsspp would be better suited to separate the tag data into lists belonging to the same PRX type.

I'm also not sure if there is additional work to be done for other executable types, but I can add other PRX types easily.

Additionally, for others outside of PPSSPP interested in using these patches, consider it "dual-licensed" under the public domain.


hrydgard:
This is very cool! I'll take a look in detail soon.

I'm not sure what you mean by separating the tag data into lists - do you mean that we should keep a list of which GameIDs require which type of encryption, or something else?

DaveeFTW:
Each tag corresponds to a certain seed/key and a certain decryption routine based on prx type. Currently, since most tags don't have the prx type defined I just go through each decryption routine until there is a successful decrypt. It's not too expensive given that decryption is pretty rare, but it does mean that `pspDecryptPRX` error values are not particularly useful.

Since keys are no longer an issue, the only errors that `pspDecryptPRX` would report are:
* PRX type isn't supported
* Missing tag in list
* File is corrupted

All of these issues won't be fixed by nicely organising lists and would all give the same error codes anyway. So, its probably not worth doing, unless you like ultra efficiency and some neat organisation :)

LunaMoo:
Really cool and something I never hoped to happen ~ does that mean game updates also no longer will need to be pre-decrypted and pre-patched?:3 I forgot which key they were using, but it was one of the biggest annoyances of PSP emulation and pretty confusing to new users as well.

 If we can decrypt Pauth files as well, we will be able to remove the old workaround and some hack used for it as well, yay:).

LunaMoo:
Hmm, "Evangelion Jo" does not work with it showing "pauth decryption failed", it uses scePauth_F7AA47F6, game returns to main menu on failed decryption(happens after first level).

DaveeFTW:
Looks like there was some incomplete keys that took precedence over the complete ones that I added. My updated commit should allow that game to decrypt pauth files.

hrydgard:
Just tried a whole bunch of stuff (that worked before), everything seems fine so far. Seeing no obvious issue with the code, I'm just gonna go ahead and merge. Thanks!

hrydgard:
Hey @DaveeFTW , appears it might not work with the God Eater series, see #12571 and #12570. 

I'll look into it myself soon too if you don't beat me to it.

DaveeFTW:
@hrydgard Yes, that issue is caused by this patch. Since we can now decrypt *all* prx, the supporting libraries included in games now no longer fail and fall through to whatever executable loading mechanism ppsspp has.
I'm not sure whether or not ppsspp's PRX loader should be able to handle these helper libraries, but an obvious solution here would be to add the offending libraries to `IsHLEVersionedModule` and let the already existing HLE handling do the work.
I will upload a patch that does this, you can decide whether or not its the way to go forward :)

hrydgard:
Right, thanks for diagnosing the problem!

Issue #12565, opened 2020-01-12
Visualise Grid when editing Touchscreen control layout


DavidTamas:
PR #12517 added the feature of snapping the controls to an invisible grid in the custom layout option under touchscreen controls.

I would like to request a feature to actually display this grid on the screen to help further with positioning the controls. Horizontal and vertical center lines would also be useful.

DavidTamas:
![index](https://user-images.githubusercontent.com/22328803/72219231-0b308980-3544-11ea-9c1f-094e1c21b84a.png)
A quick draw of what I had in mind.

DavidTamas:
It doesn't draw for me on v1.9.3-225-gdd1de41db

PR #12567, opened 2020-01-12
Use snap grid on resize as well

### LINKED ISSUE #12566: Apply snapping points to on-screen control resizing
The last third of https://github.com/hrydgard/ppsspp/issues/12361. Grid request: https://github.com/hrydgard/ppsspp/issues/12565.

I see no downside for snapping move and resize together. Having an option for each would add even more clutter when you can just toogle it.

Thanks for the snap btw, the dynamic grid is a godsend.

iota97:
Fix #12566, grid size is halved on resize as it seem more natural on the move scale.

About rendering the grid to screen as mentioned in #12565, any suggestion?


hrydgard:
The grid rendering has already been addressed (and bugfixed) :)

Issue #12337, opened 2019-09-18
PSP camera support on non-Android platforms


hrydgard:
Now that the basic infrastructure is in place thanks to the progress on @8521 , work is needed to support the camera on desktop platforms. 

On Windows, the traditional way has been DirectShow, is that still the way to go? It's a real hassle.

On Linux I really have no idea what API to use. V4L?

shenweip:
Or OpenCV?

Florin9doi:
Windows and linux are done

Panderner:
But MacOS is remaining

Florin9doi:
Systems/APIs :
- [x] basic support for Android / Windows / Linux / iOS
- [x] basic support for Qt (Linux/macOS) (#12607)
- [ ] UWP ??
- [ ] SDL:macOS ??

Features:
- [x] support for multiple resolutions (#12607)
- [ ] aspect ratio fixes/cropping
- [ ] still image capture
- [ ] save states
- [ ] lens direction simulation ??

Florin9doi:
@hrydgard, can I use some camera capture function for a different project?

hrydgard:
You can use any code from PPSSPP as long as you follow the license.

raleighlittles:
Any updates on this?

Florin9doi:
The QT version still works on 13.6.6, including Continuity Camera:
<img width="992" alt="Screenshot 2024-04-01 at 21 14 23" src="https://github.com/hrydgard/ppsspp/assets/2995486/33253a48-bb76-4a98-b88a-aa72f21fb86e">
<img width="1313" alt="Screenshot 2024-04-01 at 22 35 18" src="https://github.com/hrydgard/ppsspp/assets/2995486/c6520e02-6826-4561-8a77-f1f5df5d72e4">
<img width="1547" alt="Screenshot 2024-04-01 at 22 36 39" src="https://github.com/hrydgard/ppsspp/assets/2995486/0f509a05-4278-4d95-bbb2-46fdc92024d6">

PR #12572, opened 2020-01-13
Add camera support for windows.

shenweip:
Implement this with MF api,referenced some Microsoft docs and samples.
I tested **Invizimals**, it works fine on my laptop(with a Microsoft HD Webcam).
**Preview:**

Also has a basic support for multi camera switching.

EDIT by hrydgard: Helps #12337

### DIFF
--- a/UI/NativeApp.cpp
+++ b/UI/NativeApp.cpp
@@ -37,6 +37,7 @@
 #if defined(_WIN32)
 #include "Windows/WindowsAudio.h"
 #include "Windows/MainWindow.h"
+#include "Windows/CaptureDevice.h"
 #endif
 
 #include "base/display.h"
@@ -67,6 +68,7 @@
 #include "Common/LogManager.h"
 #include "Common/MemArena.h"
 #include "Common/GraphicsContext.h"
+#include "Common/OSVersion.h"
 #include "Core/Config.h"
 #include "Core/ConfigValues.h"
 #include "Core/Core.h"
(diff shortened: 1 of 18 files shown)


hrydgard:
Very cool! How far back in Windows versions does this API work? Is it supported on Vista? 

I was dreading doing this because of DirectShow which I still have nightmares about (only slightly kidding)...

shenweip:
Accordings to docs, most of them are supported on vista, maybe few are not supported though,like MFEnumDeviceSources function.It has a full support on win7.I think not many users on vista?

hrydgard:
It's just that I don't want a change to our minimum requirements to be accidental, it should be intentional. Currently PPSSPP supports Vista and forwards, so it would be good to be sure - if this API doesn't work there, we could simply disable it on Vista, keeping PPSSPP working for Vista users even though they won't get this feature.

shenweip:
Yeah, it's right, I will comfirm this through VMware.

UI/NativeApp.cpp
```diff
@@ -879,6 +888,11 @@ void NativeShutdownGraphics() {
 	winAudioBackend = nullptr;
 #endif
 
+#ifdef _WIN32
+	delete winCamera;
```
Florin9doi: if (winCamera) ?

shenweip: Right.

hrydgard: Not really necessary as delete handles nullptr, but it's fine either way I think.

Florin9doi:
I've tested the first commit and I've observed that the image is stretched because the games ask for 480x272, but the webcam provides 640x480 (I think) which have a different aspect ratio.
This can be fixed by passing to sws_scale/sws_getContext a smaller source height

hrydgard:
If you pass a smaller source height you should also offset, to center it vertically, I guess. Though I'm not sure how much that matters for these games...

hrydgard:
Improving stuff like that can be done later, good to get the functionality in, so I'm merging.

PR #12580, opened 2020-01-15
Add camera support for linux (V4L2)

Florin9doi:


Core/HLE/sceKernel.cpp
```diff
@@ -276,6 +276,7 @@ void __KernelDoState(PointerWrap &p)
 		__VideoPmpDoState(p);
 		__AACDoState(p);
 		__UsbGpsDoState(p);
+		// __UsbCamDoState(p); // not yet ;D
```
hrydgard: Not yet, why? Either explain in the comment, or remove the line entirely please.

Core/HLE/sceUsbCam.cpp
```diff
@@ -89,65 +126,43 @@ static int sceUsbCamReadMicBlocking(u32 bufAddr, u32 size) {
...
-	INFO_LOG(HLE, "UNIMPL sceUsbCamAutoImageReverseSW");
+static int sceUsbCamStopVideo() {
+	Camera::stopCapture();
 	return 0;
 }
 
 static int sceUsbCamReadVideoFrameBlocking(u32 bufAddr, u32 size) {
 	std::lock_guard<std::mutex> lock(videoBufferMutex);
-	u32 transferSize = std::min(videoBufferLength, size);
+	u32 transferSize = (std::min)(videoBufferLength, size);
```
hrydgard: What's with the weird parentheses?

Core/HW/Camera.h
```diff
@@ -0,0 +1,66 @@
...
+// https://github.com/hrydgard/ppsspp and http://www.ppsspp.org/.
+
+#pragma once
+
+#include "ppsspp_config.h"
+#include "Core/Config.h"
+#include "Log.h"
+
+#if PPSSPP_PLATFORM(LINUX) && !PPSSPP_PLATFORM(ANDROID)
+	#include <stdio.h>
```
hrydgard: nitpick: we don't indent includes

android/src/org/ppsspp/ppsspp/NativeActivity.java
```diff
@@ -1409,12 +1409,6 @@ public boolean processCommand(String command, String params) {
 				// Only keep the screen bright ingame.
 				window.clearFlags(WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON);
 			}
-		} else if (command.equals("event")) {
```
hrydgard: Why is this removed?

hrydgard (changes requested):

android/src/org/ppsspp/ppsspp/NativeActivity.java
Florin9doi: this is called from __UsbCamShutdown now

Core/HLE/sceKernel.cpp
Florin9doi: I find challenging to maintain backward compatibility for save states and I doubt that the current version contain enough information to cover every scenario. I'll remove it

Core/HW/Camera.h
Florin9doi: fixed

Core/HLE/sceUsbCam.cpp
Florin9doi: https://stackoverflow.com/questions/5004858/why-is-stdmin-failing-when-windows-h-is-included . I replaced it with NOMINMAX

android/src/org/ppsspp/ppsspp/NativeActivity.java
hrydgard: No not this Android mCameraHelper, as far as I can tell skimming the code...

hrydgard:
I'm fine with this now, except still unsure about that Android camerahelper thingy..

Florin9doi:
event#exitgame was used to deallocate camera resources when a game is closed.
In this patch, the same thing is done by: 
__UsbCamShutdown>>Camera::stopCapture>>System_SendMessage("stopVideo")

hrydgard:
I see! Sorry I missed that.

hrydgard (approved):

PR #12582, opened 2020-01-17
SDL app icon

Florin9doi:


iota97:
This will not work if PPSSPP is not started from the same directory containing assets (as in `./build/PPSSPPSDL`). Maybe its better to concat with `SDL_GetBasePath()`?

Florin9doi:
fixed

Issue #10654, opened 2018-02-26
Feature: Enable Vulkan on macOS/iOS using MoltenVK and the LunarG Apple SDK for Vulkan


hrydgard:
Might be interesting to see what kind of performance we could get with this stuff (Vulkan on top of Metal).

https://www.khronos.org/news/press/vulkan-applications-enabled-on-apple-platforms
https://github.com/KhronosGroup/MoltenVK
 https://www.anandtech.com/show/12465/khronos-group-extends-vulkan-portability-with-opensource

Interesting is the "Portable Subset", which I believe will not affect us very much:

https://www.khronos.org/vulkan/portability-initiative

At most we might have to emulate triangle fans and maybe swap some texture color channels.

konistehrad:
Assuming the performance impact is minimal: do you see the project moving to the Portable Subset for all Vulcan-supporting platforms, or only for platforms that require a MoltenVK translation layer?

hrydgard:
We only exceed the Portable Subset in a single place currently, we use texture channel swizzles to get the texture formats to match the PSP completely. We'll probably just add a slower path for MoltenVK that swizzles texture colors in software before upload, for simplicity, if we ever get around to trying this :)

iOS4all:
There's a new update about using vulkan rendering for apple devices Mac/iOS

Just see here on retroarch commit:
https://github.com/libretro/RetroArch/commit/0a4ff225be284744168f2e3f97b48f2df067f25b
https://github.com/libretro/RetroArch/commit/789933b333f4a3988edc20b92fac17f31feb8492

Maybe you'll find it very interesting.

hrydgard:
(Just a note, the texture swizzle issue above is no longer a concern, MoltenVK handles it transparently, and it's natively supported in newer Metal versions).

hrydgard:
I retract that - it still seems to be an issue at least on my old macbook.

Halo-Michael:
![@I(L@5K LU18RP6YQ98HCG](https://user-images.githubusercontent.com/45062120/101402006-1df73e00-390e-11eb-91b1-6eea2b966af2.png)
I am not sure if vulkan is actually used. When I switch between OpenGL and Vulkan when running a game, it seems that the frame rate has not changed significantly; in addition, I have not obtained information in the log that the engine used by the game is OpenGL or Vulkan. Maybe you need to add the currently used engine to the log?

Halo-Michael:
[...]

PR #12583, opened 2020-01-19
macOS: Initial support for vulkan on macOS ( MoltenVK )

CarterLi:
This PR tries to ~~address #10654 -~~ implement a highly ~~experimental~~ buggy ~~vulkan~~ MoltenVK support on modern macOS.

To test it:

1. Build MoltenVK yourself. Please follow [the official document](https://github.com/KhronosGroup/MoltenVK#building). `make macos` is enough. Or use the official --outdated-- homebrew version `brew install molten-vk`
2. Copy `MoltenVK/macOS/dynamic/libMoltenVK.dylib` ( or `Package/Release/MoltenVK/dylib/macOS/libMoltenVK.dylib` for higher version ) to `/usr/local/lib` or somewhere else that PPSSPP can find it. If use homebrew, skip this.
3. Build PPSSPP with SDL mode. Qt mode is NOT supported

Tested on macOS 10.15.2 (19C57)

image
image


Common/Vulkan/VulkanLoader.cpp
```diff
@@ -310,6 +317,8 @@ bool VulkanMayBeAvailable() {
 	const char * const platformSurfaceExtension = VK_KHR_WIN32_SURFACE_EXTENSION_NAME;
 #elif defined(__ANDROID__)
 	const char *platformSurfaceExtension = VK_KHR_ANDROID_SURFACE_EXTENSION_NAME;
+#elif defined(VK_USE_PLATFORM_METAL_EXT)
+  const char * const platformSurfaceExtension = VK_EXT_METAL_SURFACE_EXTENSION_NAME;
```
hrydgard: nitpicking: inconsistent indentation

SDL/SDLCocoaMetalLayer.h
```diff
@@ -0,0 +1,5 @@
+#pragma once
+
+extern "C" {
+  void *makeWindowMetalCompatible(void *window);
```
hrydgard: another indentation miss

hrydgard (commented):
Sweet!

I'm only a little concerned that it would now no longer build without the manually built dylib present. 

Do you think we could either autodetect its presence and turn off VK_USE_PLATFORM_METAL_EXT for example, or maybe supply a precompiled dylib as a submodule or something? What might be best?

It's important for the build user experience that checking out and building is smooth.

CarterLi:
> Sweet!
> 
> I'm only a little concerned that it would now no longer build without the manually built dylib present.
> 

`libMoltenVK.dylib` is dynamic loaded, like other platforms. If it's not found, PPSSPP will fall back to OGL mode.

> Do you think we could either autodetect its presence and turn off VK_USE_PLATFORM_METAL_EXT for example, or maybe supply a precompiled dylib as a submodule or something? What might be best?

`VK_USE_PLATFORM_METAL_EXT` is in vulkan header, which is [embeded into PPSSPP source tree](https://github.com/hrydgard/ppsspp/blob/master/ext/vulkan/vulkan_metal.h). Thus it should not be a problem.

> 
> It's important for the build user experience that checking out and building is smooth.

The only problem is that [#define VK_USE_PLATFORM_METAL_EXT](https://github.com/hrydgard/ppsspp/pull/12583/files#diff-96c245bf2078b113993d44e76c81973fR29) is guarded by `#elif defined(__APPLE__)`, which can affect iOS build. `VK_USE_PLATFORM_METAL_EXT` itself is both supported by macOS and iOS. The only difference is that [makeWindowMetalCompatible](https://github.com/hrydgard/ppsspp/pull/12583/files#diff-f29789b2ebcb51f76b7291ecf7b21462R7). `NSWindow` stuff are macOS specific, but should be easy to port them to iOS `UIWindow`

CarterLi:
> One problem is that #define VK_USE_PLATFORM_METAL_EXT is guarded by #elif defined(__APPLE__), which may affect iOS build. 

I can change this to PPSSPP_PLATFORM_MAC, if desired.

hrydgard:
Oh right it's a dylib not a lib... then things should be fine even when not present indeed. Both builds seem to still pass, so I don't know if you need to change that preprocessor check, but it might still be a good idea.

hrydgard:
So this is a little odd. I built Moltenvk:

```
./fetchDependencies.sh
make macos
sudo make install
```.

The last line of the output was `/bin/cp -a Package/Latest/MoltenVK/macOS/framework/MoltenVK.framework /Library/Frameworks/`

You'd think PPSSPP should be able to find it there, but it doesn't. It did however work copying it to /usr/local/lib indeed.

Anyway, happy to merge this, but seems I was wrong in the other issue that we don't need to care about swizzle anymore, this is from running Wipeout Pure:

```[mvk-error] VK_ERROR_FORMAT_NOT_SUPPORTED: VkFormat VK_FORMAT_B4G4R4A4_UNORM_PACK16 is not supported on this device.
[mvk-error] VK_ERROR_FORMAT_NOT_SUPPORTED: VkFormat VK_FORMAT_B4G4R4A4_UNORM_PACK16 is not supported on this device.
[mvk-error] VK_ERROR_FEATURE_NOT_PRESENT: The value of vkCreateImageView(VkImageViewCreateInfo::components) (VK_COMPONENT_SWIZZLE_A, VK_COMPONENT_SWIZZLE_R, VK_COMPONENT_SWIZZLE_G, VK_COMPONENT_SWIZZLE_B), when applied to a VkImageView, requires full component swizzling to be enabled both at the time when the VkImageView is created and at the time any pipeline that uses that VkImageView is compiled. Full component swizzling can be enabled via the MVKConfiguration::fullImageViewSwizzle config parameter or MVK_CONFIG_FULL_IMAGE_VIEW_SWIZZLE environment variable.
E: VulkanImage.cpp:156: vkCreateImageView failed: VK_ERROR_FEATURE_NOT_PRESENT```

So for this to be fully usable on old macs we need to do some work either enabling that feature or getting rid of our use of texture swizzle, and apparently of our use of that 16-bit texture format.

**Comment by CarterLi** [comment]
We use libMoltenVK.dylib, not the framework

`VK_FORMAT_B4G4R4A4_UNORM_PACK16 is not supported on this device.`

Yes I also encountered this error for some game, thus I marked highly experimental ;)

Have no idea how to resolve it though...

**Comment by hrydgard** [comment]
We simply have to convert that texture to another supported format. Might be able to reuse the path we have to work around the lack of support for the 16-bit formats on Windows 7 / DX11.

**Comment by CarterLi** [comment]
Add support for iOS ( completely untested )

And found more info about VK_FORMAT_B4G4R4A4_UNORM_PACK16:

https://github.com/KhronosGroup/MoltenVK/issues/204
https://github.com/KhronosGroup/MoltenVK/issues/203
https://github.com/gfx-rs/portability/issues/125

**Comment by CarterLi** [comment]
> We simply have to convert that texture to another supported format. Might be able to reuse the path we have to work around the lack of support for the 16-bit formats on Windows 7 / DX11.

I have no idea how to do this. Would you mind merging my PR first and then work around this issue?

**Comment by hrydgard** [comment]
Yeah, sure, I'm not expecting you to take care of that :)

**Comment by hrydgard** [comment]
Hm, by the way, do you know if we need to do anything to make it use Metal 3, as opposed to Metal 1 or 2? I have no idea how versioning works in Metal-land....

**Comment by CarterLi** [comment]
> Hm, by the way, do you know if we need to do anything to make it use Metal 3, as opposed to Metal 1 or 2? I have no idea how versioning works in Metal-land....

Sorry, but isn't it chosen by MoltenVK internally?

**Comment by hrydgard** [comment]
Hm, right, most probably. Was curious because Metal 3 is supposed to fix that swizzle issue and it does not seem to be enabled on my (old) macbook pro, maybe it doesn't support it (or it's just a secondary error of the texture format not being supported).

**Comment by Gcenx** [comment]
Metal level is dependent on macOS version;
- Metal (OS X El Capitan)
- Metal 2 (macOS High Sierra)
- Metal 3 (macOS Catalina)

To force software swizzle you use the following export 

```
export MVK_CONFIG_FULL_IMAGE_VIEW_SWIZZLE=1
```

That or compile MoltenVK from source and [force swizzle](https://github.com/KhronosGroup/MoltenVK/issues/203#issuecomment-457083221)

**Comment by CarterLi** [comment]
> Metal level is dependent on macOS version;
> 
> * Metal (OS X El Capitan)
> * Metal 2 (macOS High Sierra)
> * Metal 3 (macOS Catalina)
> 
> To force software swizzle you use the following export
> 
> ```
> export MVK_CONFIG_FULL_IMAGE_VIEW_SWIZZLE=1
> ```
> 
> That or compile MoltenVK from source and [force swizzle](https://github.com/KhronosGroup/MoltenVK/issues/203#issuecomment-457083221)

Tested but it doesn't work for me.

---

Just found this

https://github.com/KhronosGroup/MoltenVK/blob/fd7b69b9e5ddd4bbbc7fda5f471dfe6c53858e80/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm#L866

And this

https://github.com/KhronosGroup/MoltenVK/blob/fd7b69b9e5ddd4bbbc7fda5f471dfe6c53858e80/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm#L1953

But the log for my device reads

```
[mvk-info] GPU device:
		model: AMD Radeon R9 M370X
		type: Discrete
		vendorID: 0x1002
		deviceID: 0x6821
		pipelineCacheUUID: 00002738-0300-07D2-F981-8F0804E6E73E
	supports the following Metal Versions, GPU's and Feature Sets:
		Metal Shading Language 2.2
		GPU Family Mac 2
		GPU Family Mac 1
		GPU Family Common 3
		GPU Family Common 2
		GPU Family Common 1
		macOS GPU Family 2 v1
		macOS GPU Family 1 v4
		macOS GPU Family 1 v3
		macOS GPU Family 1 v2
		macOS GPU Family 1 v1
		macOS Read-Write Texture Tier 2
```

Seems no luck for me :(

hrydgard:
I'll open a new issue for the swizzle thing.

PR #12589, opened 2020-01-22
Don't shut down camera when exiting games.

### LINKED ISSUE #12588: Windows Camera code shuts down prematurely, causing random crash going into settings
Discovered by Sorer, Azuka and Zakiji from our Discord.

Currently the whole camera system is shut down after a game exits, from Camera::stopCapture in sceUsbCam (both STOP and SHUTDOWN is called). This is not ok because later you might go into settings, which enumerates the cameras - and if the camera system is gone, we're toast :)

@shenweip :)

unknown

shenweip:
Fixes #12588, the crash is reasonable I think, but i can't trigger it in release build, it's odd, maybe the way `CoTaskMemFree` works is not what I think?

We might also need to handle `WM_DEVICECHANGE` later I think.

### DIFF
--- a/Core/HLE/sceUsbCam.cpp
+++ b/Core/HLE/sceUsbCam.cpp
@@ -319,7 +319,6 @@ int Camera::stopCapture() {
 	#elif defined(_WIN32) && !PPSSPP_PLATFORM(UWP)
 		if (winCamera) {
 			winCamera->sendMessage({ CAPTUREDEVIDE_COMMAND::STOP, nullptr });
-			winCamera->sendMessage({ CAPTUREDEVIDE_COMMAND::SHUTDOWN, nullptr });
 		}
 	#else
 		ERROR_LOG(HLE, "%s not implemented", __FUNCTION__);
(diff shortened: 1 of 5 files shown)


hrydgard:
I don't think this is the right fix.

The issue is that when you start the emulator, the camera system launches, but then when the first game shuts down, the entire camera system also gets shut down, even though it should only have stopped recording. 

From Camera::stopCapture in sceUsbCam.cpp:
```
winCamera->sendMessage({ CAPTUREDEVIDE_COMMAND::STOP, nullptr });
winCamera->sendMessage({ CAPTUREDEVIDE_COMMAND::SHUTDOWN, nullptr });
```

Later when accessing settings, the camera system is down so we are not really able to enumerate the cameras (even though it sometimes works anyway, it's unstable).

I believe that we should only send `CAPTUREDEVIDE_COMMAND::STOP` here, and send `CAPTUREDEVIDE_COMMAND::SHUTDOWN` only when the whole emulator has exited.

Alternatively, we should start up the camera system properly in the settings menu so that the cameras can be safely enumerated.

hrydgard:
Yeah, that looks better! Thanks!

Issue #11852, opened 2019-02-27
[Feature Request] Some ideas for (touch) controls


leopard20:
I had a few ideas for the controls (esp. touch controls) and I was gonna give them a whirl myself before posting them here but I'm having issues with Android Studio at the moment so I decided to go ahead and put them here, see if anyone else is interested in them!

**1. Rapid Fire:** As the name goes, it presses a certain key for n times each T seconds in succession. 
**2. Adjustable deadzone for the touch analog stick:** Posted something similar here https://github.com/hrydgard/ppsspp/issues/10808 but I think the deadzone of the analog stick should be adjustable in general and not only for tilt input.
**3. Second analog stick** that could be mapped to either action buttons (triangle,X,square,circle) or the D-Pad.
**4. Simulate analog input for the second analog stick:** This one may not be possible as I didn't give it enough thought! But it would be very cool if it could happen.
Let me elaborate with an example. Let's say the stick is supposed to simulate analog input for the D-Pad buttons, and it is being pushed halfway toward the upper-right corner at 15 degrees from the "up" direction. To simulate an analog input, we'd have to press the keys multiple times in succession according to where the stick is located. The distance from the deadzone would be used as a "delay" parameter and the angle as the number of times each key should be pressed. E.g. since the current angle is 15 degrees from "up" and 75 degrees from "right", and the ratio of those is 1:5, then the "up" key would have to be pressed 5 times and the right direction simply once (for each timestep, of course). These could supposedly be handled by the "rapid fire" function maybe!

I think all these feature could enhance the touch control experience.

Hasster1:
The third one is what i need right now.
In star wars: battlefront 2 there's a "Retro" input style, that let you move with action buttons, and look around with analog stick.
And it's not very comfortable to walk with those buttons.

Gamemulatorer:
This wasn't cover by https://github.com/hrydgard/ppsspp/commit/57b4cd394a69029dcb8943e1bad33a4d149dcf99?

PR #12601, opened 2020-01-28
Add rapid fire to touch control

iota97:
Add a toggle rapid fire button to touch control
First point of #11852


hrydgard:
Yeah, I guess this could be useful, and the implementation is pretty clean.

PR #12611, opened 2020-02-03
Use ASharedMemory api instead of direct access to /dev/ashmem

Swordfish90:
Hi,

I wasn't able run the ppsspp libretro core in an application with targetSdk set to 29.

Direct access to /dev/ashmem has been deprecated with Android 10: https://developer.android.com/about/versions/10/behavior-changes-10#shared-memory

This fixes the issue, but sadly the clean way to use ASharedMemory directly requires bumping minSdkVersion so I had to fallback to use dlopen().

Feel free to let me know if you want me to change something in this pull request.

### DIFF
--- a/Common/MemArenaAndroid.cpp
+++ b/Common/MemArenaAndroid.cpp
@@ -30,6 +30,8 @@
 #include <cerrno>
 #include <sys/ioctl.h>
 #include <linux/ashmem.h>
+#include <dlfcn.h>
+#include <sys/system_properties.h>
 
 // Hopefully this ABI will never change...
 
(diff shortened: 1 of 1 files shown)


hrydgard:
Thanks, looking good!

Instead of your get_sdk_version, though, please include "UI/NativeApp.h" and use System_GetProperty(SYSPROP_SYSTEMVERSION). Just trying not to have too many ways of doing the same thing...

Swordfish90:
There you go. Sorry, I missed that while grepping around the code.

hrydgard:
That's alright, thanks :)

PR #12612, opened 2020-02-04
SDL analog mouse input

iota97:
Add analog mouse input to SDL, adding another platform to  #4650.

### DIFF
--- a/SDL/SDLMain.cpp
+++ b/SDL/SDLMain.cpp
@@ -67,9 +67,6 @@ static int g_DesktopWidth = 0;
 static int g_DesktopHeight = 0;
 static int g_RefreshRate = 60000;
 
-static float g_mouseDeltaX = 0;
-static float g_mouseDeltaY = 0;
-
 int getDisplayNumber(void) {
 	int displayNumber = 0;
 	char * displayNumberStr;
@@ -683,6 +680,9 @@ int main(int argc, char *argv[]) {
 	}
 	graphicsContext->ThreadStart();
 
+	float mouseDeltaX;
+	float mouseDeltaY;
+	bool mouseCaptured = false;
 	bool windowHidden = false;
 	while (true) {
 		double startTime = time_now_d();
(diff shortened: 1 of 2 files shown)


SDL/SDLMain.cpp
```diff
@@ -680,6 +680,9 @@ int main(int argc, char *argv[]) {
 	}
 	graphicsContext->ThreadStart();
 
+	float mouseDeltaX;
```
hrydgard: These should be zero-initialized, to avoid the risk of a nonsense axis event being sent when turning on mouse control.

SDL/SDLMain.cpp
```diff
@@ -894,6 +897,10 @@ int main(int argc, char *argv[]) {
 					input.id = 0;
 					NativeTouch(input);
 				}
+				if (g_Config.bMouseControl) {
```
hrydgard: You don't really need to check g_Config.bMouseControl here, doesn't seem harmful to always update the deltas.

iota97:
Fixed

iota97:
Sending KEY_UP right after was too fast for any game I tried (was working only on the bind menu), added mouse 3, 4 and 5 press as well to match windows version.

hrydgard:
Hm, what if there's not a next cycle? Maybe a timer (frame counter) to send a KEY_UP after a while?

iota97:
If in the main loop, while (true), shouldn't there be always another cycle (unless the emulator get closed at least)?

hrydgard:
Oh never mind, for some reason I thought you meant next time you get a mousewheel event.

Never mind, it seems right (but only one frame of down state might not always be enough).

iota97:
Yeah, made it 5 (may be a bit too much, but it allow to simulate a hold if scrolled fast, still can be reduced)

hrydgard:
Yeah, we can tweak that as needed. Let's merge.

PR #12617, opened 2020-02-05
libretro: Fix Android Builds

webgeek1234:
About a week ago, Android builds started failing on my libretro mirror. The weird thing is, I can't point at a commit or change in infra that caused the failures. But when I went digging, I saw that the flow of the cmake scripts was causing the standalone ui to be built into the libretro android cores, which is certainly not desired.

Secondarily, the camera changes a few days ago didn't account for libretro either. On non-mobile platforms, what was done will work since the calls can still be made, although not ideally since libretro should handle all hardware interactions with a core. But on mobile, the function calls are in the native ui code which isn't available for libretro. So for now, just adding a dummy impl to get things to compile again. Someone later may wish to do a proper impl to interface with the libretro camera api.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -718,7 +718,7 @@ if(ANDROID)
 		Common/GL/GLInterfaceBase.h
 	)
 
-	set(nativeExtra ${nativeExtra}
+	set(NativeAppSource ${NativeAppSource}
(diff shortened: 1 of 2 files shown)


hrydgard:
thanks for the fix!

PR #12621, opened 2020-02-11
Add frame time graph to develop menu

iota97:
Add a frame time graph to Developer menu

It only update the value when `CalculateFPS()` is called, I am totally NOT sure if is enough or the right place to make the calculation, in case let me know if there is a better place/way :)

I hope it can help see the frame pacing better.

Screenshot_2020-02-11_10-28-33


hrydgard:
Well, there are many ways to measure frame time, but this definitely will show some stuff...

We actually already have a similar thing that's disabled compile-time, the "profiler". I should probably put some effort into enabling that at least in buildbot builds...

iota97:
Should I just close this then? I totally missed the profiler being there...

hrydgard:
Nah, I like this as a low-overhead (but lower-information) alternative. I'm gonna go ahead and merge, and maybe if I can get the profiler to be good enough and fast enough to replace it I'll remove it again.

PR #12623, opened 2020-02-12
Add toggle flag to combo button

iota97:
Add toggle mode to combo button (the combo button it self will not be hold in down animation, as it would make little sense as other input can override the held and there might be multiple button in the combo).

Should help use cases as #10159 a bit.

EDIT: it swap button separately if one is pressed and the other not when is activated, i dunno what should be the expected behaviour with multiple key toggle tho' (remember the last status and change them all? Loop to check if all have been released by other input before?).


hrydgard:
Hm, are you happy with this as-is? Can't quite tell from the description.

iota97:
Yeah, I think even if it can be a bit weird with multiple button other solution will end up more confusing/complex with a single one (and i think one button toggle it's a more common use case than multiple ones).

If you want to check how it is now the apk mentioned here have it: https://github.com/hrydgard/ppsspp/pull/12486#issuecomment-592730796

PR #12637, opened 2020-02-23
Region flag and game ID on game selection screen

### LINKED ISSUE #12636: [Feature Request] Game Region Thumbnail Icons in Recent + Games lists
I have multiple copies of the same game, all for different regions, and they all share the same thumbnails for the "Recent" and "Games" menus in PPSSPP.
For example, I own 3 copies of Ace Combat X -- ULUS10176 (US),  UCES00423 (EU), and ULJS00086 (JP).

This makes it a bit tedious with finding out which version is which region, so I always have to right-click them to view the information page, so that I can see the game region and Game ID to identify which version I want to launch.

My suggestion? 
An optional, small visual overlay for thumbnails that changes depending on the game's detected region.
It could function similarly to the "game config" gear icon, like so:
region-flag-example
(This mockup is using free public domain 24x24 pixel region map icons)

This certainly isn't a _major_ issue by any degree, but could definitely prove useful for other users in the same boat. Thoughts and opinions?

iota97:
Add small 22x14 flags icon (found online under creative common license) based on game region if enabled in the setting (Asia region uses China flag).

Screenshot_2020-02-23_14-10-44

Fixes #12636.

PS: Atlas file have been generated on Linux, it seem to have worked fine, but it might be better rebuild it on windows for extra safety?


Abbanon:
Might I suggest adding the Korean/KO flag as well? 
They have their own specific Game ID region codes, `UCKSxxxxx`
kr

DavidTamas:
The Region of a game disc is a separate property and doesn't always correspond with the Disc_ID property.
kép

A Korean flag is indeed necessary, and I would also add something for when it couldn't be determined because the emulator seems to guess it from the third character of Disc_ID, meaning it's not gonna be accurate.
https://github.com/hrydgard/ppsspp/blob/6610cd7663811442724cf3df1179f037655f33f0/UI/GameInfoCache.cpp#L262

iota97:
Added korea region, both as flag and in game info.

About detection accuracy it seem from this line `// region = paramSFO.GetValueInt("REGION");  // Always seems to be 32768?` that it's not easy to find it and that was a workaround, not sure if there even is a better way tbh.

Abbanon:
It looks like there are at least 5 region types for PSP games:
- us US (USA) 

- eu EU (EUR) 

- jp JP (JAP)
- kr KO (KOR)

- cn AS (ASIA)

And each region has several different Game ID/UMD Serial codes.
For example, 
US can have: `ULUS`, `UCUS`, `NPUZ`, `NPUX`, `NPUF`, `NPUH`, `NPUG`
EU can have: `ULES`, `UCES`, `NPEZ`, `NPEX`, `NPEH`, `NPEG`
JP can have: `ULJS`, `ULJM`, `UCJS`, `UCJM`, `UCJB`, `NPJJ`, `NPJH`, `NPJG`
KO can have: `ULKS`, `UCKS`, `NPHH`, `NPHG`
AS can have: `ULAS`, `UCAS`
(Not sure if that's all of them, but it's all that I'm finding)

If a game has an un-identifiable region, why not leave it blank/without a region flag?

iota97:
PPSSPP detect the region (as shown on game info screen) by checking the 3rd letter of the code (notice how all the ID you listed have the same 3rd letter) as this:

U - USA
E - Europe
J - Japan
A - Asia
H - Hong kong
K - Korea (this is not detected in master, just added in this PR)

If the letter doesn't match any of those the region will be "other" and no region will be written on game info screen (and no flag will be displayed in this PR).

DavidTamas:
There is already a method in the code for detecting the region based on specific ID's but it's commented out in favor of the 3rd-character detection.
https://github.com/hrydgard/ppsspp/blob/6610cd7663811442724cf3df1179f037655f33f0/UI/GameInfoCache.cpp#L270

However, I still don't believe this is going to be accurate. I see multiple region types for NPHH ID's in a database.
kép

Perhaps we should look into how UMDGen determines the region?

Abbanon:
It looks like UMDGen determines based on the PARAM.SFO file contents.
Untitled

iota97:
NPHH will be detected as Hong kong, is that a possible region in UMDGen? 
Also if UMDGen have is source open and available it might be possible to take a look, but even then is it worth the time?
The reason of the issues was to see the region if you have multiple version of the SAME game, accuracy is better for sure, but i got the feeling we are getting away from the issues itself.

DavidTamas:
If that's the reason then the base assumption was wrong to begin with: games can have multiple versions and rereleases for the same region.

Take a loot at GTA Liberty City Stories which has:
* 3 US
* 3 European
* 2 German
* 3 Japanese
* 1 Korean
versions and that list only includes physical releases.

A better method to tell the game version would be to use a combination of Disc_ID and Disc_Version.
An even better solution would be the calculation of the CRC-32 checksum of the game upon booting it as suggested in #12469 . This would even be able to tell if the user is loading a bad dump found from the internet, which the other methods wouldn't be able to determine.

DavidTamas:
We've gotten a bit off-topic here, this PR is a nice addition for the users of the emulator. However, game region alone is not reliable for telling specific game versions apart. As for how PPSSPP determines the region should be discussed in a new issue.

Abbanon:
Okay, so looking more into UMDGen:
I didn't notice this before, but among the PARAM.SFO information it explicitly states the region.
`NPHH` shows up as `N/A`
Meanwhile, 
`ULUS` shows `Ntsc`, 
`ULES` shows `Pal`, 
`ULJS` shows `Japan`, and 
`ULKS` shows `Korea`
I have no idea how it determines this based on data, however. 
Untitled

iota97:
Added game ID and version (not sure if any game have different feature in different version in the same region tho', and that what i thought it was the main use case). Still both flag and ID are optional and disabled by default so I see no harm in having them.

LunaMoo:
@iota97 you should probably avoid those features for Homebrew as we generate fake ID for them. Pretty much everything launched from "Game" folder might have a fake ID(althrough the detection is also checking some other things, it's overally recommended for compatibility to not place any commercial games inside that folder while always placing homebrew only there), we do assign them a "HOMEBREW" region if I recall, but there's also some still unsolved race issue where homebrew is not detected at recent screen on windows when using "\" for path divider.

hrydgard:
I like the flag idea as long as we can get consensus on the region detection. Showing the game ID is maybe kinda odd but I guess sometimes useful.

How does this look in list view?

iota97:
Screenshot_2020-02-24-09-56-48-388_org ppsspp ppsspp

Abbanon:
Considering List View has such smaller game thumbnails, what if the "game config" gear and region flags were moved to the left, between the thumbnail and the edge of the application window?
I threw this mockup together to explain what I mean:
listview-example

iota97:
Scaling the font down a bit seem fine (I had forgotten to check the list view, my bad)

hrydgard:
I have to say that does look a little too cluttered. I'd prefer Abbanon's approach, but I'm also fine with just leaving them out in list view for now, and tweak that look later.

iota97:
Let's go for Abbanon way, game ID will be placed in bracket after the tittle in list view.

DavidTamas:
I like what you did with the icons in list view. Maybe we can display the game_id+version in smaller text size so it takes up less space. Could omit the brackets too that way.

For the grid view, would it be possible to highlight the text with a semi-transparent color for the lighter icons?

iota97:
I guess it's better to display the ID on the right as another field as this:





About highlight, the text already have a shadow so it look like this on light icon:

Abbanon:
The only thing I could suggest further would be to shift the List View thumbnails to the right a tiny bit, so the gear and flag icons don't seem so squished into the edge.
Otherwise I think this is perfect already!

iota97:
Added some spacing there and better aligned the ID on the grid view. Here the result:

g
l

hrydgard:
Now we're getting somewhere :)  I like it.

DavidTamas:
Had to wait for a build until I could see for myself, but what I feared about the text color not always being readable became true. This is why I suggested a highlight color, even semi-transparent could work if it's possible to do.
kép

Loving the update otherwise and the other things you've been working on too. Keep it up!

PR #12641, opened 2020-02-24
Improve stability on non-jailbroken devices

saagarjha:
This is largely untested; I have checked that it builds and doesn't crash when launching a non-jailbroken and jailbroken device and that's about it. I got some linking errors for the private functions we're using so I changed them to be looked up at runtime; I'm not sure how this was supposed to work before.


hrydgard:
Disclaimer: I don't develop actively on iOS myself and am not really familiar with these tricks. That said, it looks very plausibly correct, so let's try it!

PR #12646, opened 2020-02-26
Resizable game icons

iota97:
As mentioned in #12353 game icon size may end up looking too small.

Design wise I'm not sure where to place the "+" and "-" button (and there might be a better way to center the text on them than 2 spaces), it used the topBar enabling it partially for "recent game" with 
+, -, and grid/list mode (icon size affect grid only).

Some screenshot on a 1080p display (the scale is a bit extreme here, it's more to give an idea)

base standard size

b increased size

I also wonder if the icon look too blurry to use this way and some better scaling might be needed.

### DIFF
--- a/Core/Config.cpp
+++ b/Core/Config.cpp
@@ -450,6 +450,7 @@ static ConfigSetting generalSettings[] = {
 
 	ConfigSetting("ShowRegionOnGameIcon", &g_Config.bShowRegionOnGameIcon, false, true, true),
 	ConfigSetting("ShowIDOnGameIcon", &g_Config.bShowIDOnGameIcon, false, true, true),
+	ConfigSetting("GameGridScale", &g_Config.fGameGridScale, 1.0, true, true),
 	ConfigSetting("GridView1", &g_Config.bGridView1, true),
 	ConfigSetting("GridView2", &g_Config.bGridView2, true),
 	ConfigSetting("GridView3", &g_Config.bGridView3, false),
--- a/Core/Config.h
+++ b/Core/Config.h
@@ -211,6 +211,7 @@ public:
 	int iShowFPSCounter;
 	bool bShowRegionOnGameIcon;
 	bool bShowIDOnGameIcon;
+	float fGameGridScale;
 
 	// TODO: Maybe move to a separate theme system.
 	uint32_t uItemStyleFg;
--- a/UI/GameSettingsScreen.cpp
+++ b/UI/GameSettingsScreen.cpp
@@ -722,9 +722,6 @@ void GameSettingsScreen::CreateViews() {
 #endif
 
 	systemSettings->Add(new CheckBox(&g_Config.bCheckForNewVersion, sy->T("VersionCheck", "Check for new versions of PPSSPP")));
-	if (g_Config.iMaxRecent > 0)
-		systemSettings->Add(new Choice(sy->T("Clear Recent Games List")))->OnClick.Handle(this, &GameSettingsScreen::OnClearRecents);
-
 	const std::string bgPng = GetSysDirectory(DIRECTORY_SYSTEM) + "background.png";
 	const std::string bgJpg = GetSysDirectory(DIRECTORY_SYSTEM) + "background.jpg";
 	if (File::Exists(bgPng) || File::Exists(bgJpg)) {
@@ -987,12 +984,6 @@ UI::EventReturn GameSettingsScreen::OnSavePathOther(UI::EventParams &e) {
 }
 
 #endif
-
-UI::EventReturn GameSettingsScreen::OnClearRecents(UI::EventParams &e) {
-	g_Config.recentIsos.clear();
-	OnRecentChanged.Trigger(e);
-	return UI::EVENT_DONE;
-}
 
 UI::EventReturn GameSettingsScreen::OnChangeBackground(UI::EventParams &e) {
 	const std::string bgPng = GetSysDirectory(DIRECTORY_SYSTEM) + "background.png";
--- a/UI/GameSettingsScreen.cpp
+++ b/UI/GameSettingsScreen.cpp
@@ -722,8 +722,6 @@ void GameSettingsScreen::CreateViews() {
 #endif
 
 	systemSettings->Add(new CheckBox(&g_Config.bCheckForNewVersion, sy->T("VersionCheck", "Check for new versions of PPSSPP")));
-	systemSettings->Add(new CheckBox(&g_Config.bShowIDOnGameIcon, sy->T("Show ID on game selection screen")));
-	systemSettings->Add(new CheckBox(&g_Config.bShowRegionOnGameIcon, sy->T("Show region flag on game selection screen")));
 	if (g_Config.iMaxRecent > 0)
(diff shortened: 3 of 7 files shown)


Abbanon:
What if the Texture Scaling feature (xBRZ, Hybrid, etc) could be applied to these thumbnails at increased sizes? Maybe have them fully generate as filtered .png files upon being enabled so it won't be an active filter feature that runs in the background.

iota97:
Texture Scaling would be nice, sadly I don't think I'll be able to implement it with as my limited skill.

hrydgard:
I think the icons can generally stand a 2x scale but not very much more.

Yeah the tricky part is the UI. On PC we could support Ctrl+mousewheel which is zoom in most apps, and on mobile with multitouch, two finger drag, but there are more platforms than that, and adding buttons gets a little ugly. Maybe we could simply have yet another normal setting, with no immediate adjustment if the other options aren't available.

iota97:
Getting all the view to update seem trickier than I hoped too, currently it iterate the over the vector with the gameBrowser to call Refresh() on them all, but this miss the remote ISO screen (and maybe replace UMD too, but I'm not sure what that view is) updating the scale.

I'm a bit lost on how to handle this well too tbh.

hrydgard:
Maybe this will look awkward, but you could try
```screenManager()->topScreen()->RecreateViews();``` to recreate everything.

iota97:
screenManager() is not in scope (missing some header?), anyway i guess Draw override will make thing easier with other binding.

hrydgard:
Ah I see you solved it, never mind.

Abbanon:
I have an idea! What if...

Instead of increasing the scale of the Thumbnail Art, just increasing the scale of the thumbnail/tappable area itself by adding an Info Box?

Here are a few examples I whipped together:
(Box highlighted in blue indicates what the cursor is currently selecting)

**1x size:** (keeps thumbnail art the same, adds info box underneath)
example-1

**2x size:** (increases thumbnail art size to 2x, adds expanded info box, gear and flag are moved to info box, Game ID added to info box)
example-2

Honestly, being able to scale Thumbnail Art and Info Box sizes independently from each other would probably be a good idea too.
So you could have the expanded 2x Info Box with just the 1x Thumbnail Arts, for example.

Does this seem like a good idea?

iota97:
Info box could work, but as it's a separate thing, I think it's better to just make the current grid view scalable before adding another style.

iota97:
\+ and - button will be visible in grid mode, on top bar if present or on the side otherwise:

r
l
g

iota97:
I'm fine with the result atm (still any suggestion is welcome on both code and UI design). I'll rebase when ui atlas change get merged.

Abbanon:
It almost looks like a scaling filter is being used. Is it?

iota97:
Well if a image get scaled new pixel need a value, here should just be bilinear interpolated nothing more. I guess is just isn't nearest-neighbor, you may check a gallery [here](https://en.wikipedia.org/wiki/Comparison_gallery_of_image_scaling_algorithms).

Abbanon:
In my personal opinion, nearest-neighbour might be ideal for some users. 
Would it be too much to ask for an option to "Force nearest-neighbour when scaling game icons"?

hrydgard:
I'm starting to think that instead of the two little plus and minus button, we should just make a button that pops up a little menu with the display settings for the gamelist. They could then be removed from the System settings where they don't really belong. There we could put a checkbox for filtering. The problem with nearest here is that it's only appropriate if the scale is exactly 1x, 2x etc, but due to DPI etc that will at least on mobile rarely be the case whatever you do. On Windows though if your DPI is 100% it could work...

iota97:
Might a menu like this work? With a little menu button rather than "Game list setting one" may be.

Screenshot_2020-03-01_13-00-51

Screenshot_2020-03-01_13-01-16

hrydgard:
I thought something more lightweight, like the ingame Dev menu popup or so, not a whole screen. I don't want a major new category for this, it's too much visual space for something I think few people with bother with.

iota97:
I guess this is starting to look better.

b
a

Few note: 
- There are 3 config option for grid/list mode not sure if is better to display all, only the relevant one or merge them in "global" setting.
- Needed to change GameBrowser contructor to pass screenManager() to it.
- Using a slider pop up would make the main screen black (I guess too many stucked popup).
- Clear game list is still in system setting, gotta take a better look to that button logic to move it.
- I wonder if there is a color palette option for this pop up menu style.

hrydgard:
Hm, yeah, didn't think about the slider popup darkening thing... Maybe need to do something about that. Anyway, getting there! I think we should probably keep the old buttons for list vs icons though, in addition to this one? maybe this one can use the gear icon for now to not clash (though should think about a better one)?

Yeah, the colors on this style of popup aren't great. Can always tweak little things like that after merge though.

iota97:
Added the old button back, changed icon with the gear one and rebased on master (I squashed commit together to reduce the conflict, I still have the old history on a local branch if needed).

unknownbrackets:
I wonder why the height is filling vertically rather than wrapping content, but we can always investigate later.  I think showing the settings inline here makes a lot of sense.

-[Unknown]

iota97:
Moved the clear recent button as well and some minor renaming (no reason to say show flag on game list in a game list only option).

a

b

hrydgard:
Starting to look really really good! I'm happy to merge as-is and tweak the color scheme afterwards. Those popup windows really are ugly, matching the colors of Android 3.0 Honeycomb... :P

Abbanon:
I would suggest keeping the gear and flags from scaling with the thumbnails, and instead keeping them at their original static size since thy're already so miniscule (32x32 and 24x14, respectively)

iota97:
I think not scaling them would be worse as they will look too small over a big icon.

Drasikk:
@hrydgard You can't see the last option.

Screenshot_20200303-080236_PPSSPP

Abbanon:
Can confirm, under a certain window size the "Clear Recent games list" option is cut off from view. Can't scroll down to see it.

Additionally, there's a scroll bar appearing in Recent, and I'm able to scroll up/down, despite not having enough thumbnails to allow scrolling:
Untitled

iota97:
Mhhh, removing that weight caused some issues I didn't notice, my bad. Adding it back fix them, I'm not sure what's going on tho'.

hrydgard:
I merged your workaround, I'm gonna clean this up visually later. For now it still works pretty well.

Drasikk:
Arranged. Thank you!

Screenshot_20200303-123323_PPSSPP

unknownbrackets:
I generally like this change, but when the window is 1x or 2x (the default), I think it looks really awkward how there's three games per row and a bunch of space to the right of them.

Obviously a bunch of the settings are related to the Recent list, but I wonder if it'd be an option to hide it on the Recent tab.  Or, to put it above the game icons (like on other tabs) when we'd show less than 4 games per line.

-[Unknown]

iota97:
Putting it above will make a weird space on top sadly.

We could maye add a checkbox to hide/show the button on recent tab (both in these setting and in system one to re-enable it).

hrydgard:
@unknownbrackets It annoys me too, but I have a plan for that, stay tuned.

PR #12667, opened 2020-03-02
Chat feature based on Adenovan's Rechat branch

anr2me:
This is based on @adenovan PR https://github.com/hrydgard/ppsspp/pull/9090
I only fixed the conflicts and made minor tweaks

PS: There were errors related to missing `releaseButtons()` function, i don't know what it is used for, so i removed it LOL :D


hrydgard:
Cool, thanks for reviving this!

There's a few minor things I'd like to tweak and modernize in this code, but I'll take care of it myself and merge later this week.

adenovan:
https://github.com/adenovan/amultios-ppsspp/commit/94369d6037e373d5fbfe96bc2ab5681c8746b1f8

actually there is more good looking gui that use onscreen touch control on that commit in different branch but its need to recompile atlas tools.

and on latest i worked on the chat is on pause screen but this chat use different protocol can reuse the gui if you want
https://github.com/adenovan/amultios-ppsspp/blob/amultios-osx/UI/AmultiosOverlayScreen.cpp

hrydgard:
Well dealing with the atlas should be much easier now - the tools are now easily built.

I think let's merge this, then we can look at merging your newer GUI after that.

PR #12669, opened 2020-03-03
Scroll view fix - workaround

iota97:
As mentioned here https://github.com/hrydgard/ppsspp/pull/12646#issuecomment-593800495 it seem removing the weight make the scroll view miss the last item when scrolling, weight and wrap content make the popup all screen long, weight and a small size make the popup right in both situation (not sure why it work tbh).

There is probably some other issues underneath, dunno where tho'.


hrydgard:
I'll try to fix this properly later, but this is fine for now.

PR #12670, opened 2020-03-03
Attempts to replace 0 frame width with valid frame width.(sceMpegAvcCsc)

### LINKED ISSUE #12627: [Android][OpenGL&Vulkan][Black out] SMASH COURT TENNIS3  (ULJS00098)
### What happens?
Black out lntro movie.
Sound is Playing.
It coming title page when finish the intro movie.

### What *should* happen?
Play the movie collectly.

PPSSPP Ver. : 1.9.3-282
OS: Android 9.0
GPU: Adreno 540
Resolution: 720×1280
RAM: 4GB
ROM:32GB

shenweip:
Fixes #12627.


hrydgard:
Looks reasonable. In theory this should be tested on hardware but that might be quite difficult, and this should be quite safe so I'll just go with it.

PR #12686, opened 2020-03-08
Vulkan: Fix minor vulkan resource leak if inflight frames isn't max.

hrydgard:
This gets changed after the rendermanager is created.


unknownbrackets (approved):
Ah, I thought I'd checked and everywhere that freed was doing it like OpenGL - always using max to free.

PR #12688, opened 2020-03-08
UI: Temporary solution to the layout issue: Remove the gear button from the recent screen.

hrydgard:
Just until we can figure out a better solution, the current look isn't good.

Can always use the gamelist menu from the other screens.

I'm thinking about maybe putting it next to the tabs instead? Or in the very top right corner?


unknownbrackets (approved):
Agreed.  I thought about tabs too, but thought it'd get messy trying to put it there.  I guess it wouldn't if we anchored it...

-[Unknown]

PR #12723, opened 2020-03-17
cmake: Disable discord-rpc for libretro builds.

orbea:
I don't believe libretro needs this in ppsspp, it is already in RetroArch and should be part of the frontend.

@webgeek1234 Can you confirm?


webgeek1234:
Affirmative.

unknownbrackets (approved):

PR #12727, opened 2020-03-18
Correcting return value of some mipscalls without after action.

### LINKED ISSUE #12726: Jeanne D'arc hangs on startup
Report in https://github.com/hrydgard/ppsspp/pull/12709#issuecomment-600424103.

shenweip:
Fixes #12726, the V0 mipscall saved is the return value from last call instead of current syscall, so likely to return a wrong value. This seems the simplest way to fix, not technical enough though...


unknownbrackets (approved):
I think this is okay for now, but it's a bit of a hack.

Probably the right thing to do is move the mips call handling to an hleAfter flag, so v0 is handled automatically before it happens (like with rescheduling.)

-[Unknown]

hrydgard:
Cool. I wasn't sure whether we had a better alternative to this already..

PR #12749, opened 2020-03-22
Auto rotating analog

### LINKED ISSUE #12315: [Feature request] automatic analog rotation by pressing a key
some games demand you to rotate the analog, which in my case fail all the time on keyboard or be very frustrating whit the touch screen, i'am in gow: ghost of sparta and i can't continue because of this. so as this feature would be very helpful: you press a key and the analog rotate counter/clockwise.

iota97:
Not sure what to set the speed (both default and config limit). ~~Also I wonder if it's better to have 2 button separate for clockwise and anti-clockwise (setting negative speed will do, but some game might require both I guess).~~

Fixes #12315.


unknownbrackets (approved):

hrydgard:
Yeah this looks like a good implementation of the idea.

PR #12766, opened 2020-03-26
Rember cheat scroll position

### LINKED ISSUE #11945: Remembering cheat menu scroll state
When I'm trying out new cheats from cheats.db I have to scroll all the way through the enormous list , enable the cheat , try it and then scroll again , find it and then disable it. Would be better if the cheats menu opened up at the place you last left it.

iota97:
Closes #11945.


hrydgard:
Okay!

PR #12767, opened 2020-03-27
Sticky touch analog

iota97:
Will make the touch analog keep his position after the finger goes up, can be reset to center by tapping it.

I'm not sure how much useful this could actually be, but I guess still might have it's use cases in game where you need to perform action while using the analog at the same time without having to claw.

### DIFF
--- a/UI/GamepadEmu.h
+++ b/UI/GamepadEmu.h
@@ -174,6 +174,7 @@ private:
 
 	float centerX_;
 	float centerY_;
+	float lastTouchDownTime_;
 };
 
 //initializes the layout from Config. if a default layout does not exist,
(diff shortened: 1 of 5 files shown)


hrydgard:
What do you find it the most useful for? Because I assume you have a use case yourself :)

iota97:
Basically I need the player to keep moving while turning the camera with the D-PAD.

It take a bit to get used to it (mostly because MHFU require to have the analog all like 90% of his range for moving and if you to move again in another direction click a bit to close to the center it stop the player before moving again, that depend on the game tho') but it still play better than having to stop everytime one want to move the camera.

UI/GamepadEmu.h
```diff
@@ -174,6 +174,7 @@ class PSPStick : public GamepadView {
 
 	float centerX_;
 	float centerY_;
+	float LastTouchDownTime_;
```
hrydgard: we start member variables with lower case letter.

leopard20:
Well, if you put the Dpad on the right you can do that without going to so much trouble. That's how I play the game.

iota97:
In the end this is probably more confusing than good, as such I'm closing it.

PR #12778, opened 2020-03-30
Set target to Android 10 (SDK 29).

### LINKED ISSUE #11997: Android 12 scoped storage hell
Android Q will introduce "scoped storage", which imposes really heavy restrictions on external storage. And in the version after Q, it will be a requirement for all apps, even those who target Android-28 (which PPSSPP will do as long as possible). So we will be able to limp along for a while longer without it, but next year, new devices will start requiring it, and it'll greatly worsen the user experience. We need to find ways to make it as workable as possible.

For us, the filesystem really is part of the UX to match the real PSP with an USB connection.

Users will additionally have problems if they install the app on an older OS, upgrade to Q, and then downgrade to an older version of the app (which is done sometimes to bisect bugs, etc). 

The one bright point is that the scoped storage APIs allows getting file descriptors so I/O performance will at least not be hampered much - assuming the file descriptors are unrestricted with regards to seek etc.

https://android-developers.googleblog.com/2019/04/android-q-scoped-storage-best-practices.html

EDIT: I have requested a reprieve through AndroidManifest.xml, the hammer drops with Android 12.

hrydgard:
Includes a nasty SDK bug workaround.

Setting android:requestLegacyExternalStorage="true" so we can still access storage properly. This is the last and only SDK version in which we'll be able to do this - when we must target Android 11, or run on Android 12, we need to fix https://github.com/hrydgard/ppsspp/issues/11997.

### DIFF
--- a/android/src/org/ppsspp/ppsspp/LocationHelper.java
+++ b/android/src/org/ppsspp/ppsspp/LocationHelper.java
@@ -13,6 +13,8 @@ import android.os.Build;
 import android.os.Bundle;
 import android.util.Log;
 
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.util.Iterator;
 
 class LocationHelper implements LocationListener {
@@ -25,6 +27,8 @@ class LocationHelper implements LocationListener {
 	private GpsStatus.NmeaListener mNmeaListener;
 	private float mAltitudeAboveSeaLevel = 0f;
 	private float mHdop = 0f;
+	private Method addNmeaListenerMethod = null;
+	private Method removeNmeaListenerMethod = null;
 
 	LocationHelper(Context context) {
 		mLocationManager = (LocationManager) context.getSystemService(Context.LOCATION_SERVICE);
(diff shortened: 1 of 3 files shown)


android/src/org/ppsspp/ppsspp/LocationHelper.java
```diff
@@ -100,7 +114,16 @@ void stopLocationUpdates() {
 				mGpsStatusListener = null;
 			}
 			if (mNmeaListener != null) {
-				mLocationManager.removeNmeaListener(mNmeaListener);
+				// Use reflection to work around a bug in the Android 29 SDK.
+				// https://stackoverflow.com/questions/57975969/accessing-nmea-on-android-api-level-24-when-compiled-for-target-api-level-29
+				try {
+					if (removeNmeaListenerMethod == null) {
+						removeNmeaListenerMethod = LocationManager.class.getMethod("addNmeaListener", GpsStatus.NmeaListener.class);
```
hrydgard: oops, typo, fixing.

PR #12817, opened 2020-04-12
Unification of the menu of Linux and Windows versions

### LINKED ISSUE #10484: Unification menus on Windows and Linux versions

TotalCaesar659:
Second trying to fix https://github.com/hrydgard/ppsspp/issues/10484.

### DIFF
--- a/Core/Config.cpp
+++ b/Core/Config.cpp
@@ -423,7 +423,7 @@ static ConfigSetting generalSettings[] = {
 	ConfigSetting("FirstRun", &g_Config.bFirstRun, true),
 	ConfigSetting("RunCount", &g_Config.iRunCount, 0),
 	ConfigSetting("Enable Logging", &g_Config.bEnableLogging, true),
-	ConfigSetting("AutoRun", &g_Config.bAutoRun, false),
+	ConfigSetting("AutoRun", &g_Config.bAutoRun, true),
 	ConfigSetting("Browse", &g_Config.bBrowse, false),
 	ConfigSetting("IgnoreBadMemAccess", &g_Config.bIgnoreBadMemAccess, true, true),
 	ConfigSetting("CurrentDirectory", &g_Config.currentDirectory, ""),
--- a/Core/Config.cpp
+++ b/Core/Config.cpp
@@ -423,7 +423,7 @@ static ConfigSetting generalSettings[] = {
 	ConfigSetting("FirstRun", &g_Config.bFirstRun, true),
 	ConfigSetting("RunCount", &g_Config.iRunCount, 0),
 	ConfigSetting("Enable Logging", &g_Config.bEnableLogging, true),
-	ConfigSetting("AutoRun", &g_Config.bAutoRun, true),
+	ConfigSetting("AutoRun", &g_Config.bAutoRun, false),
 	ConfigSetting("Browse", &g_Config.bBrowse, false),
 	ConfigSetting("IgnoreBadMemAccess", &g_Config.bIgnoreBadMemAccess, true, true),
 	ConfigSetting("CurrentDirectory", &g_Config.currentDirectory, ""),
--- a/Qt/mainwindow.cpp
+++ b/Qt/mainwindow.cpp
@@ -260,6 +260,11 @@ void MainWindow::runonloadAct()
 	g_Config.bAutoRun = !g_Config.bAutoRun;
 }
 
+void MainWindow::breakonloadAct()
+{
+	g_Config.bAutoRun = !g_Config.bAutoRun;
+}
+
 void MainWindow::lmapAct()
 {
 	QFileDialog dialog(0,"Load .MAP");
@@ -292,6 +297,41 @@ void MainWindow::smapAct()
 	{
 		fileNames = dialog.selectedFiles();
 		g_symbolMap->SaveSymbolMap(fileNames[0].toStdString().c_str());
+	}
+}
+
+void MainWindow::lsymAct()
+{
+	QFileDialog dialog(0,"Load .SYM");
+	dialog.setFileMode(QFileDialog::ExistingFile);
+	QStringList filters;
+	filters << "Symbols (*.sym)" << "|All files (*.*)";
+	dialog.setNameFilters(filters);
+	dialog.setAcceptMode(QFileDialog::AcceptOpen);
+	QStringList fileNames;
+	if (dialog.exec())
+		fileNames = dialog.selectedFiles();
+
+	if (fileNames.count() > 0)
+	{
+		QString fileName = QFileInfo(fileNames[0]).absoluteFilePath();
+		g_symbolMap->LoadNocashSym(fileName.toStdString().c_str());
+	}
+}
+
+void MainWindow::ssymAct()
+{
+	QFileDialog dialog(0,"Save .SYM");
+	dialog.setFileMode(QFileDialog::AnyFile);
+	dialog.setAcceptMode(QFileDialog::AcceptSave);
+	QStringList filters;
+	filters << "Save .SYM (*.sym)" << "|All files (*.*)";
+	dialog.setNameFilters(filters);
+	QStringList fileNames;
+	if (dialog.exec())
+	{
+		fileNames = dialog.selectedFiles();
+		g_symbolMap->SaveNocashSym(fileNames[0].toStdString().c_str());
 	}
 }
 
@@ -495,20 +535,31 @@ void MainWindow::createMenus()
 		->addEventChecked(&g_Config.bAutoRun);
 
 	// Debug
-	MenuTree* debugMenu = new MenuTree(this, menuBar(),   QT_TR_NOOP("De&bug"));
-	debugMenu->add(new MenuAction(this, SLOT(lmapAct()),      QT_TR_NOOP("Load Map File...")))
+	MenuTree* debugMenu = new MenuTree(this, menuBar(),   QT_TR_NOOP("&Debug"));
+	debugMenu->add(new MenuAction(this, SLOT(breakonloadAct()),   QT_TR_NOOP("Break on load")))
+		->addEventChecked(&g_Config.bAutoRun);
+	debugMenu->add(new MenuAction(this, SLOT(ignoreIllegalAct()),  QT_TR_NOOP("&Ignore illegal reads/writes")))
+		->addEventChecked(&g_Config.bIgnoreBadMemAccess);
+	debugMenu->addSeparator();
+	debugMenu->add(new MenuAction(this, SLOT(lmapAct()),      QT_TR_NOOP("&Load MAP file...")))
 		->addDisableState(UISTATE_MENU);
-	debugMenu->add(new MenuAction(this, SLOT(smapAct()),      QT_TR_NOOP("Save Map File...")))
+	debugMenu->add(new MenuAction(this, SLOT(smapAct()),      QT_TR_NOOP("&Save MAP file...")))
 		->addDisableState(UISTATE_MENU);
-	debugMenu->add(new MenuAction(this, SLOT(resetTableAct()),QT_TR_NOOP("Reset Symbol Table")))
+	debugMenu->add(new MenuAction(this, SLOT(lmapAct()),      QT_TR_NOOP("Lo&ad SYM file...")))
+		->addDisableState(UISTATE_MENU);
+	debugMenu->add(new MenuAction(this, SLOT(smapAct()),      QT_TR_NOOP("Sav&e SYM file...")))
+		->addDisableState(UISTATE_MENU);
+	debugMenu->add(new MenuAction(this, SLOT(resetTableAct()),QT_TR_NOOP("Reset s&ymbol table")))
 		->addDisableState(UISTATE_MENU);
 	debugMenu->addSeparator();
-	debugMenu->add(new MenuAction(this, SLOT(dumpNextAct()),  QT_TR_NOOP("Dump next frame to log")))
+	debugMenu->add(new MenuAction(this, SLOT(takeScreen()),  QT_TR_NOOP("&Take screenshot"), Qt::Key_F12))
 		->addDisableState(UISTATE_MENU);
-	debugMenu->add(new MenuAction(this, SLOT(takeScreen()),  QT_TR_NOOP("Take Screenshot"), Qt::Key_F12))
+	debugMenu->add(new MenuAction(this, SLOT(dumpNextAct()),  QT_TR_NOOP("D&ump next frame to log")))
 		->addDisableState(UISTATE_MENU);
+	debugMenu->add(new MenuAction(this, SLOT(statsAct()),   QT_TR_NOOP("Show debu&g statistics")))
+		->addEventChecked(&g_Config.bShowDebugStats);
 	debugMenu->addSeparator();
-	debugMenu->add(new MenuAction(this, SLOT(consoleAct()),   QT_TR_NOOP("Log Console")))
+	debugMenu->add(new MenuAction(this, SLOT(consoleAct()),   QT_TR_NOOP("&Log console"), Qt::CTRL + Qt::Key_L))
 		->addDisableState(UISTATE_MENU);
 
 	// Game settings
--- a/Qt/mainwindow.cpp
+++ b/Qt/mainwindow.cpp
@@ -561,7 +561,7 @@ void MainWindow::createMenus()
 	// Debug
 	MenuTree* debugMenu = new MenuTree(this, menuBar(),   QT_TR_NOOP("&Debug"));
 	debugMenu->add(new MenuAction(this, SLOT(breakonloadAct()),   QT_TR_NOOP("Break on load")))
-		->addEventChecked(&g_Config.bAutoRun);
+		->addEventUnchecked(&g_Config.bAutoRun);
 	debugMenu->add(new MenuAction(this, SLOT(ignoreIllegalAct()),  QT_TR_NOOP("&Ignore illegal reads/writes")))
 		->addEventChecked(&g_Config.bIgnoreBadMemAccess);
 	debugMenu->addSeparator();
--- a/Qt/mainwindow.cpp
+++ b/Qt/mainwindow.cpp
@@ -351,9 +351,19 @@ void MainWindow::forumAct()
 	QDesktopServices::openUrl(QUrl("https://forums.ppsspp.org/"));
 }
 
+void MainWindow::goldAct()
+{
+	QDesktopServices::openUrl(QUrl("https://central.ppsspp.org/buygold"));
+}
+
 void MainWindow::gitAct()
 {
 	QDesktopServices::openUrl(QUrl("https://github.com/hrydgard/ppsspp/"));
+}
+
+void MainWindow::discordAct()
+{
+	QDesktopServices::openUrl(QUrl("https://discord.gg/5NJB6dD"));
 }
 
 void MainWindow::aboutAct()
@@ -598,10 +608,13 @@ void MainWindow::createMenus()
 
 	// Help
 	MenuTree* helpMenu = new MenuTree(this, menuBar(),    QT_TR_NOOP("&Help"));
-	helpMenu->add(new MenuAction(this, SLOT(websiteAct()),    QT_TR_NOOP("Official &website"), QKeySequence::HelpContents));
-	helpMenu->add(new MenuAction(this, SLOT(forumAct()),      QT_TR_NOOP("Official &forum")));
-	helpMenu->add(new MenuAction(this, SLOT(gitAct()),        QT_TR_NOOP("&GitHub")));
-	helpMenu->add(new MenuAction(this, SLOT(aboutAct()),      QT_TR_NOOP("&About PPSSPP..."), QKeySequence::WhatsThis));
+	helpMenu->add(new MenuAction(this, SLOT(websiteAct()),    QT_TR_NOOP("Visit www.&ppsspp.org")));
+	helpMenu->add(new MenuAction(this, SLOT(forumAct()),      QT_TR_NOOP("PPSSPP &forums")));
+	helpMenu->add(new MenuAction(this, SLOT(goldAct()),       QT_TR_NOOP("Buy &Gold")));
+	helpMenu->add(new MenuAction(this, SLOT(gitAct()),        QT_TR_NOOP("Git&Hub")));
+	helpMenu->add(new MenuAction(this, SLOT(discordAct()),      QT_TR_NOOP("Discord")));
+	helpMenu->addSeparator();
+	helpMenu->add(new MenuAction(this, SLOT(aboutAct()),      QT_TR_NOOP("&About PPSSPP...")));
 
 	retranslate();
 }
--- a/Qt/mainwindow.cpp
+++ b/Qt/mainwindow.cpp
@@ -80,14 +80,7 @@ void MainWindow::newFrame()
 
 void MainWindow::updateMenus()
 {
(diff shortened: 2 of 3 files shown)


Qt/mainwindow.cpp
```diff
@@ -598,10 +608,13 @@ void MainWindow::createMenus()
 
 	// Help
 	MenuTree* helpMenu = new MenuTree(this, menuBar(),    QT_TR_NOOP("&Help"));
-	helpMenu->add(new MenuAction(this, SLOT(websiteAct()),    QT_TR_NOOP("Official &website"), QKeySequence::HelpContents));
-	helpMenu->add(new MenuAction(this, SLOT(forumAct()),      QT_TR_NOOP("Official &forum")));
-	helpMenu->add(new MenuAction(this, SLOT(gitAct()),        QT_TR_NOOP("&GitHub")));
-	helpMenu->add(new MenuAction(this, SLOT(aboutAct()),      QT_TR_NOOP("&About PPSSPP..."), QKeySequence::WhatsThis));
+	helpMenu->add(new MenuAction(this, SLOT(websiteAct()),    QT_TR_NOOP("Visit www.&ppsspp.org")));
+	helpMenu->add(new MenuAction(this, SLOT(forumAct()),      QT_TR_NOOP("PPSSPP &forums")));
+	helpMenu->add(new MenuAction(this, SLOT(goldAct()),       QT_TR_NOOP("Buy &Gold")));
```
unknownbrackets, line 675: I have no strong opinion either way, but I don't think you can actually buy a "gold" version of PPSSPP Qt.  That said, it's just a compile flag so you could compile it...

Not sure it matters though, since it's just a donation essentially anyway.

-[Unknown]

Qt/mainwindow.cpp
```diff
@@ -598,10 +608,13 @@ void MainWindow::createMenus()
 
 	// Help
 	MenuTree* helpMenu = new MenuTree(this, menuBar(),    QT_TR_NOOP("&Help"));
-	helpMenu->add(new MenuAction(this, SLOT(websiteAct()),    QT_TR_NOOP("Official &website"), QKeySequence::HelpContents));
-	helpMenu->add(new MenuAction(this, SLOT(forumAct()),      QT_TR_NOOP("Official &forum")));
-	helpMenu->add(new MenuAction(this, SLOT(gitAct()),        QT_TR_NOOP("&GitHub")));
-	helpMenu->add(new MenuAction(this, SLOT(aboutAct()),      QT_TR_NOOP("&About PPSSPP..."), QKeySequence::WhatsThis));
```
unknownbrackets, line 604: Is removing the keyboard shortcut intentional?  I'm not sure anyone really needs a quick way to get to these, though...

-[Unknown]

Qt/mainwindow.cpp
```diff
@@ -80,48 +80,66 @@ void MainWindow::newFrame()
...
 			action->setChecked(true);
 			break;
 		}
 	}
 
-	foreach(QAction * action, screenGroup->actions()) {
-		int width = (g_Config.IsPortrait() ? 272 : 480) * action->data().toInt();
-		int height = (g_Config.IsPortrait() ? 480 : 272) * action->data().toInt();
-		if (g_Config.iWindowWidth == width && g_Config.iWindowHeight == height) {
+	foreach(QAction * action, renderingResolutionGroup->actions()) {
```
unknownbrackets, line 106: nit: Could we make this a method?  Something like...

```c++
void MainWindow::updateMenuGroupInt(QActionGroup *group, int value) {
	foreach (QAction *action, group->actions()) {
		action->setChecked(action->data().toInt() == value);
	}
}

updateMenuGroupInt(renderingResolutionGroup, g_Config.iInternalResolution);
updateMenuGroupInt(renderingModeGroup, g_Config.iRenderingMode);
```

Disclaimer: I don't really know Qt slots/etc. well.  Not sure if it's better to only set true on the one matching one?

There's a lot of int settings, so this would probably make the code more readable and easier to add/remove settings later.

-[Unknown]

Qt/mainwindow.h
```diff
@@ -90,27 +90,37 @@ private slots:
...
-	// Options
-	// Core
-	void vertexDynarecAct() { g_Config.bVertexDecoderJit = !g_Config.bVertexDecoderJit; }
-	void fastmemAct() { g_Config.bFastMemory = !g_Config.bFastMemory; }
-	void ignoreIllegalAct() { g_Config.bIgnoreBadMemAccess = !g_Config.bIgnoreBadMemAccess; }
-
-	// Video
-	void anisotropicGroup_triggered(QAction *action) { g_Config.iAnisotropyLevel = action->data().toInt(); }
+	// Game settings
+	void keepOnTopAct() { g_Config.bTopMost = !g_Config.bTopMost; }
```
unknownbrackets, line 100: I doubt this one works on Qt without implementing, but probably just need a `setWindowFlags(Qt::WindowStaysOnTopHint);` somewhere.  It needs to be in the menu switch and also on startup.

-[Unknown]

Qt/mainwindow.h
```diff
@@ -90,27 +90,37 @@ private slots:
...
+	void pauseWhenNotFocusedAct() { g_Config.bPauseOnLostFocus = !g_Config.bPauseOnLostFocus; }
```
unknownbrackets, line 101: This one also needs hook-up with Qt.  On Windows, we listen for WM_ACTIVATE, and basically call `Core_EnableStepping(true/false);` to pause or resume emulation.

-[Unknown]

Qt/mainwindow.h
```diff
@@ -122,28 +132,18 @@ private slots:
...
+	// Cheats
+	void cheatsAct() { g_Config.bEnableCheats = !g_Config.bEnableCheats; }
+
 	void fullscrAct();
 	void raiseTopMost();
 	void statsAct() {
 		g_Config.bShowDebugStats = !g_Config.bShowDebugStats;
 		NativeMessageReceived("clear jit", "");
 	}
+	void VSyncAct() { g_Config.bVSync = !g_Config.bVSync; }
```
unknownbrackets, line 150: As mentioned, this ini setting doesn't do anything without platform hookup.

This implies it can be done using a surface format setting: https://stackoverflow.com/questions/37319328/qt-5-disable-vsync

There was talk in #9736 and #12024 of potentially making it possible to disable vsync on Linux, but I'd suggest we don't tackle that yet and leave this option out for now.

-[Unknown]

Qt/mainwindow.cpp
```diff
@@ -493,118 +511,87 @@ void MainWindow::createMenus()
...
 #endif
 		->addEventChecked(&g_Config.bFullScreen);
-	optionsMenu->add(new MenuAction(this, SLOT(statsAct()),   QT_TR_NOOP("&Show debug statistics")))
-		->addEventChecked(&g_Config.bShowDebugStats);
-	optionsMenu->add(new MenuAction(this, SLOT(showFPSAct()), QT_TR_NOOP("&Show FPS")))
+	gameSettingsMenu->add(new MenuAction(this, SLOT(VSyncAct()),   QT_TR_NOOP("VS&ync")))
+		->addEventChecked(&g_Config.bVSync);
+	MenuTree* postprocessingShaderMenu = new MenuTree(this, gameSettingsMenu, QT_TR_NOOP("Postprocessin&g shader"));
+	postprocessingShaderGroup = new MenuActionGroup(this, postprocessingShaderMenu, SLOT(postprocessingShaderGroup_triggered(QAction *)),
+		QStringList() << "Off" << "FXAA Antialiasing" << "CRT scanlines" << "Natural Colors" << "Natural Colors (no blur)" << "Vignette" << "Grayscale" << "Bloom" << "Sharpen" << "Inverse Colors" << "Scanlines (CRT)" << "Cartoon" << "4xHQ GLSL" << "AA-Color" << "Spline36 Upscaler" << "5xBR" << "5xBR-lv2" << "Video Smoothing AA" << "Super Sampling AA (Gauss)",
```
unknownbrackets, line 610: On Windows, we use `GetAllPostShaderInfo()` which dynamically loads the available shaders.  This list here is hardcoded, and may not match the list of shaders the user actually has.  If we add/remove a shader, no one will remember to update this list - especially because it's automatically updated on Windows.

There's some complexity here with language translations and timing, so it might be another thing to do in a follow-up issue/pull.

-[Unknown]

Qt/mainwindow.h
```diff
@@ -90,27 +90,37 @@ private slots:
...
+	void moreSettingsAct() { NativeMessageReceived("settings", ""); }
 
 	void bufferRenderAct() {
 		g_Config.iRenderingMode = !g_Config.iRenderingMode;
 		NativeMessageReceived("gpu_resized", "");
 	}
 	void linearAct() { g_Config.iTexFiltering = (g_Config.iTexFiltering != 0) ? 0 : 3; }
 
-	void screenGroup_triggered(QAction *action) { SetWindowScale(action->data().toInt()); }
+	void postprocessingShaderGroup_triggered(QAction *action) { g_Config.sPostShaderName = action->data().toInt(); }
```
unknownbrackets, line 113: Need to call `NativeMessageReceived("gpu_resized", "");` after this.  Also, can't set `g_Config.sPostShaderName` to an integer?  It expects a string (name.)

-[Unknown]

Qt/mainwindow.h
```diff
@@ -90,27 +90,37 @@ private slots:
...
+	void renderingResolutionGroup_triggered(QAction *action) { g_Config.iInternalResolution = action->data().toInt(); }
```
unknownbrackets, line 114: Need to call `NativeMessageReceived("gpu_resized", "");` afterward.  Also, if `g_Config.iTexScalingLevel == TEXSCALING_AUTO` then we also need to call `NativeMessageReceived("gpu_clearCache", "");`.

-[Unknown]

Qt/mainwindow.h
```diff
@@ -90,27 +90,37 @@ private slots:
...
+	void windowGroup_triggered(QAction *action) { SetWindowScale(action->data().toInt()); }
 
 	void displayLayoutGroup_triggered(QAction *action) {
 		g_Config.iSmallDisplayZoomType = action->data().toInt();
 		NativeMessageReceived("gpu_resized", "");
 	}
+	void renderingModeGroup_triggered(QAction *action) { g_Config.iRenderingMode = action->data().toInt(); }
```
unknownbrackets, line 121: Skip buffer effects (FB_NON_BUFFERED_MODE) needs to turn off auto frameskip (`g_Config.bAutoFrameSkip = false;`) because it won't work.  We also need to call `NativeMessageReceived("gpu_resized", "");` after changing the rendering mode.

-[Unknown]

Qt/mainwindow.h
```diff
@@ -90,27 +90,37 @@ private slots:
...
+	void autoframeskipAct() { g_Config.bAutoFrameSkip = !g_Config.bAutoFrameSkip; }
```
unknownbrackets, line 122: If `g_Config.iRenderingMode` is `FB_NON_BUFFERED_MODE`, needs to change it to FB_BUFFERED_MODE` and issue `NativeMessageReceived("gpu_resized", "");`.  Otherwise auto frameskip won't work.

-[Unknown]

Qt/mainwindow.h
```diff
@@ -90,27 +90,37 @@ private slots:
...
+	void frameSkippingGroup_triggered(QAction *action) { g_Config.iFrameSkip = action->data().toInt(); }
+	void frameSkippingTypeGroup_triggered(QAction *action) { g_Config.iFrameSkipType = action->data().toInt(); }
+	void textureFilteringGroup_triggered(QAction *action) { g_Config.iTexFiltering = action->data().toInt(); }
+	void screenScalingFilterGroup_triggered(QAction *action) { g_Config.iBufFilter = action->data().toInt(); }
+	void textureScalingLevelGroup_triggered(QAction *action) { g_Config.iTexScalingLevel = action->data().toInt(); }
+	void textureScalingTypeGroup_triggered(QAction *action) { g_Config.iTexScalingType = action->data().toInt(); }
+	void deposterizeAct() { g_Config.bTexDeposterize = !g_Config.bTexDeposterize; }
```
unknownbrackets, line 129: `NativeMessageReceived("gpu_clearCache", "");` afterward.  For `iTexScalingType` and `iTexScalingLevel` also.

-[Unknown]

Qt/mainwindow.h
```diff
@@ -122,28 +132,18 @@ private slots:
...
+	void cheatsAct() { g_Config.bEnableCheats = !g_Config.bEnableCheats; }
+
 	void fullscrAct();
 	void raiseTopMost();
 	void statsAct() {
 		g_Config.bShowDebugStats = !g_Config.bShowDebugStats;
 		NativeMessageReceived("clear jit", "");
 	}
+	void VSyncAct() { g_Config.bVSync = !g_Config.bVSync; }
 	void showFPSAct() { g_Config.iShowFPSCounter = !g_Config.iShowFPSCounter; }
```
unknownbrackets, line 151: Ideally, this should toggle between 0 and 3, not 0 and 1.  That's what it does on Windows - 3 means show FPS and Speed.

-[Unknown]

Qt/mainwindow.cpp
```diff
@@ -495,20 +535,31 @@ void MainWindow::createMenus()
...
+	debugMenu->add(new MenuAction(this, SLOT(breakonloadAct()),   QT_TR_NOOP("Break on load")))
+		->addEventChecked(&g_Config.bAutoRun);
+	debugMenu->add(new MenuAction(this, SLOT(ignoreIllegalAct()),  QT_TR_NOOP("&Ignore illegal reads/writes")))
+		->addEventChecked(&g_Config.bIgnoreBadMemAccess);
+	debugMenu->addSeparator();
+	debugMenu->add(new MenuAction(this, SLOT(lmapAct()),      QT_TR_NOOP("&Load MAP file...")))
+		->addDisableState(UISTATE_MENU);
+	debugMenu->add(new MenuAction(this, SLOT(smapAct()),      QT_TR_NOOP("&Save MAP file...")))
+		->addDisableState(UISTATE_MENU);
+	debugMenu->add(new MenuAction(this, SLOT(lmapAct()),      QT_TR_NOOP("Lo&ad SYM file...")))
```
unknownbrackets, line 572: Should this be lsymAct?  And below ssymAct?

-[Unknown]

Qt/mainwindow.cpp
```diff
@@ -495,20 +535,31 @@ void MainWindow::createMenus()
 		->addEventChecked(&g_Config.bAutoRun);
 
 	// Debug
-	MenuTree* debugMenu = new MenuTree(this, menuBar(),   QT_TR_NOOP("De&bug"));
-	debugMenu->add(new MenuAction(this, SLOT(lmapAct()),      QT_TR_NOOP("Load Map File...")))
+	MenuTree* debugMenu = new MenuTree(this, menuBar(),   QT_TR_NOOP("&Debug"));
+	debugMenu->add(new MenuAction(this, SLOT(breakonloadAct()),   QT_TR_NOOP("Break on load")))
+		->addEventChecked(&g_Config.bAutoRun);
```
unknownbrackets, line 564: This is inversed - `g_Config.bAutoRun == true` means it should be unchecked.

-[Unknown]

Qt/mainwindow.cpp
```diff
@@ -508,21 +522,28 @@ void MainWindow::createMenus()
 {
 	// File
 	MenuTree* fileMenu = new MenuTree(this, menuBar(),    QT_TR_NOOP("&File"));
-	fileMenu->add(new MenuAction(this, SLOT(openAct()),       QT_TR_NOOP("&Open..."), QKeySequence::Open))
+	fileMenu->add(new MenuAction(this, SLOT(loadAct()),       QT_TR_NOOP("&Load...")))
```
unknownbrackets, line 525: Hm, seems like we'd still want `QKeySequence::Open` here?  No?

-[Unknown]

Qt/mainwindow.cpp
```diff
@@ -508,21 +522,28 @@ void MainWindow::createMenus()
...
 		->addDisableState(UISTATE_MENU);
-	fileMenu->add(new MenuAction(this, SLOT(lstateAct()),     QT_TR_NOOP("&Load State File...")))
+	fileMenu->add(new MenuAction(this, SLOT(lstateAct()),     QT_TR_NOOP("&Load state file...")))
 		->addDisableState(UISTATE_MENU);
-	fileMenu->add(new MenuAction(this, SLOT(sstateAct()),     QT_TR_NOOP("&Save State File...")))
+	fileMenu->add(new MenuAction(this, SLOT(sstateAct()),     QT_TR_NOOP("&Save state file...")))
 		->addDisableState(UISTATE_MENU);
 	fileMenu->addSeparator();
-	fileMenu->add(new MenuAction(this, SLOT(exitAct()),       QT_TR_NOOP("E&xit"), QKeySequence::Quit));
+	fileMenu->add(new MenuAction(this, SLOT(exitAct()),       QT_TR_NOOP("E&xit"), Qt::ALT + Qt::Key_F4));
```
unknownbrackets, line 546: Here I definitely think we should keep `QKeySequence::Quit`.  Alt-F4 is a Windows thing, I think it's usually Ctrl-Q on Linux.

-[Unknown]

unknownbrackets (changes requested):

Qt/mainwindow.cpp
TotalCaesar659: Why not? If somebody wants to donate, he may buy vesion for other platform. There is no reason not to show it and no other way to do it (like Patreon or smth).

Qt/mainwindow.cpp
TotalCaesar659: Yes, Windows version doesn't have them too. I can get them back.

Qt/mainwindow.cpp
unknownbrackets: Sounds fair.

-[Unknown]

Qt/mainwindow.cpp
TotalCaesar659: Then I'll remove it, thanks.

Qt/mainwindow.h
TotalCaesar659: In Windows version, menu doesn't correspond internal settings: it has 2 states and turns both Speed and FPS when active.


It would be better to fix then both in further PRs.

unknownbrackets: Well, I think the idea is - it's not really that necessary to be able to select between all four options in the menu.  It (and other settings) are available in the cross-platform UI within the emulator if you need it.

For example, we definitely wouldn't want to make the Windows menu a maze with every single possible setting shown there.

That's why we don't show all the possible FPS options either.

-[Unknown]

hrydgard: I do think that if we show a setting in the platform-specific UI, it's nice if the choices match those in the cross-platform one - this kind of inconsistency isn't super elegant. But there's indeed also some merit in keeping things simple.

TotalCaesar659: The idea of it is reasonable, but disrepance of these menus is a little confusing. And the top menu is already the maze (compared to RPCS3, it looks more like LibreOffice). It wouldn't hurt to slim it down, I guess, but it's the subject of another discussion.
OK, I'll change it to Windows-specific way.

Qt/mainwindow.cpp
TotalCaesar659: It doesn't compile and shows something like `error: prototype for ‘void MainWindow::updateMenuGroupInt(QActionGroup*, int)’ does not match any in class ‘MainWindow’`.

Qt/mainwindow.h
TotalCaesar659: Removed for the future.

Core/Config.cpp
```diff
@@ -423,7 +423,7 @@ static ConfigSetting generalSettings[] = {
 	ConfigSetting("FirstRun", &g_Config.bFirstRun, true),
 	ConfigSetting("RunCount", &g_Config.iRunCount, 0),
 	ConfigSetting("Enable Logging", &g_Config.bEnableLogging, true),
-	ConfigSetting("AutoRun", &g_Config.bAutoRun, true),
+	ConfigSetting("AutoRun", &g_Config.bAutoRun, false),
```
unknownbrackets, line 426: We definitely don't want to change the default like this, or change its meaning.  People will get super confused if games don't start by default.

That's what this setting does: it's mainly for debugging.  If you're a rom hacker, you do want the game to wait for you to do stuff in the debugger before it starts running.  But that's not most people.

-[Unknown]

Qt/mainwindow.h
```diff
@@ -110,23 +108,47 @@ private slots:
...
 	void linearAct() { g_Config.iTexFiltering = (g_Config.iTexFiltering != 0) ? 0 : 3; }
 
-	void postprocessingShaderGroup_triggered(QAction *action) { g_Config.sPostShaderName = action->data().toInt(); }
-	void renderingResolutionGroup_triggered(QAction *action) { g_Config.iInternalResolution = action->data().toInt(); }
+	void renderingResolutionGroup_triggered(QAction *action) {
+		g_Config.iInternalResolution = action->data().toInt();
+		NativeMessageReceived("gpu_resized", "");
+		if (g_Config.iTexScalingLevel == TEXSCALING_AUTO) {
+			NativeMessageReceived("gpu_clearCache", "");
+	}
```
unknownbrackets, line 116: nit: Indentation fuax pas.

-[Unknown]

Qt/mainwindow.h
```diff
@@ -110,23 +108,47 @@ private slots:
...
+		g_Config.iRenderingMode = action->data().toInt();
+		g_Config.bAutoFrameSkip = false;
+		NativeMessageReceived("gpu_resized", "");
+	}
+	void autoframeskipAct() {
+		g_Config.bAutoFrameSkip = !g_Config.bAutoFrameSkip;
+		if (g_Config.iRenderingMode == FB_NON_BUFFERED_MODE) {
+			g_Config.iRenderingMode = FB_BUFFERED_MODE;
+			NativeMessageReceived("gpu_resized", "");
+	}
```
unknownbrackets, line 134: nit: Indentation fuax pas.

-[Unknown]

Qt/mainwindow.h
```diff
@@ -272,6 +277,8 @@ public slots:
 	void update() {
 		if (_eventCheck)
 			setChecked(*_eventCheck);
+		if (_eventUncheck)
+			setChecked(false);
```
unknownbrackets, line 281: Hm, why not `!*_eventUncheck` instead of `false`?

-[Unknown]

Qt/mainwindow.h
hrydgard: nit: faux pas doesn't mean mistake or slip-up, it means something embarrassing or socially unacceptable - though maybe indentation mistakes are ;)

unknownbrackets: Yep, I say this just like I say "nit: Tabs and spaces living together in despicable harmony."  I'm not being serious or accusing anyone of some actual unacceptable act.

-[Unknown]

unknownbrackets (approved):

Qt/mainwindow.cpp
unknownbrackets: Hm, it sounds like you might've skipped adding it to the header?  But no big deal anyway.  I think it's fine.

-[Unknown]

hrydgard:
Nice work both reviewing and fixing :)  Looks pretty good now, let's merge.

Emulatorer:
not added to v1.10.0 milestone?

PR #12818, opened 2020-04-13
Add option to prevent Mipmaps from being dumped

NR74W:
WipEout Pure and Pulse use mipmapping for just about anything;
it is a huge waste of time to sort / blacklist every mip levels to prevent the 'new' folder from being cluttered.

This small change will be very useful to texture modders for games relying massively on mipmaps.

I'm not sure about the name of the setting though, so perhaps anyone has a suggestion ?

### DIFF
--- a/Core/TextureReplacer.cpp
+++ b/Core/TextureReplacer.cpp
@@ -80,7 +80,8 @@ bool TextureReplacer::LoadIni() {
 	allowVideo_ = false;
 	ignoreAddress_ = false;
 	reduceHash_ = false;
-	disallowMipmap_ = false;
+	// Prevents dumping the mipmaps.
+	ignoreMipmap_ = false;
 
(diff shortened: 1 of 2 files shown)


JuliusSenkus:
@NR74W Suggestion `ignoreMipmaps`?

JuliusSenkus:
Also just a recommendation, you might leave short comment in front of the boolean variable, why it was added, so someone accidentally would not remove it (don't know exact this repository code recommendations, so just my opinion here).

NR74W:
This is the first time I'm contributing; is it better like this ?

efonte:
You have to add the default value of the new variable

https://github.com/hrydgard/ppsspp/blob/master/Core/TextureReplacer.cpp#L695

hrydgard:
@efonte  Done.

Issue #8664, opened 2016-03-25
Street Fighter Alpha 3 Max graphical glitches (android)


RaduNastase:
I am on Android 4.1.2 (Samsung Galaxy S2 with Dorimanx 8.43), using ppsspp 1.2.2. The lines are both vertical (like in the pics) and horizontal, sometimes forming full squares. I changed all settings back and forth but the lines do not go away. Strangely they do however start to diminish after a few stages of gameplay.

![screenshot_2016-03-25-02-44-13](https://cloud.githubusercontent.com/assets/18061144/14035325/75519ea4-f235-11e5-8e57-d03edd6935ae.png)
![screenshot_2016-03-25-02-42-08](https://cloud.githubusercontent.com/assets/18061144/14035333/921afe90-f235-11e5-83ad-03d22b170250.png)

unknownbrackets:
Switch to 1x render resolution and make sure vertex cache is disabled.

-[Unknown]

RaduNastase:
no change. The errors are still there. There is something strange seeing as all other games i have are glitch free (Fate, DBZ, Strikers plus, BlazBlue, GGXX).

zminhquanz:
@RaduNastase which app monitor your CPU and ram is it ?

RaduNastase:
it's Cool Tool for memory and Cpu usage and CpuTemp for cpu temp :P

RaduNastase:
Little update: I tried the game in Retroarch ppsspp core and while it runs much slower than in PPSSPP 1.2.2, it has the exact same errors regardless of settings.

zminhquanz:
thank you for your app , it better than other app , ex : System Monitor , TinyCore

RaduNastase:
you're welcome!

Jinzombie:
I have the same issue Im using the fast one ppsspp 1.6.2
![screenshot_2018-06-01-08-11-33](https://user-images.githubusercontent.com/37603562/40814571-75c5bca6-6530-11e8-8071-e573269cc4bb.png)

RaduNastase:
Ok, so i used to have this issue with my old Mali GPU powered smartphone (Samsung S2), but after i got an Adreno GPU phone (Note 3), the errors were gone. I also would recommend putting all filters to "nearest" and also playing at native psp resolution without any upscale of any kind.

Jinzombie:
[...]

PR #12821, opened 2020-04-13
#8714 Unified resolution scaling for all platforms.

JuliusSenkus:
Related to #8714. Removed for all platforms, but can apply that only for android platform if needed. But repository owners should decide if it is needed and if it should be global.


Emulatorer:
This pr can affect this game #8664

JuliusSenkus:
@Emulatorer how? `1x PSP` is preserved, what this change does, it only adds more rendering options. Can you expand in what way it would impact it?

unknownbrackets (approved):

Emulatorer:
@JuliusSenkus thnks 4 the explanation 😅

PR #12823, opened 2020-04-14
Auto rotate over a square

iota97:
At first I thought the analog would get only to 1/sqrt(2) at the corner as the HW was a circle, after seeing #11552 this is probabbly safer for QTE.

### DIFF
--- a/UI/EmuScreen.cpp
+++ b/UI/EmuScreen.cpp
@@ -1186,12 +1186,13 @@ void EmuScreen::update() {
 
 	if (autoRotatingAnalogCW_) {
 		const float now = time_now_d();
-		__CtrlSetAnalogX(cos(now*-g_Config.fAnalogAutoRotSpeed), 0);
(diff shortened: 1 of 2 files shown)


hrydgard:
```
/home/travis/build/hrydgard/ppsspp/UI/GamepadEmu.cpp:206:71: error: no matching function for call to ‘max(float, double)’
   __CtrlSetAnalogX(std::min(1.0f, std::max(-1.0f, 1.42f*cos(now*speed))), 0);
```
The two arguments to std::min (and std::max) need to be of the same type, easily fixed by changing cos to cosf and sin to sinf.

fu8765:
v1.9.3-664-g30eebe2f1 doesn't fix it for me. (if that was the intention)

iota97:
This is for the auto analog rotation only (it allow to rotate the analog with a key press to help with QTE on keyboard and touch control), so it's expected not to fix that, sorry for the confusion :)

PR #12833, opened 2020-04-17
Improve code readability

TotalCaesar659:
@unknownbrackets [proposed](https://github.com/hrydgard/ppsspp/pull/12817#discussion_r407236037) that.


unknownbrackets (approved):
I'm obviously biased but I think this is much easier to read.

-[Unknown]

hrydgard:
Bias or not, of course this is better.

PR #12834, opened 2020-04-18
Add "Switch UMD" menu point to Linux version

TotalCaesar659:


Qt/mainwindow.cpp
```diff
@@ -141,7 +142,7 @@ void MainWindow::closeAct()
 void MainWindow::openmsAct()
 {
 	QString confighome = getenv("XDG_CONFIG_HOME");
-	QString memorystick = confighome + "/ppsspp/PSP";
+	QString memorystick = confighome + "/ppsspp";
```
hrydgard, line 145: What's the motivation for this change?

TotalCaesar659: Just unification with behavior of Windows version. Besides I've found there's more folders in memstick than just PSP.
https://doodypsp.blogspot.com/2007/10/psps-directory-structure-explained_06.html

hrydgard: Oh right, forgot that's how we define it there. Good one.

PR #12837, opened 2020-04-19
Update VS projects

Florin9doi:
Remove an unused include (WindowsInclude); Include ffmpeg/arm* for arm* instead of x86_64; Add missing libs for Headless and UnitTests; Reorder libs for consistency


hrydgard:
Thanks!

PR #12839, opened 2020-04-19
GLES: Disable range culling on old Tegras, they seem to misbehave

hrydgard:
See issue #12838.

This is a blind fix - couldn't find my old Shield Portable :(

EDIT: Found it! But it doesn't start. Let's see if we charge it for a while...


unknownbrackets (approved):

hrydgard:
Confirmed working on hardware, merging.

PR #12852, opened 2020-04-24
Update <Share> dialog on Android 10; Fixes #12846

### LINKED ISSUE #12846: Update "Share" dialog on Android 10
PPSSPP uses old "Share" dialog on Android 10, that supports new one (https://developer.android.com/training/sharing/send).

Florin9doi:
Fixes #12846 

Screenshot_1587762653

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@
 *.exp
 *.qdact
 controls.ini
+desktop.ini
 Debug
 Release
 Windows/.vs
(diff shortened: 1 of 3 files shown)


Florin9doi:
Should we change http to https?

hrydgard:
Yes. Also, Intent.createChooser only exists from API 22 and forwards. Please add a fall back path for older API versions.

Florin9doi:
There are 2 methods; 
- public static Intent createChooser (Intent target, CharSequence title) - Added in API level 1
- public static Intent createChooser (Intent target, CharSequence title, IntentSender sender) - Added in API level 22

Should be ok

hrydgard:
Oh sorry, I missed that. Should be OK then.

PR #12889, opened 2020-05-12
Improved compatibility of sceGeListEnQueue: verify that stackDepth < 256

sum2012:
https://github.com/jpcsp/jpcsp/commit/f9a116f78a939f27440e5edd5fd0c70cddc78537

Thanks gid15

May help #11660 crash


hrydgard:
Looks fairly safe and straightforward, should really hardware test this, but ...

sum2012:
gid15 already do the hardware test ,but of course we can also test.
https://www.emunewz.net/forum/showthread.php?tid=242609&pid=364785#pid364785

PR #12920, opened 2020-05-17
Remove the Audio Resampling setting (now always on).

hrydgard:
Having it off is not useful because it'll never sync up perfectly over time.


unknownbrackets (approved):

hrydgard:
@orbea Only use case I could think of is if Retro has its own thing that does the same thing. Two of them on top of each other might be a little much. In that case, we could change this to a libretro-specific #ifdef instead.

PR #12948, opened 2020-05-22
[Camera/V4L] Add support for jpeg and multiple buffers

Florin9doi:
required for Sony EyeToy


hrydgard:
Looks good. Indeed appropriate to support using the EyeToy as a camera for PPSSPP :)

PR #12965, opened 2020-05-24
DarkStalkers: Handle the "normal" screen stretch too, not just "wide".

hrydgard:
To avoid a surprising performance drop just from tweaking in-game settings. Plus, you could argue that the game is meant to be played this way anyway.

A followup to #12443 .

@LunaMoo  FYI :) Anything else it's missing?

### DIFF
--- a/GPU/Common/PresentationCommon.cpp
+++ b/GPU/Common/PresentationCommon.cpp
@@ -570,6 +570,13 @@ void PresentationCommon::CopyToOutput(OutputFlags flags, int uvRotation, float u
 		}
 	}
 
+	if (flags & OutputFlags::PILLARBOX) {
+		for (int i = 0; i < 4; i++) {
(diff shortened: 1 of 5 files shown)


Panderner:
How do you feel for this pull request on a real PSP or PS Vita Screen?

hrydgard:
I don't understand that question at all.

Gamemulatorer:
Maybe this pr could help the performance drops during fighting..

Introduction fight this game is smooth
Screenshot_20200524-235131
but during fight scene the performance drastically drops
Screenshot_20200524-235153

hrydgard:
Yeah there's still more work to be done on performance, independently from this change. For now, it's at least OK on PCs and high power phones.

Gamemulatorer:
I feel a little improvement thanks :)

PR #12984, opened 2020-05-29
add raspberry pi 64 bit build files

kaiomatico:
Adds the new raspberry pi os 64 bit support to the b.sh script and adds the cmake file. Tested on raspberry pi 4 with the latest apt updates


kaiomatico:
If you are happy with this I can also update the build instructions on the wiki

hrydgard:
Yeah, please do.

kaiomatico:
done :+1:

PR #13005, opened 2020-06-07
Wrap some SysMemForKernel

sum2012:
Used in "Hatsune Miku Project Diva Extend" chinese patched version
Prepare real support the chinese patched game.


sum2012:
oop

PR #13039, opened 2020-06-21
Add a sceOpenPSID_driver nid

sum2012:
Used in  "Shin Sangoku Musou 6 special" chinese patched game

### DIFF
--- a/Core/HLE/sceOpenPSID.cpp
+++ b/Core/HLE/sceOpenPSID.cpp
@@ -35,6 +35,21 @@ static int sceOpenPSIDGetOpenPSID(u32 OpenPSIDPtr)
 	return 0;
 }
 
+static int sceOpenPSID_driver_0x19D579F0(u32 OpenPSIDPtr,u32 unknown)
+{
+	ERROR_LOG(HLE, "UNTESTED sceOpenPSID_driver_0x19D579F0(%d,%d)", OpenPSIDPtr,unknown);
+	u8 dummyOpenPSID[16] = { 0x10, 0x02, 0xA3, 0x44, 0x13, 0xF5, 0x93, 0xB0, 0xCC, 0x6E, 0xD1, 0x32, 0x27, 0x85, 0x0F, 0x9D };
+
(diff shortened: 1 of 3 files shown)


Core/HLE/sceOpenPSID.cpp
```diff
@@ -35,6 +35,21 @@ static int sceOpenPSIDGetOpenPSID(u32 OpenPSIDPtr)
 	return 0;
 }
 
+static int sceOpenPSID_driver_0x19D579F0(u32 OpenPSIDPtr,u32 unknown)
+{
+	ERROR_LOG(HLE, "UNTESTED sceOpenPSID_driver_0x19D579F0(%d,%d)", OpenPSIDPtr,unknown);
+	u8 dummyOpenPSID[16] = { 0x10, 0x02, 0xA3, 0x44, 0x13, 0xF5, 0x93, 0xB0, 0xCC, 0x6E, 0xD1, 0x32, 0x27, 0x85, 0x0F, 0x9D };
```
hrydgard, line 41: What is this value?

hrydgard: Looks like OpenPSID is some device-unique value on the PS3? https://www.psdevwiki.com/ps3/OpenPSID  

Is it the same on PSP, or does it even matter? Something used for multiplayer?

vnctdj: Not sure if useful, but, for the record, I've already talked about this variable with @unknownbrackets [a long time ago](https://github.com/hrydgard/ppsspp/issues/4366).

hrydgard: That was indeed a long time ago!

Still curious where these numbers come from, but I'm fine merging a static value here until it's been proven to matter for anything...

sum2012: jpcsp ref 's code
https://github.com/jpcsp/jpcsp/blob/master/src/jpcsp/HLE/modules/sceOpenPSID.java#L45

PR #13041, opened 2020-06-21
Simplify Vulkan initialization a little, don't recreate surface on resize

hrydgard:
Except on Android, where we still do since the surface is lost on defocus/resize.

Might help #13009.

Not yet tested with SDL, will do shortly.

### DIFF
--- a/Common/Vulkan/VulkanContext.cpp
+++ b/Common/Vulkan/VulkanContext.cpp
@@ -267,6 +267,11 @@ void VulkanContext::DestroyInstance() {
 		}
 	}
 
+	if (surface_) {
+		vkDestroySurfaceKHR(instance_, surface_, nullptr);
+		surface_ = VK_NULL_HANDLE;
+	}
+
 	vkDestroyInstance(instance_, nullptr);
 	VulkanFree();
 	instance_ = VK_NULL_HANDLE;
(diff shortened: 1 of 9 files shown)


hrydgard:
Tried this on Ubuntu 20.04. Resizing the window works great, however toggling the fullscreen option on and then off crashes with:

```
W: VulkanRenderManager.cpp:1031: VK_SUBOPTIMAL_KHR returned - ignoring
INTEL-MESA: error: ../src/intel/vulkan/anv_batch_chain.c:1697: execbuf2 failed: Invalid argument (VK_ERROR_DEVICE_LOST)
46:24:310 Common/MsgHandler.cpp:45 E[SYSTEM]: Critical: Lost the Vulkan device! If this happens again, switch Graphics Backend from Vulkan to OpenGL
E: MsgHandler.cpp:72: Lost the Vulkan device! If this happens again, switch Graphics Backend from Vulkan to OpenGL
```

Not sure whether I'm doing things wrong there..

hrydgard:
Just realized there's a pretty bad race condition though, since we don't acquire/present on the thread that gets the resize message. Will fix.

hrydgard:
Well actually we do handle that by joining the thread. hm.

hrydgard:
The crash here might be related to the one from #13009  ?  Very odd.

Doesn't happen on master.

iota97:
FWIW this PR is working fine on Manjaro with NVIDIA.

hrydgard:
It works fine on my ubuntu laptop too except when coming back from full screen, which is odd.

hrydgard:
Closed in favor of #13164.

PR #13049, opened 2020-06-23
Update Lang

sum2012:
Mainly update zh_CN.ini and zh_TW.ini


sum2012:
I don't know that why ppsspp 's lang repository and the buildbot of file is not the same .

sum2012:
I find the problem now.
Tonight I will do again translate

PR #13071, opened 2020-06-28
D3D9: Fix a sign mistake generating the projection matrix.

### LINKED ISSUE #13063: Viewport regression in D3D9
Since D3D9 sees little use nowadays, not sure if it belongs in v1.10.1, let's see.

Either way, got reported on [Reddit](https://www.reddit.com/r/emulation/comments/hgtjc8/ppsspp_version_110_is_now_officially_released/fw8vbdp?utm_source=share&utm_medium=web2x) that the viewport in Star Ocean is way off in some views:

Ex: Star Ocean: https://www.screenshotcomparison.com/comparison/3939

Tales of The World: https://www.screenshotcomparison.com/comparison/3938

hrydgard:
Fixes #13063

I hope we don't break something else...

We really need some kind of regression tests for this stuff..

### DIFF
--- a/GPU/Directx9/ShaderManagerDX9.cpp
+++ b/GPU/Directx9/ShaderManagerDX9.cpp
@@ -251,9 +251,9 @@ static void ConvertProjMatrixToD3D(Matrix4x4 &in, bool invertedX, bool invertedY
 
 	float yoff = -1.0f / gstate_c.curRTRenderHeight;
 	if (invertedY) {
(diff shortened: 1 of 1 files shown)


unknownbrackets (approved):
At least, this seems like a straight-forward mistake in 41a6b97.

I think we could probably auto-run tests to compare GE dumps, although I worry about slight pixel differences.  I feel like Dolphin's stuff got quite complicated to handle that...

-[Unknown]

PR #13082, opened 2020-07-02
Fix reliability of the CMake generated version

vnctdj:
steps to reproduce :

git checkout v1.10.1~3
git describe --always
--> v1.10-19-g9aab3d986
[build now]
PPSSPPSDL --version
--> v1.10-19-g9aab3d986

git checkout v1.10.1
git describe --always
--> v1.10.1
[build now]
PPSSPPSDL --version
--> v1.10-19-g9aab3d986

Indeed, the "v1.10.1" regex matches the "v1.10-19-g9aab3d986" string. How unfortunate!
So I just deleted that check, building the same commit twice should not happen that often :)

### DIFF
--- a/git-version.cmake
+++ b/git-version.cmake
@@ -23,13 +23,6 @@ if(EXISTS ${GIT_VERSION_FILE})
 	if(NOT ${match} EQUAL "")
 		set(GIT_VERSION_UPDATE "0")
 	endif()
-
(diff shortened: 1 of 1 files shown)


hrydgard:
Hmm, ideally the dots in the regex should be escaped I guess.. but this should do.

PR #13096, opened 2020-07-05
Revert "Remove the I/O on Thread option - treat it as always on."

hrydgard:
This reverts commit 9c40381db090285aa7848bfee93dca6c4a220b62 which removed the option.

Got a complaint that disabling this was useful to fix some Adhoc problem. Until we figure that out, let's restore the option.


EX-e-LD:
Hi,me again.Yes,i can confirm this state,i work on revision of GuenosNoLife's Ad-Hoc list,and surprisingly,lot of games there work with I/O thread **disabled**,like PixelJunk Monsters Deluxe that doesn't work if the game is actually on enabled I/O thread.

PR #13100, opened 2020-07-05
Android: Add option to ignore camera notches when sizing the game display

hrydgard:
This is generally what you want, at least on phones with small camera notches.

Not sure about the most intuitive polarity and name of the setting.. bit of a shame to create another setting with a negation. But most people should probably ignore this.

Also not sure whether to squeeze it into 1.10.2 or not... judging from email, there's quite some demand for this. Unfortunately not enough time for translations to come in...


iota97:
Just wondering if is better to just call it "notch" rather than "inset" in the UI as it's a more common used term while "camera inset" hardly find anything even on google.

hrydgard:
hmm right. too late for 1.10.2 though, will change it later.

PR #13117, opened 2020-07-10
(Libretro) Buildfixes for Windows

inactive123:
Hi there, we'd appreciate if this could be merged as soon as possible. It would fix the Windows build again on our end.

Also, looking at the code, perhaps it would be worthwhile if in the future we could build around PPSSPP_PLATFORM(LIBRETRO) in the codebase ( I assume this doesn't work yet), since for some of these platform-specific bits, a libretro-specific implementation (usually implemented on the frontend side) is preferred.


hrydgard:
Yes, will add that.

twinaphex:
Thanks!

Issue #10698, opened 2018-03-08
Ridge Racer 2 adhoc not working due to channel mismatch, includes workaround


DarkArchonMind:
What happens?

Ridge Racer 2 adhoc doesn't work because of channel mismatch
After calling sceUtilityGetSystemParamInt, since default value in ppsspp.ini is 0, RR2 decides to set the channel to 11 and caches it in memory

Later, when executing sceNetAdhocctlGetScanInfo, after getting the group result, the game looks in the groups for ones which are on the same channel, which fails, since all the group channels are set to 1 here
https://github.com/hrydgard/ppsspp/blob/97ab91c3d5b74d8afdeaf55b6360f40906192624/Core/HLE/sceNetAdhoc.cpp#L966

Change both ppsspp config files ppsspp.ini to set WlanAdhocChannel = 1 fixes this, but I'm wondering if there could be a way to detect and patch this, Or, maybe set  WlanAdhocChannel = 1 by default instead of 0, especially since so much of the networking code tries to force things into channel 1.

What *should* happen?

Should connect and work correctly.

What hardware and operating system are you running PPSSPP on? GPU might matter if it's a graphical issue.

Windows

sm32x:
Does this actually fix networking for Ridge Racer 2? I tested this and while I can host OK on the instance running the adhoc server, the client instance either says "communication error" or "no games found." It also sometimes gets stuck at "searching for a race open for entries." If I try to host a game with a client instance, it just says communication error. Tested on v1.6 with EUR RR2.

adenovan:
does the game call this hle function ?

https://github.com/hrydgard/ppsspp/blob/master/Core/HLE/sceUtility.cpp#L675 ? 

double check on host and client.

if yes maybe we should emulate it properly by implementing the actual channel instead of force it to 1 , on SceUtilitySetParamInt And SceUtilityGetParamInt , its already has commented line on link to the ini file , try to uncomment that and make the options in networking settings.

there is 4 adhoc channel in psp , actually 3 because automatic will choose 1 of the 3 channel from psp sdk depend on the game call like this case.


`#define PSP_SYSTEMPARAM_ADHOC_CHANNEL_AUTOMATIC  0`
`#define PSP_SYSTEMPARAM_ADHOC_CHANNEL_1		1`
`#define PSP_SYSTEMPARAM_ADHOC_CHANNEL_6		6`
`#define PSP_SYSTEMPARAM_ADHOC_CHANNEL_11	11`

[...]

Issue #8756, opened 2016-05-19
[GUI] Networking


Anuskuss:
This isn't an real issue but it would be convenient:

In **Networking** when you select **Enable built-in PRO ad hoc server** it should automatically change **Change PRO ad hoc server iP adress** to your local ip adress. This would be one less step for people who just want to play locally.

vnctdj:
I don't agree with this idea.
Lots of people don't use their local IP but the IP of a network adapter (provided by a software like Tunngle for example).

kelichu:
Then add an option in "Networking" called "Set Server IP to my Local IP".

Anuskuss:
I was referring to the **Built-in PRO ad hoc server** which AFAICT always runs on the local machine so my suggestion still stands. When that box is ticked, change the IP adress to `localhost` / `127.0.0.1` (and that option should be grayed out).

PR #13132, opened 2020-07-12
Adhoc fix + UPnP + Multiple Instance

anr2me:
These might be able to fix more games, but it's experimental (based on trial-and-error method without confirming it on real hardware), dirty, and badly written.

Win32&64 Builds: https://www.dropbox.com/s/i6699cyf1tzu29b/PPSSPP_1.10.3-adhocfix_Win32x64.zip?dl=0
Android ARMv7 Build: https://www.dropbox.com/s/dmz6jvesmm3uwp2/PPSSPP_1.10.3-adhocfix_ARMv7.apk?dl=0

I haven't been able to continue working on this for months and probably not in the future either.
Since i won't be updating this anymore (other than making sure it can be build successfully), here are the TODO list i remembered and was planning to do:
1). Grouping related functions into classes (AdhocControl, AdhocMatching, AdhocPDP, AdhocPTP,etc like what JPCSP did) for cleaner and easier to work with future improvements.
2). Replacing some real threads with fake PSP threads (ie. AdhocMatchingEvent, AdhocMatchingInput, etc) since some of them are using sceNet functions (like PdpSend/PdpRecv) which may cause issue when blocking-socket implementation are being done in the future (ie. using hle functions)
3). Implements blocking-socket in similar way to sceIo's Read/Write.
4). Replaces state polling inside sceNet function (to solve issues related to high latency over the internet) with hleDelayResult+CoreTiming::ScheduleEvent (similar to blocking-socket implementation) instead of polling it with sleep_ms which cause stutters and emulation performance degradations.
5). Add an option to show all players latency (current & highest) and packet lost rate (similar to FPS rate)

Hopefully someone can make a cleaner and easier to understand adhoc code and contribute it to official PPSSPP so that new/other contributors/devs can continue working on adhoc code without getting a headache too much LOL

### DIFF
--- a/Core/HLE/proAdhoc.cpp
+++ b/Core/HLE/proAdhoc.cpp
@@ -102,27 +102,14 @@ bool isPTPPortInUse(uint16_t port) {
 	return false;
 }
 
-char* mac2str(SceNetEtherAddr* mac) {
-#if defined(_WIN32)
-	static __declspec(thread) char str[18] = ":::::";
-#elif !PPSSPP_PLATFORM(MAC) && !PPSSPP_PLATFORM(IOS)
-	static __thread char str[18] = ":::::";
-#else
-	// Temporary hack to avoid huge rebase conflicts. Remove this when applying the mac2str rewrite.
-	static char str[18] = ":::::";
-#endif
-	if (mac == NULL) return str;
-	snprintf(str, sizeof(str), "%02x:%02x:%02x:%02x:%02x:%02x", mac->data[0], mac->data[1], mac->data[2], mac->data[3], mac->data[4], mac->data[5]);
+std::string mac2str(SceNetEtherAddr* mac) {
+	char str[18] = ":::::";
 
-	return str;
-}
+	if (mac != NULL) {
+		snprintf(str, sizeof(str), "%02x:%02x:%02x:%02x:%02x:%02x", mac->data[0], mac->data[1], mac->data[2], mac->data[3], mac->data[4], mac->data[5]);
+	}
 
-char* mac2str(SceNetEtherAddr* mac, char* str, size_t size) {
-	if (mac == NULL || str == NULL || size < 18) return NULL;
-
-	snprintf(str, size, "%02x:%02x:%02x:%02x:%02x:%02x", mac->data[0], mac->data[1], mac->data[2], mac->data[3], mac->data[4], mac->data[5]);
-
-	return str;
+	return std::string(str);
 }
 
 SceNetAdhocMatchingMemberInternal* addMember(SceNetAdhocMatchingContext * context, SceNetEtherAddr * mac) {
@@ -131,8 +118,7 @@ SceNetAdhocMatchingMemberInternal* addMember(SceNetAdhocMatchingContext * contex
 	SceNetAdhocMatchingMemberInternal * peer = findPeer(context, mac);
 	// Already existed
 	if (peer != NULL) {
-		char tmpmac[18];
-		WARN_LOG(SCENET, "Member Peer Already Existed! Updating [%s]", mac2str(mac, tmpmac));
+		WARN_LOG(SCENET, "Member Peer Already Existed! Updating [%s]", mac2str(mac).c_str());
 		peer->lastping = CoreTiming::GetGlobalTimeUsScaled();
 	}
 	// Member is not added yet
@@ -158,9 +144,8 @@ void addFriend(SceNetAdhocctlConnectPacketS2C * packet) {
 	SceNetAdhocctlPeerInfo * peer = findFriend(&packet->mac);
 	// Already existed
 	if (peer != NULL) {
-		char tmpmac[18];
 		u32 tmpip = packet->ip;
-		WARN_LOG(SCENET, "Friend Peer Already Existed! Updating [%s][%s][%s]", mac2str(&packet->mac, tmpmac), inet_ntoa(*(struct in_addr*)&tmpip), packet->name.data); //inet_ntoa(*(in_addr*)&packet->ip)
+		WARN_LOG(SCENET, "Friend Peer Already Existed! Updating [%s][%s][%s]", mac2str(&packet->mac).c_str(), inet_ntoa(*(struct in_addr*)&tmpip), packet->name.data); //inet_ntoa(*(in_addr*)&packet->ip)
 		peer->nickname = packet->name;
 		peer->mac_addr = packet->mac;
 		peer->ip_addr = packet->ip;
@@ -348,9 +333,8 @@ void deleteFriendByIP(uint32_t ip) {
 			else prev->next = peer->next;
 			*/
 
-			char tmpmac[18];
 			u32 tmpip = peer->ip_addr;
-			INFO_LOG(SCENET, "Removing Friend Peer %s [%s]", mac2str(&peer->mac_addr, tmpmac), inet_ntoa(*(struct in_addr *)&tmpip)); //inet_ntoa(*(in_addr*)&peer->ip_addr)
+			INFO_LOG(SCENET, "Removing Friend Peer %s [%s]", mac2str(&peer->mac_addr).c_str(), inet_ntoa(*(struct in_addr *)&tmpip)); //inet_ntoa(*(in_addr*)&peer->ip_addr)
 
 			// Free Memory
 			//free(peer);
@@ -501,8 +485,7 @@ void postAcceptAddSiblings(SceNetAdhocMatchingContext * context, int siblingcoun
 			// Spawn Established Event
 			spawnLocalEvent(context, PSP_ADHOC_MATCHING_EVENT_ESTABLISHED, &sibling->mac, 0, NULL);
 
-			char tmpmac[18];
-			INFO_LOG(SCENET, "Accepting Peer %s", mac2str(&sibling->mac, tmpmac));
+			INFO_LOG(SCENET, "Accepting Peer %s", mac2str(&sibling->mac).c_str());
 		}
 	}
 }
@@ -628,8 +611,7 @@ void deletePeer(SceNetAdhocMatchingContext * context, SceNetAdhocMatchingMemberI
 			// Beginning Item
 			else context->peerlist = item->next;
 
-			char tmpmac[18];
-			INFO_LOG(SCENET, "Removing Member Peer %s", mac2str(&peer->mac, tmpmac));
+			INFO_LOG(SCENET, "Removing Member Peer %s", mac2str(&peer->mac).c_str());
 		}
 
 		// Free Peer Memory
@@ -904,8 +886,7 @@ void handleTimeout(SceNetAdhocMatchingContext * context)
 				spawnLocalEvent(context, PSP_ADHOC_MATCHING_EVENT_TIMEOUT, &peer->mac, 0, NULL); // This is the only code that use PSP_ADHOC_MATCHING_EVENT_TIMEOUT, should we let it timedout?
 			}
 
-			char tmpmac[18];
-			INFO_LOG(SCENET, "TimedOut Member Peer %s (%lldms)", mac2str(&peer->mac, tmpmac), (context->timeout/1000));
+			INFO_LOG(SCENET, "TimedOut Member Peer %s (%lldms)", mac2str(&peer->mac).c_str(), (context->timeout/1000));
 
 			// Delete Peer from List
 			deletePeer(context, peer);
@@ -1236,8 +1217,7 @@ int friendFinder(){
 						// Cast Packet
 						SceNetAdhocctlConnectBSSIDPacketS2C* packet = (SceNetAdhocctlConnectBSSIDPacketS2C*)rx;
 
-						char tmpmac[18];
-						INFO_LOG(SCENET, "FriendFinder: Incoming OPCODE_CONNECT_BSSID [%s]", mac2str(&packet->mac, tmpmac));
+						INFO_LOG(SCENET, "FriendFinder: Incoming OPCODE_CONNECT_BSSID [%s]", mac2str(&packet->mac).c_str());
 						// From JPCSP: Some games have problems when the PSP_ADHOCCTL_EVENT_CONNECTED is sent too quickly after connecting to a network. The connection will be set CONNECTED with a small delay (200ms or 200us?)
 						/*if (adhocctlCurrentMode == ADHOCCTL_MODE_GAMEMODE) {
 							setState(ADHOCCTL_STATE_GAMEMODE);
@@ -1316,7 +1296,7 @@ int friendFinder(){
 
 						// Log Incoming Peer
                         u32_le ipaddr = packet->ip;
-						INFO_LOG(SCENET, "FriendFinder: Incoming OPCODE_CONNECT [%s][%s][%s]", mac2str(&packet->mac), inet_ntoa(*(in_addr*)&ipaddr), packet->name.data);
+						INFO_LOG(SCENET, "FriendFinder: Incoming OPCODE_CONNECT [%s][%s][%s]", mac2str(&packet->mac).c_str(), inet_ntoa(*(in_addr*)&ipaddr), packet->name.data);
 
 						// Add User
 						addFriend(packet);
--- a/Core/HLE/proAdhoc.cpp
+++ b/Core/HLE/proAdhoc.cpp
@@ -1182,279 +1182,292 @@ int friendFinder(){
 		// Acquire Network Lock
 		//_acquireNetworkLock();
 
-		// Ping Server
-		now = real_time_now() * 1000000.0; // Use real_time_now()*1000000.0 instead of CoreTiming::GetGlobalTimeUsScaled() if the game gets disconnected from AdhocServer too soon when FPS wasn't stable
-		if (now - lastping >= PSP_ADHOCCTL_PING_TIMEOUT) { // We may need to use lower interval to prevent getting timeout at Pro Adhoc Server through internet
-			// original code : ((sceKernelGetSystemTimeWide() - lastping) >= ADHOCCTL_PING_TIMEOUT)
-			// Update Ping Time
-			lastping = now;
-
-			// Prepare Packet
-			uint8_t opcode = OPCODE_PING;
-
-			// Send Ping to Server, may failed with socket error 10054/10053 if someone else with the same IP already connected to AdHoc Server (the server might need to be modified to differentiate MAC instead of IP)
-			int iResult = send(metasocket, (const char *)&opcode, 1, 0);
-			/*if (iResult == SOCKET_ERROR) {
-			ERROR_LOG(SCENET, "FriendFinder: Socket Error (%i) when sending OPCODE_PING", errno);
-			//friendFinderRunning = false;
-			}*/
+		// Reconnect when disconnected while Adhocctl is still inited
+		if (metasocket == (int)INVALID_SOCKET && netAdhocctlInited) {
+			if (g_Config.bEnableWlan && initNetwork(&product_code) == 0) {
+				networkInited = true;
+			}
 		}
 
-		// Wait for Incoming Data
-		int received = recv(metasocket, (char *)(rx + rxpos), sizeof(rx) - rxpos, 0);
+		if (networkInited) {
+			// Ping Server
+			now = real_time_now() * 1000000.0; // Use real_time_now()*1000000.0 instead of CoreTiming::GetGlobalTimeUsScaled() if the game gets disconnected from AdhocServer too soon when FPS wasn't stable
+			if (now - lastping >= PSP_ADHOCCTL_PING_TIMEOUT) { // We may need to use lower interval to prevent getting timeout at Pro Adhoc Server through internet
+				// original code : ((sceKernelGetSystemTimeWide() - lastping) >= ADHOCCTL_PING_TIMEOUT)
+				// Update Ping Time
+				lastping = now;
 
-		// Free Network Lock
-		//_freeNetworkLock();
+				// Prepare Packet
+				uint8_t opcode = OPCODE_PING;
 
-		// Received Data
-		if (received > 0) {
-			// Fix Position
-			rxpos += received;
-
-			// Log Incoming Traffic
-			//printf("Received %d Bytes of Data from Server\n", received);
-			INFO_LOG(SCENET, "Received %d Bytes of Data from Adhoc Server", received);
-		}
-
-		// Handle Packets
-		if (rxpos > 0) {
-			// BSSID Packet
-			if (rx[0] == OPCODE_CONNECT_BSSID) {
-				// Enough Data available
-				if (rxpos >= (int)sizeof(SceNetAdhocctlConnectBSSIDPacketS2C)) {
-					// Cast Packet
-					SceNetAdhocctlConnectBSSIDPacketS2C * packet = (SceNetAdhocctlConnectBSSIDPacketS2C *)rx;
-
-					char tmpmac[18];
-					INFO_LOG(SCENET, "FriendFinder: Incoming OPCODE_CONNECT_BSSID [%s]", mac2str(&packet->mac, tmpmac));
-					// From JPCSP: Some games have problems when the PSP_ADHOCCTL_EVENT_CONNECTED is sent too quickly after connecting to a network. The connection will be set CONNECTED with a small delay (200ms or 200us?)
-					/*if (adhocctlCurrentMode == ADHOCCTL_MODE_GAMEMODE) {
-						setState(ADHOCCTL_STATE_GAMEMODE);
-						notifyAdhocctlHandlers(ADHOCCTL_EVENT_GAME, 0);
-					}
-					else {
-						setState(ADHOCCTL_STATE_CONNECTED);
-						notifyAdhocctlHandlers(ADHOCCTL_EVENT_CONNECT, 0);
-					}*/
-					
-					// Update User BSSID
-					parameter.bssid.mac_addr = packet->mac; // This packet seems to contains Adhoc Group Creator's BSSID (similar to AP's BSSID) so it shouldn't get mixed up with local MAC address
-					// Notify Event Handlers
-					//notifyAdhocctlHandlers(ADHOCCTL_EVENT_CONNECT, 0);
-					// Change State
-					threadStatus = ADHOCCTL_STATE_CONNECTED;
-					// Give time a little time
-					//sceKernelDelayThread(adhocEventDelayMS * 1000);
-					//sleep_ms(adhocEventDelayMS);
-					
-					// Move RX Buffer
-					memmove(rx, rx + sizeof(SceNetAdhocctlConnectBSSIDPacketS2C), sizeof(rx) - sizeof(SceNetAdhocctlConnectBSSIDPacketS2C));
-
-					// Fix RX Buffer Length
-					rxpos -= sizeof(SceNetAdhocctlConnectBSSIDPacketS2C);
+				// Send Ping to Server, may failed with socket error 10054/10053 if someone else with the same IP already connected to AdHoc Server (the server might need to be modified to differentiate MAC instead of IP)
+				int iResult = send(metasocket, (const char*)&opcode, 1, 0);
+				if (iResult == SOCKET_ERROR) {
+					ERROR_LOG(SCENET, "FriendFinder: Socket Error (%i) when sending OPCODE_PING", errno);
+					networkInited = false;
+					shutdown(metasocket, SD_BOTH);
+					closesocket(metasocket);
+					metasocket = (int)INVALID_SOCKET;
 				}
 			}
 
-			// Chat Packet
-			else if (rx[0] == OPCODE_CHAT) {
-				// Enough Data available
-				if (rxpos >= (int)sizeof(SceNetAdhocctlChatPacketS2C)) {
-					// Cast Packet
-					SceNetAdhocctlChatPacketS2C * packet = (SceNetAdhocctlChatPacketS2C *)rx;
-					INFO_LOG(SCENET, "FriendFinder: Incoming OPCODE_CHAT");
-					
-					// Fix strings with null-terminated
-					packet->name.data[ADHOCCTL_NICKNAME_LEN - 1] = 0;
-					packet->base.message[ADHOCCTL_MESSAGE_LEN - 1] = 0;
+			// Wait for Incoming Data
+			int received = recv(metasocket, (char*)(rx + rxpos), sizeof(rx) - rxpos, 0);
 
-					// Add Incoming Chat to HUD
-					NOTICE_LOG(SCENET, "Received chat message %s", packet->base.message);
-					incoming = "";
-					name = (char *)packet->name.data;
-					incoming.append(name.substr(0, 8));
-					incoming.append(": ");
-					incoming.append((char *)packet->base.message);
-					chatLog.push_back(incoming);
-					//im new to pointer btw :( doesn't know its safe or not this should update the chat screen when data coming
-					if (chatScreenVisible) {
-						updateChatScreen = true;
-					}
-					else {
-						if (newChat < 50) {
-							newChat += 1;
+			// Free Network Lock
+			//_freeNetworkLock();
+
+			// Received Data
+			if (received > 0) {
+				// Fix Position
+				rxpos += received;
+
+				// Log Incoming Traffic
+				//printf("Received %d Bytes of Data from Server\n", received);
+				INFO_LOG(SCENET, "Received %d Bytes of Data from Adhoc Server", received);
+			}
+
+			// Handle Packets
+			if (rxpos > 0) {
+				// BSSID Packet
+				if (rx[0] == OPCODE_CONNECT_BSSID) {
+					// Enough Data available
+					if (rxpos >= (int)sizeof(SceNetAdhocctlConnectBSSIDPacketS2C)) {
+						// Cast Packet
+						SceNetAdhocctlConnectBSSIDPacketS2C* packet = (SceNetAdhocctlConnectBSSIDPacketS2C*)rx;
+
+						char tmpmac[18];
+						INFO_LOG(SCENET, "FriendFinder: Incoming OPCODE_CONNECT_BSSID [%s]", mac2str(&packet->mac, tmpmac));
+						// From JPCSP: Some games have problems when the PSP_ADHOCCTL_EVENT_CONNECTED is sent too quickly after connecting to a network. The connection will be set CONNECTED with a small delay (200ms or 200us?)
+						/*if (adhocctlCurrentMode == ADHOCCTL_MODE_GAMEMODE) {
+							setState(ADHOCCTL_STATE_GAMEMODE);
+							notifyAdhocctlHandlers(ADHOCCTL_EVENT_GAME, 0);
 						}
+						else {
+							setState(ADHOCCTL_STATE_CONNECTED);
+							notifyAdhocctlHandlers(ADHOCCTL_EVENT_CONNECT, 0);
+						}*/
+
+						// Update User BSSID
+						parameter.bssid.mac_addr = packet->mac; // This packet seems to contains Adhoc Group Creator's BSSID (similar to AP's BSSID) so it shouldn't get mixed up with local MAC address
+						// Notify Event Handlers
+						//notifyAdhocctlHandlers(ADHOCCTL_EVENT_CONNECT, 0);
+						// Change State
+						threadStatus = ADHOCCTL_STATE_CONNECTED;
+						// Give time a little time
+						//sceKernelDelayThread(adhocEventDelayMS * 1000);
+						//sleep_ms(adhocEventDelayMS);
+
+						// Move RX Buffer
+						memmove(rx, rx + sizeof(SceNetAdhocctlConnectBSSIDPacketS2C), sizeof(rx) - sizeof(SceNetAdhocctlConnectBSSIDPacketS2C));
+
+						// Fix RX Buffer Length
+						rxpos -= sizeof(SceNetAdhocctlConnectBSSIDPacketS2C);
 					}
-
-					// Move RX Buffer
-					memmove(rx, rx + sizeof(SceNetAdhocctlChatPacketS2C), sizeof(rx) - sizeof(SceNetAdhocctlChatPacketS2C));
-
-					// Fix RX Buffer Length
-					rxpos -= sizeof(SceNetAdhocctlChatPacketS2C);
 				}
-			}
 
-			// Connect Packet
-			else if (rx[0] == OPCODE_CONNECT) {
-				// Enough Data available
-				if (rxpos >= (int)sizeof(SceNetAdhocctlConnectPacketS2C)) {
-					// Cast Packet
-					SceNetAdhocctlConnectPacketS2C * packet = (SceNetAdhocctlConnectPacketS2C *)rx;
+				// Chat Packet
+				else if (rx[0] == OPCODE_CHAT) {
+					// Enough Data available
+					if (rxpos >= (int)sizeof(SceNetAdhocctlChatPacketS2C)) {
+						// Cast Packet
+						SceNetAdhocctlChatPacketS2C* packet = (SceNetAdhocctlChatPacketS2C*)rx;
+						INFO_LOG(SCENET, "FriendFinder: Incoming OPCODE_CHAT");
 
-					// Fix strings with null-terminated
-					packet->name.data[ADHOCCTL_NICKNAME_LEN - 1] = 0;
+						// Fix strings with null-terminated
+						packet->name.data[ADHOCCTL_NICKNAME_LEN - 1] = 0;
+						packet->base.message[ADHOCCTL_MESSAGE_LEN - 1] = 0;
 
-					// Log Incoming Peer
-					INFO_LOG(SCENET, "FriendFinder: Incoming OPCODE_CONNECT [%s][%s][%s]", mac2str(&packet->mac), inet_ntoa(*(in_addr*)&packet->ip), packet->name.data);
+						// Add Incoming Chat to HUD
+						NOTICE_LOG(SCENET, "Received chat message %s", packet->base.message);
+						incoming = "";
+						name = (char*)packet->name.data;
+						incoming.append(name.substr(0, 8));
+						incoming.append(": ");
+						incoming.append((char*)packet->base.message);
+						chatLog.push_back(incoming);
+						//im new to pointer btw :( doesn't know its safe or not this should update the chat screen when data coming
+						if (chatScreenVisible) {
+							updateChatScreen = true;
+						}
+						else {
+							if (newChat < 50) {
+								newChat += 1;
+							}
+						}
 
-					// Add User
-					addFriend(packet);
+						// Move RX Buffer
+						memmove(rx, rx + sizeof(SceNetAdhocctlChatPacketS2C), sizeof(rx) - sizeof(SceNetAdhocctlChatPacketS2C));
 
-					/* // Make sure GameMode participants are all joined (including self MAC)
-					if (adhocctlCurrentMode == PSP_ADHOCCTL_MODE_GAMEMODE) {
-						// From JPCSP: Join complete when all the required MACs have joined
-					}*/
-
-					// Update HUD User Count
-					incoming = "";
-					incoming.append((char *)packet->name.data);
-					incoming.append(" Joined ");
-					//do we need ip?
-					//joined.append((char *)packet->ip);
-					chatLog.push_back(incoming);
-					//im new to pointer btw :( doesn't know its safe or not this should update the chat screen when data coming
-					if (chatScreenVisible) {
-						updateChatScreen = true;
+						// Fix RX Buffer Length
+						rxpos -= sizeof(SceNetAdhocctlChatPacketS2C);
 					}
-				
+				}
+
+				// Connect Packet
+				else if (rx[0] == OPCODE_CONNECT) {
+					// Enough Data available
+					if (rxpos >= (int)sizeof(SceNetAdhocctlConnectPacketS2C)) {
+						// Cast Packet
+						SceNetAdhocctlConnectPacketS2C* packet = (SceNetAdhocctlConnectPacketS2C*)rx;
+
+						// Fix strings with null-terminated
+						packet->name.data[ADHOCCTL_NICKNAME_LEN - 1] = 0;
+
+						// Log Incoming Peer
+                        u32_le ipaddr = packet->ip;
+						INFO_LOG(SCENET, "FriendFinder: Incoming OPCODE_CONNECT [%s][%s][%s]", mac2str(&packet->mac), inet_ntoa(*(in_addr*)&ipaddr), packet->name.data);
+
+						// Add User
+						addFriend(packet);
+
+						/* // Make sure GameMode participants are all joined (including self MAC)
+						if (adhocctlCurrentMode == PSP_ADHOCCTL_MODE_GAMEMODE) {
+							// From JPCSP: Join complete when all the required MACs have joined
+						}*/
+
+						// Update HUD User Count
+						incoming = "";
+						incoming.append((char*)packet->name.data);
+						incoming.append(" Joined ");
+						//do we need ip?
+						//joined.append((char *)packet->ip);
+						chatLog.push_back(incoming);
+						//im new to pointer btw :( doesn't know its safe or not this should update the chat screen when data coming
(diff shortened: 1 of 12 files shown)


boy13epic:
are You AdamN?

hrydgard:
Quite an impressive set of changes! I wish I knew enough about AdHoc to properly review this code and work on it...

I do agree that we should finally get the change that allows multi instance on one PC in.

I see that there are details of fixed games in the commits, but do you have a more organized list of what games are actually expected to work with this change in? Are you aware of any regressions?

The timing changes (from PSP time to real time in a couple of places) do make sense in a networking context but also do feel a little scary.

MojoJojoDojo:
Me and some friends made a list of games that work on both single PC and over the internet :
https://docs.google.com/spreadsheets/d/1qXXAAE_HBxa4hlYVUcltZdpHlxsUDkTsWkvK3X6d-4o/edit#gid=0
If someone wanna know the changes and regressions (some of them are based on an older build though). 
I didn't test every game with adhoc thought but I think it's enough.

EX-e-LD:
@ANR2ME @hrydgard
Hi,it is me again.I have been wondering,if code was to say rewritten,would it mean it can break some games that worked before?I'm curious at this point because it would be lot of trouble to have games that worked,fix if they would be broken upon merging this.I would double test this because lot of games that work could be broken with this.I don't need theory,we need proof.

TIA

adenovan:
when its merged make sure be ready to host #12340 as well

Core/HLE/proAdhoc.cpp
```diff
@@ -76,7 +76,7 @@ bool isLocalServer = false;
 uint8_t PPSSPP_ID = 0;
 sockaddr LocalhostIP;
 sockaddr LocalIP;
-int defaultWlanChannel = 1;
+int defaultWlanChannel = PSP_SYSTEMPARAM_ADHOC_CHANNEL_1; // Don't put 0(Auto) here, it needed to be a valid/actual channel number
```
adenovan, line 80: better use 11 on this one to match psp behavior #10698

Common/Log.h
```diff
@@ -91,7 +91,7 @@ void GenericLog(LogTypes::LOG_LEVELS level, LogTypes::LOG_TYPE type,
 bool GenericLogEnabled(LogTypes::LOG_LEVELS level, LogTypes::LOG_TYPE type);
 
 #if defined(LOGGING) || defined(_DEBUG) || defined(DEBUGFAST) || defined(_WIN32)
-#define MAX_LOGLEVEL DEBUG_LEVEL
+#define MAX_LOGLEVEL VERBOSE_LEVEL
```
unknownbrackets, line 94: Not sure if we want to change this?

-[Unknown]

unknownbrackets (commented):
It looks like the only timeout changes are network related, right?

-[Unknown]

Core/HLE/sceNetAdhoc.cpp
```diff
@@ -885,15 +1065,23 @@ static int sceNetAdhocctlScan() {
 				return ERROR_NET_ADHOCCTL_DISCONNECTED; // ERROR_NET_ADHOCCTL_BUSY 
 			}
 
+			// Does Connected Event's mipscall need be executed after returning from sceNetAdhocctlScan ?
```
adenovan, line 1068: yep we can only do scan before updating Adhocctlhandler into connected state , scanning on CONNECT STATE lead to crash on real psp

Core/HLE/sceNetAdhoc.cpp
```diff
@@ -1107,62 +1287,94 @@ static u32 sceNetAdhocctlDisconnect() {
...
 		}
+
+		// Clear Peer List
+		int32_t peercount = 0;
+		freeFriendsRecursive(friends, &peercount);
+		INFO_LOG(SCENET, "Cleared Peer List (%i)", peercount);
+		// Delete Peer Reference
+		friends = NULL;
 		//May also need to clear Handlers
 		adhocctlHandlers.clear();
```
adenovan, line 1328: will be my next note on testing with auto test does term really removing the adhocctl handler?

Core/HLE/sceNetAdhoc.cpp
```diff
@@ -2004,21 +2323,22 @@ static int sceNetAdhocPtpConnect(int id, int timeout, int flag) {
...
-								INFO_LOG(SCENET, "sceNetAdhocPtpConnect[%i:%u]: Established (%u.%u.%u.%u:%u)", id, socket->lport, pip[0], pip[1], pip[2], pip[3], socket->pport);
+								INFO_LOG(SCENET, "sceNetAdhocPtpConnect[%i:%u]: Established (%s:%u)", id, socket->lport, inet_ntoa(peer.sin_addr), socket->pport);
 								
 								// Success
 								return 0;
 							}
 							
 							// Timeout occured
-							return ERROR_NET_ADHOC_TIMEOUT;
+							return ERROR_NET_ADHOC_CONNECTION_REFUSED; // ERROR_NET_ADHOC_TIMEOUT;
```
adenovan, line 2333: this one looks scary and probably can be regression over internet and real psp using aemu plugin , note this one if some game compability broken or not in high latency

Common/Log.h
ANR2ME: I couldn't see any verbose log when debugging without it, any better way?

Core/HLE/sceNetAdhoc.cpp
ANR2ME: As i remembered i took that return code from JPCSP, it seems they only use a few error codes instead of the whole set like we do.

ANR2ME:
> 
> 
> Quite an impressive set of changes! I wish I knew enough about AdHoc to properly review this code and work on it...
> 
> I do agree that we should finally get the change that allows multi instance on one PC in.
> 
> I see that there are details of fixed games in the commits, but do you have a more organized list of what games are actually expected to work with this change in? Are you aware of any regressions?
> 
> The timing changes (from PSP time to real time in a couple of places) do make sense in a networking context but also do feel a little scary.

There might be some regression over the internet i think, as i remembered someone mentioning Monster Hunter games, since i couldn't test it over the internet i never know what happened.

Core/HLE/proAdhoc.cpp
ANR2ME: i was using 11 when i find out JPCSP using 11 as default, but later changed it back to 1 just to be different LOL

Core/HLE/sceNetAdhoc.cpp
adenovan: we have a whole set of error code but its not precise like real console return so its broke some register often , im trying to increase the precision of the error by messing with the control and especially this api on digimon games , due to aemu code often make the game timeout to reduce latency to trick some games this is need a proper testing from the community that use upnp to play.

adenovan: embarking quest on monster hunter does relying a lot in this api before the game decide to drop a peer.

ANR2ME:
For the rest of the compilation error,
on Android seems to be related to CMakeLists.txt which i'm not familiar with :( i guess i won't be able to fix it.

on Windows it seems to be related to Windows SDK version, since i don't have the latest one i can't change it either, i'll let you guys change it.

ANR2ME:
> 
> 
> It looks like the only timeout changes are network related, right?
> 
> -[Unknown]

yes, only timeout used in adhoc codes.

Core/HLE/sceNetAdhoc.cpp
```diff
@@ -1945,15 +2263,21 @@ static int sceNetAdhocPtpConnect(int id, int timeout, int flag) {
...
+					// Switch to Nonblocking Behaviour. Forcing blocking behaviour on the first connect may fix connection issue on GvG Next Plus, But i don't like using blocking socket with infinite timeout if the game it self were asking for non-blocking behaviour :(
+					/*if (nbio == 0) 
+					{
 						// Overwrite Socket Option
 						changeBlockingMode(socket->id, 1);
-					}
+					}*/
 					
 					// Connect Socket to Peer (Nonblocking)
+					// NOTE: Based on what i read at stackoverflow, The First Non-blocking POSIX connect will always returns EAGAIN/EWOULDBLOCK because it returns without waiting for ACK/handshake, But GvG Next Plus is treating non-blocking PtpConnect just like blocking connect, May be on a real PSP the first non-blocking sceNetAdhocPtpConnect can be successfull?
```
ANR2ME, line 2280: @adenovan  you might want to confirm this on real PSP.
Based on how GvG Next Plus is using non-blocking PtpConnect like a blocking one making me to have suspicion, that a non-blocking PtpConnect might have the possibility to be successfully connected on the first attempt.

While a non-blocking POSIX connect are not possible to be successful on the first attempt due to returning without waiting for ACK (at least based on what i read at stackoverflow, never tested this my self)

ANR2ME:
Okay i'm in a mess now :( i solve a simple conflict on proAdhoc.h through github directly, which apparently creating a merge commit instead of rebase, and ended making a lot of conflicts with my local branch, so i deleted my local branch and re-pull from adhoc_fix at github, and now i don't know how to rebase it to make my commits to be on top :( without deleting the branch on github that is

unknownbrackets:
Even in the worst case, you can always use `git reflog` to get back to your old commits.

All you need to do though is the following:

```bash
git fetch upstream
git rebase upstream/master
git push origin -f adhoc_fix
```

This assumes "upstream" is the name of the hrydgard remote and "origin" is the ANR2ME remote, which is the common convention.  The important thing here is `-f` to tell git it's okay to overwrite the branch head even though it no longer has the merge in it.

-[Unknown]

EX-e-LD:
@ANR2ME 
Hi,me again.If i understand correctly Mon Hun games won't work like they did before the fix?If so,then what is point of this fix if it broke Mon Hun games,which run best on Ad-Hoc;tested myself every game for PSP.Can we talk about this,how it can really break something.

TIA

MojoJojoDojo:
> @ANR2ME
> Hi,me again.If i understand correctly Mon Hun games won't work like they did before the fix?If so,then what is point of this fix if it broke Mon Hun games,which run best on Ad-Hoc;tested myself every game for PSP.Can we talk about this,how it can really break something.
> 
> TIA

I think that monster hunter might work on the new build because I tried Worms Open Warfare 2 and it works and it uses the same adhoc functions as MH.

EX-e-LD:
Ad-Hoc is my temporary to-do project anyways,i just want to play the best games and finish them,same what i did with GBA,might take lot of months,but i hope i will do all i plan without any bugs in the way.Anyways,thank you.

EX-e-LD:
@MojoJojoDojo 
Hi,me again.I don't want to sound rude but comparing what emu-mp.net and GuenosNoLife,you just scratched tip of an iceberg regarding testing.I won't say 78 is not much,but comparing what people did with emulator itself,i think games are based on interest of play,not to be sure what is good and what not.

MojoJojoDojo:
> @MojoJojoDojo
> Hi,me again.I don't want to sound rude but comparing what emu-mp.net and GuenosNoLife,you just scratched tip of an iceberg regarding testing.I won't say 78 is not much,but comparing what people did with emulator itself,i think games are based on interest of play,not to be sure what is good and what not.

That list you mentioned is very old I have an even better list but its private until I conclude things more.
But I dont think anyone else would have done what me and others did there you should respect the efforts being done.

boy13epic:
> @ANR2ME @hrydgard
> Hi,it is me again.I have been wondering,if code was to say rewritten,would it mean it can break some games that worked before?I'm curious at this point because it would be lot of trouble to have games that worked,fix if they would be broken upon merging this.I would double test this because lot of games that work could be broken with this.I don't need theory,we need proof.
> 
> TIA

stop trashing sorer’s list.

EX-e-LD:
@boy13epic
Hi.Who said i was trashing MojoJojoDojo's list,do you get that out of context?I said that his list can be better if he goes by testing everything he wants,not he finds interest in.Please if you found it offensive i will delete my message,but if it doesn't don't say something offensive just because you read between lines.

TIA

hrydgard:
Let's stop talking about lists. 

I'm not quite sure where to start here - I'm thinking about pulling out pieces of this and merging them one by one, so we can reduce any breakage to a minimum through careful testing of each part.

For example, the UPNP stuff should be mergable mostly on its own.

Does that sound resonable?

EX-e-LD:
@hrydgard
I'm not developer by any means,but does that mean we would have and fixes if we found MH games broken?That would be good as it sounds,but that is going to be lot of work.

boy13epic:
this is resonable

ANR2ME:
> 
> 
> Let's stop talking about lists.
> 
> I'm not quite sure where to start here - I'm thinking about pulling out pieces of this and merging them one by one, so we can reduce any breakage to a minimum through careful testing of each part.
> 
> For example, the UPNP stuff should be mergable mostly on its own.
> 
> Does that sound resonable?

That would be great! I realized too late for making a mistake by mixing bug fix with feature addition on the same branch until it's getting to large :(

Btw, some of the commits might includes minor changes that might not have anything todo with what's written in the comment, so you might want to remove those unrelated parts when cherry picking it.

ANR2ME:
> 
> 
> @ANR2ME
> Hi,me again.If i understand correctly Mon Hun games won't work like they did before the fix?If so,then what is point of this fix if it broke Mon Hun games,which run best on Ad-Hoc;tested myself every game for PSP.Can we talk about this,how it can really break something.
> 
> TIA

Actually you just need to test out the builds i provides on top of this page to see whether it really breaks any previously working games or not? if you're really curious, and don't forget to let us know the result :-)

If you're able to play multiplayer games for more than 5 minutes without problem, it will most likely have no issue (unless if single player was also having an issue)

EX-e-LD:
@ANR2ME
Issue is,UPnP kinda pulls issues on MHFU?Could you see why it is like that?

ANR2ME:
> 
> 
> @ANR2ME
> Issue is,UPnP kinda pulls issues on MHFU?Could you see why it is like that?

What kind of issue?
If you see a red message on screen saying "Failed/Unable to detected UPnP device" (or something like that) on screen, it means you don't have UPnP capable router (or haven't enable UPnP on your router)

If you're playing over the internet with other people using a different build than this build, you need to disable the "UPnP use original port" setting, it's enabled by default to maintain compatibility with real PSP and other/non-PPSSPP emulators not supporting port offset (only effective if you have port offset > 0), so that real PSP or non-PPSSPP emulators will see it as their own over the internet (external port doesn't use port offset), but locally still mapped to ports using port offset.

EX-e-LD:
@ANR2ME
UPnP is enabled,but it just not performes same like it does before your build.Tested on neighbor's PC across 2 **different** networks and it is somehow really different,i can't explain.

ANR2ME:
Hmm.. this is odd... why am i seeing these commits from @unknownbrackets after rebasing from my branch (not really sure why i need to pull before i can push to my own branch)

CMakeLists.txt
```diff
@@ -1818,6 +1821,80 @@ if(USE_DISCORD AND NOT IOS AND NOT LIBRETRO)
...
+	#  find_library (RESOLV_LIBRARY NAMES resolv)
+	#  set (LDLIBS ${SOCKET_LIBRARY} ${NSL_LIBRARY} ${RESOLV_LIBRARY} ${LDLIBS})
+	endif()
+	if (MSVC)
+		# Suppress noise warnings
+		target_compile_definitions(miniupnpc _CRT_SECURE_NO_WARNINGS _WINSOCK_DEPRECATED_NO_WARNINGS)
+	endif()
+	if (UPNPC_BUILD_STATIC)
+		add_library(miniupnpc STATIC ${MINIUPNPC_SOURCES})
+		target_link_libraries(${CoreLibName} miniupnpc ${LDLIBS})
```
ANR2ME, line 1893: Does anyone know why Android LIBRETRO is the only successful build on travisci while the rest of Android compiles failed to recognize PortManager.h/cpp and getting these kind of error?
```
[arm64-v8a] SharedLibrary  : libppsspp_jni.so

./obj/local/arm64-v8a/libppsspp_core.a(sceNet.o): In function `__NetInit()':

/home/travis/build/hrydgard/ppsspp/android/jni/../../Core/HLE/sceNet.cpp:256: undefined reference to `g_PortManager'
```

hrydgard: I'll look into it later today.

EX-e-LD:
@ANR2ME 
Hi,me again.I noticed that you were also present when GuenosNoLife / emu-mp.net made their list.So can i ask you (not my business and i will delete if you/owner want) what made you quit after all those years?

TIA

hrydgard:
Please keep irrelevant communication off here, this is about the code.

CMakeLists.txt
hrydgard: This is really weird. I was able to build it just fine locally, but on CI, I don't see PortManager.cpp in the build log so it seems it just isn't building it?? Bizarre...

hrydgard: As for the Windows build, in miniupnpc.vcxproj, you're gonna have to add this workaround in a few places (manually, using a text editor):

```
    <Import Project="..\Windows\fix_2017.props" />
```

See ext/glslang.vcxproj .

hrydgard: Ah I just realized what's wrong with Android.

We still have an old Android.mk based build which we are using on CI. It's probably time to rip it out, but in the meantime, you need to add PortManager.cpp to android/jni/Android.mk . You'll figure out how.

ANR2ME: I've done both of them, but on android it still have the same error, while on Windows... i think it's getting worse LOL

ANR2ME: do i need to add these kind of line on Android.mk ?
```
$(call import-module,miniupnpc)
```

and also need to add miniupnpc to Locals.mk ?

hrydgard: Oh, right hehe. Yes you might have to do that and also add an Android.mk for it (alternatively, just add the files directly from one of the other Android.mk that are sprinkled around).

Really we probably should just kill it, though will need to reconfigure that build bot. Will open an issue.

ANR2ME: One of the source file need to be generated like this on CMakeLists.txt
```
configure_file(${MINIUPNP_DIR}/miniupnpcstrings.h.cmake ${CMAKE_CURRENT_BINARY_DIR}/miniupnpcstrings.h) # by default miniupnp repo doesn't contains miniupnpcstrings.h and need to be generated
```
Does Android.mk can also generate that file?

hrydgard:
Two things left to fix the Windows build:

* Still something different about the vcproj, not sure exactly what but I'll help compare to the others later, causing these weird `can't find <string>` things 

* The UWP vcxprojs also need the new PortManager.cpp added.

Yes it's a royal pain with all these build systems...

ANR2ME:
> 
> 
> Two things left to fix the Windows build:
> 
>     * Still something different about the vcproj, not sure exactly what but I'll help compare to the others later, causing these weird `can't find <string>` things
> 
>     * The UWP vcxprojs also need the new PortManager.cpp added.
> 
> 
> Yes it's a royal pain with all these build systems...

I've tried the 1st method from here (Ignore Precompiled Headers #1) but still no luck https://stackoverflow.com/questions/44219901/cannot-open-include-file-stdio-h-visual-studio-community-2017-c-error

Btw, how do i tried to build the UWP projects locally? i never use UWP before

Anuskuss:
@ANR2ME Gave this a try and YGO:TF2 (only game I really play adhoc) works perfectly but I had to change the address from `localhost` to my actual private ip. Would be nice if you could either fix it or drop a little warning for people like me 😄.

ANR2ME:
> 
> 
> @ANR2ME Gave this a try and YGO:TF2 (only game I really play adhoc) works perfectly but I had to change the address from `localhost` to my actual private ip. Would be nice if you could either fix it or drop a little warning for people like me 😄.

If you're playing on LAN or over the internet, DO NOT use localhost, localhost is only used for multiple instance on one PC.
If you put "localhost" as Adhoc Server it will detects it as 127.0.0.x thus you can't see any other players who connects to you using LAN or public IP from 127.0.01 (assuming you host the adhoc server locally)

Linux/Android (not sure about iOS/MacOS) will always translate localhost to 127.0.0.1, while Windows can translate localhost to LAN IP but it's inconsistent, so it's better to explicitly tells PPSSPP which IP you want to use to connect to adhoc server, even when you're hosting the adhoc server your self if you want to see other players over LAN or internet.

One PC can have multiple IP especially if you have VPN, LAN through cable, and also LAN through wifi all together, PPSSPP doesn't know which IP you'll be using before you're connected to the adhoc server, however, there are codes that need to know the IP before connecting to adhoc server as i remembered (ie. for binding purpose), and the multiple instance feature need to bind only to a single IP instead of the whole existing IP (0.0.0.0) which is why i differentiates it by detecting "localhost"/127.0.0.x as multiple instance.

Anuskuss:
Many years ago I've read somewhere that you should use `localhost` instead of `127.0.0.1` so the application (I think it was MySQL) would use a UNIX socket instead of TCP/IP which was a sort of "free" performance improvment. I think this doesn't apply here but I've requested for this setting to be handled automatically [years ago](https://github.com/hrydgard/ppsspp/issues/8756), so I just left my server at `localhost` and my clients at it's IP address (or rather local domain name) and called it a day.
I don't play over the internet so a change was never necessary (I would've modified it in that case anyway). If this PR finally gets merged, there should atleast be a check in place (maybe change the string from `Network initialized` to `Network initialized (<IP>)` and if it contains a loopback/private IP append a `(same computer only)`/`(local network only)`. This could be really hard to debug because some people might leave that setting alone, expecting it to continue to work. Just a suggestion though, I'm sure they'll find this convo it they search long enough 😉

EX-e-LD:
> Many years ago I've read somewhere that you should use `localhost` instead of `127.0.0.1` so the application (I think it was MySQL) would use a UNIX socket instead of TCP/IP which was a sort of "free" performance improvment. I think this doesn't apply here but I've requested for this setting to be handled automatically [years ago](https://github.com/hrydgard/ppsspp/issues/8756), so I just left my server at `localhost` and my clients at it's IP address (or rather local domain name) and called it a day.
> I don't play over the internet so a change was never necessary (I would've modified it in that case anyway). If this PR finally gets merged, there should atleast be a check in place (maybe change the string from `Network initialized` to `Network initialized (<IP>)` and if it contains a loopback/private IP append a `(same computer only)`/`(local network only)`. This could be really hard to debug because some people might leave that setting alone, expecting it to continue to work. Just a suggestion though, I'm sure they'll find this convo it they search long enough 
> 
> P.S. Since I'm able to talk to the guy who was responsible for the original implementation years ago, I can finally ask: If I don't plan to use Adhoc, do I need to disable both _Enable networking_ and _Enable built-in adhoc server_ or is _Enable networking_ enough? I know it's a dumb question (I've disabled both every time) but I was always wondering if _Enable built-in adhoc server_ would do something without _Enable networking_

Not him,and bit off-topic,but when you want to disable networking,turning off enable networking is enough,from my experience.

ANR2ME:
> 
> 
> Many years ago I've read somewhere that you should use `localhost` instead of `127.0.0.1` so the application (I think it was MySQL) would use a UNIX socket instead of TCP/IP which was a sort of "free" performance improvment. I think this doesn't apply here but I've requested for this setting to be handled automatically [years ago](https://github.com/hrydgard/ppsspp/issues/8756), so I just left my server at `localhost` and my clients at it's IP address (or rather local domain name) and called it a day.
> I don't play over the internet so a change was never necessary (I would've modified it in that case anyway). If this PR finally gets merged, there should atleast be a check in place (maybe change the string from `Network initialized` to `Network initialized (<IP>)` and if it contains a loopback/private IP append a `(same computer only)`/`(local network only)`. This could be really hard to debug because some people might leave that setting alone, expecting it to continue to work. Just a suggestion though, I'm sure they'll find this convo it they search long enough 😉
> 
> P.S. Since I'm able to talk to the guy who was responsible for the original implementation years ago, I can finally ask: If I don't plan to use Adhoc, do I need to disable both _Enable networking_ and _Enable built-in adhoc server_ or is _Enable networking_ enough? I know it's a dumb question (I've disabled both every time) but I was always wondering if _Enable built-in adhoc server_ would do something without _Enable networking_ 😄

As i said before, translating "localhost" into the correct IP is not a simple thing especially if you have multiple IP on your computer (LANs+VPNs+public IP)

For example, you run your own (built-in) adhoc server, and you also play as a client in that same computer and connect to that adhoc server using "localhost", the Adhoc Server will recognize it as IP 127.0.0.1 and it will broadcast that IP to any other players (including players on different PC) and when these players on different PC receive that broadcast data saying your IP is 127.0.0.1 and try to communicate to your PC, it will try to talk to it's self instead because 127.0.0.1 is their own PC from their perspective.
Players who connects using LAN IP can't communicate with players using public IP either, because remote players over the internet will try to talk to their own LAN instead, although most people only have 1 internet connection where LAN IP can be translated to public IP, but it's doesn't mean a PC can't have multiple internet connection (i used to have 2 internet connection in the old days, a USB modem as a backup internet when the LAN one is down)

Even if someone made an algorithm to try to guess the correct IP and it works for one person, it doesn't mean that guessing method will works on other people, after all "guessing" is just an estimation which may not be always be accurate.
An accurate one will be by explicitly use the correct IP/network adapter so Adhoc Server can recognize the correct IP.

devinprater:
Is there a way to select a network adaptor to use, instead of writing an IP address? Like "computer only" mapped to localhost, "local network on adaptor x", and "Over the Internet?" I know many people who I'm trying to get into emulation where typing in an IP address will be hard to get passed. Then again, emulation is an advanced topic.

MojoJojoDojo:
> Is there a way to select a network adaptor to use, instead of writing an IP address? Like "computer only" mapped to localhost, "local network on adaptor x", and "Over the Internet?" I know many people who I'm trying to get into emulation where typing in an IP address will be hard to get passed. Then again, emulation is an advanced topic.

I think that someone similar is done on JPCSP its almost like what you said here.

Anuskuss:
@ANR2ME I think I get what you're trying to say, it just doesn't make sense to me how `localhost` used to work before. But I'll follow your advice on this one and change my server to it's local domain name :relaxed:

> Is there a way to select a network adaptor to use, instead of writing an IP address? Like "computer only" mapped to localhost, "local network on adaptor x", and "Over the Internet?" I know many people who I'm trying to get into emulation where typing in an IP address will be hard to get passed. Then again, emulation is an advanced topic.

@devinprater +1

ANR2ME:
> 
> 
> @ANR2ME I think I get what you're trying to say, it just doesn't make sense to me how `localhost` used to work before. But I'll follow your advice on this one and change my server to it's local domain name ☺️
> 
> > Is there a way to select a network adaptor to use, instead of writing an IP address? Like "computer only" mapped to localhost, "local network on adaptor x", and "Over the Internet?" I know many people who I'm trying to get into emulation where typing in an IP address will be hard to get passed. Then again, emulation is an advanced topic.
> 
> @devinprater +1

it's because of Multiple Instance implementation is exclusively making use of "localhost" and treats it differently.

ANR2ME:
> 
> 
> Is there a way to select a network adaptor to use, instead of writing an IP address? Like "computer only" mapped to localhost, "local network on adaptor x", and "Over the Internet?" I know many people who I'm trying to get into emulation where typing in an IP address will be hard to get passed. Then again, emulation is an advanced topic.

I've added a quick IP selection for Adhoc Sever IP address settings today.
You can try it from the download link on top.

devinprater:
I had to use the INI file for configuring PPSSPP, since the UI doesn't work with screen readers... I'll make an issue about maybe commenting the INI to tell what options are used for and what acceptable parameters are. I just know other users would rather have a selection list than having to figure out an IP.

ANR2ME:
> 
> 
> @ANR2ME
> Issue is,UPnP kinda pulls issues on MHFU?Could you see why it is like that?

I tried to minimize the lags/stutters caused by UPnP today by doing UPnP stuff on a separate thread, could you test it and see whether UPnP is broken or not :D LOL

because, apparently the only MiFi i had that support UPnP is kinda broken (i haven't use it for months and now it won't turn on), so i can no longer test UPnP now :( hopefully UPnP is not broken tho

MojoJojoDojo:
> > @ANR2ME
> > Issue is,UPnP kinda pulls issues on MHFU?Could you see why it is like that?
> 
> I tried to minimize the lags/stutters caused by UPnP today by doing UPnP stuff on a separate thread, could you test it and see whether UPnP is broken or not :D LOL
> 
> because, apparently the only MiFi i had that support UPnP is kinda broken (i haven't use it for months and now it won't turn on), so i can no longer test UPnP now :( hopefully UPnP is not broken tho
> 
> PS: The download links are still the same, on top of this page.

Yes this eliminated most of the stutters and lag from booting up the game and going to multiplayer lobbies.
It also works with other people too.
So I think it's now fine to have it enabled by default now? it seems like an important thing to enable for matchmaking and playing with others.

ANR2ME:
> 
> 
> > > @ANR2ME
> > > Issue is,UPnP kinda pulls issues on MHFU?Could you see why it is like that?
> > 
> > 
> > I tried to minimize the lags/stutters caused by UPnP today by doing UPnP stuff on a separate thread, could you test it and see whether UPnP is broken or not :D LOL
> > because, apparently the only MiFi i had that support UPnP is kinda broken (i haven't use it for months and now it won't turn on), so i can no longer test UPnP now :( hopefully UPnP is not broken tho
> > PS: The download links are still the same, on top of this page.
> 
> Yes this eliminated most of the stutters and lag from booting up the game and going to multiplayer lobbies.
> It also works with other people too.
> So I think it's now fine to have it enabled by default now? it seems like an important thing to enable for matchmaking and playing with others.

Thanks for testing it.

But i don't think it's wise to enable it by default, because not everyone have UPnP-capable router (especially those who only plays single player), which will resulting a continuous "Failed to detect UPnP device" message on screen which can be annoying and taking CPU time for keep trying to connect to UPnP router.

MojoJojoDojo:
> > Is there a way to select a network adaptor to use, instead of writing an IP address? Like "computer only" mapped to localhost, "local network on adaptor x", and "Over the Internet?" I know many people who I'm trying to get into emulation where typing in an IP address will be hard to get passed. Then again, emulation is an advanced topic.
> 
> I've added a quick IP selection for Adhoc Sever IP address settings today.
> You can try it from the download link on top.
> image
I have some more suggestions if its ok.
Have a button that gives you the private IP of the device.
Change or add more to the "enable built-in proadhocserver" because I think its kinda lacking (maybe say that only 1 of the users have to enable it?).

ANR2ME:
> 
> 
> > > Is there a way to select a network adaptor to use, instead of writing an IP address? Like "computer only" mapped to localhost, "local network on adaptor x", and "Over the Internet?" I know many people who I'm trying to get into emulation where typing in an IP address will be hard to get passed. Then again, emulation is an advanced topic.
> > 
> > 
> > I've added a quick IP selection for Adhoc Sever IP address settings today.
> > You can try it from the download link on top.
> > image
> 
> I have some more suggestions if its ok.
> Have a button that gives you the private IP of the device.
> Change or add more to the "enable built-in proadhocserver" because I think its kinda lacking (maybe say that only 1 of the users have to enable it?).

Those IP shown on the list are all private IP found on your PC/device, even my android phone have multiple private IP. 
Although i've decided to removes IP prefixed with "127.", "169.254.", and "0." since they could also be found on some platform but either redundant or unusable for multiplayer.

It's obvious that you only need to use 1 adhoc server as a gathering place, otherwise players won't be able to meet each other if they use different adhoc server each. 
Although most of the time enabling adhoc server on all players have no problem (other than a minor CPU usage for listening a socket). 
What's important is that all players need to use the same "Adhoc Server IP address" in order to meet each other.

hrydgard:
Let's start trying to get this in, piece by piece. I broke out the first part, multi-instance, for merge, in #13169 . Please have a look and test it :)

hrydgard:
I have merged #13169, so please go ahead and rebase this on top of master so we can extract the next part.

You'll get a few conflicts but they should be very minor and easy to resolve.

hrydgard:
Found that I could easily extract a couple of more commits without rebase, so I made #13173 .

ANR2ME:
Apparently the only conflicts i had was within __NetShutdown function LOL pretty cool :)

I should've re-arranged all of those commits to move that "Fix Local IP detection on non-Windows system" commit closer to multiple instance commits, but i guess it's too late now LOL

hrydgard:
Nice!

I liked my #13172 change much that I'm gonna merge that early once CI passes, it makes testing this stuff a lot more pleasant, but I think it won't conflict at all really, unless you did some major changes to the multi instance stuff in some commit I haven't noticed.

hrydgard:
By the way, there's a little thing you do here a lot (and other contributors have too) where you replace some code (for example with UPnP) but leave the old code around in a comment. Most of the time it makes more sense to just delete the old lines, you can get them back from git anyway. Just some feedback :)

ANR2ME:
> 
> 
> By the way, there's a little thing you do here a lot (and other contributors have too) where you replace some code (for example with UPnP) but leave the old code around in a comment. Most of the time it makes more sense to just delete the old lines, you can get them back from git anyway. Just some feedback :)
 
Yeah, most of the commented code were because the changes was still experimental and i wasn't sure which one is better.
In the case i did to UPnP changes recently it was because i couldn't test the new code my self (no longer have a working UPnP device)

hrydgard:
Merged that last bunch of commits. Probably good to do another rebase - there will be some minor conflicts from the reordering I did, but shouldn't be a big deal hopefully.

You'll also get the new convenient "Open New Instance" command in the file menu :)

Anuskuss:
@hrydgard I've built this branch again (with #13169 in master) and the second instance shows `instance: 3` for me 😄. Anyway, this wasn't that PR but I'd like to keep everything in one place so I wanna make a suggestion: for every new instance PPSSPP should also generate a new MAC address.
@ANR2ME I don't play a lot of games that have Adhoc support but one game I do like but which has never worked before is DBZ:SB. So thank you for fixing that 👍
I don't know how difficult it is but could you also include the machine/host name in the IP list? I know it will probably be annoying to support it on all platforms, so feel free to reject.
Can you please explain what exactly _Minimum Timeout_ does? Any why is there a separate _UPnP use original port_ option when there's already _Port offset_?

Speaking of UPnP, is hosting your own server and playing over the internet supported? I've setup my server (pointing to the private instead of the public IP because that didn't work), activated UPnP and I can clearly see two new entries (the second `:PPSSPP` in the service name is kinda redundant):

Service | LAN IP | Protocol | Port | WAN Port
-- | -- | -- | -- | --
PPSSPP:ULES00925:PPSSPP | 192.168.0.181 | TCP | 27312 | 27312
PPSSPP:ULES00925:PPSSPP | 192.168.0.181 | UDP | 1 | 1

I've setup my phone (which isn't on this branch btw) to point to my public IP address (with the help of dynamic DNS) and the network is successfully initialized, but the devices can't see each other. I've tried playing around with port offset (which changes the local and public port of the UDP service) and _UPnP use original port_ (which only changes the local) but to no avail. Do I need to build this branch for my phone or am I doing something wrong?

ANR2ME:
> 
> 
> @hrydgard I've built this branch again (with #13169 in master) and the second instance shows `instance: 3` for me 😄. Anyway, this wasn't that PR but I'd like to keep everything in one place so I wanna make a suggestion: for every new instance PPSSPP should also generate a new MAC address.
> @ANR2ME I don't play a lot of games that have Adhoc support but one game I do like but which has never worked before is DBZ:SB. So thank you for fixing that 👍
> I don't know how difficult it is but could you also include the machine/host name in the IP list? I know it will probably be annoying to support it on all platforms, so feel free to reject.
> Can you please explain what exactly _Minimum Timeout_ does? Any why is there a separate _UPnP use original port_ option when there's already _Port offset_?
> 
> Speaking of UPnP, is hosting your own server and playing over the internet supported? I've setup my server (pointing to the private instead of the public IP because that didn't work), activated UPnP and I can clearly see two new entries (the second `:PPSSPP` in the service name is kinda redundant):
> Service 	LAN IP 	Protocol 	Port 	WAN Port
> PPSSPP:ULES00925:PPSSPP 	192.168.0.181 	TCP 	27312 	27312
> PPSSPP:ULES00925:PPSSPP 	192.168.0.181 	UDP 	1 	1
> 
> I've setup my phone (which isn't on this branch btw) to point to my public IP address (with the help of dynamic DNS) and the network is successfully initialized, but the devices can't see each other. I've tried playing around with port offset (which changes the local and public port of the UDP service) and _UPnP use original port_ (which only changes the local) but to no avail. Do I need to build this branch for my phone or am I doing something wrong?

The instance number will keep increasing on new instance as long there is one PPSSPP running, it will reset after all PPSSPP is closed down.
That instance number will be used to assign the localhost IP (ie. 127.0.0.x) and the MAC address (ie. 02:02:02:02:02:02) so it's easier to recognize while debugging or watching the logs to investigate networking behavior.

Retrieving host name can be platform-specific, the posix/universal way (ie. gethostname) seems to only works well on Windows, while on linux/unix/android(and probably macOS/iOS too) will only shows "localhost".
However, it's better to choose it by IP instead of hostname, since most of the time (as a player who joins) you will need to edit the last number in the IP to match the IP who runs the adhoc server on the same network.

Minimum Timeout is used to override the timeout used by the game, this should prevent disconnections due to high latency (was tested by brazillian players who have high ping, and they said it worked well).
However, there will be noticeable delays in exchange of disconnection when using large minimum-timeout. 
And overriding socket's timeout like this only works on games that use blocking sockets, while games that use non-blocking sockets usually have their own elapsed-time measurement as timeout (probably using the scaled time runs on PSP-side which can be unstable during loading scene due to unstable FPS) thus can't be overridden.

I never had an internet connection capable of port-forwarding (3G/4G provider doesn't allows this) so i never tested hosting built-in adhoc server and using the public IP to connect to it from the same PC.
Theorically as long you can see your public IP recognized as a player in the Adhoc Server logs it should works, but if you see a LAN or localhost IP it will most-like not gonna works, because the IP you see in the Adhoc Server logs will be given to all the players where they will be connecting directly, since all the other players are using public IP they won't be able to connects to your LAN/localhost directly from their side.

Not, it's not redundant on UPnP mapped pors, in your example there are 2 different protocols and different ports. TCP port 27312 is for Adhoc Server, while UDP port 1 is probably the port used by the game, there will be more ports depend on the game.

If both your phone and PC are using the same internet connection it will not works, because:
1). UPnP or port-forwarding in general can only maps 1 public/external port to a single LAN port, playing the same game on 2 device using the same router will cause both of them to fight for the same port, thus not gonna works.
2). Adhoc Server can only assign 1 IP per player, if both players are using the same internet (have the same public IP) one of them will be dropped/kicked from adhoc server.
In this case, using VPN is the only solution, where each players will have their own private IP, but UPnP are not used/needed for LAN or VPN connection.

Anuskuss:
There were a bunch of misunderstanding lol

> The instance number will keep increasing on new instance as long there is one PPSSPP running, it will reset after all PPSSPP is closed down. That instance number will be used to assign the localhost IP (ie. 127.0.0.x) and the MAC address (ie. 02:02:02:02:02:02) so it's easier to recognize while debugging or watching the logs to investigate networking behavior.

No I meant that it's showing as `instance: 3` while it should've been `instance: 2`. Also rn it's using the same MAC address.

> However, it's better to choose it by IP instead of hostname, since most of the time (as a player who joins) you will need to edit the last number in the IP to match the IP who runs the adhoc server on the same network.

Hostname works just fine and will continue to work if the DHCP server assigns a new IP to my server.

> Minimum Timeout is used to override the timeout used by the game

So what does _1_ refer to? 1 ms? That's way too low for a default.

> Not, it's not redundant on UPnP mapped ports

No, I meant the service **name**. Should be `PPSSPP:ULES00925` instead of `PPSSPP:ULES00925:PPSSPP`.

> If both your phone and PC are using the same internet connection it will not works

No, it's not using the same internet connection. I'm talking about my server using my home connection (and public IP) and my phone connecting to it using LTE.

hrydgard:
That first thing sounds wrong - you mean you really start only two instances but still get instance: 3 on the second one? What does the first one show, instance: 1 ?

LunaMoo:
Maybe it's because gpu api validation, isn't PPSSPP booting twice for that?

hrydgard:
Oh, interesting thought...

hrydgard:
Interesting, I only tested in debug.

I wouldn't worry too much about it for now, it's gonna work anyway, but I'm gonna investigate and fix it in a bit.

hrydgard:
Extracted the initial UPnP support commits into #13177. Getting that to build should also fix the build for the rest once you rebase I think, since that's the last of the commit here that has build system changes, right?

ANR2ME:
Yes, i think so, because apparently i won't be able to download UWP framework anytime soon so won't be able to test it.
Just don't forge to add the command line to run the vbs script to generate a header file required by miniupnpc (that header doesn't came by default, they only provide the script)

Anuskuss:
@ANR2ME Thank you for trying to answer my question but next time please just create a new comment instead of editing your previous one because that makes it much easier to read.

> Minimum Timeout is used to override the timeout used by the game, this should prevent disconnections due to high latency (was tested by brazillian players who have high ping, and they said it worked well).

> The minimum timeout on Windows is 1ms while on linux/unit it's 1 usec (1/1000 ms) while default socket's timeout is usually 5 seconds, but since we're talking about minimum number we should avoid using larger number than what the game is using (for compatibility reason) so 1ms is the most compatible one across platforms (because 0 = infinite timeout)

Okay, if I understand correctly if the game goes over the minimum timeout it'll disconnect, right? Why isn't it called just _Timeout_ then and why isn't is set to something realistic like 300 ms?

> I never had an internet connection capable of port-forwarding (3G/4G/LTE provider doesn't allows port-forwarding since it's shared IP) so i never tested hosting built-in adhoc server and using the public IP to connect to it from the same PC.

Hmm, I was under the impression that only the server needs port forwarding and the client is fine to just point to the IP address and connect (even for LTE). How did it work in the past when you were able to just connect to e.g. SushiCat without manually port forwarding/UPnP? And would it work to connect a client with a different IP to my server with the client having port forwarding capabilities?

> PS: If you're using a mobile data (3G/4G/LTE connection) it probably doesn't support port-forwarding as it's shared IP (will need to ask permission to your data provider to open some ports for you), in this case VPN is the only way since it doesn't need port-forwarding.

With VPN you mean a generic one (e.g. NordVPN) or actually running your own VPN server (e.g. WireGuard) and connecting to it from outside?

> In the description it's marked as: PPSSPP:GameID:PlayerName

Yeah, that makes sense since in the default configuration the user is indeed called _PPSSPP_. You can leave it that way but be careful that the username might contain characters with could mess up UPnP or even the router interface.

ANR2ME:
> 
> 
> @ANR2ME Thank you for trying to answer my question but next time please just create a new comment instead of editing your previous one because that makes it much easier to read.
> 
i'm just feeling lazy to quote too many questions on a single post, and we usually use forums for these kind of general questions.

> Okay, if I understand correctly if the game goes over the minimum timeout it'll disconnect, right? Why isn't it called just _Timeout_ then and why isn't is set to something realistic like 300 ms?
> 
It's "Minimum" because we only want to override it if the game is trying to use lower value than our minimum value, if the game want to use higher value we certainly don't want to override it will lower value and ended causing problem to the game.
Most of the time overriding timeout with higher value will cause unnecessary delays if the game already have their own timeout handling procedure, so using the lowest value as the default is the best choice for compatibility reason.

> Hmm, I was under the impression that only the server needs port forwarding and the client is fine to just point to the IP address and connect (even for LTE). How did it work in the past when you were able to just connect to e.g. SushiCat without manually port forwarding/UPnP? And would it work to connect a client with a different IP to my server with the client having port forwarding capabilities?
> 
Adhoc is similar to peer2peer so both side will bind and listen to ports thus both will need port-forwarding, it doesn't works like regular client-server, even when the game show "Host" or "Join" on screen under the hood the one who's closer to being a server (the one who listen to a port) is actually the "Join" one on most adhoc games.

> With VPN you mean a generic one (e.g. NordVPN) or actually running your own VPN server (e.g. WireGuard) and connecting to it from outside?
> 
Any VPN will do if the adhoc server resides within the VPN network, but if you use public adhoc server like sushicat all VPN clients will need to have a different public IP, most VPN have multiple IP so it's rare for 2 people to have the same IP.
I even heard that free VPN like Speedify on android can be used to play on sushicat (it never worked for me tho)

> > In the description it's marked as: PPSSPP:GameID:PlayerName
> 
> Yeah, that makes sense since in the default configuration the user is indeed called _PPSSPP_. You can leave it that way but be careful that the username might contain characters with could mess up UPnP or even the router interface.
True, i do think PlayerName should be removed since it can be customized by player.

OOT: I even surprised to find out that an SSID can use custom font like this O.o
image

hrydgard:
Alright, merged that =)  Looks like you're up for some new little conflicts to solve, but they should be very easy..

Anuskuss:
> i'm just feeling lazy to quote too many questions on a single post, and we usually use forum or discord for these kind of general questions.

PPSSPP definitely needs a wiki entry or some guide for all the network related settings because they're quite difficult to understand. I doubt the average user knows what UPnP or Minimum Timeout is (me included).

hrydgard:
Definitely. Feel free to start writing one :)

ANR2ME:
Btw after rebasing and try to build the Windows release x64 i saw this in the output logs:
```
16>C:\Users\AdamN\Documents\Projects\ppsspp\ext\miniupnpc.vcxproj(158,5): warning MSB4011: "C:\Users\AdamN\Documents\Projects\ppsspp\Windows\fix_2017.props" cannot be imported again. It was already imported at "C:\Users\AdamN\Documents\Projects\ppsspp\ext\miniupnpc.vcxproj (102,5)". This is most likely a build authoring error. This subsequent import will be ignored.
```

hrydgard:
Extracted and merged some more commits (#13179 ), and then extracted yet another one (#13181). Rebased that one myself on master, should be fine to later rebase this on master after merging it.

hrydgard:
You need to delete the commit "Added uPNP support", it now only contains a downgrade of the miniupnp module :)

ANR2ME:
> 
> 
> You need to delete the commit "Added uPNP support", it now only contains a downgrade of the miniupnp submodule :)

How do i delete an old commit?

hrydgard:
git rebase -i [some commit before the commit you need to delete]

Then in the editor that comes up, just delete the line, save and exit.

hrydgard:
Doesn't look like I'm gonna do a lot more reordering, so you don't really need to rebase after every single one of my merges - won't be much conflicts. Looking at 2 or 3 more merges.

ANR2ME:
I'll removes some of these commits that are no longer needed

Lemoncak3:
Hello may i ask about dynasty warrior strikeforce? The adhoc works fine and we can enter online city together but whenever we send item trade, friend request, mission request other players got nothing so we only can wander in the city.

ANR2ME:
> 
> 
> Hello may i ask about dynasty warrior strikeforce? The adhoc works fine and we can enter online city together but whenever we send item trade, friend request, mission request other players got nothing so we only can wander in the city.

Is this a regression that was working before? or it's been like that for a long time?

Lemoncak3:
I habe tried this almost a year ago with same issue though. Because i really want to play with my brpther and sis. Today i tried it again with the latest build of ppsspp.

ANR2ME:
> 
> 
> I habe tried this almost a year ago with same issue though. Because i really want to play with my brpther and sis. Today i tried it again with the latest build of ppsspp.

Well if it's not working properly since last year you should create a new issue/bug report (if it's not already existed yet) and wait until someone could investigate it.

ANR2ME:
Hmm.. when i tried to run Patapon 3 using Debug build i'm getting this crash:
image
Release build doesn't have this issue.

hrydgard:
That would seem to be a bug in the D3D11 backend that should be reported separately - that can't really have anything directly to do with this PR.

hrydgard:
It is done!

This can be closed now. I hope we didn't break too many things - but at least now we can narrow down the range a lot easier by directing people to buildbot builds.

ANR2ME:
Awesome! Thanks for the hard work in splitting this up :D

hrydgard:
Oh by the way I think you're only seeing it in a debug build because D3D11 validation is only active in debug.

hrydgard:
@ANR2ME fix submitted to master.

Common/Log.h
ANR2ME: btw @hrydgard / @unknownbrackets is there anyway we can see Verbose logs in PPSSPP Debug Console Log without changing the source code like this? 

Because changing the Logging Channel of SCENET to "Verb" from PPSSPP Settings doesn't seems to be working :( (was it a bug?)
And changing the source code + rebuild whenever we want to see verbose log is kinda troublesome isn't?

PR #13134, opened 2020-07-12
(Libretro) Add custom GLEW - interfaces with libretro get_proc_address -

inactive123:
We need this for compatibility with DRM/KMS and Wayland contexts - otherwise we're limited to GLX context.

Libretro cores typically implement their own extension wrangler for GL called glsym, making use of the get_libretro_proc_address  callback through which we can grab symbols.

To make sure we stay in lockstep with upstream instead of going our own way with the extension wrangling, what I did here was take glew.c as used in PPSSPP, make a copy of it and put it in the libretro dir, rip out the platform-specific code such as WGL/GLX/etc., and simply rely on the get_libretro_proc address. I also had to make some other changes to glew.c to get this to work. I did not edit any other files beyond this. 

The alternative would have been to make deep edits to glew.c and I'm doubtful anybody in PPSSP upstream would have been a fan of that. This way, the maintenance of this file falls squarely in our lap, and by my estimation this file shouldn't really need to be updated much beyond this point. It should not be a maintenance burden by any means.

Hopefully you guys can merge this. Right now, we're unable to use the GL renderer with DRM/KMS ( a popular target for RetroArch on Linux), and pretty much requires X11 for it to run.


hrydgard:
Hm, I'll look closer tomorrow (github won't even show the new file due to number of lines!) but I think this is probably the right way to go indeed.

twinaphex:
Thank you.

i30817:
Is this supposed to be only for vulkan? I tried to build it here (libretro build, release, retroarch with only wayland aand no x11 built) and while the crash appears to have changed, the renderer chosen is still the null one.

https://gist.github.com/i30817/65f6f12d33b5d9fb9a6d1784674b2739

twinaphex:
It fixes DRM/KMS right now, Wayland still has to be researched as to what might be going wrong. For whatever reason its not creating any FBOs there and just exits.

i30817:
Maybe i did something wrong, it crashes in KMS here too in the same way (by logout of gnome and login into a tty). Oh well, unless you want to ask something specific, i'll let you continue debugging this and use the RA x11 version with the wayland support stripped that forces Xwayland for ppsspp.

PR #13137, opened 2020-07-13
(Libretro) Add optimal input read codepath - bitmasks

inactive123:
This is an optimal codepath for libretro that allows us to grab the entire button state (sans analogs) in one function call instead of having to query a function callback for each individual button. It also has a fallback path for libretro frontends that don't implement this environment  callback.

### DIFF
--- a/libretro/Makefile
+++ b/libretro/Makefile
@@ -376,6 +376,113 @@ else ifneq (,$(findstring windows_msvc2017,$(platform)))
 	GL_LIB := opengl32.lib
 	LDFLAGS += ws2_32.lib user32.lib shell32.lib avcodec.lib avutil.lib swresample.lib swscale.lib avformat.lib advapi32.lib winmm.lib gdi32.lib d3d9.lib d3dx9.lib
 
+# Windows MSVC 2019 all architectures
+else ifneq (,$(findstring windows_msvc2019,$(platform)))
+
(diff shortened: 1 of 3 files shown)


hrydgard:
I don't mind this at all, but if you check the diff here on github, you see that there's something quite wrong with the tabs and spaces here. Fix it up and I'll merge :)

twinaphex:
We follow different code formatting conventions for libretro - 3 spaces for tabs et all. I have this setup in Vim and all.

Would you mind if I just change everything in libretro.cpp to follow this same convention then?

hrydgard:
I'm ok with a separate convention as long as it stays in the retro folder. But I'd suggest maybe converting on a function by function basis as you touch them, to not make huge diffs. Every other line in a different convention makes the diffs too hard to read.

twinaphex:
Fair point. I'll follow the style conventions of PPSSPP in all files outside of the libretro/ dir.

twinaphex:
I'll change the other files in libretro/ on an as-per needed basis, i.e. when I need to make edits to them.

Hopefully this is sufficient for now.

hrydgard:
I'm ok with that.

Though honestly it's not that hard in vim to setup to have different conventions in different folders...

twinaphex:
We'll definitely do that for the files outside the libretro folder.

twinaphex:
So to confirm, this is ready to go or does something still need to be done on my end?

PR #13138, opened 2020-07-13
Vulkan: Make sure textures/samplers are unbound at the end of PresentationCommon::CopyToOutput

hrydgard:
Validation caught an issue (and I had a crash in a Debug build, too) where old stuff lingered in sampler 1 and texture 1.

Bug probably introduced in #12921, but could also be others.

This could maybe be done more elegantly in some other way, like unbinding stuff when we bind a new pipeline that uses less samplers, or something like that. But we don't have a clear view how many samplers a pipeline uses here.

Also started doing some similar cleanup in the other backends but likely not necessary.


hrydgard:
Gonna merge this prematurely since it's kinda critical and I'm doing stuff on top of it.

PR #13099, opened 2020-07-05
Link PR and issues in README

iota97:
This will make them clickable for now, not sure if is worth the trouble tho'.

PR #13153, opened 2020-07-15
Update README following unknows markup way

iota97:
Using https://github.com/hrydgard/ppsspp/pull/13099#issuecomment-658454819 way we can just add all the URL/tittle and the end of README.md, so it can be read again in raw form.

The python script automated the fetch of title and everything, kinda hacky tho' (but by the time README gonna need to be updated for v1.11 I'm not even sure what github gonna do anyway)

PS: this restore to no markup from my previous PR way
```
#!/bin/python

import re
f = open("README.md", "r+")
cont = f.read()
updated = re.sub(r"\[#(\d+)]\(https://github.com/hrydgard/ppsspp/issues/\d+\)", r"#\1", cont)
f.seek(0)
f.write(updated)
f.truncate()
f.close()
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,55 +27,55 @@ For game compatibility, see [community compatibility feedback](https://report.pp
 
 What's new in 1.10.3
 --------------------
-* Fix for control layout editor ([#13125](https://github.com/hrydgard/ppsspp/issues/13125))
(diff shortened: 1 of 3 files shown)


hrydgard:
That's pretty neat!

PR #13172, opened 2020-07-20
Generalized multi-instance

hrydgard:
Takes #13169 and extracts it to a separate file, and changes how it's called.

Instance counter is now initialized on startup, enabling us to do the following things for secondary instances:

* They don't save their configs
* They are silent (can be overridden at runtime if desired).
* The title bar shows the instance number (Windows only so far).

Which makes the experience a lot better.

Possible alternatives would be to have a separate configs: ppsspp2.ini, ppsspp3.ini and so on, but felt this was simpler.

### DIFF
--- a/Core/HLE/sceNet.cpp
+++ b/Core/HLE/sceNet.cpp
@@ -42,6 +42,7 @@
 #include "Core/HLE/sceNetAdhoc.h"
 #include "Core/HLE/sceNet.h"
 #include "Core/Reporting.h"
+#include "Core/Instance.h"
 
 static bool netInited;
 static bool netInetInited;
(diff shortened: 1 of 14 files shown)


LunaMoo:
Since PPSSPP will now oficially support multiply instances on same pc, it's worth noting that some people were requesting to disable(on windows) titlebar and borders(as they play like that with a single display).

 I made options for that long ago:
https://github.com/LunaMoo/ppsspp/commit/3c49485f7c2922ab8c514c2802df45b22de761f0
https://github.com/LunaMoo/ppsspp/commit/9b0b23934a958b08e9f46ef0e66f74934a19b0a1
https://github.com/LunaMoo/ppsspp/commit/06e90a74a03813f41bf94a080b879afabf1e7028
 could open a PR with those options, through not sure which menu should it go, it might also need some more "automatic" function for placing windows(since they can't be moved around with titlebar disabled) considering the new way doesn't save configs. IE it could just place each instance window in different corner up to 4 or something.

hrydgard:
I can see how those options would be useful. I think we could make things even more convenient with some sort of auto-tile that just does all that in fullscreen mode automatically if there are multiple instances around - then you'd need no options, you'd just go into fullscreen and it would automatically tile the two instances borderlessly. And the windows would be restored coming out of full screen. Would just need a little window message trickery ...

Core/HLE/sceNet.cpp
```diff
@@ -232,10 +115,6 @@ static void __ResetInitNetLib() {
 
 void __NetInit() {
 	portOffset = g_Config.iPortOffset;
-	if (PPSSPP_ID == 0) // For persistent ID/IP, New instances may have a possibility to have the same ID/IP if PPSSPPIDCleanup() being called on every reset
-	{
-		PPSSPP_ID = getPPSSPPInstanceNumber(); // This should be called when program started instead of when the game started/reseted
-	}
 
 	net::Init();
```
ANR2ME, line 119: I also need to remove this `net::Init()` from `__NetInit()` while rebasing because `net::Shutdown()` have been removed from `__NetShutdown()` so the number of WSAStartup and WSAShutdown will match.

hrydgard: Ah, okay.

LeoKids:
This might help me getting problems using Netplay on Naruto Ultimate Ninja Heroes.

PR #13173, opened 2020-07-20
Make sure to run callback on AdhocThread (extracted from #13132).

hrydgard:
Two more commits extracted from #13132 .

### DIFF
--- a/Core/HLE/proAdhoc.cpp
+++ b/Core/HLE/proAdhoc.cpp
@@ -935,11 +935,13 @@ void AfterMatchingMipsCall::run(MipsCall &call) {
 	}
 	context->eventlock->unlock();  //peerlock.unlock();
 	//call.setReturnValue(v0);
+	if (Memory::IsValidAddress(bufAddr)) userMemory.Free(bufAddr);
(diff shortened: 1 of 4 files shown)


Core/HLE/proAdhoc.cpp
```diff
@@ -949,41 +951,45 @@ void AfterMatchingMipsCall::SetContextID(u32 ContextID, u32 eventId) {
 void notifyAdhocctlHandlers(u32 flag, u32 error) {
 	__UpdateAdhocctlHandlers(flag, error);
 	// TODO: We should use after action instead of guessing the time like this
-	sleep_ms(20); // Ugly workaround to give time for the mips callback to fully executed, usually only need <16ms
+	//sleep_ms(20); // Ugly workaround to give time for the mips callback to fully executed, usually only need <16ms
```
hrydgard, line 954: @ANR2ME  Why was this commented out?

ANR2ME: because it was delayed to make sure previous callback is already fully executed, but how much time a callback required to be done is just a guess, so i decide to use After action instead t be exact, but later i decide to give the delay (hleDelayResult) within `__NetTriggerCallbacks` where it process a single callback once per HLE call instead of multiple callbacks.

PR #13174, opened 2020-07-20
Extract another few commits from #13132

hrydgard:
From #13132. This one is huge so I want it separate so we get a build. Ideally this should probably be split into a few individual changes, but...

"Improved multiplayer compatibility on some games (ie. GTA VCS, Naruto Ultimate Ninja Heroes 3, DBZ Shin Budokai 2, Power Stone Collection, .hack//Link, etc). "

### DIFF
--- a/Core/HLE/proAdhoc.cpp
+++ b/Core/HLE/proAdhoc.cpp
@@ -106,8 +106,11 @@ bool isPTPPortInUse(uint16_t port) {
 char* mac2str(SceNetEtherAddr* mac) {
 #if defined(_WIN32)
 	static __declspec(thread) char str[18] = ":::::";
-#else
+#elif !PPSSPP_PLATFORM(MAC) && !PPSSPP_PLATFORM(IOS)
 	static __thread char str[18] = ":::::";
+#else
+	// Temporary hack to avoid huge rebase conflicts. Remove this when applying the mac2str rewrite.
+	static char str[18] = ":::::";
 #endif
 	if (mac == NULL) return str;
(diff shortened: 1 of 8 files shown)


hrydgard:
Turns out this one doesn't build as-is, probably gonna have to pick a few more into this round after all...

hrydgard:
Yeah this has too many platform issues, I'm gonna have to grab a few more commits. Was hoping to avoid bringing in UPnP in this one so that can be merged separately, let's see...

hrydgard:
@ANR2ME  I picked off a few more commits, let's see if we can get it to build on all platforms now. A couple from after the UPnP addition so had to make a small tweak. But really, this contains everything from where we left off up to the UPnP addition, which will be the next merge.

Core/HLE/proAdhoc.cpp
```diff
@@ -82,32 +87,62 @@ int isLocalMAC(const SceNetEtherAddr * addr) {
...
 	// Iterate Sockets
-	int i = 0; for(; i < 255; i++) if(ptp[i] != NULL && ptp[i]->lport == port) return 1;
+	for(int i = 0; i < 255; i++) if(ptp[i] != NULL && ptp[i]->lport == port) return true;
 	
 	// Unused Port
-	return 0;
+	return false;
+}
+
+char* mac2str(SceNetEtherAddr* mac) {
```
ANR2ME, line 106: Ah i already replace this function with a new one, returning std::string instead of char* so it's no longer platform specific and thread-safe

PR #13181, opened 2020-07-22
Fix Blinking game room issue on NFS Most Wanted (extracted from #13132 by ANR2ME)

hrydgard:
From #13132 , this will also fix lobby issue on Ridge Racer 2, Naruto Kizuna Drive, etc.


hrydgard:
Cherry-picked a buildfix from later in the sequence.

PR #13182, opened 2020-07-22
Fixes for GvG Next Plus, Phantasy Star Portable 2 extracted from #13132

hrydgard:
More stuff from #13132. 

One step at a time :)


lightfirst:
Gundam Vs Gundam Next Plus adhoc working fine on windows, but when play whith android, disconnect appeared in the middle of the multiplayer battle. (same as my hack builds)

leoxxx:
@hrydgard 
Sometimes display network error when playing Gundam Vs Gundam Next Plus in v1.10.3-810-gef08c0237.
It's unstable. I just test that PC(cable) as server and android phone(WIFI) as client in a lan.
It can link more game. I hope it will be better(stable).

PR #13198, opened 2020-07-25
AdhocMatching fix 3rd Player issue

anr2me:
An attempt to fix an issue when there are more than 2 players joined, some other players are thinking that the new player is the host and behave abnormally on Patapon 3.

This should also affects other games that use AdhocMatching and were having similar issue (if any)


MojoJojoDojo:
Which other games use adhocmatching?

ANR2ME:
> 
> 
> Which other games use adhocmatching?

There are too many of them LOL 
If you see "sceNetAdhocMatching..." in the Log console, that means they are using it

Issue #13205, opened 2020-07-26
Shin Sangoku Musou 6 Special chinese patch cannot switch umd itself


sum2012:
The iso is combine disc 1 and disc 2

I think that we try to imp kernel module  instead directly modify sceumd
Please note that jpcsp emulator already done it.

18:22:888 hle\scekernelmodule.cpp:753 I[LOADER]: Function (sceMeCore_driver,67f0ed84) unresolved, storing for later resolving
18:22:889 hle\scekernelmodule.cpp:753 I[LOADER]: Function (sceMeCore_driver,c287ad90) unresolved, storing for later resolving
18:22:889 hle\scekernelmodule.cpp:753 I[LOADER]: Function (sceMeCore_driver,d857cf93) unresolved, storing for later resolving
18:22:889 hle\scekernelmodule.cpp:753 I[LOADER]: Function (sceMeCore_driver,8988ad49) unresolved, storing for later resolving
18:22:889 hle\scekernelmodule.cpp:753 I[LOADER]: Function (sceMeCore_driver,051c1601) unresolved, storing for later resolving
18:22:889 hle\scekernelmodule.cpp:753 I[LOADER]: Function (sceMeCore_driver,99e4dbfa) unresolved, storing for later resolving
18:22:889 hle\scekernelmodule.cpp:753 I[LOADER]: Function (InterruptManagerForKernel,f987b1f0) unresolved, storing for later resolving
18:22:889 hle\scekernelmodule.cpp:753 I[LOADER]: Function (InterruptManagerForKernel,8b61808b) unresolved, storing for later resolving
18:22:889 hle\scekernelmodule.cpp:753 I[LOADER]: Function (semaphore,4c537c72) unresolved, storing for later resolving
18:22:889 hle\scekernelmodule.cpp:753 I[LOADER]: Function (semaphore,77e97079) unresolved, storing for later resolving
18:22:889 hle\scekernelmodule.cpp:753 I[LOADER]: Function (SystemCtrlForKernel,159af5cc) unresolved, storing for later resolving
18:22:889 hle\scekernelmodule.cpp:753 I[LOADER]: Function (SystemCtrlForKernel,1c90becb) unresolved, storing for later resolving
18:22:889 hle\scekernelmodule.cpp:751 W[LOADER]: Unknown syscall in known module 'ThreadManForKernel': 0x28b6489c
[...]

PR #13206, opened 2020-07-26
Imp sceUtilsBufferCopyWithRange and sceUtilsBufferCopyByPollingWithRange

sum2012:
Used in Shin Sangoku Musou 6 Special chinese patch #13205
Not sure "type cast" is right

### DIFF
--- a/Core/HLE/sceKernelSemaphore.cpp
+++ b/Core/HLE/sceKernelSemaphore.cpp
@@ -461,16 +461,15 @@ int sceKernelPollSema(SceUID id, int wantedCount)
 static u32 hleUtilsBufferCopyWithRange(u32 outAddr, int outSize, u32 inAddr, int inSize, int cmd)
 {
 		int temp = sceUtilsBufferCopyWithRange((u8*)outAddr, outSize, (u8*)inAddr, inSize, cmd);
-		INFO_LOG_REPORT_ONCE(hleUtilsBufferCopyWithRange,SCEKERNEL, "sceUtilsBufferCopyWithRange(%08x,%i,%08x,%i,%x) should return %i",outAddr,outSize,inAddr,inSize,cmd,temp);
-		// Fake a successful operation
+		if (temp != 0) {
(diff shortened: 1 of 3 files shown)


Core/HLE/sceKernelSemaphore.cpp
```diff
@@ -457,3 +458,26 @@ int sceKernelPollSema(SceUID id, int wantedCount)
 	}
 }
 
+static u32 hleUtilsBufferCopyWithRange(u32 outAddr, int outSize, u32 inAddr, int inSize, int cmd)
+{
+		sceUtilsBufferCopyWithRange((u8*)outAddr, outSize, (u8*)inAddr, inSize, cmd);
+		// Fake a successful operation
+		return 0;
```
hrydgard, line 465: Please log something when it goes wrong even if it helps to fake a successful operation here.

(also, the indentation is off by one).

Core/HLE/sceKernelSemaphore.cpp
```diff
@@ -457,3 +458,29 @@ int sceKernelPollSema(SceUID id, int wantedCount)
 	}
 }
 
+static u32 hleUtilsBufferCopyWithRange(u32 outAddr, int outSize, u32 inAddr, int inSize, int cmd)
+{
+		int temp = sceUtilsBufferCopyWithRange((u8*)outAddr, outSize, (u8*)inAddr, inSize, cmd);
+		INFO_LOG_REPORT_ONCE(hleUtilsBufferCopyWithRange,SCEKERNEL, "sceUtilsBufferCopyWithRange(%08x,%i,%08x,%i,%x) should return %i",outAddr,outSize,inAddr,inSize,cmd,temp);
```
hrydgard, line 464: No sorry, I mean to log something special if there's an error, like
```
if (temp != 0) {
  ERROR_LOG(SCEKERNEL, "hleUtilsBufferCopyWithRange: Failed with %d", temp);
}
return 0;
```
I don't think we need to use REPORT for this, it's quite obscure.

Also the indentation is still one level too much, look at the diff here.

sum2012:
@hrydgard  done

PR #13207, opened 2020-07-27
Added more stub for some unimplemented sceNetApctl* functions

anr2me:
Scrapped from all over the internet, some from jpcsp, some from @LunaMoo gist https://gist.github.com/LunaMoo/ec340ddf0f6bc7c7277e7b9603061a8e (not sure if the one from LunaMoo really have that many function arguments or not tho)

We will probably need to separate APctl into a separate file such as sceNetAP/sceNetApctl .cpp/h in the future since it's getting bigger.

### DIFF
--- a/Core/HLE/sceNet.cpp
+++ b/Core/HLE/sceNet.cpp
@@ -48,6 +48,7 @@
 static bool netInited;
 bool netInetInited;
 bool netApctlInited;
+int netApctlStatus;
 u32 netDropRate = 0;
 u32 netDropDuration = 0;
 u32 netPoolAddr = 0;
@@ -57,6 +58,8 @@ u32 netThread2Addr = 0;
 static struct SceNetMallocStat netMallocStat;
 
 static std::map<int, ApctlHandler> apctlHandlers;
+
+static struct SceNetApctlInfoInternal netApctlInfo;
 
 static int InitLocalhostIP() {
 	// find local IP
@@ -152,6 +155,7 @@ static void __UpdateApctlHandlers(int oldState, int newState, int flag, int erro
 
 	for(std::map<int, ApctlHandler>::iterator it = apctlHandlers.begin(); it != apctlHandlers.end(); ++it) {
 		args[4] = it->second.argument;
+		// TODO: Need to make sure netApctlState is updated before calling the callback's mipscall so the game handler's subroutine can Get and make use the new state/info
 		hleEnqueueCall(it->second.entryPoint, 5, args);
 	}
 }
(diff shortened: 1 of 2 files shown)


Core/HLE/sceNet.h
```diff
@@ -85,10 +147,39 @@ typedef struct ProductStruct { // Similar to SceNetAdhocctlAdhocId ?
...
 
+typedef void(*sceNetApctlHandler)(int oldState, int newState, int event, int error, void* pArg);
+
+#define APCTL_PROFILENAME_MAXLEN 64
+#define APCTL_SSID_MAXLEN 32
+#define APCTL_IPADDR_MAXLEN 16
+#define APCTL_URL_MAXLEN 128
+typedef struct SceNetApctlInfoInternal { // Use struct instead of union for internal purpose
+	char 			name[APCTL_PROFILENAME_MAXLEN];
+	unsigned char 	bssid[ETHER_ADDR_LEN];
```
hrydgard, line 163: I think you can safely declare these as char instead of unsigned char, so you then can use string functions without casting.

Core/HLE/sceNet.cpp
```diff
@@ -449,22 +548,135 @@ int sceNetInetTerm() {
...
 
-static int sceNetApctlInit() {
+static int sceNetApctlInit(int stackSize, int initPriority) {
 	ERROR_LOG(SCENET, "UNIMPL sceNetApctlInit()");
 	if (netApctlInited)
 		return ERROR_NET_APCTL_ALREADY_INITIALIZED;
+
+	memset(&netApctlInfo, 0, sizeof(netApctlInfo));
+	// Set default values
+	strncpy((char*)&netApctlInfo.name, "Wifi", sizeof(netApctlInfo.name));
```
hrydgard, line 558: In Common/StringUtil.h we have truncate_cpy which does the right thing with zero-termination, unlike strncpy.

hrydgard:
Looks good. Now that your PRs are a bit smaller I might end up reviewing style slightly more though :)

ANR2ME:
I'm still adding some more functions/nid that existed in jpcsp but doesn't exist yet in ppsspp

Core/HLE/sceNet.cpp
ANR2ME: hmm.. i can't find Common/StringUtil.h , may be you meant "base/stringutil.h"?

Core/HLE/sceNet.h
ANR2ME: Ah okay i'll change fields that contains null-terminated string to char, except that bssid since it's 6-byte of mac address and not null terminated one i should use memcpy instead

ANR2ME:
This should be all i could find on the internet related to Apctl

PR #13209, opened 2020-07-27
Add some threadman's nids

sum2012:
used in #13205


hrydgard:
Weird that these would show up in ThreadManForUser, when they already also are in UtilsForUser.

But if it works, it works...

PR #13210, opened 2020-07-27
Fixes for texpack installation

### LINKED ISSUE #13199: Crash while install texture pack. (Android)
### What happens?
PPSSPP crash after press texture pack zip file. (Android only)

First break: v1.10.3-91-gfe91f62f9
Last work: v1.10.3-73-g2916e5b67

### What *should* happen?
Can install texture pack zip file via PPSSPP. :)

Huawei P9 Plus (Kirin 955, Mali T-880)

Logcat: [ppsspp-logcat.zip](https://github.com/hrydgard/ppsspp/files/4975050/ppsspp-logcat.zip)

hrydgard:
Fixes #13199 .

There are some deeper bugs with initialization error handling left, but this avoids any chance of getting into that when installing stuff from zip files at least.


Panderner:
This will break all games because it crash when i trying to boot the game.

Issue #13157, opened 2020-07-17
[Android][GLES2][Mali] Sreen flickering some games using latest git build of ppsspp


Gamemulatorer:
On the latest git build of ppsspp the screen is flickering in this game and I also notice it on other game(s) intro video

### screen recording
•buffered rendering
•frameskipping off
•2x rendering resolution
•vertex cache off

[Phineas and Ferb PSP.zip](https://github.com/hrydgard/ppsspp/files/4931192/Phineas.and.Ferb.PSP.zip)

### ge dump
[Phineas&Ferb.zip](https://github.com/hrydgard/ppsspp/files/4931194/Phineas.Ferb.zip)

### What *should* happen?
No screen flickering using ppsspp v1.10.3 stable

### What hardware, operating system, and PPSSPP version?  On desktop, GPU matters for graphical issues.
Android 5.1
Mali-450 GPU
PPSSPP v1.10.3-107-g6e6cac6b9

Gamemulatorer:
Issue started in https://github.com/hrydgard/ppsspp/commit/9ebb402e4fa9900a9b94ed64814c4386e50c0437 latst work no screen flickering build https://github.com/hrydgard/ppsspp/commit/bb6219e40202383f38f3439d7a3d78e2389024c9

Panderner:
No flickering at all running on Realme C2 on latest git build.
![Screenshot_2020-07-17-19-10-54-22_2f85358b2198d26f8aca533d68bee793](https://user-images.githubusercontent.com/37503397/87780437-4e581600-c861-11ea-9bcd-fc829e3bb14d.png)

iota97:
Mhhhh, between the two commit there is this addition:

https://github.com/hrydgard/ppsspp/blob/6e6cac6b9b975819032b2911194ea8038a77a67b/ext/native/thin3d/thin3d_gl.cpp#L622-L629

boundTextures_ is an array, but boundSamplers_ is a vector instead, may need to use clear() as code check its size for loops?

Gamemulatorer:
07-19 11:14:26.650 I/PPSSPP  (20381): Creating 480 x 272 FBO using no depth
07-19 11:14:26.695 I/PPSSPP  (20381): Creating 480 x 272 FBO using no depth
07-19 11:14:44.935 I/PPSSPP  (20381): Creating 960 x 544 FBO using DEPTH24_STENCIL8 texture
[...]

PR #13212, opened 2020-07-27
Remove unnecessary vector from thin3d_gl texture handling.

hrydgard:
Don't really expect this to fix anything but the code looked messy.


Gamemulatorer:
Maybe this can help #13157

hrydgard:
I doubt it, but you never know..

PR #13229, opened 2020-07-31
Add basic Microphone support on Windows.

shenweip:
Helps #12336,  #10970, tested **TalkMan**, **Audio Mechanica V8d**.
**TalkMan** works now.

In **Audio Mechanica V8d**, I can hear the playback of the recording, but I don't see the dots jumping on the screen like a real PSP, it may be a issue apart.
There are still a bunch of HLE functions about microphone not implemented, also need to support on other platforms.

### DIFF
--- a/Core/HLE/sceKernel.cpp
+++ b/Core/HLE/sceKernel.cpp
@@ -77,12 +77,13 @@
 #include "sceImpose.h"
 #include "sceUsb.h"
 #include "sceUsbGps.h"
+#include "sceUsbCam.h"
+#include "sceUsbMic.h"
 #include "scePspNpDrm_user.h"
 #include "sceVaudio.h"
 #include "sceHeap.h"
 #include "sceDmac.h"
 #include "sceMp4.h"
-#include "sceUsbCam.h"
 
 #include "../Util/PPGeDraw.h"
 
@@ -146,6 +147,7 @@ void __KernelInit()
 	__VideoPmpInit();
 	__UsbGpsInit();
 	__UsbCamInit();
+	__UsbMicInit();
 	
 	SaveState::Init();  // Must be after IO, as it may create a directory
 	Reporting::Init();
@@ -170,6 +172,7 @@ void __KernelShutdown()
 	kernelObjects.Clear();
 
 	__UsbCamShutdown();
+	__UsbMicShutdown();
 	__UsbGpsShutdown();
 
 	__AudioCodecShutdown();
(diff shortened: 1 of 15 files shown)


Core/HLE/sceKernel.cpp
```diff
@@ -269,6 +272,7 @@ void __KernelDoState(PointerWrap &p)
 		__UmdDoState(p);
 		__UtilityDoState(p);
 		__UsbDoState(p);
+		__UsbMicDoState(p);
```
Florin9doi, line 275: DoState should be added at the end

shenweip: Yes, I missed that note, thanks.

Core/HLE/sceUsbMic.h
```diff
@@ -16,4 +16,61 @@
...
+void __UsbMicDoState(PointerWrap &p);
+
+struct MicWaitInfo {
+	SceUID threadID;
+	u32 addr;
+	u32 needSize;
+	u32 sampleRate;
+};
+
+class QueueBuf {
```
hrydgard, line 36: We have a few similar queues around the code base - any particular reason none of them could be re-used? 

I do understand how it can be easiest to just make another one that does exactly what you need though, and I am fine with that, at least temporarily - just wondering if you looked around at the others and found something.

hrydgard:
I had always thought we'd end up implementing this using the recording capabilities of the audio APIs (DSound, WASAPI) but I guess using the capture device API makes sense too, especially when we're recording video+audio like in Invizimals.

Has it been tested with that game, by the way? (of course not a requirement, just curious)

Core/HLE/sceUsbMic.h
shenweip: I hope to have a queue buffer that can automatically increase its capacity, because the data is dynamic, which can ensure that the buffer capacity is sufficient and secure. However, I have not found any similar queue, or maybe I missed it?

shenweip:
I have not tested Invizimals, I forget this game also use microphone:p.

Core/HLE/sceUsbMic.h
hrydgard: Let's merge and maybe look at this later.

jordenlang32robloxwikia:
I'll merge this to my fork of ppsspp.

zackzick2:
Amazing bro. We will waiting improvement progress to playing invizimals with microphone perfectly ,👍

Issue #13214, opened 2020-07-27
[Update Suggestion]Make UI for Ad-Hoc more organized


EX-e-LD:
**(Sorry for first closed one,accidentally hit enter)**

Hi,me again.Well,UI of PPSSPP is very unique and awesome,also good organized,well,except Ad-Hoc.With new UPnP addition,Ad-Hoc menu got really messy,and it is really hard to find what you're looking for.My opinion is it should be revamped in for example sections,where one could find what he actually needs.This is just suggestion,not something that should be your priority but my idea in general,because you did same thing with the Graphics and System,and that is really neat,and can be lot better.

TIA

HiImBrala:
Not too bad idea,i have seen couple emulators and games and network settings now are really messed up.

hrydgard:
Did #13232 improve it enough?

EX-e-LD:
@hrydgard 
Thank you!!This looks really good,and it is more polished now,exactly way i was thinking.

hrydgard:
Well, thank @ANR2ME for the actual work. I'll close it for you.

PR #13232, opened 2020-08-01
Reorganized Networking UI

anr2me:
Reorganized Networking UI as requested at https://github.com/hrydgard/ppsspp/issues/13214


hrydgard:
Don't really have any strong opinions here - from the diff, looks fine to me. Let's go with it and polish further if needed.

PR #13233, opened 2020-08-01
Add some comments and const-correctness, and a pointer type fix, to the "semaphore" crypto stuff.

hrydgard:
@sum2012 A followup to your change #13206.

Renamed the function internally to avoid the name clash.

I think it's supposed to work like this, hope it helps.


sum2012 (approved):
Thanks,I will avoid the bug again.

Issue #13236, opened 2020-08-02
Savestate failed


Lemoncak3:
I download v1.10.3-276-g4766e4fcb from githyb but when i use load state it says load state failed.

hrydgard:
Getting this too. I'm on it

ANR2ME:
@Lemoncak3 v1.10.3-276-g4766e4fcb doesn't have the default value on DoState, have you try to load it using v1.10.3-282-g914b6ce24 ?

Lemoncak3:
Okay will try it, wait

Lemoncak3:
![Screenshot_20200802-182846](https://user-images.githubusercontent.com/67026697/89122012-0b976e80-d4ee-11ea-8241-32d53fd55c4b.png)
Still has the same issue

ANR2ME:
May i know what game is that? and what was the version the save state created with? i'll try to create a new save state with that version

Lemoncak3:
i think i use this one v1.10.3-260-g70802e61d before updating to v1.10.3-276-g4766e4fcb

ANR2ME:
Could you provide your digimon/growlanser save state? i couldn't reproduce it using new save state created with older version

btw does growlanser have multiplayer? i didn't see in it in main menu

hrydgard:
It's an issue in the sceUsbMic savestate handling, it turns out. PR coming.

hrydgard:
Please confirm that it's fixed in new builds.

Lemoncak3:
It worked now, thanks a lot

PR #13237, opened 2020-08-02
Fix error in sceUsbMic savestate handling (need to accept old savestates without the section)

hrydgard:
Hopefully helps #13236.


hrydgard:
Not gonna bother waiting for CI for this one, heh.

PR #13241, opened 2020-08-03
Added some missing OpenPSID nids and some stub functions

anr2me:
Based on: 
https://playstationdev.wiki/pspprxlibraries/5.00/kd/openpsid.xml
https://uofw.github.io/uofw/group__OpenPSID.html

### DIFF
--- a/Core/HLE/HLETables.cpp
+++ b/Core/HLE/HLETables.cpp
@@ -304,6 +304,7 @@ void RegisterAllModules() {
 	Register_sceUsbMic();
 	Register_sceOpenPSID_driver();
 	Register_semaphore();
+	Register_sceDdrdb();
 	// add new modules here.
 }
 
--- a/Core/HLE/sceKernel.cpp
+++ b/Core/HLE/sceKernel.cpp
@@ -85,6 +85,7 @@
 #include "sceHeap.h"
 #include "sceDmac.h"
 #include "sceMp4.h"
+#include "sceOpenPSID.h"
 
 #include "../Util/PPGeDraw.h"
 
@@ -150,6 +151,7 @@ void __KernelInit()
 	__UsbGpsInit();
 	__UsbCamInit();
 	__UsbMicInit();
+	__OpenPSIDInit();
 	
 	SaveState::Init();  // Must be after IO, as it may create a directory
 	Reporting::Init();
@@ -173,6 +175,7 @@ void __KernelShutdown()
 	hleCurrentThreadName = NULL;
 	kernelObjects.Clear();
 
+	__OpenPSIDShutdown();
 	__UsbCamShutdown();
 	__UsbMicShutdown();
 	__UsbGpsShutdown();
--- a/Core/HLE/sceOpenPSID.cpp
+++ b/Core/HLE/sceOpenPSID.cpp
@@ -18,44 +18,69 @@
 #include "Core/HLE/HLE.h"
 #include "Core/HLE/FunctionWrappers.h"
 #include "Core/HLE/sceOpenPSID.h"
-#include "Core/MemMap.h"
+#include "Core/MemMapHelpers.h"
 #include <Core/HLE/proAdhoc.h>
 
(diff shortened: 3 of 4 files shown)


hrydgard:
Did you find any games that use these?

I'm fine with adding it, just curious.

ANR2ME:
Nope, i'm not familiar with games that use these functions, probably only used for PSN/infrastructure mode or internal use only.

unknownbrackets:
There's a huge number of driver and other functions that games are literally not allowed to call, accessible only from kernel code.  I don't know that we should go down the path of creating all these - it's a big time investment for probably almost no gain.

At best it'd enable running the actual VSH, but most likely at the cost of performance and significant time investment.  Much of the kernel code requires different design for some basic things than our HLE uses.  We'd end up needing to rewrite a lot of things.

Whenever you see "_driver", you're most likely looking at a kernel only module.  Our current HLE code (including networking) is a small percentage of all functions that'd need to be implemented if you count all kernel functions.

-[Unknown]

hrydgard:
Right, yeah, _driver might be one step too far, though we do have to go a little outside what is normally allowed for games to support some homebrew and language patches, apparently.

ANR2ME:
Btw, talking about NID, i wonder... is there anyway to find out the nid from unknown syscall like this? or at least know which module/library it tried to use.
image
We probably won't be implementing these, i'm just being curious what it is :) because this kinda log is too general and not informative enough (in the case it was something needed). 
Or are you guys usually check the logs on real psp to find more info of unknown syscall?

hrydgard:
Those will happen when we missed looking up some NIDs earlier during start, so filling in NIDs with dummy functions will help with that.

HiImBrala:
> There's a huge number of driver and other functions that games are literally not allowed to call, accessible only from kernel code. I don't know that we should go down the path of creating all these - it's a big time investment for probably almost no gain.
> 
> At best it'd enable running the actual VSH, but most likely at the cost of performance and significant time investment. Much of the kernel code requires different design for some basic things than our HLE uses. We'd end up needing to rewrite a lot of things.
> 
> Whenever you see "_driver", you're most likely looking at a kernel only module. Our current HLE code (including networking) is a small percentage of all functions that'd need to be implemented if you count all kernel functions.
> 
> Basically: internally, PPSSPP is an "electric car" and doesn't process petrol or have internal combustion happening. But the road can't tell, because its wheels still spin. These functions you're importing here are "spark plugs" and "timing belts", "ignition coils", etc. The road (games) can't even tell if we're using them.
> 
> -[Unknown]

Woah,Unknown,never knew you were a car guy/petrolhead.I'm not only one anymore :(

Does this help with games that crash like Ratchet & Clank/Kao Challengers?

boy13epic:
@hrydgard i have found fat princess,socom,outrun 2006

hrydgard:
@boy13epic but those games don't use the _driver functions, right?

@ANR2ME  I'm happy to merge the rest of this PR if you get rid of the _driver ones for now (we'll consider them later if we find a need).

ANR2ME:
How about "sceOpenPSID_driver" ? it's already there in master but using a dummy function name instead of the correct name, all i did was changed the function name and add the duplicate function (`sceOpenPSIDGetOpenPSID`)

hrydgard:
Let's keep that one for now then.

boy13epic:
is this for psn connectivity

ANR2ME:
> 
> 
> is this for psn connectivity

It's used for identification to differentiate each PSP, and i think PSN using it to check for blacklisted (or whitelisted?) PSPs

HiImBrala:
Infrastructure multiplayer won't be thing on PPSSPP,safe to say forever,because it would need complete network code re-write,as well dedicated server that wouldn't run on @MrColdbird's code,but it would be rewritten as well.Developers said million times now Ad-Hoc is not priority,because there are games which don't boot at all,and for **fast and portable emulator** ,it wouldn't be goal,simple as that.

Pik-a-boo:
PSN emulation might be possible for games that use matchmaking only.
Rpcs3 already got a solution for that.

hrydgard:
Forgot about this one, merged.

PR #13251, opened 2020-08-05
Fit viewport to scissor

### LINKED ISSUE #4845: Dante's Inferno - Graphics issue - Bloom
The Game, Dante's inferno, plays fine but there's graphics issue:

The background seems to be mirrored in transparency, in the 1st screeshot, we can see the burning torches are shown, in the 2nd the "cross" on the wall is overlaid.

screenshot from 2013-12-15 20 52 04
screenshot from 2013-12-15 20 50 13

Here's video too!
http://www.youtube.com/watch?v=AhJspRA6Og4

hrydgard:
The PSP's GPU doesn't have any real triangle clipping to the sides - all it does is scissor off the pixels outside. Thus, the viewport only positions and scales the rendering, it doesn't restrict it. So if a game sets a viewport that would result in a PC GPU viewport rectangle that's smaller than the scissor rectangle, we need to expand the viewport to fill it up.

Fortunately we already had code to do the opposite - cut the viewport down in case it was outside the framebuffer and scale the projection matrix to compensate. This logic can easily do the reverse, too.

Dante's Inferno's bloom effect used a weird viewport when clearing the bloom subtract buffer, where the viewport didn't end up covering the whole thing. That resulted in #4845, which this PR fixes. Who knows what else it might fix...

Fixes #4845.

I have not yet written a hardware test to confirm this (though I plan to), but it makes sense and if it's not like this, the bug in Dante's Inferno makes zero sense.


hrydgard:
I mean, thinking about it some more, it pretty much has to be like this. The viewport you set up on the PSP is just a center and a scale factor, you don't set any boundaries, and it's unlikely the hardware would compute the boundaries just to add another scissor inside the already existing scissor rectangle feature.

http://lukasz.dk/mirror/forums.ps2dev.org/viewtopic7cdf.html?t=3644

PR #13258, opened 2020-08-07
Remove double disable function on frame duplication

iota97:
In df6f2160560f0fa00f55a52bcced243e30fcf972 it was added a second call to SetEnabledFunc rather than changing the previous one, code work fine, just a bit confusing.

### DIFF
--- a/UI/GameSettingsScreen.cpp
+++ b/UI/GameSettingsScreen.cpp
@@ -381,15 +381,12 @@ void GameSettingsScreen::CreateViews() {
 #endif
 
 	CheckBox *frameDuplication = graphicsSettings->Add(new CheckBox(&g_Config.bRenderDuplicateFrames, gr->T("Render duplicate frames to 60hz")));
-	frameDuplication->SetEnabledFunc([] {
-		return g_Config.iRenderingMode != FB_NON_BUFFERED_MODE || (g_Config.bSoftwareRendering && g_Config.iFrameSkip != 0);
-	});
 	frameDuplication->OnClick.Add([=](EventParams &e) {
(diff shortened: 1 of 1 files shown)


hrydgard:
Hm, actually, I'd say the old one was correct.. We can't do frame duplication in non-buffered mode.

iota97:
As #13068 happened in buffered rendering (reading his settings) it should be  `g_Config.iRenderingMode != FB_NON_BUFFERED_MODE && g_Config.iFrameSkip == 0;` then.

I don't really get the part of Software rendering in the old one tbh, is fine only with frameskipping enabled, kinda odd.

Gamemulatorer:
This could also help https://github.com/hrydgard/ppsspp/issues/13157#issuecomment-670344167?

hrydgard:
@Gamemulatorer no.

Gamemulatorer:
But this is accidentally fixed this issue I mentioned here lol
> Just ot @hrydgard that in the latest git build render duplicate to 60hz is not highlighted when set to skip-buffered rendering. #12724
> Screenshot_2020-08-07-14-03-36

unknownbrackets:
Unless, frameskip is enabled, software rendering ALWAYS renders at 60 FPS (well, NTSC.)  Essentially, "render duplicate frames" is always enabled when using the software renderer, unless frameskip is enabled.

It reads out from VRAM, so it's not doing any detection to "guess" the game FPS unless it needs to for frameskip (which will ignore CPU writes.)  That's why this feature isn't really relevant to the software renderer, normally.

-[Unknown]

PR #13268, opened 2020-08-10
Don't include sstream from stringutil.h or i18n.h

hrydgard:
Haven't done any measurements, but should help compile speed a little according to #13267.


hrydgard:
Both stringutil.h and i18n.h should be a lot faster with this, would be cool if you could confirm. I suppose I should also try to set that analyzer up...

PR #13278, opened 2020-08-12
An attempt to fake PSN authentication

anr2me:
I only tested this on Patapon 3, it somehow managed to by pass the "authenticating" screen, and the next missing NID i bumped into was `sceNpMatching2Init` which is the start of the actual multiplayer session (probably similar functionality to sceNetAdhocMatching that runs on it's own thread). 
But i won't be implementing this online matchmaking otherwise i'll ended creating my own private matchmaking server LOL we don't even have the basic TCP/UDP/HTTP[S] sceNetInet* functions implemented at all

### DIFF
--- a/Core/HLE/sceNet.cpp
+++ b/Core/HLE/sceNet.cpp
@@ -44,6 +44,7 @@
 #include "Core/HLE/proAdhoc.h"
 #include "Core/HLE/sceNetAdhoc.h"
 #include "Core/HLE/sceNet.h"
+#include "Core/HLE/sceNp.h"
 #include "Core/Reporting.h"
 #include "Core/Instance.h"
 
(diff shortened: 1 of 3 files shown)


boy13epic:
maybe it’s the last game that still supports online

boy13epic:
even in 2020

ANR2ME:
> even in 2020

i think they brought back the server online last year

boy13epic:
im willing to try this on patapon 3

hrydgard:
Merging this doesn't really fix anything then, it just gets a little further? Still, this implements and documents more functions so I think I want to merge this anyway?

ANR2ME:
Yeah, some of this implementation are not even available on JPCSP yet

hrydgard:
Seems OK to me. If you make it actually start connecting to the real thing, then we need to figure out how far we should go heh, though the PSP stuff is kinda.. old now.

ANR2ME:
True, even PS3 is starting to shutdown their servers and people are starting to create private servers for them. 
It's probably too late to create private servers for PSP since their servers already down thus can no longer capture the packets.
But games that only use PSN for matchmaking might still works if PSP have similarity to PS3 matchmaking API

PR #13282, opened 2020-08-13
Fix CMake build for Windows (VS and clang-cl)

Kingcom:
Some notes:

- the two .rc files were encoded in UTF-16, which Clang doesn't support. I've converted them to UTF-8 and added a code-page pragma. The about dialog looked fine with both compilers
- The SSE level detection defaulted to 4.2 on Windows. Clang advertises its support level and expects that to be respected, so I swapped the checks around a little
- the enum types are due to negative constants being used in switch cases with an unsigned value

lld doesn't work yet, as it doesn't support some of the pragmas used, but it Clang builds with link.exe now. At least aside from basically always doing a fresh build, but that's a different problem (I suspect this is due to -D_DEBUG being appended to  CMAKE_CXX_FLAGS_DEBUG on every CMake run).


unknownbrackets (approved):
I'm slightly scared of MSVC surprise breaking the rc files again, but we really already have problems when you use the design view to edit the rc files anyway...

We should update the editorconfig though:
https://github.com/hrydgard/ppsspp/blob/master/.editorconfig#L29-L30

-[Unknown]

hrydgard:
Updated .editorconfig for you in a followup commit.

PR #13283, opened 2020-08-13
Fix menu navigation from axis in linux

mrfixit2001:
Up is down and down is up - mimic the change in android for linux to compensate and correct. Tested and working.


hrydgard:
Yeah I guess let's just go ahead and do that.

PR #13288, opened 2020-08-15
Move timeutil to Common. (Experiment to see how much work it is to move these).

hrydgard:
I really would like to get rid of "native" eventually, but it seems to be a lot of work hehe. Gonna have to move groups of files though, this is impractical to do one by one. PR-ing this one anyway just because.


hrydgard:
Fails weirdly only on CI, don't feel like debugging it - closing for now, not worth the trouble at this time.

PR #13292, opened 2020-08-16
Fix undesired re-builds after runnning CMake

Kingcom:
CMake seems to cache values that PPSSPP appends to, so the same flags keep getting added at the end of the flags, leading to almost a full rebuild after every run of CMake. To work around this,  store the initial state of the flags in the cache and restore it before appending further options.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -240,6 +240,18 @@ endif()
 if(ARM64)
 	message("Generating for ARMv8, ${CMAKE_BUILD_TYPE}")
 endif()
+
+# It looks like the flags for the selected build type are written to the cache after each run, which causes some of the operations
+# below to keep expanding them with the same flags over and over on every run, leading to a rebuild of the majority of the files.
+# To work around this, remember the initial state of the variables from the first run and reset the variables to that.
+# TODO: Setting the attributes per target would probably be a better solution.
+foreach (LANGUAGE C CXX)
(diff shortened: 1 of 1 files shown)


hrydgard:
Seems like an okay workaround but feel free to make the proper fix too ;)

I'm not seeing this issue on Linux though, no unwanted rebuilds there.

Kingcom:
Hm, weird... maybe it's a bug in the version I'm using then. I'll try some newer/older versions. I've used 3.16.2 with the Ninja generator.

The proper fix would be nice, but would also require quite a major overhaul of the CMake project. There'd be a lot of clarity to be gained from it though, so it could be worthwhile.

Kingcom:
I've tested 3.18..1 and 3.10.2 as well, both show the same behavior. Interestingly all three versions only write the updated variable of the chosen build type in the cache (debug in this case):
```
//Flags used by the CXX compiler during DEBUG builds.
CMAKE_CXX_FLAGS_DEBUG:STRING=/MDd /Zi /Ob0 /Od /RTC1 -D_DEBUG -D_DEBUG -D_DEBUG -D_DEBUG -D_DEBUG -D_DEBUG -D_DEBUG

//Flags used by the CXX compiler during RELEASE builds.
CMAKE_CXX_FLAGS_RELEASE:STRING=/MD /O2 /Ob2 /DNDEBUG
```
...even though -D_NDEBUG is also appended to CMAKE_CXX_FLAGS_RELEASE in the MSVC case. Weird behavior...

hrydgard:
Wacky...

PR #13304, opened 2020-08-18
Add a check in sceKernelExitDeleteThread

sum2012:
Based on https://github.com/jpcsp/jpcsp/blob/master/src/jpcsp/HLE/modules/ThreadManForUser.java#L3949

I don't know that which game need it.


hrydgard:
Makes sense and seems consistent with the other one. Nice.

PR #13318, opened 2020-08-21
Simulate blocking behavior using non-blocking on networking

anr2me:
An attempt to improve multiplayer performance/reduce lags/stutters due to blocking behavior.
This should fixes games that get heavy performance degradation (ie. Power Stone Collection/2), and games that almost permanently frozen (ie. Hotshots Tennis, Digimon World Re:Digitize, etc).
Also bumped into possible crash related to AdhocMatching while testing this, thus fixing it along with this.

PS: This is my first time using ScheduleEvent so there are many questions and assumption :)

Here are the test builds:
Win32 & Win64: https://www.dropbox.com/s/kbwwx0q0w4ecm2d/PPSSPP_1.10.3-perftest_Win32x64.zip?dl=0
Android ARMv7 32bit: https://www.dropbox.com/s/3gl942pt5n91fer/PPSSPP_1.10.3-perftest_ARMv7.apk?dl=0

### DIFF
--- a/Core/Dialog/PSPNetconfDialog.cpp
+++ b/Core/Dialog/PSPNetconfDialog.cpp
@@ -18,17 +18,20 @@
 #if defined(_WIN32)
 #include "Common/CommonWindows.h"
 #endif
+#include <TimeUtil.h>
 #include "i18n/i18n.h"
 #include "Common/Serialize/Serializer.h"
 #include "Common/Serialize/SerializeFuncs.h"
 #include "Core/Config.h"
 #include "Core/MemMapHelpers.h"
 #include "Core/Util/PPGeDraw.h"
+#include "Core/HLE/sceKernelMemory.h"
 #include "Core/HLE/sceCtrl.h"
 #include "Core/HLE/sceUtility.h"
-#include <Core/HLE/sceNet.h>
+#include "Core/HLE/sceNet.h"
+#include "Core/HLE/sceNetAdhoc.h"
 #include "Core/Dialog/PSPNetconfDialog.h"
-#include <ext/native/util/text/utf8.h>
+#include "ext/native/util/text/utf8.h"
 
 
 #define NETCONF_CONNECT_APNET 0
@@ -44,6 +47,13 @@ static const float FONT_SCALE = 0.65f;
 const static int NET_INIT_DELAY_US = 300000;
 const static int NET_SHUTDOWN_DELAY_US = 26000;
 const static int NET_RUNNING_DELAY_US = 1000000; // KHBBS is showing adhoc dialog for about 3-4 seconds, but feels too long, so we're faking it to 1 sec instead to let players read the text
+const static int NET_CONNECT_TIMEOUT = 5000000;
+
+struct ScanInfos {
+	s32_le sz;
+	SceNetAdhocctlScanInfoEmu si;
+} PACK;
+
 
 PSPNetconfDialog::PSPNetconfDialog() {
 }
@@ -77,6 +87,11 @@ int PSPNetconfDialog::Init(u32 paramAddr) {
 		okButtonFlag = CTRL_CROSS;
 		cancelButtonFlag = CTRL_CIRCLE;
 	}
+
+	connResult = -1;
+	scanInfosAddr = 0;
+	scanStep = 0;
+	startTime = (u64)(real_time_now() * 1000000.0);
 
 	StartFade(true);
 	return 0;
@@ -114,7 +129,7 @@ void PSPNetconfDialog::DisplayMessage(std::string text1, std::string text2a, std
 
 	// Without the scrollbar, we have 350 total pixels.
 	float WRAP_WIDTH = 300.0f;
-	if (UTF8StringNonASCIICount(text1.c_str()) >= text1.size() / 4) {
+	if (UTF8StringNonASCIICount(text1.c_str()) >= (int)text1.size() / 4) {
 		WRAP_WIDTH = 336.0f;
 		if (text1.size() > 12) {
 			messageStyle.scale = 0.6f;
@@ -182,12 +197,20 @@ void PSPNetconfDialog::DisplayMessage(std::string text1, std::string text2a, std
 
 	PPGeScissor(0, (int)(centerY - h2 - 2), 480, (int)(centerY + h2 + 2));
 	PPGeDrawTextWrapped(text1.c_str(), 240.0f, centerY - h2 - scrollPos_, WRAP_WIDTH, 0, messageStyle);
-	if (text2a != "")
-		PPGeDrawTextWrapped(text2a.c_str(), 240.0f - 5.0f, centerY - h2 - scrollPos_ + totalHeight1 + marginTop, WRAP_WIDTH, 0, messageStyleRight);
+	if (text2a != "") {
+		if (text2b != "")
+			PPGeDrawTextWrapped(text2a.c_str(), 240.0f - 5.0f, centerY - h2 - scrollPos_ + totalHeight1 + marginTop, WRAP_WIDTH, 0, messageStyleRight);
+		else
+			PPGeDrawTextWrapped(text2a.c_str(), 240.0f, centerY - h2 - scrollPos_ + totalHeight1 + marginTop, WRAP_WIDTH, 0, messageStyle);
+	}
 	if (text2b != "")
 		PPGeDrawTextWrapped(text2b.c_str(), 240.0f + 5.0f, centerY - h2 - scrollPos_ + totalHeight1 + marginTop, WRAP_WIDTH, 0, messageStyleLeft);
-	if (text3a != "")
-		PPGeDrawTextWrapped(text3a.c_str(), 240.0f - 5.0f, centerY - h2 - scrollPos_ + totalHeight1 + totalHeight2 + marginTop, WRAP_WIDTH, 0, messageStyleRight);
+	if (text3a != "") {
+		if (text3b != "")
+			PPGeDrawTextWrapped(text3a.c_str(), 240.0f - 5.0f, centerY - h2 - scrollPos_ + totalHeight1 + totalHeight2 + marginTop, WRAP_WIDTH, 0, messageStyleRight);
+		else
+			PPGeDrawTextWrapped(text3a.c_str(), 240.0f, centerY - h2 - scrollPos_ + totalHeight1 + totalHeight2 + marginTop, WRAP_WIDTH, 0, messageStyle);
+	}
 	if (text3b != "")
 		PPGeDrawTextWrapped(text3b.c_str(), 240.0f + 5.0f, centerY - h2 - scrollPos_ + totalHeight1 + totalHeight2 + marginTop, WRAP_WIDTH, 0, messageStyleLeft);
 	PPGeScissorReset();
@@ -229,6 +252,7 @@ int PSPNetconfDialog::Update(int animSpeed) {
 	UpdateButtons();
 	auto di = GetI18NCategory("Dialog");
 	auto err = GetI18NCategory("Error");
+	u64 now = (u64)(real_time_now() * 1000000.0);
 	
 	// It seems JPCSP doesn't check for NETCONF_STATUS_APNET
 	if (request.netAction == NETCONF_CONNECT_APNET || request.netAction == NETCONF_STATUS_APNET || request.netAction == NETCONF_CONNECT_APNET_LAST) {
@@ -310,6 +334,7 @@ int PSPNetconfDialog::Update(int animSpeed) {
 	}
 	else if (request.netAction == NETCONF_CONNECT_ADHOC || request.netAction == NETCONF_CREATE_ADHOC || request.netAction == NETCONF_JOIN_ADHOC) {
 		int state = NetAdhocctl_GetState();
+		bool timedout = (state == ADHOCCTL_STATE_DISCONNECTED && now - startTime > NET_CONNECT_TIMEOUT);
 
 		UpdateFade(animSpeed);
 		StartDraw();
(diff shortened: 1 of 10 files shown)


MojoJojoDojo:
Warriors Orochi 2 : in-game when playing on adhoc would be too slow with multiple instances of PPSSPP.
Didn't try over the internet.
Timeout option wont affect FPS.

Core/Dialog/PSPNetconfDialog.cpp
```diff
@@ -374,11 +456,30 @@ int PSPNetconfDialog::Shutdown(bool force) {
...
+		Do(p, connResult);
+	}
+	else {
+		scanInfosAddr = 0;
+		scanStep = 0;
+		connResult = -1;
+	}
+
+	if (p.mode == p.MODE_READ) {
+		// Reallocate if it wasn't the correct memory block
```
unknownbrackets, line 476: You definitely shouldn't need to do this, unless `scanInfosAddr` is 0.  The memory blocks are all loaded as well, although you shouldn't depend on the sequence of loading.

For example, `PSPNetconfDialog::DoState` may actually be called before `userMemory.DoState` (I think this is not the case, but I don't think the code should depend on the order.)  You really should not interact with `userMemory` within this function at all.

Once all DoState functions finish, you can depend on `userMemory` looking exactly like it did when the user saved state.  So the address will still be valid if it was valid before.

When a save state is "upgraded" then you need a fallback, but that's it.  And in that case, `scanStep` will also be 0 so we should be safe.

-[Unknown]

Core/HLE/proAdhoc.cpp
```diff
@@ -200,6 +203,15 @@ SceNetAdhocctlPeerInfo * findFriend(SceNetEtherAddr * MAC) {
 	return peer;
 }
 
+int getNonBlockingFlag(int fd) {
+#ifdef _WIN32
```
unknownbrackets, line 207: Hm, you could use `ioctlsocket`.  What is this used for?

-[Unknown]

Core/HLE/proAdhocServer.cpp
```diff
@@ -2075,7 +2075,7 @@ int server_loop(int server)
 		sleep_ms(1);
```
unknownbrackets, line 2075: This should really be a `select()` and there should be no sleeps.

In case it's helpful, see this code that does a bunch of non-blocking socket operations:
https://github.com/hrydgard/ppsspp/blob/master/ext/native/net/websocket_server.cpp

-[Unknown]

Core/HLE/sceNetAdhoc.cpp
```diff
@@ -113,8 +120,457 @@ void __NetAdhocShutdown() {
...
+static void __AdhocctlNotify(u64 userdata, int cyclesLate) {
+	SceUID threadID = userdata >> 32;
+	int uid = (int)(userdata & 0xFFFFFFFF);
+
+	s64 result = 0;
+	u32 error = 0;
+
+	SceUID waitID = __KernelGetWaitID(threadID, WAITTYPE_NET, error);
+	if (waitID == 0 || error != 0)
+		return; // FIXME: Is it safe to exit here like this without re-scheduling the event? Will this event be triggered again? What will happen to the result i might want to change if exited here?
```
unknownbrackets, line 132: Well, it won't automatically.  But if the thread is not waiting no that, something has gone wrong and you probably don't want it to.

Most common cause would be the game calling `sceKernelReleaseWaitThread()`.

-[Unknown]

Core/HLE/sceNetAdhoc.cpp
```diff
@@ -113,8 +120,457 @@ void __NetAdhocShutdown() {
...
+
+	int waitVal = __KernelGetWaitValue(threadID, error); // FIXME: Is this value going to be a valid value if waitID == 0? or it's a value belonged to other event?
```
unknownbrackets, line 134: If you got an invalid wait ID, it could be this thread could even be waiting on something totally different now.  So no, the waitVal is probably not interesting in that case.

-[Unknown]

Core/HLE/sceNetAdhoc.cpp
```diff
@@ -113,8 +120,457 @@ void __NetAdhocShutdown() {
...
+				return;
+			}
+			else
+				result = 0; // ERROR_NET_ADHOCCTL_BUSY
+		}
+		else
+			result = 0; // Faking successfully connected to adhoc server
+	}
+
+	//HLEKernel::ResumeFromWait(threadID, WAITTYPE_NET, uid, result); // FIXME: This won't do anything if waitID == 0, not sure what kind of value returned from the HLE which i might want to change here.
```
unknownbrackets, line 151: Hm, it seems like it'd be better to put the desired adhocctlState in waitID, and keep the waitVal for the return value.  Might be tricky for save states.

You could also keep an unordered_map of thread id -> return value, and store the value there when scheduling the wait.

-[Unknown]

unknownbrackets (commented):
For any FPS drop, is the game FPS down or is PPSSPP having trouble?

The way to tell is VPS, if we're still flipping as often, then yeah - we're delaying something too much, or maybe even not enough (allowing the game to get stuck in a loop that doesn't reschedule and cause another PSP thread to run.)

-[Unknown]

Core/HLE/proAdhocServer.cpp
ANR2ME: it was the old code from aemu project, i didn't touch the logic much

actually, `select` with timeout will also sleeping under the hood, since it's blocking the thread :)

Core/HLE/sceNetAdhoc.cpp
ANR2ME: During my test it was woken-up by hleDelayResult on the AdhocMatching loop hle, so i'm replacing it with sceKernelDelayThread and didn't get woken-up randomly anymore.

What i want to know it, in the case the waiting sceNet function got woken up by something else, what kind of return value will it get? was it the normal return value from that function will be used?

For example:
```
int sceNetSomething(){
...
	CoreTiming::ScheduleEvent(usToCycles(100), adhocSocketNotifyEvent, param);
	__KernelWaitCurThread(WAITTYPE_NET, socketId, 0, 0, false, reason);

	// Fallback return value?
	return ERROR_NET_ADHOC_TIMEOUT;
}
```
i assume the ERROR_NET_ADHOC_TIMEOUT will be used as return value in the case the thread got woken up unexpectedly, is that a correct assumption?

unknownbrackets: The "normal" return value is discarded.  Nothing will use that.

`SCE_KERNEL_ERROR_RELEASE_WAIT` is the return value if `sceKernelReleaseWaitThread` is used.  If you use `hleDelayResult` on a thread that was already waiting on something else, its new result will be whatever you specified in the `hleDelayResult()` call.

`__KernelGetWaitID()` actually returns -1 on error, so technically 0 just means you specified 0 as the wait id in `__KernelWaitCurThread()` or etc.

PSP threads won't be woken unexpectedly.  There's no such thing as spurious wakeup in our HLE thread management.

-[Unknown]

Core/HLE/proAdhocServer.cpp
unknownbrackets: Right, but performance should be better with `select()` since it'll act on new network data much quicker.

-[Unknown]

Core/HLE/sceNetAdhoc.cpp
ANR2ME: Well i'm trying to use something unique as waitID, since windows socket is kinda unique and there is only one socket that communicate with adhoc server so i'm using it, it's also easier to use compared to create yet another global vector/map :D

Core/HLE/proAdhoc.cpp
ANR2ME: We are currently switching the socket blocking & non-blocking.

From what i read at stackoverflow, on Windows there is no way to get the blocking state of a socket, so this getNonBlockingFlag is kinda broken on windows, and i guess it won't be needed anymore since this PR is simulating the blocking process so we can assume it will always be in non-blocking mode.

Core/HLE/sceNetAdhoc.cpp
```diff
@@ -113,8 +120,457 @@ void __NetAdhocShutdown() {
...
+		target.sin_addr.s_addr = peer->ip;
+		target.sin_port = htons(peer->port + ((isOriPort && !isPrivateIP(peer->ip)) ? 0 : portOffset));
+
+		int ret = sendto(pdpsocket->id, (const char*)req.buffer, targetPeers.length, 0, (sockaddr*)&target, sizeof(target));
+		int sockerr = errno;
+
+		if (ret >= 0) {
+			DEBUG_LOG(SCENET, "sceNetAdhocPdpSend[%i:%u](B): Sent %u bytes to %s:%u\n", uid, getLocalPort(pdpsocket->id), ret, inet_ntoa(target.sin_addr), ntohs(target.sin_port));
+			// Remove successfully sent to peer to prevent sending the same data again during a retry
+			peer = targetPeers.peers.erase(peer);
```
ANR2ME, line 247: I'm a bit worried with this `erase`, i wanted to erase the content of the global map, but not sure whether this code will affect globally or this `peers` is only a copy, i only want to retry sending on failed peers and not resending to the one already successful on the next rescheduled event

unknownbrackets: `targetPeers` is a reference, so any changes to `targetPeers.peers` will affect that reference.

`foo.erase()` mutates foo, you'd need to do `std::deque<AdhocSendTarget>(foo).erase` or something to make it use a copy, and even then `peer` would not be a valid iterator to the copy, I assume.

-[Unknown]

ANR2ME:
> 
> 
> Warriors Orochi 2 : in-game when playing on adhoc would be too slow with multiple instances of PPSSPP.
> Didn't try over the internet.
> Timeout option wont affect FPS.
> This PR fixes the game though.

Btw is there any regression on games that was working fine before?

MojoJojoDojo:
> > Warriors Orochi 2 : in-game when playing on adhoc would be too slow with multiple instances of PPSSPP.
> > Didn't try over the internet.
> > Timeout option wont affect FPS.
> > This PR fixes the game though.
> 
> Btw is there any regression on games that was working fine before?
> 
> PS: If you are using an older version of test build, you might want to redownload the test build since i updated it yesterday to implement blocking simulation on PdpSend

Yes DBZ Tag Team would sometimes disconnect in the middle or the beginning of the battle.
I dont remember disconnections before this PR.

cahvireyu:
I can't see other players on Power Stone Collection using the test build but I can using automated builds. I'm using Zerotier on 2 Android devices. Anyway thank for your work.

Core/Dialog/PSPNetconfDialog.cpp
ANR2ME: Btw, does kernelMemory blocks also saved just like userMemory? i remembered long time ago i use kernelMemory for the loop code instead of userMemory, may be i should change that to userMemory too

ANR2ME:
> 
> 
> > > Warriors Orochi 2 : in-game when playing on adhoc would be too slow with multiple instances of PPSSPP.
> > > Didn't try over the internet.
> > > Timeout option wont affect FPS.
> > > This PR fixes the game though.
> > 
> > 
> > Btw is there any regression on games that was working fine before?
> > PS: If you are using an older version of test build, you might want to redownload the test build since i updated it yesterday to implement blocking simulation on PdpSend
> 
> Yes DBZ Tag Team would sometimes disconnect in the middle or the beginning of the battle.
> I dont remember disconnections before this PR.
> Ratchet and Clank Size Matters didnt exactly worked in-game but now it regressed even further because I cant go further in the menus when I try to join.
> Tried both with multiple instances and timeout = 1.

I just realized there is a bug with communication to the adhoc server while watching the logs on KHBBS, sometimes the Ping is getting socket error 10053 and got disconnected from adhoc server, so may be the disconnection issue was related to that bug.

Well i've fixed it on my side, will push it later after i'm done removing memory allocation from savesate, probably on weekend since i need to prepare for a job interview first.

ANR2ME:
Okay, this should fix:
1). An issue with PSPNetconfDialog where the joining player could stuck with repeated GetScanInfo if the Host created the group too late (ie. after the scan already completed)
2). An issue with loosing connectivity to the Adhoc Server in the middle of multiplayer (ie. when you're seeing "Network Initialized" in the middle of multiplayer it got disconnected due to this bug and tries to reconnect)

PS: How do i test multiplayer with Warrior Orrochi 2 ? the 2nd player keeps getting "channel already used"

PPS: Ratchet and Clank Size Matters sometimes crashed with WriteUncheckedU32 when both players are ready and the multiplayer game started, not sure if this is related to adhoc or not because i don't see anything strange with the sceNet logs, but there are a few untested module/disp hle before the crash. 
During the crash `currentMIPS->pc` shown on VS2019 watch pointing at 83B5FEB1 which doesn't have any content in the memory view of PPSSPP.
```
48:32:374 level load & E[SCEMODULE]: HLE\sceKernelModule.cpp:2496 UNTESTED sceKernelGetModuleIdList(09fcd540, 1024, 09fcd940)
48:32:446 rcp level ma E[SCEMODULE]: HLE\sceKernelModule.cpp:2496 UNTESTED sceKernelGetModuleIdList(09fcd0f0, 1024, 09fcd4f0)
48:34:625 rcp level ma E[SCEDISP]: HLE\sceDisplay.cpp:1006 80000104=sceDisplaySetFrameBuf(04000000, 0, 0, 1): invalid stride
```

Sometimes it crashed even before using any networking hle (during the rocket flying to a planet animation when choosing Multiplayer Game from start menu)
image
currentMIPS->pc pointing at FFFFFFFF during this crash

zakilj3:
Is the updated build already available for some quicktest?

ANR2ME:
> 
> 
> Is the updated build already available for some quicktest?

not yet, i'm building it right now, i'll post again after the test build got uploaded

MojoJojoDojo:
> Okay, this should fix:
> 1). An issue with PSPNetconfDialog where the joining player could stuck with repeated GetScanInfo if the Host created the group too late (ie. after the scan already completed)
> 2). An issue with loosing connectivity to the Adhoc Server in the middle of multiplayer (ie. when you're seeing "Network Initialized" in the middle of multiplayer it got disconnected due to this bug and tries to reconnect)
> 
> PS: How do i test multiplayer with Warrior Orrochi 2 ? the 2nd player keeps getting "channel already used"
> 
> PPS: Ratchet and Clank Size Matters sometimes crashed with WriteUncheckedU32 when both players are ready and the multiplayer game started, not sure if this is related to adhoc or not because i don't see anything strange with the sceNet logs, but there are a few untested/error module/disp hle before the crash.
> During the crash `currentMIPS->pc` shown on VS2019 watch pointing at 83B5FEB1 which doesn't have any content in the memory view of PPSSPP.
> 
> ```
> 48:32:374 level load & E[SCEMODULE]: HLE\sceKernelModule.cpp:2496 UNTESTED sceKernelGetModuleIdList(09fcd540, 1024, 09fcd940)
> 48:32:446 rcp level ma E[SCEMODULE]: HLE\sceKernelModule.cpp:2496 UNTESTED sceKernelGetModuleIdList(09fcd0f0, 1024, 09fcd4f0)
> 48:34:625 rcp level ma E[SCEDISP]: HLE\sceDisplay.cpp:1006 80000104=sceDisplaySetFrameBuf(04000000, 0, 0, 1): invalid stride
> ```
> 
> Sometimes it crashed even before using any networking hle (during the rocket flying to a planet animation when choosing Multiplayer Game from start menu), may be it have something to do with the animation?
> image
> currentMIPS->pc pointing at FFFFFFFF during this crash
Actually Ratchet & Clank Size Matters is another game like KHBBS that only works with JPCSP.
But yeah it never worked on PPSSPP for some reason and always crashed.
Good job on fixing the other issues though.

ANR2ME:
> 
> Actually Ratchet & Clank Size Matters is another game like KHBBS that only works with JPCSP.
> But yeah it never worked on PPSSPP for some reason and always crashed.
> Good job on fixing the other issues though.

Ratchet & Clank Size Matters worked when i tested it on localhost, it's just that sometimes it crashed, but i haven't tested it with D3D11, vulkan often cause a crash on my end :( like when i load savestate on Bleach heat the soul 7 it crashed when i use vulkan but didn't crash with D3D11 (a few days ago that is, not sure about now tho, may be there are more commits since then)

Regarding Kingdom Hearts BBS, even with JPCSP multiplayer only works when using decrypted prx files, just by removing the utility.prx to use the built-in sceUtility HLE KHBBS will no longer works, so i suspects it have something to do with NetconfDialog but can't see anything different other than it uses a lot of undocumented sceNetAdhocAuth* functions (especially the one that looked like have a priority & stacksize in the argument making me curious what kind of thread did it creates and used for)

MojoJojoDojo:
Tried Ratchet & Clank again but the Client is still stuck at connecting and "Network Initialized" is appearing over and over again.
Wanted to know if i can reach ingame lol.

zakilj3:
I tried ratchet and clank size matter and I came to 2 conclusion :

1)its the pal(europe) version that works for multiplayer while the USA version doesn't

2) it appears that it works on pc but not on Android as it crash on Android, on pc I had some kind of fixed screen while the rest of the games works fine(I don't know if its related to my settings for the game or smth else)

Saramagrean:
> PS: How do i test multiplayer with Warrior Orochi 2 ? The 2nd player keeps getting "channel already used" message. 

Change WLAN channel to 1 or 6. :)

ANR2ME:
> 
> 
> > PS: How do i test multiplayer with Warrior Orochi 2 ? The 2nd player keeps getting "channel already used" message.
> 
> Change WLAN channel to 1 or 6. :)

So you can play together even with different channel? O.o i thought you need to use the same channel in order to be able to meet/see each other (like most wireless adapter usually do)

@MojoJojoDojo check the log, what kind of SCENET error did you get?

@zakilj3 if this crash on android only happened on Ratchet & Clank or other game too?
If it's only on Ratchet & Clank it's probably the same crashed i had on PC (which doesn't seem to be network related, since it's also happening before i choose adhoc/infrastructure selection screen), this crash is a bit random.

ANR2ME:
Updated because i forgot another condition on NetconfDialog where the Joining side could get an endless repetition of GetScanInfo if the Host is creating the group too late.

The rest of the changes are minor changes trying to get the logs & outputs in similar condition with the working KHBBS when using prx files.
However, these changes are probably won't affect any game at all, including KHBBS LOL

Anyway, test builds have been updated also.

zakilj3:
The crash on Android only affected size matter, even when playing alone in multiplayer on Android (also infrastructures doesn't work when I try it, it get a message saying that PPSSPP can't go online as it require using a psn account and other stuff as its real online and not ad-hoc)

ANR2ME:
> 
> 
> The crash on Android only affected size matter, even when playing alone in multiplayer on Android (also infrastructures doesn't work when I try it, it get a message saying that PPSSPP can't go online as it require using a psn account and other stuff as its real online and not ad-hoc)
> 
> Edit : for me the crash on android happen when I m about to launch the game, After the loading it goes black screen and crash(and in USA version it crash like that on both Android and pc so you must use the PAL version)

What is the code for the PAL version?

Does the PAL version also have graphical issue like this on android? or was it because i was using remote disc (which shows a lot of Error writing to socket) ?
image

There is also Profile issue i'm experiencing with US version where sometimes i can't "Select profile" and "Create profile" didn't works either because it keeps asking for password when i wanted to save the profile, when this happened the only solution i know was to delete the savedata and startover :(
Is this profile issue also happened on PAL version?

zakilj3:
Sorry what do you mean by code? Cheking the log or the game ID?

I didn't had graphical issue on my side so maybe its due to remote disc

And the only issue I had once was that I couldn't load a profile or create one because I speed up too much the game that it didn't got time to load the preexisting  profile(but when I reopened the game and didn't speed up, it went back to normal) my crash for size matter is pretty much the same as before when I start the game and there is the loading screen, it crash before starting so I could join my pc's lobby

ANR2ME:
> 
> 
> Sorry what do you mean by code? Cheking the log or the game ID?
> 
> I didn't had graphical issue on my side so maybe its due to remote disc
> 
> And the only issue I had once was that I couldn't load a profile or create one because I speed up too much the game that it didn't got time to load the preexisting profile(but when I reopened the game and didn't speed up, it went back to normal) my crash for size matter is pretty much the same as before when I start the game and there is the loading screen, it crash before starting so I could join my pc's lobby

Ah yes, i meant the game/title ID

zakilj3:
The game ID is UCES00420

MojoJojoDojo:
I got the european version of Ratchet and Clank to work but it will just freeze after a game starts.
It will freeze and the music will continue but you cant do anything.

zakilj3:
I also have that but it appear more that's the image that's fix, I can hear my character walk or swimp and I can hear the pause menu when I push it

ANR2ME:
> 
> 
> I also have that but it appear more that's the image that's fix, I can hear my character walk or swimp and I can hear the pause menu when I push it

Have you tried changing the graphics setting from Buffered to non-buffered?
I think that game still have some graphical issue, probably not a good time to test multiplayer on it yet :D

zakilj3:
I might try on non buffered to see how it goes

MojoJojoDojo:
There is no graphical issues it's only the multiplayer that got that issue lol.
Single player works fine.
It might not be adhoc related dunno.

zakilj3:
I'd say that its probably the multiplayer mode itself that is bugged and not the connection (at least not with multiple instance on pc)

ANR2ME:
In my case it even crashed after choosing "Multiplayer Mode" which doesn't do anything network related yet, it crashed in the middle of the rocket flying animation LOL

Anyway, did you guys still seeing "Network Initialized" message on screen in the middle of multiplayer?

ANR2ME:
Updated to fix an issue with Naruto Shippuden Ultimate Ninja Heroes 3 that got broken on previous changes.

Also added a 5 seconds timeout on NetconfDialog to prevent it from retrying forever when loosing connectivity to the adhoc server.
It will shows something like this when getting timedout:
image

PS: Test builds have been updated

ANR2ME:
After rebasing this it seems Kingdom Hearts BBS have a working Arena VS mode O.o
image

hrydgard:
Ooh, very cool!

Is this branch ready to go now?

zakilj3:
Interesting, I should try it with 2 windows and also with android(but I feel there will likely be issues with Android just like with bleach and some other games)

ANR2ME:
> 
> 
> Ooh, very cool!
> 
> Is this branch ready to go now?

I think so, i haven't found anymore issue so far, at least no regression.

ANR2ME:
> 
> 
> Interesting, I should try it with 2 windows and also with android(but I feel there will likely be issues with Android just like with bleach and some other games)

Android build is not updated yet, i'm still building it.

MojoJojoDojo:
This got anything to do with the Gran Turismo fix?
Maybe its the key to fix the coop mode which is the only one that got left lol.

ANR2ME:
> 
> 
> This got anything to do with the Gran Turismo fix?
> Maybe its the key to fix the coop mode too which is the only one that got left lol.

High probability, because that PR fixed a lot of things on games that use AdhocMatching, while this PR is improvements with side effects of fixing games that was frozen due to blocking socket blocking the emulation.

MojoJojoDojo:
> > This got anything to do with the Gran Turismo fix?
> > Maybe its the key to fix the coop mode too which is the only one that got left lol.
> 
> High probability, because that PR fixed a lot of things on games that use AdhocMatching, while this PR is improvements with side effects of fixing games that was frozen due to blocking socket that blocks the CPU emulation.
> 
> Btw, which one is the coop mode? i only tested it on Versus and Arena mode

I fogot that Arena is the coop mode lol.
Yeah it means all the important modes work now.

ANR2ME:
> 
> 
> Interesting, I should try it with 2 windows and also with android(but I feel there will likely be issues with Android just like with bleach and some other games)

This even works better than JPCSP with prx files LOL with JPCSP after you're done and return to mirage area, one of the player (the joining one) is always got kicked out from mirage arena or frozen, while with PPSSPP it can go back to mirage area successfully without any issue.

Regarding Bleach 7, i feels there is timing issue because the host need to accept joining player as soon as possible, delaying for a short time will automatically cancel the join request, and it only happened on P2P mode, while Host/Join mode worked without problem.
Well at least one of multiplayer (Host/Join mode) on Bleach 7 is working properly now :D

zakilj3:
I ve tested Android with pc, both vs and arena mode works flawly I ll be infinitely grateful to U ANR2ME

Lemoncak3:
Phantasy star portable 2 have worked wonderfully again now. I hope you will have a way to fix another multiplayer game too. Thanks a lot

Issue #11836, opened 2019-02-23
(Libretro) Crashes with vulkan when started


orbea:
### What happens?

When running the libretro core in RetroArch with vulkan it segfaults immediately when starting.
```
Thread 1 "retroarch" received signal SIGSEGV, Segmentation fault.
0x0000000000000000 in ?? ()
(gdb) bt
#0  0x0000000000000000 in ?? ()
#1  0x00007fffeb5eb5b3 in VulkanContext::CreateInstance (this=0x1eaf7c0, 
    info=...)
    at /home/orbea/gittings/forks/ppsspp/Common/Vulkan/VulkanContext.cpp:245
#2  0x00007fffeb21f0f7 in create_device (context=0x7fffffffcf80, 
    instance=0x1c4ffa0, gpu=0x0, surface=0x1, 
    get_instance_proc_addr=0x7fffe836b6b0 <vkGetInstanceProcAddr>, 
    required_device_extensions=0xf04230 <device_extensions>, 
    num_required_device_extensions=1, 
    required_device_layers=0xf04238 <device_layers>, 
    num_required_device_layers=1, required_features=0x7fffffffcfb0)
    at /home/orbea/gittings/forks/ppsspp/libretro/LibretroVulkanContext.cpp:41
#3  0x0000000000749c15 in vulkan_context_init_device (vk=0x1794c00)
    at gfx/common/vulkan_common.c:1683
#4  0x000000000074b1c1 in vulkan_surface_create (vk=0x1794c00, 
    type=VULKAN_WSI_XCB, display=0x1795580, surface=0xfbffd8 <g_x11_win>, 
    width=748, height=211, swap_interval=0) at gfx/common/vulkan_common.c:2483
#5  0x00000000007185f0 in gfx_ctx_x_set_video_mode (data=0x1794bb0, 
    video_info=0x7fffffffd570, width=1680, height=1050, fullscreen=true)
    at gfx/drivers_context/x_ctx.c:906
#6  0x00000000004a932f in video_context_driver_set_video_mode (
    mode_info=0x7fffffffda64) at gfx/video_driver.c:3264
#7  0x0000000000740f0e in vulkan_init (video=0x7fffffffdac0, 
    input=0xf5bf90 <current_input>, input_data=0xf5bf98 <current_input_data>)
    at gfx/drivers/vulkan.c:1206
#8  0x00000000004a50df in video_driver_init_internal (
[...]

PR #13328, opened 2020-08-23
Fix libretro with Vulkan

hrydgard:
This gets it going again, although not completely stable and it has some other minor issues.

However, it's a start.

Includes a new README_WINDOWS.txt file for getting started with building and debugging for libretro on Windows.


orbea:
I suppose I'll link this here for future reference.

https://github.com/hrydgard/ppsspp/issues/11836

Common/Vulkan/VulkanContext.cpp
```diff
@@ -146,10 +146,14 @@ VkResult VulkanContext::CreateInstance(const CreateInfo &info) {
 		}
 	}
 
+	// Temporary hack for libretro. For some reason, when we try to load the functions from this extension,
+	// we get null pointers when running libretro. Quite strange.
```
aliaspider, line 150: `vkCreateInstance` is sort of stubbed with libretro since the frontend already created it, which means `instance_extensions_enabled_` is currently being ignored. 
I can't think of an easy way to solve this other than having the frontend always load the extension when available, the current libretro_vulkan api lacks the ability to allow the core to request specific instance extensions.

hrydgard: Okay, that does make sense at least. We'll just have to deal with that then, there's no urgent need for that extension anyway.

hrydgard:
I've fixed the screen size issue, making it a bit more playable.

There are some weird stability issues when unthrottling (space), and the sizing is not 100% accurate - there are some ugly duplicated lines. Not sure whose fault those are...

hrydgard:
Yeah run at 1x and play a game with detailed graphics, you'll notice that a couple of lines are missing or duplicated, it's like the vertical size is just a pixel or two off. 

But at least it's kinda usable now. I'm gonna merge.

aliaspider:
ah okay that is what you meant.

truns out `System_GetPropertyFloat` was defaulting to `-1` for unimplemented `SYSPROP*` which had some bad side effects.
#13548 should take care of it.

PR #13360, opened 2020-08-30
GE dump loading: Try to use the correct GameID so compat.ini flags apply.

hrydgard:
compat.ini flags can sometimes impact rendering, such as BlockTransferAllowCreateFB (virtual framebuffer copies), so when replaying GE dumps, we should use the same compat.ini settings as the game. This change extracts the game ID from the dump filename, so if you want to avoid this, rename the GE dump.

I thought this might be why GE dumps from Test Drive like 
[ULES00637_0003.zip](https://github.com/hrydgard/ppsspp/files/5146610/ULES00637_0003.zip) didn't render correctly, but that did not fix it.

The problem is block transfer related, at least.

### DIFF
--- a/Core/ELF/ParamSFO.h
+++ b/Core/ELF/ParamSFO.h
@@ -22,6 +22,7 @@
 #include <vector>
 
 #include "Common/CommonTypes.h"
+#include "Common/Log.h"
 
 class ParamSFOData
 {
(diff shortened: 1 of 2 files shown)


unknownbrackets (approved):
Currently the remote/web GE debugger does not include the game ID in the filename, but this is probably fine for now.

-[Unknown]

Issue #2451, opened 2013-06-24
Crazy Taxi: Fare Wars


HelicopterP:
The game freezes at loading after selecting Crazi Taxi 1. The log is really huge but it goes something like this:
a few MB of this lines -
36:53:926 idle0        W[HLE]: HLE\sceKernelThread.cpp:2598 sceKernelSuspendThread(314)
36:53:926 user_main    W[HLE]: HLE\sceKernelThread.cpp:1926 800201a4=sceKernelStartThread(thread=314, argSize=16, argPtr=0986be40): thread already running
36:53:926 user_main    W[HLE]: HLE\sceKernelThread.cpp:2629 sceKernelResumeThread(314)
36:53:926 idle0        W[HLE]: HLE\sceKernelThread.cpp:2598 sceKernelSuspendThread(314)
36:53:926 user_main    W[HLE]: HLE\sceKernelThread.cpp:1926 800201a4=sceKernelStartThread(thread=314, argSize=16, argPtr=0986be40): thread already running
36:53:926 user_main    W[HLE]: HLE\sceKernelThread.cpp:2629 sceKernelResumeThread(314)
36:53:926 idle0        W[HLE]: HLE\sceKernelThread.cpp:2598 sceKernelSuspendThread(314)

than this:

36:53:943 user_main    I[HLE]: HLE\sceKernel.cpp:461 KO 297: Alarm "[Alarm]": -
36:53:944 user_main    I[HLE]: HLE\sceKernel.cpp:461 KO 298: Alarm "[Alarm]": -
36:53:944 user_main    I[HLE]: HLE\sceKernel.cpp:461 KO 299: Thread "idle0": 

than some stuff mentioning sceMp3 than a lot of Alarms again than crash.

Here is a full log: http://www.sendspace.com/file/fwuiss

sum2012:
v0.9.9.1-1503-gc250074 same problem
ct1 debug log:
https://drive.google.com/file/d/0B3OaSdeV0L8kQWk2SE5zTHAwbDA/view?usp=sharing

ct2 debug log:
https://drive.google.com/file/d/0B3OaSdeV0L8kcXIxU1pzdzBnMEk/view?usp=sharing

ppmeis:
Tested with latest build. It crash in both CT1 and CT2 after a while. Maybe related to #3854?¿

sum2012:
@ppmeis It doesn't work with interpreter mode,so that it is not related to #3854

[...]

PR #13368, opened 2020-09-01
Reschedule after resuming thread from suspend.

shenweip:
This might help #2451 , a test result see [here](https://github.com/hrydgard/pspautotests/pull/208).


hrydgard:
Interesting find, your tests seems to prove that this is correct indeed. Let's try it!

PR #13385, opened 2020-09-05
Adhoc Socket Reindexing - Fix Crash issue due to out of bound when indexing

anr2me:
I forgot to reindex PDP socket on the blocking socket simulation PR after rebasing from Gran Turismo fix PR which have been reindexed :( 
it's so hard to works on multiple branch if they are using the same things... because the blocking simulation PR was completed since days ago i forgotten about it LOL

This should fix crash issue due to out of bound while indexing PDP array (was having this issue when testing Monster Hunter Portable 3rd with 4 players in localhost)


MojoJojoDojo:
Hopefully Kingdom Hearts status will stay the same after all the fixes lol.

PR #13405, opened 2020-09-08
Allow blocking socket simulation on multiple threads simultaneously

anr2me:
It seems it's allowed to have blocking Recv on one thread, while having the same socket to Send on another thread.


hrydgard:
Ready to go? I can't say I understand everything that's going on here exactly, but looks sensible :)

ANR2ME:
I think it's ready, although i wasn't sure whether the flags should be reset after the alert triggered or not, but doesn't seems to have any issue so far.

Issue #13398, opened 2020-09-07
PPSSPP lag at exit menu after playing network


Lemoncak3:
I have tried with the latest build and play together with friends, when want to exit to menu it's so lag and some game seems ro be stuck after pressing exit menu. When i play game normally without network, this issue didn't happen though. Tried it in phantasy star portable 2 and another game that have network mode. Not tested to can play together or not just test until network initialized thing and i press exit to menu and the issue happens.

MojoJojoDojo:
Does not happen on Linux if I am not mistaken so its Windows only.
On Linux PPSSPP might crash if you connecting to an Ad-hoc server that does not exist (exclusive to that OS for now).

Lemoncak3:
I am using android though, and have tested it with few game with network initialized worked, and when press the exit to menu either it's very lag or i am just stuck in there sometimes ppsspp will force close.

ANR2ME:
When exiting a game it need to cleanup some resources, and if you're using UPnP it might takes longer because it need to close open ports or restore it back to the original owner of all the ports used by the game on-by-one (unfortunately there isn't a batching way to do this from miniUPnP)

It's also possible for the lag to be caused by crashes, for example when i had the crash related to D3D11 i reported recently, i noticed a longer time to close down PPSSPP, but that's related to GPU so it's probably different issue with this one.

Anyway, it's hard to investigate this without checking the logs, even better if debugged directly from android studio to catch any crash.

Unfortunately, my laptop can no longer detects my android phone (not even as storage) when i plugged the usb cable, not sure why :( but i think it's started since i updated android studio/sdk tools that have a new google driver, and haven't found a way to solve this. So i can't test android anymore.

MojoJojoDojo:
There is also that issue of Linux crashing when connecting to a non existing adhoc server .
Is there a way to get a log after a crash though?

ANR2ME:
> 
> 
> There is also that issue of Linux crashing when connecting to a non existing adhoc server .
> Is there a way to get a log after a crash though?
> @Lemoncak3 can you check if that happens on Android too?
> connect to a random IP address.



> 
> 
> There is also that issue of Linux crashing when connecting to a non existing adhoc server .
> Is there a way to get a log after a crash though?
Well, i'm not familiar with linux (never use one) so not sure about it either.

I wonder if the crash on linux might be related to the SIGPIPE thing? have you tried to build it again with the latest commits which includes the SIGPIPE PR?

[...]

PR #13408, opened 2020-09-09
Stop waiting for connection on Exit to Menu to prevent causing lags.

anr2me:
This should fix the lag issue on https://github.com/hrydgard/ppsspp/issues/13398

### DIFF
--- a/Core/HLE/proAdhoc.cpp
+++ b/Core/HLE/proAdhoc.cpp
@@ -1974,12 +1974,19 @@ int initNetwork(SceNetAdhocctlAdhocId *adhoc_id){
 	iResult = connect(metasocket, (sockaddr*)&server_addr, sizeof(server_addr));
 	errorcode = errno;
 
-	if (iResult == SOCKET_ERROR && errorcode != EISCONN && (IsSocketReady(metasocket, true, true, nullptr, adhocDefaultTimeout * 1000) <= 0)) {
-		char buffer[512];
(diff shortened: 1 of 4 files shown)


Core/HLE/proAdhoc.cpp
```diff
@@ -1973,12 +1973,19 @@ int initNetwork(SceNetAdhocctlAdhocId *adhoc_id){
...
-		ERROR_LOG(SCENET, "%s", buffer);
-		host->NotifyUserMessage(n->T("Failed to connect to Adhoc Server"), 1.0f, 0x0000ff);
-		return iResult;
+	if (iResult == SOCKET_ERROR && errorcode != EISCONN) {
+		u64 startTime = (u64)(real_time_now() * 1000.0);
+		while (IsSocketReady(metasocket, true, true) <= 0) {
+			u64 now = (u64)(real_time_now() * 1000.0);
+			if (coreState == CORE_POWERDOWN) return iResult;
+			if (now - startTime > adhocDefaultTimeout) break;
+			sleep_ms(1);
```
hrydgard, line 1982: Maybe increase to 10 or maybe 50? Not that it probably matters much here, but generally, it's not good for power consumption to poll stuff as fast as at a 1ms interval.

ANR2ME: Actually, it's quite common to sleep for 1ms while waiting for as state change on non-blocking socket on many networking code example, probably to simulate blocking behavior on the system level just like when using blocking socket.

But since there are cases where the Adhoc Server may not even exist/unreachable here, it will probably ended being used more often due to a retry on disconnection detection, so i guess i should raise the sleep to 10ms to lower power CPU usage.

ANR2ME:
Hmm.. looks like there is another lags issue when it failed to resolve the Adhoc Server DNS when i tested it with random domain for the server.

PR #13411, opened 2020-09-10
Add option for the clear-fb-on-first-use speedhack.

hrydgard:
See #13295.

End goal is probably to remove it. Defaults to false so we'll see if anyone notices any speed difference.


unknownbrackets (approved):

Gamemulatorer:
I didn't feel any improvement when this option enable.

fahadrashid7705:
yes not feel any improvement too

Issue #13186, opened 2020-07-23
Dynasty Warrior strikeforce multiplayer issue


Lemoncak3:
I have tested that we can enter the city just fine, even can see each other but when send item trade or friend request and also accept request, other player can't accept because there is no windows or notifications so we can't enter the mission only stuck in the city.

adenovan:
log please

Lemoncak3:
How to get a log in PPSSPP? Never get one before..

adenovan:
Settings->Tools->Developer Tools->Debug Logging

Saramagrean:
Try change WLAN channel to 1, It work for WO2 (but gameplay so slow),  think it will work? because it is the Musou series from the same game developer.
![WO2](https://user-images.githubusercontent.com/26355472/89121004-48129c80-d4e5-11ea-8484-3c8d8aeaaa0f.png)

Lemoncak3:
Have tested with all wlan channel with still same issue 
![Screenshot_20200816-042839](https://user-images.githubusercontent.com/67026697/90322086-20f2ab00-df7a-11ea-841a-fbd53f6b8549.png)
My brother can join just fine but when i send item trade request and friend request, he didn't receive anything.
![Screenshot_20200816-042855](https://user-images.githubusercontent.com/67026697/90322092-3b2c8900-df7a-11ea-9c9b-3ec7b1efcbb0.png)

Lemoncak3:
Okay will download the latest build and will inform you the result

ANR2ME:
> 
> 
> Okay will download the latest build and will inform you the result

it's not built yet, you need to wait until that PR appeared on https://buildbot.orphis.net/ppsspp/

Lemoncak3:
Have tested it but from client side it shows this error
![IMG-20200912-WA0000](https://user-images.githubusercontent.com/67026697/92969473-41ddeb80-f4a7-11ea-9957-783d89bb6b5d.jpg)
and from my side as the server tried to accept quest, the client side still haven't receive anything
![Screenshot_20200912-031604](https://user-images.githubusercontent.com/67026697/92969550-5fab5080-f4a7-11ea-9aef-02586c608f84.png)
And when i send item trade or friend request this error shows in my screen 
![Screenshot_20200912-030241](https://user-images.githubusercontent.com/67026697/92969593-76ea3e00-f4a7-11ea-8fa9-8da51cede3bd.png)

ANR2ME:
[...]

PR #13414, opened 2020-09-11
Allows reusing the same port for PtpListen and PtpOpen

anr2me:
When testing Shin Sangoku Musou Multi Raid 2 using JPCSP + prx files, it seems PTP Ports are reusable.
This fixes Shin Sangoku Musou Multi Raid 2, and probably other games with similar issue (ie. Dynasty Warrior Strikeforce https://github.com/hrydgard/ppsspp/issues/13186)

PS: Not sure if PDP Ports also reusable like this or not, will need someone to test this on homebrew.

### DIFF
--- a/Core/HLE/sceNetAdhoc.cpp
+++ b/Core/HLE/sceNetAdhoc.cpp
@@ -2605,130 +2605,119 @@ static int sceNetAdhocPtpOpen(const char *srcmac, int sport, const char *dstmac,
 			// Random Port required
 			if (sport == 0) {
 				isClient = true;
-				// Find unused Port
-				// while (sport == 0 || _IsPTPPortInUse(sport)) {
-				// 	// Generate Port Number
-				// 	sport = (uint16_t)_getRandomNumber(65535);
-				// }
-			}
-			
-			// Valid Ports
-			if (!isPTPPortInUse(sport) /*&& dport != 0*/) {
 				//sport 0 should be shifted back to 0 when using offset Phantasy Star Portable 2 use this
-				if (sport == 0) sport = -(int)portOffset;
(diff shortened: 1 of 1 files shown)


Core/HLE/sceNetAdhoc.cpp
```diff
@@ -2605,130 +2605,119 @@ static int sceNetAdhocPtpOpen(const char *srcmac, int sport, const char *dstmac,
...
-						
-						// Binding Information for local Port
-						sockaddr_in addr;
-						// addr.sin_len = sizeof(addr);
-						addr.sin_family = AF_INET;
-						addr.sin_addr.s_addr = INADDR_ANY;
-						if (isLocalServer) {
-							getLocalIp(&addr);
+					// Disable Nagle Algo to send immediately. Or may be we shouldn't disable Nagle since there is PtpFlush function?
+					if (g_Config.bTCPNoDelay) setSockNoDelay(tcpsocket, 1);
```
hrydgard, line 2636: Just thought of a coding style thing - when you're changing or adding code, could you put the statement after an if on its own line? I think that reads a lot better. 

Often it can even be nice to put that single line in brackets - makes it easier to add more code later (though for things where that's not likely, omitting the brackets is ok too).

Don't have to fix this now in this PR, but in the future.

ANR2ME: Yes, i usually put it on next line when writing new code since it's easier to put breakpoints that way, but if it was an old code... i leave it as is LOL

hrydgard: Ah didn't realize that line was just moved :)

PR #13418, opened 2020-09-12
IO:Don't read data more than data size of pgd files.

### LINKED ISSUE #6025: Dream Club Portable crash after select girl
v0.9.8-588-gdd10207
info log:
https://gist.github.com/sum2012/100dd16e1236c2e8d635
Debug log:
https://drive.google.com/file/d/0B3OaSdeV0L8kU2tnZ0toSXpJTWc/edit?usp=sharing

shenweip:
Fixes #6025 , matches a PSP.

### DIFF
--- a/Core/HLE/sceIo.cpp
+++ b/Core/HLE/sceIo.cpp
@@ -967,6 +967,8 @@ static u32 npdrmRead(FileNode *f, u8 *data, int size) {
 	block  = pgd->file_offset/pgd->block_size;
 	offset = pgd->file_offset%pgd->block_size;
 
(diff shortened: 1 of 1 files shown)


shenweip:
[sceIo.cpp#L2418](https://github.com/hrydgard/ppsspp/blob/766bd8e6bb182f4ef6523083efaea251114cd646/Core/HLE/sceIo.cpp#L2418)
This was added in 8021e49c, I not sure whether this is related to PGD fonts, but I notice that PSP has no this behavior when reading a normal PGD file.

PR #13426, opened 2020-09-13
Texture filtering cleanups

hrydgard:
Some code cleanup, preparing for bigger cleanups and the addition of some new features soon (automatic top mipmap quality mode, etc).

Also remove the unclear "Linear on FMV" option. NEAREST filtering on video makes very little sense so now it's always on if resolution is 2x or higher, except in force-nearest mode.


GPU/Common/TextureCacheCommon.cpp
```diff
@@ -154,46 +154,48 @@ void TextureCacheCommon::GetSamplingParams(int &minFilt, int &magFilt, bool &sCl
...
 			magFilt |= 1;
 			minFilt |= 1;
 		}
-	}
-	bool forceNearest = g_Config.iTexFiltering == TEX_FILTER_NEAREST;
-	// Force Nearest when color test enabled and rendering resolution greater than 480x272
-	if ((gstate.isColorTestEnabled() && !IsColorTestTriviallyTrue()) && g_Config.iInternalResolution != 1 && gstate.isModeThrough()) {
+	} else if (g_Config.iTexFiltering == TEX_FILTER_NEAREST ||
+		(gstate.isModeThrough() && g_Config.iInternalResolution != 1 &&
+		gstate.isColorTestEnabled() && !IsColorTestTriviallyTrue() && gstate.getColorTestRef() != 0)) {
```
unknownbrackets, line 195: My opinion is that multiline ifs, and multiline statements in general, is just a sign of hard to read code.

I don't think anyone will look at this if statement and suggest that its logic is obvious or clear.  Adding more newlines, parenthesis, or playing with indentation would not help that at all.  If you can't get your point across in a normal sentence, making it longer without a period isn't going to improve the situation.

Generally, I think the same about function arguments (calls or defs), etc.  If your function takes 10 arguments, you should use a struct/class, if your if has 10 clauses you should use temporaries, funcs, or lambdas, etc.

-[Unknown]

hrydgard: Yeah I'm honestly a bit torn here. Sequences of only && are pretty easy to read chained but as soon as there's an ||, you start to need the parenthesis and it suddenly becomes a lot less readable.

I'm going to change it to a kind of middle ground where each if has only one type of operator. That should take care of it.

GPU/Common/TextureCacheCommon.cpp
```diff
@@ -154,46 +154,48 @@ void TextureCacheCommon::GetSamplingParams(int &minFilt, int &magFilt, bool &sCl
...
 			magFilt |= 1;
 			minFilt |= 1;
 		}
 	}
-	if (g_Config.iTexFiltering == TEX_FILTER_LINEAR && (!gstate.isColorTestEnabled() || IsColorTestTriviallyTrue())) {
-		if (!gstate.isAlphaTestEnabled() || IsAlphaTestTriviallyTrue()) {
+
+	// Filtering overrides
+	if (g_Config.iTexFiltering == TEX_FILTER_LINEAR) {
+		// Only override to linear filtering if there's no alpha or color testing going on.
```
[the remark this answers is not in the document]
hrydgard: Just trying to explain what's going on, not changing the logic. Not quite sure what "without entirely needing to" refers to?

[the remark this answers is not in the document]
unknownbrackets: Oh, I thought this changed the rules.  I guess we didn't end up allowing force linear for color test = 0.  I remembered that longstanding issue for Tekken and was thinking of that mainly.

-[Unknown]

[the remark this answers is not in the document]
hrydgard: Yeah I'm not changing any rules here at all, except the video filtering.

hrydgard:
Rewrote the logic again, I think it's a ton more readable now.

GPU/Common/TextureCacheCommon.cpp
```diff
@@ -177,27 +177,39 @@ void TextureCacheCommon::GetSamplingParams(int &minFilt, int &magFilt, bool &sCl
...
 	}
 
 	// Filtering overrides
-	if (g_Config.iTexFiltering == TEX_FILTER_FORCE_LINEAR) {
-		// Only override to linear filtering if there's no alpha or color testing going on.
+	switch (g_Config.iTexFiltering) {
+	case TEX_FILTER_AUTO:
+		// Follow what the game wants. We just do a single heuristic change to avoid bleeding of wacky color test colors
+		// in higher resolution (used by some games for sprites, and they accidentally have linear filter on).
+		// This does duplicate the body of the above if, but it's clearer this way.
```
unknownbrackets, line 190: No longer above, but the nearest case?  Not an important distinction.

-[Unknown]

unknownbrackets (approved):

GPU/Common/TextureCacheCommon.cpp
hrydgard: Woops, it went through a couple iterations.

Issue #13431, opened 2020-09-14
Graphical Regression in Burnout games Since v1.10.3-661-ga6084f6fe


Panderner:
It seems #13420 broke Burnout games here's a list of bugs

Burnout Legends:
in Car select screen in legends the car textures are black:
![Screenshot_2020-09-15-00-09-43-22_2f85358b2198d26f8aca533d68bee793](https://user-images.githubusercontent.com/37503397/93111802-b86a2b80-f6e9-11ea-89b2-c91138fe2a63.png)

Some tracks gave big performance killing problem (winter city, silver lake, waterfront, etc)
Before v1.10.3-661-ga6084f6fe:
![Screenshot_2020-09-15-00-07-04-60_2f85358b2198d26f8aca533d68bee793](https://user-images.githubusercontent.com/37503397/93112425-91602980-f6ea-11ea-88fd-fc4c1d634416.png)

After v1.10.3-661-ga6084f6fe:
![Screenshot_2020-09-15-00-10-01-75_2f85358b2198d26f8aca533d68bee793](https://user-images.githubusercontent.com/37503397/93112458-a0df7280-f6ea-11ea-9c3d-2443fe8ecb6b.png)

During in-game the police cars textures became black:
![Screenshot_2020-09-15-00-11-45-02_2f85358b2198d26f8aca533d68bee793](https://user-images.githubusercontent.com/37503397/93112711-e439e100-f6ea-11ea-9a64-f9c64f0d8caf.png)

The Flags are corrupted (as i reported in r/EmulationOnAndroid in Reddit)
![df7e6a3](https://user-images.githubusercontent.com/37503397/93113069-4eeb1c80-f6eb-11ea-9cb5-d96f98288c41.jpg)

In Burnout Dominator I don't notice any Graphical glitches but big performance bottleneck:

Before v1.10.3-661-ga6084f6fe (i turned off simulate block transfer effects):
![Screenshot_2020-09-15-00-07-59-02_2f85358b2198d26f8aca533d68bee793](https://user-images.githubusercontent.com/37503397/93113246-85289c00-f6eb-11ea-8e07-434a8e990e25.png)

After v1.10.3-661-ga6084f6fe:
![Screenshot_2020-09-15-00-32-49-16_2f85358b2198d26f8aca533d68bee793](https://user-images.githubusercontent.com/37503397/93113416-bef9a280-f6eb-11ea-81e8-4b62fd8054e9.png)

Panderner:
[...]

PR #13436, opened 2020-09-15
Texture invalidation fixes

hrydgard:
TextureCache: Don't drastically over-invalidate in NotifyFramebuffers. Burnout has a case where a framebuffer that's updated a lot (the lens flare effect virtual framebuffer) has a misdetected height and overlaps a lot of textures, so cap to make sure that when creating framebuffers, we only really kill off cached textures that start at or nearly at the same address.

Should help at least the performance part of #13431. Can't quite account for the broken graphics yet, which don't repro on NV. (however there's a related lost-device issue on Intel...).

Snuck in a couple of unrelated bugfixes and some more asserts.


hrydgard:
Removed the header cleanup stuff from this branch, will do that separately later.

PR #13440, opened 2020-09-16
Warning fixes.

shenweip:
Fixes warning C4267, warning C4819, warning C4018 in VS.


hrydgard:
Not sure if we should fix warnings in ext dependencies like Snappy or just find a way to disable them instead... hm

hrydgard:
Merged this manually, along with some extra warning fixes. closing.

PR #13450, opened 2020-09-20
Implementing Adhoc GameMode

anr2me:
This should fix multiplayer on BattleZone. 
PS: I only tested this on BattleZone because it's the only game i knew that uses GameMode API


sum2012:
You can use INFO_LOG_REPORT_ONCE so that you will see it in https://report.ppsspp.org/logs/kinds/
You might will know other game that uses GameMode API

hrydgard:
Very cool, code looks good (skimmed through it though).

Yes it's a good idea by sum2012 to add some reporting for this.

ANR2ME:
Hmm.. There is only 1 example of `INFO_LOG_REPORT_ONCE` example in the solution:
```
INFO_LOG_REPORT_ONCE(displayoffset, FRAMEBUF, "Rendering from framebuf with offset %08x -> %08x+%dx%d", addr, vfb->fb_address, offsetX, offsetY);
```
But i don't understand what is `displayoffset` used for? it doesn't seems to be declared/defined anywhere.

PR #13457, opened 2020-09-21
Make universal binary version of ffmpeg work.

MaddTheSane:
This will make the compiler know where to look for FFmpeg if it's compiled as a universal binary for Arm64 and x86_64.

Note: you will need Xcode 12.2 or later (12.2 is currently downloadable as a beta) in order to compile Arm64 Mac code.


hrydgard:
Took some hackery but:

image

PR #13470, opened 2020-09-23
Adhoc Socket - An attempt to improve connection stability that was broken recently.

anr2me:
It's seems my previous PR to fix performance on Warriors Orochi 2 was causing connection instability due to high rate of dropped packets.

### DIFF
--- a/Core/HLE/sceNet.cpp
+++ b/Core/HLE/sceNet.cpp
@@ -712,7 +712,7 @@ static void sceNetEtherStrton(u32 bufferPtr, u32 macPtr) {
 
 // Write static data since we don't actually manage any memory for sceNet* yet.
 static int sceNetGetMallocStat(u32 statPtr) {
-	DEBUG_LOG(SCENET, "UNTESTED sceNetGetMallocStat(%x)", statPtr);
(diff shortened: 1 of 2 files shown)


Lemoncak3:
So will this fix make warrior orochi lag again? I haven't tried it before though.

MojoJojoDojo:
I guess you should just stay on that specific version if you want orochi to have no lag.

kage2051:
Or better yet, upgrading to a PC or device with higher end specs might be your solution.

hrydgard:
What kind of a silly comment is that? In networking code, performance problems are rarely due to CPU...

ANR2ME:
> So will this fix make warrior orochi lag again? I haven't tried it before though.

This should also fix the issue where Warriors Orochi often failed to Join. 
On localhost Warriors Orochi should be running in full speed, but on LAN there might be some slowdowns sometimes (especially if there are too many movements), but not as bad as a slide show.

kage2051:
For some reason I misread when I posted my original reply.

But yeah, network lag can also be produced by other protocols using 2.4GHz channel if WLAN is used on that same channel as well. This on Android multiplayer.

PR #13473, opened 2020-09-24
SSE/NEON-optimize IndexGenerator::AddStrip. 

hrydgard:
Did some profiling and this one was pretty high in several games.

Shaves about half a percent of frame time from GoW so not a huge success but something at least.


kage2051:
How does it improve performance in general? Really don't quite understand this PR.

hrydgard:
The PSP's GPU is happy doing very small draws, such as triangle strips of say 20 triangles - and lots of those. Modern GPUs are much more parallel and "heavy" and want their work in bigger chunks so we try to put together many small triangle strips into large lists of indexed triangles. Generating the index buffer for that is what this function is used for.

In this change, we use [SSE2](https://en.wikipedia.org/wiki/SSE2) instructions to process the indexing for eight triangles at a time,  which is much faster than the old code.

Overall though the performance improvement is small, but measurable for heavy games like God of War, around half a percent.

PR #13483, opened 2020-09-27
glcore context switching

barbudreadmon:
preparatory patch for https://github.com/libretro/RetroArch/pull/11368

### DIFF
--- a/libretro/CMakeLists.txt
+++ b/libretro/CMakeLists.txt
@@ -3,6 +3,7 @@ set(LIBRETRO_SRCS
 	libretro.cpp
 	LibretroGraphicsContext.cpp
 	LibretroGLContext.cpp
+	LibretroGLCoreContext.cpp
 	LibretroVulkanContext.cpp
 	libretro_vulkan.cpp)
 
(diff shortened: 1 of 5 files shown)


hrydgard:
I'm fine with this, not really sure what's going on in the discussion in libretro/RetroArch#11368 though?

barbudreadmon:
It was mostly discussed at libretro's discord, the status is that @twinaphex must test the https://github.com/libretro/RetroArch/pull/11368 PR since he was the only one who reported issues with it a few months ago. Every core using hardware rendering needs a bit of support code before it's merged, otherwise the video driver will always be forced to whatever the core requests, or the core won't launch if the video driver mismatch and the "Allow cores to switch video driver" is toggled off.
Those PRs i sent to most of those cores should be harmless with current RetroArch, so i think they should be merged as soon as possible.

barbudreadmon:
Also, i wonder if vulkan shouldn't be disabled in the libretro port for now, since it's crashing and i believe it's not an issue on my side ?

barbudreadmon:
Other core PRs were already merged since they aren't breaking anything with current RA, so if you could merge this one too it would be perfect.
Most likely the RA PR will be merged after every core needing this are ready (i still have a few cores to deal with).

barbudreadmon:
thanks :)

PR #13486, opened 2020-09-28
Add PSP color shader

ghost:
This adds the PSP color shader ported from RetroArch courtesy from @jdgleaver (https://forums.ppsspp.org/showthread.php?tid=6594&pid=130635#pid130635) to PPSSPP

Also updated with up to date changes found here (https://github.com/libretro/glsl-shaders/blob/master/handheld/shaders/color/psp-color.glsl)

### DIFF
--- a/assets/shaders/psp_color.fsh
+++ b/assets/shaders/psp_color.fsh
@@ -1,0 +1,58 @@
+/*
+	psp_color vertex shader
+	
+	Original code written by hunterk, modified by Pokefan531 and
+	released into the public domain
+	
+	'Ported' (i.e. copy/paste) to PPSSPP format by jdgleaver
+	
+	This program is free software; you can redistribute it and/or modify it
+	under the terms of the GNU General Public License as published by the Free
+	Software Foundation; either version 2 of the License, or (at your option)
+	any later version.
+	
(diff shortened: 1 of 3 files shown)


RokkumanX:
Hahaha damn, that sure was quick :)
Thanks a lot!

I've been looking for this for ages and I was thrilled that I could add this to the standalone PPSSPP, now if I only could find out how to add motionblur/response time and I would die a happy man.

hrydgard:
For simulating LCD response time we'd need to keep around at least the last frame's image. Will need a bit of plumbing.

jdgleaver:
I don't know if it helps at all, but this is how I do it (purely software based) in the mGBA RA core: https://github.com/libretro/mgba/blob/683ba98787e05c1016570f0c5278127c654fbb0a/src/platform/libretro/libretro.c#L320-L771. It's not particularly fast, but it's a fairly easy sort of thing to plug in.

(A hardware/shader based approach is of course far better for performance...)

unknownbrackets:
Shader post processing now supports basic inter-frame blending, but only going one frame back.  It's possible to give more frames backward, mostly with a cost to VRAM, but that's not implemented yet.  If anyone wants to try to model the PSP's LCD response (which is likely different between PSP-1000, PSP-2000, and PSP-3000), let me know what samples are needed.

-[Unknown]

PR #13497, opened 2020-10-01
Move native util to common

hrydgard:
Move a bunch of stuff to Common/Data and subdirectories.


hrydgard:
Looks like some kind of link order problem. Bleh.

hrydgard:
Too many build systems :/

PR #13505, opened 2020-10-04
Special hack of Dissidia Final Fantasy Chinese patch

sum2012:
#13204
No need turn off fast memory now.
Also it seems our sceKernelFindModuleByName is wrong

### DIFF
--- a/Core/HLE/sceKernelModule.cpp
+++ b/Core/HLE/sceKernelModule.cpp
@@ -21,6 +21,7 @@
 
 #include "zlib.h"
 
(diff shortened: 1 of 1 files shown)


sum2012:
I redo pull request after clean up

hrydgard:
Well, this is very small so will be possible to merge anyway.

I have an objection though - we should make it a compat.ini setting instead. I've decided that all game specific hacks/changes must go through that mechanism so they don't spread uncontrollably throughout the codebase.

PR #13507, opened 2020-10-04
Savedata:Improves  the performances according to some tests.

shenweip:
This mainly does the things as follow:

1. Improves the deleting functions.
2. Supports empty file name.
3. Corrects some performances when the data is broken.

Should help #13365, #7685.


hrydgard:
This is really nice, and the test is well written too.

I'm gonna go ahead and merge - this will not be affected by my flood of cleanup commits, it doesn't touch this area.

PR #13512, opened 2020-10-05
Add compat.ini setting to work around memstick size math problem in Harry Potter - Goblet of Fire

hrydgard:
Fixes issue #13266.

Should we raise the "normal" cap?


Panderner:
How about Street Fighter Alpha 3 Max? (#10462)

hrydgard:
Right, I'll add that one too.

Issue #13289, opened 2020-08-15
Lego indiana jones 2 network issue


Lemoncak3:
I tried the ad hoc network in this game with my brother, in channel 1 and 11 i can see my brother's name but after confirm then pop up notif box saying there is network error connection.
![Screenshot_20200816-040959](https://user-images.githubusercontent.com/67026697/90322131-9494b800-df7a-11ea-9d65-ac2587fbdeb6.png)
![Screenshot_20200816-041016](https://user-images.githubusercontent.com/67026697/90322133-978fa880-df7a-11ea-9c8a-03b511e4c181.png)
Sorry i can't provide any logs since i am using android phone..

ANR2ME:
@Lemoncak3 have you tried this game again with the latest dev build at https://buildbot.orphis.net/ppsspp/ ?

Lemoncak3:
Sorry bit busy will try it later. Thanks a lot :) @hrydgard and @ANR2ME

Lemoncak3:
I am using this build v1.10.3-634-g9814f0ed6 and after network initialized the in the next screen it shows this error. 
![Screenshot_20200912-031716](https://user-images.githubusercontent.com/67026697/92969248-caa85780-f4a6-11ea-8574-f6ca470642b0.png)

Lemoncak3:
Still same error 
![Screenshot_20200912-033455](https://user-images.githubusercontent.com/67026697/92971541-d6961880-f4aa-11ea-9c94-f89d03152592.png)
![Screenshot_20200912-033147](https://user-images.githubusercontent.com/67026697/92971546-dac23600-f4aa-11ea-8de2-a2aae1a7c6d9.png)
try with all wlan channel still failed.
And after the adhoc server exit PPSSPP before client does, client's PPSSPP got crashed.

ANR2ME:
That socket error 32 will shows up on clients if Adhoc Server already shutdown, it's normal, it's an error if a client tries to send something to server while the server already closed the socket.

Btw, what you mean by crash? was it forced closing PPSSPP or just showing red message in game?

Lemoncak3:
But after that error 32 PPSSPP will force closed on it's own. Well after network initialized thing we tried to connect but failed but it seems we already connected so when i am as the adhoc server closed the game before the client, the client's PPSSPP will force close on it's own. The same thing happen to dynasty strikeforce

ANR2ME:
On windows, it worked without any problem
[...]

PR #13519, opened 2020-10-07
Adhoc Fix - various fix

anr2me:
Hopefully each commits can be compiled without error, because it was a pile of uncommitted changes where one unrelated stuff can get mixed up due to sharing the same variable.

Here are the main changes:

1. Removal of hardcoded workaround that was used to fix the stuck MessageBox on Naruto Shippuden Ultimate Ninja Heroes 3 and .hack//Link where the workaround have a side effect of 0.5 second lag when creating/connecting/joining a group.
2. Implement Adhocctl (and Apctl) state change timings properly to fix the stuck MessageBox on Naruto SUNH 3 and .hack//Link, which apparently also fix the joining issue on LEGO Indiana Jones 2 https://github.com/hrydgard/ppsspp/issues/13289
3. Making a better message transition on NetconfDialog using the new Apctl and Adhocctl timings.
4. Updated GameMode to fix Bomberman, and improved performance on BattleZone (currently only these 2 games are known to use  GameMode API)
5. Fix games developed by Falcom which have similar issue (Ys vs. Sora no Kiseki, Vantage Master Portable, and probably more since their netcode looked similar) https://github.com/hrydgard/ppsspp/issues/13493
6. An attempt to reduce stutters related to socket buffers.
7. Making the workaround that originally fix GvG Next Plus to be Optional (previously hardcoded) since this workaround affects/fix connecting issue on many games, and able to improve connectivity on high latency networks (ie. internet)
8. Games that are known to be affected by GvG workaround and have been fixed properly are:  (there should be more games but need to wait for reports from players)
- Gundam vs Gundam Next Plus
- Bleach Heat the Soul 7
- Dissidia 012
- Carnage Heart EXA
- Fate Unlimited Codes
9. Added socom.cc as an alternative Public Adhoc Server, due to similarity to myneighborsushicat.com (ie. have statistic web page)
~~10. Added an Optional workaround to slowdown FPS on games that runs too fast in multiplayer (ie. Fate Unlimited Codes, Bleach Heat the Soul 7) https://github.com/hrydgard/ppsspp/issues/13452~~

**Test builds:**
Win32&64: https://www.dropbox.com/s/i6699cyf1tzu29b/PPSSPP_1.10.3-adhocfix_Win32x64.zip?dl=0
Android(ARMv7): https://www.dropbox.com/s/dmz6jvesmm3uwp2/PPSSPP_1.10.3-adhocfix_ARMv7.apk?dl=0

### DIFF
--- a/Core/Config.cpp
+++ b/Core/Config.cpp
@@ -974,8 +974,9 @@ static ConfigSetting networkSettings[] = {
 	ConfigSetting("PortOffset", &g_Config.iPortOffset, 0, true, true),
 	ConfigSetting("MinTimeout", &g_Config.iMinTimeout, 1, true, true),
 	ConfigSetting("TCPNoDelay", &g_Config.bTCPNoDelay, false, true, true),
+	ConfigSetting("ForcedFirstConnect", &g_Config.bForcedFirstConnect, false, true, true),
 	ConfigSetting("EnableUPnP", &g_Config.bEnableUPnP, false, true, true),
-	ConfigSetting("UPnPUseOriginalPort", &g_Config.bUPnPUseOriginalPort, true, true, true),
+	ConfigSetting("UPnPUseOriginalPort", &g_Config.bUPnPUseOriginalPort, false, true, true),
 
 	ConfigSetting("EnableNetworkChat", &g_Config.bEnableNetworkChat, false, true, true),
 	ConfigSetting("ChatButtonPosition",&g_Config.iChatButtonPosition,BOTTOM_LEFT,true,true),
--- a/Core/Config.h
+++ b/Core/Config.h
@@ -422,6 +422,7 @@ public:
 	bool bTCPNoDelay;
 	bool bEnableUPnP;
 	bool bUPnPUseOriginalPort;
+	bool bForcedFirstConnect;
 	int iPortOffset;
 	int iMinTimeout;
 	int iWlanAdhocChannel;
--- a/Core/Dialog/PSPNetconfDialog.cpp
+++ b/Core/Dialog/PSPNetconfDialog.cpp
@@ -44,9 +44,8 @@
 static const float FONT_SCALE = 0.65f;
 
 // Needs testing.
-const static int NET_INIT_DELAY_US = 300000;
-const static int NET_SHUTDOWN_DELAY_US = 260000;
-const static int NET_RUNNING_DELAY_US = 1000000; // KHBBS is showing adhoc dialog for about 3-4 seconds, but feels too long, so we're faking it to 1 sec instead to let players read the text
+const static int NET_INIT_DELAY_US = 200000; 
+const static int NET_SHUTDOWN_DELAY_US = 200000; 
 const static int NET_CONNECT_TIMEOUT = 15000000; // Using 15 secs to match the timeout on Adhoc Server side (SERVER_USER_TIMEOUT)
 
 struct ScanInfos {
@@ -306,27 +305,24 @@ int PSPNetconfDialog::Update(int animSpeed) {
 			}
 			DisplayButtons(DS_BUTTON_CANCEL, di->T("Cancel"));
 
-			// The Netconf dialog stays visible until the network reaches
-			// the state PSP_NET_APCTL_STATE_GOT_IP.			
+			// The Netconf dialog stays visible until the network reaches the state PSP_NET_APCTL_STATE_GOT_IP.			
 			if (state == PSP_NET_APCTL_STATE_GOT_IP) {
 				if (pendingStatus != SCE_UTILITY_STATUS_FINISHED) {
-					ChangeStatus(SCE_UTILITY_STATUS_FINISHED, NET_RUNNING_DELAY_US);
-				}
-				else if (GetStatus() == SCE_UTILITY_STATUS_FINISHED) {
-					// We are done!
 					StartFade(false);
+					ChangeStatus(SCE_UTILITY_STATUS_FINISHED, NET_SHUTDOWN_DELAY_US);
 				}
 			}
 
-			else if (state == PSP_NET_APCTL_STATE_GETTING_IP) {
+			else if (state == PSP_NET_APCTL_STATE_JOINING) {
 				// Switch to the next message
 				StartFade(true);
 			}
 
 			else if (state == PSP_NET_APCTL_STATE_DISCONNECTED) {
-				// When connecting with infrastructure, simulate a connection
-				// using the first network configuration entry.
-				connResult = sceNetApctlConnect(1);
+				// When connecting with infrastructure, simulate a connection using the first network configuration entry.
+				if (connResult < 0) {
+					connResult = sceNetApctlConnect(1);
+				}
 			}
 		}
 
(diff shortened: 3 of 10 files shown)


hrydgard:
So the bSlowdownNetworkFPS one, since it's apparently only needed in one game, do you think it might be a candidate for a compat.ini flag instead of an obscure .ini option_

Pik-a-boo:
> So the bSlowdownNetworkFPS one, since it's apparently only needed in one game, do you think it might be a candidate for a compat.ini flag instead of an obscure .ini option?

It's needed for more then 2 games even because a Burnout game also got that issue in addition.
There could be even more games like that.

ANR2ME:
> 
> 
> So the bSlowdownNetworkFPS one, since it's apparently only needed in one game, do you think it might be a candidate for a compat.ini flag instead of an obscure .ini option?

Well it's just a temporary workaround until a proper fix can be found, so it will most likely to be removed in the future, as it could have side effect of slowing down FPS in the wrong scene (which already runs at normal speed)

Based on the timestamp on the log when using JPCSP + prx files the AdhocPollSocket function returns within 1ms most of the time, so making it 7.5ms all the time is kinda absurd, amongst the 5 functions that were used during battle:
```
sceNetAdhocGetPtpStat(09fbf9f8, 09713900)
sceNetAdhocPollSocket(09fbfa0c, 2, 0, 1)
sceNetAdhocPtpAccept(1, [09fbfa54]=00:00:00:00:00:00, [09fbfa5a]=0, 0, 1)
sceNetAdhocPtpRecv(2,08bbb520,09fbfa5c,0,1)
sceNetAdhocPtpSend(2,08bbc060,09fbfa5c,0,1)
```
PtpSend have the most delay but varies (1~10ms) and i couldn't tell why it was varies, and GetPtpStat didn't get logged when using prx files so it could also have more than 1ms delay, but i couldn't get a stable 30 FPS by increasing the delay on these functions, while it was pretty stable on AdhocPollSocket function (only the characters intro scenes have the side effect of slower than 30 FPS)

PS: i can only use prx files as reference because i couldn't benchmark these functions on real hardware.

ANR2ME:
> 
> 
> > So the bSlowdownNetworkFPS one, since it's apparently only needed in one game, do you think it might be a candidate for a compat.ini flag instead of an obscure .ini option?
> 
> It's needed for more then 2 games even because a Burnout game also got that issue in addition.
> There could be even more games like that.

i've updated the first post with test builds, have you test the workaround on Burnout games? does it gets a stable 30 FPS on that game?

Pik-a-boo:
> > > So the bSlowdownNetworkFPS one, since it's apparently only needed in one game, do you think it might be a candidate for a compat.ini flag instead of an obscure .ini option?
> > 
> > 
> > It's needed for more then 2 games even because a Burnout game also got that issue in addition.
> > There could be even more games like that.
> 
> i've updated the first post with test builds, have you test the workaround on Burnout games? does it gets a stable 30 FPS on that game?

Didn't try specifically with that  but downclocking the emulated CPU might fix these speed-up issues anyway (as it fixed 1 game already).

ANR2ME:
I'll remove the SlowdownNetworkFPS workaround as it's pretty useless without Force real clock sync :( it barely have any effect and the result are too varies on different hardware configuration.

It will probably takes a much longer time to investigate & fix the too fast FPS issue during multiplayer. Actually, i couldn't even tell whether it was related to networking code or not, because all those 5 networking functions that were used during battle doesn't seems to have any kind of delaying mechanism on JPCSP HLE, buat able to run at normal speed 30 fps.

ANR2ME:
> 
> 
> > > > So the bSlowdownNetworkFPS one, since it's apparently only needed in one game, do you think it might be a candidate for a compat.ini flag instead of an obscure .ini option?
> > > 
> > > 
> > > It's needed for more then 2 games even because a Burnout game also got that issue in addition.
> > > There could be even more games like that.
> > 
> > 
> > i've updated the first post with test builds, have you test the workaround on Burnout games? does it gets a stable 30 FPS on that game?
> 
> Didn't try specifically with that but downclocking the emulated CPU might temporarily fix these speed-up issues anyway (as it fixed 1 game already).

Does down clocking the CPU to 55 mhz gives you a stable 30 fps even during lobby or menu? 
if it's it might be worth investigated if the game actually try to change the clock during the transition between 30 & 60 fps.

Pik-a-boo:
The transition between menus was 60 fps for some reason when I downclocked.
But yeah it seems like 30 fps on menus and ingame.
And without sound issues as well.

somepunkid:
Got it set to 222 currently. In same machine/2 instances/MP matches, I have stable sound and 30fps. Well then.

ANR2ME:
> 
> 
> Got it set to 222 currently. In same machine/2 instances/MP matches, I have stable sound and 30fps. Well then.

So basically Fate Unlimited Codes only need a fixed 222 mhz clock? After investigating it, this game doesn't seems to use any function that can change CPU clock/frequency, so may be we can just use compat.ini for Fate Unlimited Codes and Bleach Heat the Soul 7 to set a fixed clock

Pik-a-boo:
Seems like downclocking is different between hardware too though.
Although I had force real clock enabled here.

somepunkid:
Yeah I just tried Bleach HTS7 with the same settings and was still getting 60fps. Seems like an issue.

ANR2ME:
> 
> 
> Yeah I just tried Bleach HTS7 with the same settings and was still getting 60fps. Seems like an issue. Tried Fate again with and without clock sync/222 and got 40fps, just total inconsistency across the board. Somethings missing.

You need to disable the SlowdownNetworkFPS when using fixed clock, that things is not useful because the result too varies on different hardware.

Anyway, the fix for the FPS issue on Fate Unlimited Codes and Bleach Heat the Soul 7 will be made on a different PR as it will probably takes a really long time to have a proper fix, since it even hard to make a consistent result like this LOL.

hrydgard:
OK, so you consider this PR finished? Almost wanna break it up like your previous massive multi-PR but not sure I can be bothered :)

ANR2ME:
Yes, it's finished now :)
I tried to group related changes into a single commit to prevent getting compile errors, but didn't actually tried to compile each commits tho.

ANR2ME:
Okay, it's done, each commits can be compiled now.

Pik-a-boo:
Didn't you said you gonna remove the workaround?

hrydgard:
I'll update the gradle thing :)

hrydgard:
Thanks @ANR2ME , gonna go ahead and merge.

Shawnjius:
@ANR2ME I did some more testing with Ys vs. and Vantage Master, and both worked with no issue between my PC and other players who are also on PC. All of that was with WLAN, Ad-Hoc, and upnp enabled; using myneighborsushicat.com. PC to PC works great.

However, I ran into issues with the Android apk. Although I can connect to others, I'll get a wi-fi disconnected error after playing for a few minutes. I also seem to have a problem with the Android apk consistently crashing on me. I have no clue why, as it always happens under different circumstances. It might just be a "me" problem, though. I'm running it on a Galaxy S10.

Another thing I can't figure out, is that while one friend was able to connect multiple instances locally, I can't seem to make the same thing happen on my end, or set up a match between my phone and my PC. This was using 127.0.0.1 and 127.0.0.2 as the proAdhocServer addresses respectively. When I attempted to connect two different instances on PC, one would always get an 80410b10 error.

Anyway, just wanted to share my findings.

ANR2ME:
That error 80410b10 is ERROR_NET_ADHOCCTL_BUSY, usually only occurred within an instance if it tried to access Adhocctl again while previous access is still in progress (ie. still waiting for reply from AdhocServer), but shouldn't have anything to with other instances.

Anyway, we should probably discuss it at https://github.com/hrydgard/ppsspp/issues/13493 instead of here.

Issue #13510, opened 2020-10-05
Some character fail to display (Gundam vs Gundam Next Plus)


leoxxx:
I convert sarasa-mono-slab-sc-regular.ttf to jpn0.pgf, then replace the default  jpn0.pgf(Microsoft YaHei Regular) in ppsspp.
It can be displayed. 

replace before
![QQ截图20201005183142](https://user-images.githubusercontent.com/5364709/95086865-90ec0900-0753-11eb-9656-3b31b9824cbc.png)
![QQ截图20201005183244](https://user-images.githubusercontent.com/5364709/95086922-a2351580-0753-11eb-9237-2f18e3d7808b.png)
![QQ截图20201005183330](https://user-images.githubusercontent.com/5364709/95086934-a7926000-0753-11eb-84fc-23002a498139.png)
![QQ截图20201005183408](https://user-images.githubusercontent.com/5364709/95086947-abbe7d80-0753-11eb-8472-39e9c40c0dbf.png)
![QQ截图20201005183520](https://user-images.githubusercontent.com/5364709/95086965-b11bc800-0753-11eb-8af0-733a78cfa030.png)

replace after
![QQ截图20201005212635](https://user-images.githubusercontent.com/5364709/95087025-c264d480-0753-11eb-8c68-5cdac3fccd33.png)
![QQ截图20201005212725](https://user-images.githubusercontent.com/5364709/95087034-c55fc500-0753-11eb-8eda-df28e7c73dff.png)
![QQ截图20201005212800](https://user-images.githubusercontent.com/5364709/95087042-c7c21f00-0753-11eb-94d4-6febe491eb10.png)
![QQ截图20201005212830](https://user-images.githubusercontent.com/5364709/95087052-ca247900-0753-11eb-8ab2-23fda353bccd.png)
![QQ截图20201005212908](https://user-images.githubusercontent.com/5364709/95087065-cd1f6980-0753-11eb-9cba-513634aaa60e.png)

Here is the font file. 
[jpn0.zip](https://github.com/hrydgard/ppsspp/files/5327744/jpn0.zip)

hrydgard:
did you mean to post screenshot?

leoxxx:
Yes. Have some network error. The pictures and font upload later.

Panderner:
@leoxxx what game are you using?

leoxxx:
@Panderner
You can see the game in the picture title.

leoxxx:
[...]

PR #13522, opened 2020-10-09
Change jpn0.pgf for a new one generated by nassau-tk

hrydgard:
See #13510 .


hrydgard:
Just gonna merge, let's see how this goes.

PR #13523, opened 2020-10-09
Workaround for ULJS00218 until a more correct VFPU sin/cos implementation can be found.

hrydgard:
See issue #12900 for the broken game and rationale.

Basically we need a more accurate version of this function. This old one is probably actually slightly higher precision than what we're using now in other games.

JIT only for now.

### DIFF
--- a/Core/Compatibility.cpp
+++ b/Core/Compatibility.cpp
@@ -71,6 +71,7 @@ void Compatibility::CheckSettings(IniFile &iniFile, const std::string &gameID) {
 	CheckSetting(iniFile, gameID, "ForceSoftwareRenderer", &flags_.ForceSoftwareRenderer);
 	CheckSetting(iniFile, gameID, "DarkStalkersPresentHack", &flags_.DarkStalkersPresentHack);
 	CheckSetting(iniFile, gameID, "ReportSmallMemstick", &flags_.ReportSmallMemstick);
+	CheckSetting(iniFile, gameID, "UseOldSinCos", &flags_.UseOldSinCos);
 }
 
 void Compatibility::CheckSetting(IniFile &iniFile, const std::string &gameID, const char *option, bool *flag) {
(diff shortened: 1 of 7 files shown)


somepunkid:
Tested the changes, this workaround doesn't work. It is reversed just like before with this set of changes in #13523

From earlier

I have isolated the first issue to two changes listed in

master...unknownbrackets:sincos-log

Under file: Core/MIPS/MIPSVFPUUtils.h

image

Specifically reverting just lines lines 41 and and 54 of this file, the game stops reversing/has normal behavior/physics and effects don't reverse. It should be noted that reverting just one or the other does not resolve the behavior, only both. I tested each instance. This is to isolate the bug that was originally submitted. Tested with build ppsspp-v1.10.3-698-gccea07ab9-windows-amd64.

I'm unsure if any of the other modifications of Core/MIPS/MIPSVFPUUtils.h are necessary/have an effect. Just modifying these two led to visibly normal behavior and play. I've yet to find another issue, besides the secondary listed issue which persists, so this may be related to something else.

I noticed that the changes I made in the above thread were not implemented (for Core/MIPS/MIPSVFPUUtils.h ) in this PR. Perhaps it's this? I tested revering each in master...unknownbrackets:sincos-log and the only two that made a difference were just reverting 41 and 54. When I did my local tests, that's all I changed! specifically back to - float checkAngle = angle - floorf(angle * 0.25f) * 4.f;

ANR2ME:
May not be related, just some possible condition on floating point:
https://www.appinf.com/download/FPIssues.pdf
https://floating-point-gui.de/errors/comparison/

Btw, you guys aren't making use Epsilon for these  `angle -= floorf(angle * 0.25f) * 4.f` kind of thing?

hrydgard:
#13526 seems like a more promising direction, closing.

Issue #13464, opened 2020-09-22
[Android] [Vulkan] Constant FPS drop in Final Fantasy Type-0


kage2051:
When using 2x or higher rendering resolution and Vulkan backend, I get constant FPS drop in the game mentioned above, even with block transfer effects set to off.

On GL though, I can run it decently as long as I use any resolution not higher than 3x.

Device tested:
Redmi Note 6 Pro (SD636, 4GB RAM)
Android 9
Latest development build of PPSSPP. Tried going back to stable but it just makes the issue even worse.

Other settings shown below:
![Screenshot_2020-09-21-21-17-55-806_org ppsspp ppsspp](https://user-images.githubusercontent.com/25725203/93916809-52703c00-fcc7-11ea-8421-fe34df5391a2.jpg)

Panderner:
@kage2051 Frame rate comparison for both backends?

kage2051:
OpenGL, 3x rendering resolution, no texture scaling:
![Screenshot_2020-09-22-12-28-01-209_org ppsspp ppsspp](https://user-images.githubusercontent.com/25725203/93922606-628c1980-fccf-11ea-8875-a64105b8ed20.jpg)

Vulkan, 3x rendering resolution, no texture scaling:
![Screenshot_2020-09-22-12-29-04-712_org ppsspp ppsspp](https://user-images.githubusercontent.com/25725203/93922665-7c2d6100-fccf-11ea-880c-44693b306915.jpg)

As you can see, the framerate difference is brutal in that game.

hrydgard:
That is quite surprising but an interesting find. Needs investigation.

hrydgard:
Does by any chance the new option "clear framebuffer on first use" make any measurable difference here?

kage2051:
> Does by any chance the new option "clear framebuffer on first use" make any measurable difference here?

That option was checked during the tests to make sure.

kage2051:
Another thing that is worth mentioning here: sometimes when Ace (one of the characters) performs a one-hit KO with his card attack, framerate drops to 23. Also that some cutscenes tend to suffer from frame drops as well.

All of them reproduced on the same device.

MojoJojoDojo:
This game FPS is horrible on PC as well using the Vulkan renderer and an AMD Gpu

hrydgard:
Is this all through the game? Got a savegame at some particularly bad spot?

MojoJojoDojo:
[...]

PR #13530, opened 2020-10-10
Create a safer variant of BlockTransferAllowCreateFB, enable for FF Type 0

hrydgard:
Should give it a considerable boost by avoiding readbacks, see #13464 .

New compat setting is called IntraVRAMBlockTransferAllowCreateFB.

This one only allows intra-VRAM block transfers to happen as copies on-GPU, such as the one this game uses for bloom. Not readbacks to RAM. Good chances to avoid missing stuff that actually requires real readback, if there's any.


Panderner:
Screenshot_2020-10-11-09-00-55-02_2f85358b2198d26f8aca533d68bee793
After this PR it did not improves the performance for this game on lower-end devices. I'm using Realme C2 in OpenGL

Pik-a-boo:
Didn't see any improvement either , unless I didn't have it enabled or something.

hrydgard:
Okay, seems the readback I got rid of was mostly affecting cutscenes.

I have some more ideas of what might be slowing this down, stay tuned.

kage2051:
And the game itself crashes upon loading on the latest build from today. Git 894 doesn't suffer from such issue.

hrydgard:
Oh yeah, I think I know what that is. I'll fix it.

Issue #13501, opened 2020-10-02
something's wrong with network


LuisFellp:
### What happens?
when i try to play mgs:pw multilayer, using hotspot, it doesn't work. it gives a error. 
on buildbot version (1.10.3-784) it gives a red text
![Screenshot_20201002-085700_PPSSPP](https://user-images.githubusercontent.com/33043774/94926137-4d14ad80-048e-11eb-9173-07dac5b96a55.jpg)
i don't know the exact version were it started, but since 1.10.3 from play store, it doesn't work anymore 

### What *should* happen?
work

### What hardware, operating system, and PPSSPP version?  On desktop, GPU matters for graphical issues.

Exynos 7 Octa (7904), Android 10, 1.10.3-784

Pik-a-boo:
You are using Android which can't use port offsets below 1024.
So this issue is not vaild.
Make sure its above it.

LuisFellp:
@Pik-a-boo how i change it?

LuisFellp:
nvm
i found it in seettings.ini, changed to 1025 and now is giving this error
![Screenshot_20201002-100232_PPSSPP](https://user-images.githubusercontent.com/33043774/94931935-61f53f00-0496-11eb-83fd-b2045c00256e.jpg)

ANR2ME:
You need to use the same networking settings with the one that worked the last time for you, may i know which PPSSPP version was the last time working for you for this game?

Pik-a-boo:
MGS PW works good even now on the latest PPSSPP.
Could be an Android only issue because I only tried on PC.

LuisFellp:
@ANR2ME on 1.9.4 it's working fine. used same settings that i used on 1.10.3 and 1.10.3-784
![Screenshot_20201003-165619_PPSSPP](https://user-images.githubusercontent.com/33043774/95001611-84bc4c00-0599-11eb-8e11-51f786b6d99d.jpg)

ANR2ME:
Hmm.. that is strange. 
It will probably takes time before i can test android since my phone is practically dead now, i will probably have a new phone within a week or two (waiting for a discounted price on Redmi 9 or Realme C12)

For the mean time it would be nice if you can pinpoint the last APK that worked for you, you can download them from the full list here https://buildbot.orphis.net/ppsspp/index.php?m=fulllist
[...]

PR #13532, opened 2020-10-11
Adhoc fix - Fix disconnection issue due to unexpected timeout

anr2me:
I didn't expect for `CoreTiming::GetGlobalTimeUsScaled()` to return smaller number than previous one, so this was unexpected.
Hopefully this fix disconnection issue on games that use AdhocMatching, like MGS:PW https://github.com/hrydgard/ppsspp/issues/13501


hrydgard:
Huh, that does sound suspicious. Might have found a core timing bug :P 

I'm gonna file an issue to investigate, in the meantime this workaround should be fine.

PR #13548, opened 2020-10-14
Libretro: add SYSPROP_DISPLAY_SAFE_INSET* properties. fixes screen size.

aliaspider:


hrydgard:
Good catch!

libretro/libretro.cpp
```diff
@@ -888,11 +888,16 @@ float System_GetPropertyFloat(SystemProperty prop)
...
+      case SYSPROP_DISPLAY_SAFE_INSET_RIGHT:
+      case SYSPROP_DISPLAY_SAFE_INSET_TOP:
+      case SYSPROP_DISPLAY_SAFE_INSET_BOTTOM:
+         return 0.0f;
       default:
          break;
    }
 
-   return -1;
+   return 0.0f;
```
hrydgard, line 900: I think we should stick to -1 here for now.

aliaspider: allright.

PR #13549, opened 2020-10-15
Fix "naturalA" - Natural Colors (no blur) - postprocessing shader

Leopard20:
This fixes the shaking issue on Vulkan and OpenGL in: https://github.com/hrydgard/ppsspp/issues/13374
The issue with D3D9 still remains (it is not related to the PP shaders)

### DIFF
--- a/assets/shaders/defaultshaders.ini
+++ b/assets/shaders/defaultshaders.ini
@@ -17,7 +17,7 @@ Vertex=natural.vsh
 [NaturalA]
 Name=Natural Colors (No Blur)
 Fragment=naturalA.fsh
-Vertex=natural.vsh
+Vertex=naturalA.vsh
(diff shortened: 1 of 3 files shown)


hrydgard:
Thanks. Not entirely sure why this actually stops it from shaking, but the change is simple enough. I'm going to have to investigate this in more detail later.

leopard20:
It was related to:
```
uniform vec2 u_texelDelta;
```
I think it was trying to interpolate the values between neighboring texels or something when it did this: (it interpolated between 4 vertex positions to smooth the edges):
``` 
v_texcoord0=a_texcoord0.xyxy+vec4(-0.5,-0.5,-1.5,-1.5)*u_texelDelta.xyxy;
```
but no clue why it is shimmering.

PR #13551, opened 2020-10-17
Shader generation: Start sharing some parts between backends

hrydgard:
I plan to add some new capabilities to shader blend replacement, to fix some issues with masking for example (I think I'll be able to solve Outrun's car reflection color, for example). To be able to only (mostly) do these changes in one place, I here split out that part from all the shader generators so it can be shared between them.

This requires injecting a few compatibility defines and renaming some uniforms, but I think it'll be worth it.


hrydgard:
Got a bit scared of breakage here, so made a brute force shader generator verifier (D3D11/Vk only since they have offline compilers that don't need a device) and added to the unit test project. It found some bugs and might find more once I add the various GLSL and D3D9 paths...

PR #13562, opened 2020-10-19
PsmfPlayer:The status will not be changed until the delay result is finished.

shenweip:
These funcs are done in a thread blocking way, and they change status and end delay after completing the func. Before, the status was changed too early. Should helps #11975.


hrydgard:
Nice! Don't forget to handle the event properly in DoState, see for example __GeDoState.

PR #13574, opened 2020-10-22
Fix a section of the translatable message

lightfinger:
I don't know this is OK. Please check it.


hrydgard:
Yeah, I agree that this string doesn't belong in dev.

PR #13579, opened 2020-10-23
Adhoc fix - Updating Adhocctl behavior after doing some small test

anr2me:
- Fixed invalid group mode (fixes unable to see any room issue on Air Conflicts - Aces Of World War 2, should also fix other games with similar issue)
- Fixed possible memory and port leaks when loading SaveState in the middle of multiplayer.
- Fixed Socket Alerts implementation.
- Implementing Adhocctl Busy state, and Updated Adhocctl behavior.
- Fixed Pocket Pool.
- Fixed Star Wars: The Force Unleashed
- Fixed possible desync issue on GameMode multiplayer games when mission is being started (due to some initial data didn't get received sometimes).

### DIFF
--- a/Core/HLE/proAdhoc.cpp
+++ b/Core/HLE/proAdhoc.cpp
@@ -1405,6 +1405,12 @@ int friendFinder(){
 				}
 			}
 
+			// Calculate EnterGameMode Timeout to prevent waiting forever for disconnected players
+			if (isAdhocctlBusy && adhocctlState == ADHOCCTL_STATE_DISCONNECTED && adhocctlCurrentMode == ADHOCCTL_MODE_GAMEMODE && netAdhocGameModeEntered && static_cast<s64>(now - adhocctlStartTime) > netAdhocEnterGameModeTimeout) {
+				netAdhocGameModeEntered = false;
+				notifyAdhocctlHandlers(ADHOCCTL_EVENT_ERROR, ERROR_NET_ADHOC_TIMEOUT);
+			}
+
 			// Handle Packets
 			if (rxpos > 0) {
 				// BSSID Packet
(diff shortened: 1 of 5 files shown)


hrydgard (approved):

somepunkid:
Reporting a major performance regression in Adhoc on v1.10.3-1044-ga4a0b4ba0
Game -ULJS00218 - Hitman Reborn Battle Arena 2

Major performance differences between this build and the previous release build v1.10.3-1026-g36a614072. Fluctuates between 50-55 frames in menu and games, randomly dipping higher and lower. Tested on multiple previously stable systems/configs. Sometimes locks to 60 and then dips between rounds/in game. Just in general, everything is slower/chunkier in adhoc mode only now. This game still requires the "force real clock sync" option to lock to 60 in adhoc play. Otherwise, no other default setting was changed besides network settings.

 Let me know if anything is needed.

ANR2ME:
I'll check it out later. 
Are both the 60 FPS and 50-ish FPS uses the same "force real clock sync" setting ? because that setting could cause FPS drops.

Btw, is this game have english patch?

somepunkid:
> I'll check it out later.
> Are both the 60 FPS and 50-ish FPS uses the same "force real clock sync" setting ? because that setting could cause FPS drops.
> 
> Btw, is this game have english patch?

There is no english patch, sadly. Option 3 on the main menu leads to 1p vs 2p(adhoc), which should be easy to navigate from there.

They are both using "Force real clock sync"

If this option is not selected, this game defaults to slow framerates as soon as the fight starts, it will cap at 30fps. Force clock sync allows it to run correctly/at 60fps. This game has only worked with this option enabled.

If I disabled force clock sync, as soon as a player is in a match, gameplay slows to a crawl.

ANR2ME:
Ugh Hitman Reborn Battle Arena is using the smallest timeout i have ever seen so far :( and it's abit tricky to fix it.
This game only use 2 SCNET function during battle, `sceNetAdhocPdpSend` and `sceNetAdhocPdpRecv` both with blocking mode.

One of the problem was what happen to the PdpRecv, it using 50 usec timeout, and when there are no data available to be received it enters blocking simulation and the timeout were overriden by the minimum timeout workaround (which is defaulted to 1 ms or 1000 usec) so it's taking too much time to hit the timeout.
I tried changing the minimum timeout to 0 but based on this test, there are times where the interval goes nearly 1ms between each reScheduled Event :(
image

PS: when the game runs 60 FPS each PdpRecv calls never (or very rare) enter blocking simulation because for some reason there is always data ready to be received (might be just by luck that both side were in sync when sending and receiving data?), so it can receive and return immediately.

somepunkid:
Created this issue. Once again, this game worked almost perfectly before this recent change, even with Force Clock Sync. Quite interesting it would be impacted this much! Reference https://github.com/hrydgard/ppsspp/issues/13605

PR #13580, opened 2020-10-23
Merge GLSL vertex shader generators

hrydgard:
Similar to #13576, but for the vertex shader generators.

When it comes to code, less is more, because shorter code with less duplication is easier to change. This deletes the Vulkan-specific vertex shader generator, modifying the regular GLSL one to output Vulkan dialect GLSL in addition to OpenGL dialect GLSL.


hrydgard:
Converted to draft because I realize I haven't checked HW tesselation.

PR #13586, opened 2020-10-26
Update jpn0.pgf

nassau-tk:
I fixed manually all of "JIS Kanji-Level1" (2946character).
It's working like same as Real PSP jpn0.pgf font.
So,I finished work that modify for jpn0.pgf.

Sorry, I use Github Desktop firsttime.
How to use the push and the pull can't be understood right.

If there is a mistake, please tell me.


nassau-tk:
Sorry,I miss the push file.

PR #13592, opened 2020-10-27
IMP sceKernelFindModuleByUID and sceKernelFindModuleByName.

shenweip:
This should help some patched games like #13205, #13204. I think @sum2012 would be interested in this.

### DIFF
--- a/Core/HLE/sceKernelModule.cpp
+++ b/Core/HLE/sceKernelModule.cpp
@@ -2422,7 +2422,7 @@ u32 sceKernelFindModuleByUID(u32 uid)
 		ERROR_LOG(SCEMODULE, "0 = sceKernelFindModuleByUID(%d): Module Not Found or Fake", uid);
 		return 0;
 	}
-	INFO_LOG(SCEMODULE, "sceKernelFindModuleByUID(%d)", uid);
+	INFO_LOG(SCEMODULE, "%d = sceKernelFindModuleByUID(%d)", module->modulePtr, uid);
 	return module->modulePtr;
 }
 
--- a/Core/HLE/sceKernelModule.cpp
+++ b/Core/HLE/sceKernelModule.cpp
@@ -168,6 +168,7 @@ struct NativeModule {
 	char name[28];
 	u32_le status;
 	u32_le unk1;
+	u32_le modid; // 0x2C
 	u32_le usermod_thid;
 	u32_le memid;
 	u32_le mpidtext;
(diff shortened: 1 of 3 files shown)


Core/HLE/sceKernelModule.cpp
```diff
@@ -263,11 +269,23 @@ class PSPModule : public KernelObject {
...
+			NativeModule *pnm = &nm;
+			char *ptemp = temp;
+			DoArray(p, ptemp, 0xC0);
+			memcpy(pnm, ptemp, 0x2C);
+			pnm->modid = GetUID();
+			pnm += 0x30;
+			ptemp += 0x30;
+			memcpy(pnm, ptemp, 0xC0 - 0x2C);
+		}
 		Do(p, memoryBlockAddr);
```
aliaspider, line 289: why not just put 'modid' last in 'NativeModule' to keep the struct backwards compatible and avoid this hack ? or is 'NativeModule'  something the real PSP OS uses and it just had wrong offsets before ?

shenweip: NativeModule is a native struct of PSP, its members has fixed order.

sum2012:
Well done @shenweip 
jpcsp use 3 tables to make conversion
https://github.com/jpcsp/jpcsp/blob/master/src/jpcsp/HLE/kernel/managers/ModuleManager.java
But we have another method

Core/HLE/sceKernelModule.cpp
aliaspider: ok I see.
why are you increasing both `pnm`and `ptemp` by `0x30` though ? shouldn't `ptemp`be increased by `0x2C` instead ?

shenweip: Oops, thanks.

sum2012:
Well done @shenweip Fixed #13205 #13204 

I just want a small change
INFO_LOG(SCEMODULE, "sceKernelFindModuleByUID(%d)", uid); change to
INFO_LOG(SCEMODULE, "%d = sceKernelFindModuleByUID(%d)", module->modulePtr, uid);

Core/HLE/sceKernelModule.cpp
hrydgard: Turns out the math was wrong - have to cast the pointers to bytes first :)

Fixed in 2f4dfed

shenweip: Oops again, thanks:p

PR #13598, opened 2020-10-29
Fix issue in #12145

tbocek:
* fixed issue with: [libretro ERROR] [G3D] glewInit() failed 
* No need to call video_cb if there is no image

Please test first!

### DIFF
--- a/libretro/LibretroGLContext.cpp
+++ b/libretro/LibretroGLContext.cpp
@@ -16,18 +16,26 @@ bool LibretroGLContext::Init() {
 }
 
 void LibretroGLContext::CreateDrawContext() {
-	if (!glewInitDone) {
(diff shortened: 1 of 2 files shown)


hrydgard (approved):

hrydgard:
Tested enough? should I go ahead and merge?

tbocek:
My last tests were back in May, I would like to get feedback from @soredake @i30817 before merging -> (https://github.com/hrydgard/ppsspp/issues/12145)

twinaphex:
This would be very welcome if it fixes Wayland context (which previously was not working) while also still working with DRM/KMS context mode.

hrydgard:
I'm just gonna merge it, we can always revert if it causes problems.

PR #13606, opened 2020-10-31
[Adhoc] - Fix Significant FPS drops on Hitman Reborn Battle Arena 2

### LINKED ISSUE #13605: ULJS00218 - Hitman Reborn Battle Arena 2 - Adhoc performance issues after latest test build
Reporting a major performance regression in Adhoc on v1.10.3-1044-ga4a0b4ba0
Game -ULJS00218 - Hitman Reborn Battle Arena 2
VID: https://youtu.be/LzogWD5YW7Q (watch the frame counter, primarily. Left - few dips, right -lots of dips)

Performance differences between this build and the previous release build v1.10.3-1026-g36a614072. Fluctuates between 50-55 frames in menu in games, randomly dipping higher and lower. Tested on multiple previously stable systems/configs. Locks to 60 and then dips between rounds/in game. Just in general, everything is slower/chunkier in adhoc mode only now. The base game is fine outside of adhoc.

This game still requires the "force real clock sync" option to lock to 60 in adhoc play. Otherwise, no other default setting was changed besides network settings to allow connection.

This was contrasted directly to the previous build on the same machine (2 instances, same machine, see vid). This takes input buffer into consideration, so yes, there will be desyncs caused by this in the video (using a 360 stick). What should be watched is the FPS marker.

EDIT: Added vid. Tested using both instances on same machine for visible differences, set all to 1x resolution. It's possible the newer instance started running slightly better in game after first boot/use/texture loads? Either way, even with this, seems to be a really big difference in menus/screen transitions, and sometimes in the form of dropped frames in game. This was running rock solid 60 in most instances, previously, with an occasional small dip between screens.

There is no english patch, sadly. Option 3 on the main menu leads to 1p vs 2p(adhoc), which should be easy to navigate from there for testing

They are both using "Force real clock sync" in the video test, as the game won't run correctly any other way.
If this option is not selected, this game defaults to slow framerates as soon as the fight starts, it will cap at 30fps. Force clock sync allows it to run correctly/at 60fps. This game has only worked with this option enabled, in adhoc. It works both ways outside of adhoc.

If I disabled force clock sync, as soon as a player is in a match, gameplay slows to a crawl, regardless of the ppsspp version.(in adhoc)

Regardless of this, it still runs great/consistent 60fps on previous builds with force clock sync enabled, even on less capable hardware (tablets/surface tested) so this hasnt been an issue.

ADDED NOTES FROM CHANGE #13579 that changed this games performance.

ANR2ME

Ugh Hitman Reborn Battle Arena 2 is using the smallest timeout i have ever seen so far :( and it's abit tricky to fix it.
This game only use 2 SCENET function during battle, `sceNetAdhocPdpSend` and `sceNetAdhocPdpRecv` both with blocking mode.

One of the problem was what happen to the PdpRecv, it using 50 usec timeout, and when there are no data available to be received it enters blocking simulation and the timeout were overriden by the minimum timeout workaround (which is defaulted to 1 ms or 1000 usec) so it's taking too much time to hit the timeout.
I tried changing the minimum timeout to 0 but based on this test, there are times where the interval (Timestamp Delta) goes nearly 1ms between each reScheduled Event :( So there might be FPS drops during this long delays.
image

PS: when the game runs 60 FPS each PdpRecv calls never (or very rare) enter blocking simulation because for some reason there is always data ready to be received (might be just by luck that both side were in sync when sending and receiving data?), so it can receive and return immediately.

Other than this, there should be more issue during lobby/character selection, probably related to `hleEatMicro` i added in some functions.

Btw, @somepunkid could you open an new issue for this?

_Originally posted by @ANR2ME in https://github.com/hrydgard/ppsspp/pull/13579#issuecomment-719887236_

anr2me:
Fix https://github.com/hrydgard/ppsspp/issues/13605

### DIFF
--- a/Core/Config.cpp
+++ b/Core/Config.cpp
@@ -972,7 +972,7 @@ static ConfigSetting networkSettings[] = {
 	ConfigSetting("EnableAdhocServer", &g_Config.bEnableAdhocServer, false, true, true),
 	ConfigSetting("proAdhocServer", &g_Config.proAdhocServer, "myneighborsushicat.com", true, true),
 	ConfigSetting("PortOffset", &g_Config.iPortOffset, 0, true, true),
-	ConfigSetting("MinTimeout", &g_Config.iMinTimeout, 1, true, true),
+	ConfigSetting("MinTimeout", &g_Config.iMinTimeout, 0, true, true),
 	ConfigSetting("TCPNoDelay", &g_Config.bTCPNoDelay, false, true, true),
 	ConfigSetting("ForcedFirstConnect", &g_Config.bForcedFirstConnect, false, true, true),
 	ConfigSetting("EnableUPnP", &g_Config.bEnableUPnP, false, true, true),
--- a/Core/HLE/sceNetAdhoc.cpp
+++ b/Core/HLE/sceNetAdhoc.cpp
@@ -2553,7 +2553,7 @@ int sceNetAdhocctlGetPeerInfo(const char *mac, int size, u32 peerInfoAddr) {
 			// Multithreading Unlock
 			peerlock.unlock();
 		}
-		hleEatMicro(1000);
+		hleEatMicro(100);
 		return retval;
 	}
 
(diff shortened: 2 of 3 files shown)


MojoJojoDojo:
Why timeout is 0 default now?

ANR2ME:
Because that setting is a workaround, and the default should be what being used on a real PSP, which is not being overriden.

Issue #13607, opened 2020-10-31
Syphon Fillter Logan's Shadow & Resistance: Retribution Freeze upon entering multiplayer mode


MojoJojoDojo:
### What happens?
The game will freeze when you try to enter the multiplayer mode from the main menu

### What *should* happen?
Game should continue past it with no issues.

Logs :

At least on Logan's Shadow : 
`user_main    E[SCEMODULE]: HLE\sceKernelModule.cpp:2245 UNIMPL sceKernelStopUnloadSelfModuleWithStatus(00000001, 00000000, 00000000, 00000000, 00000000): game may have crashed`

https://forums.ppsspp.org/attachment.php?aid=16891

As noted on AdamN on PPSSPP forums the issue is not related to networking - it will even happen if you disable WLAN on PPSSPP.
More detailed log here : 
[ppsspplog.txt](https://github.com/hrydgard/ppsspp/files/5468875/ppsspplog.txt)
Maybe someone like @sum2012 could fix it because I always see him/her making changes related to Kernel (it seems like a kernel issue?).

MojoJojoDojo:
[Uploading ppsspplog.txt…]()
A more detailed log here.

ANR2ME:
You mentioned 2 different games but only provides 1 log file?

MojoJojoDojo:
It's the same issue between both games @ANR2ME so no point.
It's also by the same developer too.

ANR2ME:
But it could cause confusion when the dev tries to analyze the log file.
Anyway, the log file you provides doesn't seems to be coming from from Syphon Fillter Logan's Shadow.
So i'll attach the log for that game here:
[ppsspplog.txt](https://github.com/hrydgard/ppsspp/files/5468914/ppsspplog.txt)

MojoJojoDojo:
it seems to be related to issue #4626 but I dont know if it counts as a duplicate.

MojoJojoDojo:
Also the freeze on main menu also happens if you try to play the game on a PS Vita but not a PSP...

shenweip:
`47:00:002 idle0        E[SCEKERNEL]: HLE\sceKernelThread.cpp:2293 sceKernelDeleteThread(285): thread not dormant`
This is odd.

shenweip:
Please try #13608.

Pik-a-boo:
How can I download it?
Can you give a test build here?

MojoJojoDojo:
Can you give a test build here?
[...]

PR #13608, opened 2020-10-31
When stops and unloads self module, make sure that the thread is properly deleted.

shenweip:
Uses sceKernelTerminateDeleteThread instead of sceKernelDeleteThread here, or this will just return as `thread not dormant`.
Should help #13607.


hrydgard:
Interesting! Does look quite plausible.

MojoJojoDojo:
Forget to mention that this also fixes the Combat Ops Sneak Preview that Syphon Fillter Logan's Shadow got.

PR #13611, opened 2020-11-01
Merge GLSL and HLSL vertex shader generators

hrydgard:
Finally there is only one! Although with a lot of if statements and switches :)

Needs a bit more testing before merge.

I love deleting code - 500 lines gone.


hrydgard:
Seems to work now! Will test a little more on Android and then merge.

PR #13613, opened 2020-11-02
Shader writer utility, other cleanups

hrydgard:
A nicer and faster replacement for the WRITE( ) macro, that also writes some compatibility defines and stuff. Will later use this for shader depal shaders, target reinterpret shaders etc.

Redefined WRITE to use it, to avoid losing too much "blame" history in the generators.

Also some assorted cleanup.

### DIFF
--- a/GPU/Common/ColorReinterpret.cpp
+++ b/GPU/Common/ColorReinterpret.cpp
@@ -1,0 +1,49 @@
+#include <cstdarg>
+
+#include "GPU/Common/ColorReinterpret.h"
+#include "GPU/Common/ShaderWriter.h"
+
+// TODO: We could have an option to preserve any extra color precision. But gonna start without it.
+// Requires full size integer math.
+bool GenerateReinterpretFragmentShader(char *buffer, GEBufferFormat from, GEBufferFormat to, const ShaderLanguageDesc &lang) {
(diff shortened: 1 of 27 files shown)


hrydgard:
Woops, ColorReinterpret.cpp/h were not supposed to be included. Oh well, they're currently unused.

GPU/Common/ColorReinterpret.cpp
```diff
@@ -0,0 +1,49 @@
...
+	}
+	ShaderWriter writer(buffer, lang, ShaderStage::Fragment);
+
+	switch (from) {
+	case GE_FORMAT_4444:
+		writer.W("  uint color = uint(in.r * 15.99) | (uint(in.g * 15.99) << 4) | (uint(in.b * 15.99) << 8) | (uint(in.a * 15.99) << 12);\n");
+		break;
+	case GE_FORMAT_5551:
+		writer.W("  uint color = uint(in.r * 31.99) | (uint(in.g * 31.99) << 5) | (uint(in.b * 31.99) << 10);\n");
+		writer.W("  if (in.a > 128.0) color |= 0x8000;\n");
```
aliaspider, line 20: `in.a > 0.5` ?

PR #13632, opened 2020-11-06
Fix Android Studio 4.1 + Gradle Plugin 6.5

anr2me:
Not for merging, because the Timestamp usually causes a conflict on people who already upgrade their project to the latest gradle.

Just to investigate if you guys still generate an APK using the latest Android Studio that failed to be installed while it worked fine on my build https://github.com/hrydgard/ppsspp/issues/13577

Here is the test APK (ARMv7) i build based on v1.10.3-1139 https://www.dropbox.com/s/yuvjcd63gw5k03b/PPSSPP_v1.10.3-1139_ARMv7.apk?dl=0
PS: If it failed to be installed, try uninstalling PPSSPP first, as i remembered it will refuse to install if signature were different.

PPS: This was the only uncommited changes i found after upgrading

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,7 +5,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:4.0.2'
+        classpath 'com.android.tools.build:gradle:4.1.0'
     }
 }
 
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,6 +1,6 @@
-#Thu May 28 22:53:27 CEST 2020
+#Sat Nov 07 00:43:49 ICT 2020
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.1.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.5-bin.zip


hrydgard:
This is exactly the same change as https://github.com/hrydgard/ppsspp/commit/624998becd7f689b15e54873095607923886a354 which didn't work, so I don't know how this could be any different..

ANR2ME:
I see

jcchikikomori:
> Android Studio 4.1 has some really annoying bugs anyway with switching configurations, so I'm sticking to 4.0 anyway until 4.2 is out, hopefully that will work better.

Indeed. I work on other Android projects and I am also facing this especially on Windows & Linux versions

ANR2ME:
Btw, as i remembered PPSSPP Debug version was broken for quite some time (i think since Android Studio 4.0), it can't be build/compiled anymore so i can't put breakpoints anymore when debugging on Android.

I hope it will be fixed on Android Studio 4.2 too (or may be something on PPSSPP got broken?)

hrydgard:
@ANR2ME Hm? I just hit a breakpoint with no issues (except slow to load). Android studio 4.0.

ANR2ME:
Hmm.. that is odd, i remembered unable to debug it for a few months, looks like i'll need to install v4.0 again

ANR2ME:
Hmm.. after investigating the error.
I was having error:
```
Failed to transform com.bda.controller.jar to match attributes {...} Error while dexing.
```
Looks like adding `compileOptions {...}` inside `android {...}` of `build.gradle` fix the error when building Debug  variant
```
android {
...
  compileOptions {
        sourceCompatibility JavaVersion.VERSION_1_8
        targetCompatibility JavaVersion.VERSION_1_8
    }
...
}
```

Btw, i'm using AdoptOpenJDK (OpenJ9 v11.0.7.10 x64) as default (JAVA_HOME Environment Variable), i wonder if it have something todo with it, but it's strange that only Debug variant were getting this error.

PS: I try to fix this on v4.1 but v4.0 were having similar issue as i remembered, so i guess it wasn't related to Android Studio

PR #13640, opened 2020-11-08
Fix car reflections in Outrun by implementing per-bit color masking

hrydgard:
Followup to #13636 , implementing the fix for the yellow/blue reflection issue that surfaced after merging that. Just like that one, it's only enabled for Outrun for now, but I'll enable it wider once I gain confidence in it.

See https://github.com/hrydgard/ppsspp/issues/11358#issuecomment-720160221 for technical details - basically, the PSP can't just mask off the R,G,B and A channels when rendering, it can also mask off individual bits, and Outrun uses this to perform an intricate trick. 

We do this masking in the shader. Note: This fix only works on Vulkan, GLES 3+ and desktop GL and D3D11. Not D3D9 and GLES 2.

ULES00262_00015

@thedax :)


thedax:
The reflections appear to be fixed in both OpenGL & Vulkan (unable to test Direct3D stuff anymore since I can't realistically compile Windows builds here without a VM, which I'm not keen on doing :P).

At any rate, here are a couple shots from Vulkan:
image

image

Nice to see another long-standing issue getting some attention :)

PR #13643, opened 2020-11-09
Reorganize end of fragment shader, move shader translator to Common

hrydgard:
Addresses https://github.com/hrydgard/ppsspp/commit/b4a76a9f093157c667282982ff78e27f9177103a#r43990408 , thanks!


unittest/TestShaderGenerators.cpp
```diff
@@ -352,20 +352,18 @@ bool TestShaderGenerators() {
 	init_glslang();
 	LoadD3DCompilerDynamic();
 
-	if (!TestReinterpretShaders()) {
+	if (!TestFragmentShaders()) {
 		return false;
 	}
+	return true;
```
iota97, line 358: Is this return intentional or left from testing?

hrydgard: not intentional to leave behind, but inconsequential since this is just test code that doesn't get compiled into the app.

Issue #13639, opened 2020-11-08
Option to disable shader compile errors on-screen


Panderner:
Hi i wanted to add a option to disable shader compile errors on-screen.

In Downstream Panic it displays GL_INVALID_OPERATION on-screen but I don't see graphical glitches in this game and it's playable.
![Screenshot_2020-11-09-06-39-50-98_2f85358b2198d26f8aca533d68bee793](https://user-images.githubusercontent.com/37503397/98486216-fbaec980-2256-11eb-9b02-575756352d0b.png)
![Screenshot_2020-11-09-06-40-04-50_2f85358b2198d26f8aca533d68bee793](https://user-images.githubusercontent.com/37503397/98486220-010c1400-2257-11eb-87d6-68d1a4aa7830.png)

hrydgard:
That's not a shader compiler error, it's some different error (which I also just enabled display for). Unfortunately it's not very detailed. I'm gonna make a better version of this later that's more specific, in the meantime I'll disable it.

But we should fix whatever error this is. Could you take a GE dump and remind me what hardware this is?

And if you find more games with this type of issue, let me know.

Panderner:
i tried to make the GE dump but it crashed for me when i minimize PPSSPP. I'm using Realme C2.

hrydgard:
Hm, ok. Anyway, I've disabled the checks for now - they'll be back later as an option.

Panderner:
But how about Outrun 2006?
![Screenshot_2020-11-09-17-42-46-12_2f85358b2198d26f8aca533d68bee793](https://user-images.githubusercontent.com/37503397/98525179-157dfa00-22b3-11eb-95a0-8f010573fc79.png)

hrydgard:
That we should simply fix. That's bad - and having it cover the screen gets people like you to report it :)

hrydgard:
I don't think I'm gonna add such an option, I'll just fix errors as they are reported. Besides, the error won't (or at least isn't supposed to) stay on screen for long.

PR #13646, opened 2020-11-09
Framebuffer fetch fixes

hrydgard:
Fix issue with EXT_shader_framebuffer_fetch on OpenGL ES 3.0.

Remove support for NV_shader_framebuffer_fetch, extremely rare and old.

Should help with the error seen in #13639 .


Panderner (commented):
Does it break any games that uses NV_shader_framebuffer_fetch?

hrydgard:
NV_framebuffer_fetch is an extension a device can expose for PPSSPP and other apps to use. For PPSSPP it's just an optimization, and the only device I'm aware of that exposes it is the very old NVIDIA Shield handheld.

PR #13661, opened 2020-11-15
Fix crash in sceusbmic on load old staus then save again

### LINKED ISSUE #13658: Save state issue after update to build 1196
I tried with build 1196, whenever i load state and tried to save state the ppsspp force close. 
Screenshot_20201112-175654
Screenshot_20201112-175704
My save data is from around september 2020, have tried three games with same issue ( P3P, Digimon english patched, and growlanser)

sum2012:
Fix #13658 again

Can we directly add a check in the https://github.com/hrydgard/ppsspp/blob/master/Core/CoreTiming.cpp#L177 
for fix unknown problem ?


shenweip (approved):

hrydgard:
I'll consider that, in the meantime, let's merge this.

PR #13665, opened 2020-11-15
Ace Combat X/Joint Assault: Avoid slow readback for lens flare effect.

hrydgard:
Turns on the good old BlockTransferAllowCreateFB setting for these games, which allows PPSSPP to create a GPU surface instead of reading back to the CPU when a block transfer happens to memory where we didn't already have a framebuffer registered.

(I was able to confirm that there's a readback happening on the lens flare effect in #12870, slowing things down on some devices. This should eliminate that.)

(Yeah I know this is the type of change Unknown doesn't like because you don't know what else it will break in the game so someone will need to give the games some solid testing before we can safely merge... or we merge and wait for complaints. At least it's not an RPG so all scenes have one sun I guess... should be pretty uniform.).

@LunaMoo wanna try the Ace Combat games with this and see if there are no glitches?

### DIFF
--- a/assets/compat.ini
+++ b/assets/compat.ini
@@ -552,6 +552,24 @@ ULUS10249 = true
 ULES00637 = true
 ULKS46126 = true
 
+# Ace Combat X: Skies of Deception
+# Lens flare effect 
+ULUS10176 = true
+ULJS00086 = true
+UCKS45067 = true
+UCES00423 = true
+ULJS19018 = true
+
+# Ace Combat: Joint Assault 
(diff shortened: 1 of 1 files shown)


LunaMoo:
Well I tested it shortly and it bugs out he lens flare/sun effects. Normally it disappears when something like for example a plane is between the sun and the camera, with this hack it's constantly visible even through the solid objects.

hrydgard:
That's slightly unexpected, but yeah, will just close then.

Issue #13652, opened 2020-11-10
Chinese font missing issue


sum2012:
I use World chinese patch of NeverlandNaruru Oukoku Monogatari for example

![1](https://user-images.githubusercontent.com/2177532/98684437-2147ea00-23a1-11eb-889b-89eb395c883f.png)

edit:wrong picture
![2](https://user-images.githubusercontent.com/2177532/98685830-d3cc7c80-23a2-11eb-9ce8-ea7c83b771e5.png)

Old version of jpn0.pgf  more support chinese font

@hrydgard this must solve in v1.10 otherwise chinese people (special using android) will use older version forever.

Do we detect lang using zh_CN.INI or zh_TW.ini use older version of jpn0.pgf ?
Or use eboot.* ? (This game use disc0:/PSP_GAME/SYSDIR/EBOOT.LEI )
Or detect using kernel function ?

sum2012:
upload again bad picture
![2](https://user-images.githubusercontent.com/2177532/98686038-0fffdd00-23a3-11eb-8e7c-b43146ca1bae.png)

sum2012:
@nassau-tk Can your jpn0.pgf add more support of chinese font ?

hrydgard:
Yup, that would seem to be the problem...

nassau-tk:
> @ nassau-tk jpn0.pgfで中国語フォントのサポートを追加できますか？

This is a heavy ploblem!
If add chinese font then I don't have converter for including Chinese. 

@hrydgard 
What kind of converter used "Old  jpn0.pgf"?
And,Do you have it?

hrydgard:
I haven't touched that font stuff for many years, I don't remember :(

sum2012:
First I need to list out games that must use current jpn0.pgf
GUNDAM VS GUNDAM NEXT PLUS (NPJH50107) 
Neon Genesis Evangelion 2 ULJS00064
@nassau-tk do you know must games ?

nassau-tk:
@sum2012 

janne d'arc
Monster Hunter Seriese (MHP/MHP2/MHP2G/MHP3)
Assassin's Creed : Blood Line
QIX+
EXIT
Kangaeru EXIT (EXIT2)
Tantei Jinguji Saburo -Hai to Diamond- (maybe?)

I can  remember only these now.

I will try think about some way to solve font problems this week end.

By the way, How many chinese patched game is there?
Do you know?
It's many? or a little?

[...]

PR #13666, opened 2020-11-16
Add temp chinese font support

sum2012:
Temp work around for #13652
We can wait until nassau-tk make chinese compatible 's jpn0.pdf

@shenweip What do you think ?
Do we also blacklist known game which need use nassau-tk's font ?
Qix++ (NPJH50199)
GUNDAM VS GUNDAM NEXT PLUS (NPJH50107)
Neon Genesis Evangelion 2 \Shinseiki Evangelion 2 (ULJS00064)

### DIFF
--- a/Core/HLE/sceFont.cpp
+++ b/Core/HLE/sceFont.cpp
@@ -98,6 +98,10 @@ struct FontRegistryEntry {
 static const FontRegistryEntry fontRegistry[] = {
 	// This was added for Chinese translations and is not normally loaded on a PSP.
 	{ 0x288, 0x288, 0x2000, 0x2000, 0, 0, FONT_FAMILY_SANS_SERIF, FONT_STYLE_DB, 0, FONT_LANGUAGE_CHINESE, 0, 1, "zh_gb.pgf", "FTT-NewRodin Pro DB", 0, 0, 1581700, 145844, true },
+	
+	// This is also added for Chinese translations and is not normally loaded on a PSP.	
+	{ 0x288, 0x288, 0x2000, 0x2000, 0, 0, FONT_FAMILY_SANS_SERIF, FONT_STYLE_DB, 0, FONT_LANGUAGE_CHINESE, 0, 1, "chn.pgf", "FTT-NewRodin Pro DB", 0, 0, 1581700, 145844 },
+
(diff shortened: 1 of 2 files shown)


Core/HLE/sceFont.cpp
```diff
@@ -98,6 +98,10 @@ struct FontRegistryEntry {
 static const FontRegistryEntry fontRegistry[] = {
 	// This was added for Chinese translations and is not normally loaded on a PSP.
 	{ 0x288, 0x288, 0x2000, 0x2000, 0, 0, FONT_FAMILY_SANS_SERIF, FONT_STYLE_DB, 0, FONT_LANGUAGE_CHINESE, 0, 1, "zh_gb.pgf", "FTT-NewRodin Pro DB", 0, 0, 1581700, 145844, true },
+	
+	// This is also added for Chinese translations and is not normally loaded on a PSP.	
+	{ 0x288, 0x288, 0x2000, 0x2000, 0, 0, FONT_FAMILY_SANS_SERIF, FONT_STYLE_DB, 0, FONT_LANGUAGE_CHINESE, 0, 1, "chn.pgf", "FTT-NewRodin Pro DB", 0, 0, 1581700, 145844 },
```
shenweip, line 103: I'm a little confused. Is there any difference "zh_gb.pgf" and "chn.pgf"?

Core/HLE/sceFont.cpp
```diff
@@ -1171,6 +1175,11 @@ static int sceFontFindOptimumFont(u32 libHandle, u32 fontStylePtr, u32 errorCode
 	if (nearestFont) {
 		optimumFont = nearestFont;
 	}
+	//Special chinese font support
+	if ((g_Config.iLanguage == 11) || (g_Config.iLanguage == 10)) {
+		if (optimumFont == internalFonts[2])
+			optimumFont = internalFonts[0];
```
shenweip, line 1181: If "zh_gb.pgf" exits, use "zh_gb.pgf" instead of "chn.pgf"?

shenweip (changes requested):

Core/HLE/sceFont.cpp
sum2012: if (optimumFont == internalFonts[2]) should be return false

Core/HLE/sceFont.cpp
sum2012: chn.pgf use older ppsspp version 's jpn0.pgf ( v1.10.3-206-g9f88a8835)

Core/HLE/sceFont.cpp
shenweip: But if "zh_gb.pgf" exits, this index would be wrong.

shenweip:
I'm not sure if we need to have two versions of the font.

sum2012:
nassau-tk have made chinese compatible 's jpn0.pdf
Closing this

PR #13670, opened 2020-11-16
[Adhoc] Fix PSP2i unable to enter dungeon when using RadminVPN issue.

### LINKED ISSUE #13648: Phantasy Star Portable 2 Infinity Adhoc Multi Mode Issue
### What happens?
When attempting to start a mission in Phantasy Star Portable 2 Infinity only the host is able to enter the dungeon. Other players attempting to enter the dungeon are stuck on loading screen. This issue is not present in the latest stable version.

### What *should* happen?
When attempting to start a mission in Phantasy Star Portable 2 Infinity all players should be able to enter the dungeon.

Both computers I'm testing on have nearly identical hardware. Only difference is that one machine has a 2060 SUPER and the other has a 1070.
Operating System
	Windows 10 Enterprise 64-bit
CPU
	Intel Core i7 7700 @ 3.60GHz
	Kaby Lake 14nm Technology
RAM
	32.0GB
Motherboard
	Micro-Star International Co. Ltd. Z390-A PRO (MS-7B98) (U3E1)
Graphics
	VG248 (1920x1080@144Hz)
	4095MB NVIDIA GeForce GTX 1070 (EVGA)
Storage
	465GB Samsung SSD 850 EVO 500GB (SATA (SSD))
	465GB Samsung SSD 850 EVO 500GB (SATA (SSD))
	2794GB TOSHIBA HDWD130 (SATA )
Optical Drives
	ASUS DRW-24B1ST j
Audio
	Realtek High Definition Audio

anr2me:
Fix https://github.com/hrydgard/ppsspp/issues/13648

### DIFF
--- a/Core/HLE/proAdhoc.cpp
+++ b/Core/HLE/proAdhoc.cpp
@@ -1918,6 +1918,11 @@ int setSockBufferSize(int sock, int opt, int size) { // opt = SO_RCVBUF/SO_SNDBU
 	return setsockopt(sock, SOL_SOCKET, opt, (char *)&n, sizeof(n));
 }
 
+int setSockMSS(int sock, int size) {
+	int mss = size; // 1460;
+	return setsockopt(sock, IPPROTO_TCP, TCP_MAXSEG, (char*)&mss, sizeof(mss));
+}
+
 int setSockTimeout(int sock, int opt, unsigned long timeout_usec) { // opt = SO_SNDTIMEO/SO_RCVTIMEO
 	if (timeout_usec > 0 && timeout_usec < minSocketTimeoutUS) timeout_usec = minSocketTimeoutUS; // Override timeout for high latency multiplayer
 #if defined(_WIN32)
--- a/Core/HLE/proAdhoc.cpp
+++ b/Core/HLE/proAdhoc.cpp
@@ -1940,8 +1940,21 @@ int getSockNoDelay(int tcpsock) {
 	return opt;
 }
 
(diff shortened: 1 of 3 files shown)


ANR2ME:
Hmm.. multiplying the socket buffer size like this approach might not be the best solution for VPN issue due to small buffer size, since there are games that use buffer size as small as 4 bytes.

Anuskuss:
@ANR2ME Why is this (or more specifically `TCP_NODELAY`) not enabled by default? I understand that Nagle's is supposed to improve throughput but for Adhoc it feels like getting the packets ASAP is much more important (even over the internet).

ANR2ME:
> 
> 
> @ANR2ME Why is this (or more specifically `TCP_NODELAY`) not enabled by default? I understand that Nagle's is supposed to improve throughput but for Adhoc it feels like getting the packets ASAP is much more important (even over the internet).

Because PSP have syscall to Flush PTP (TCP) connection, it means Nagle algo wasn't enabled by default, otherwise Flushing won't be needed.

Unlike most PC games where nagle algo usually being disabled for performance reason.

Anuskuss:
> it means Nagle algo wasn't disabled by default on real PSP

That's tricky. I feel like it's probably better to leave it enabled then to correctly emulate the PSP but on the other side it would probably be beneficial and improve Adhoc (like most advanced features like resolution scaling and shaders which aren't technically correct either but improve the experience). Well, I'm glad I don't have to make such decisions 😛

PR #13690, opened 2020-11-21
Add ini setting to disable on screen message

### LINKED ISSUE #13682: Disable on screen messages
Found no way to disable on-screen messages, such as from folder/filename after a capture or state saved/loaded when using savestates. Is there any?

Thanks.

iota97:
Fix #13682, ini setting only as I think it could create quite some confusion in general.

While taking a look at the ini I noticed PSP Color had a space in the name, not really a issues but while I was at it I cleaned it up a little bit :)


hrydgard:
Sorry, didn't notice this one when I made https://github.com/hrydgard/ppsspp/pull/13695 :)

I'll merge your other changes.

Issue #13698, opened 2020-11-23
[Windwos][FPS goes down]Collision? Conflict? with debug Logging.[Muso Orochi (ULJM05314)]


nassau-tk:
### What happens?
This games FPS goes down by "collision ? conflict? with debug log" if debug logging turn on.
It's happen at only startup.
It will happen on Windows10(64-bit).
But,Android will not happen. 

### What *should* happen?
It's should be work smooth even if turn on the debug logging.

### What hardware, operating system, and PPSSPP version?  On desktop, GPU matters for graphical issues.
Windwos PPSSPP Ver.：1.10.3-1234
OS：Windows10（64-bit）Japanese
CPU：Ryzen7 3800X（8-core）
GPU：RADEON RX5700XT
Resolution：2560×1440
RAM：32GB

Android  PPSSPP Ver.：1.10.3-1234
OS：Android 10.0
DeviceName：Samusung Galaxy S20 / SC-51A
CPU：Snapdragon865（Octa-core）
GPU：Adreno 650
Resolution：3200 x 1440 (Max)
RAM：12GB
ROM：128GB
ExternalStrage：microSDXC 512GB

[First appearing the "Muso Orochi Logo" then FPS goes down.]
![image](https://user-images.githubusercontent.com/48179091/99923123-9fc86280-2d77-11eb-9395-6e6c1258fd62.png)

[Next,Appear the "KOEI Logo" then fps become smooth.]
![image](https://user-images.githubusercontent.com/48179091/99923230-2aa95d00-2d78-11eb-9c11-77df01ac426a.png)

[This is Log when FPS going down. ]
![image](https://user-images.githubusercontent.com/48179091/99923459-30537280-2d79-11eb-8190-06f8ae491ba6.png)

hrydgard:
Can you confirm if #13703 made a difference?

nassau-tk:
Yes,It's solved!!
Thank you!!

PR #13703, opened 2020-11-23
Guess at a fix for #13698 - sceKernelThreadGetExitStatus probably takes some cycles.

hrydgard:
We could also just turn off the logging, but I think this might help enough.

Should help #13698 .


Panderner:
this will fix performance yet?

hrydgard:
It will probably fix or at least reduce that specific problem reported in #13698

PR #13704, opened 2020-11-23
Remember game list scroll

### LINKED ISSUE #13699: [Featuer Request] A choice point is memorized in "Game List".
### What happens?
Large capacity device and low-price Memory card are the mainstream recently.
So,A lot of users are carring around many games on self device.

So, I have a suggestion.
If choose a game In the "Game List" then the game will start.
And, If you finish and back to the "Game List" then the "top of the game list" will appear.
I think at point of here.

I think about that the "Game List" should memory the "last choosen point."
"Recentry"?
um... It's not same meaning.
Different approach to userbility?I think.

iota97:
Fix #13699


hrydgard:
Looks good. Was first a little afraid of going out of bounds in case games disappear from a folder but I think the existing mechanisms in ScrollView will just fix that right up.

PR #13713, opened 2020-11-24
Mpeg:Correctly handle stuffing data of pack header.

shenweip:
This seems to be a potential issue. Maybe just a mistake?


hrydgard:
Looks like a mistake indeed!

Issue #13717, opened 2020-11-26
Ultimate Ghouls'n Ghosts (and Goku Makaimura) with graphics problems.


vanderleivieirainacio:
The Ultimate Ghouls'N Ghosts game (and its Japanese versions Goku Makaimura and Goku Makaimura Kai) are presenting graphical problems in all APIs (OpenGL, Vulkan, DirectX 11, DirectX 9) where textures disappear and a black background is left with some objects from game, and even in the Vulkan API, error messages appear several times presented by the emulator.

![Captura de tela 2020-11-23 161755](https://user-images.githubusercontent.com/46516229/100330955-57df5080-2fae-11eb-9503-06a75244b6e8.png)
![Captura de tela 2020-11-23 161702](https://user-images.githubusercontent.com/46516229/100330970-5c0b6e00-2fae-11eb-9d05-d9aa909d0039.png)

My PC:
Graphics: Sapphiretech Radeon RX 580 4GB
Operating System: Windows10 20H2 64bit

vanderleivieirainacio:
It worked great for Ultimate Ghouls'n Ghosts and Goku Makaimura, thank you very much! However, the Japanese special edition Goku Makaimura Kai still presents the problems I described.

hrydgard:
@vanderleivieirainacio Can you check its "game id"? like NPJH12345 or whatever.

vanderleivieirainacio:
Ah yes! Goku Makaimura Kai's ID is ULJM05265.

vanderleivieirainacio:
Just to make clear the ID I referred to, Goku Makaimura Kai (ULMJ05265) is the one that is having problems. The problems still remain in that ID.

vanderleivieirainacio:
Thank you!! I was rushed and didn't realize that you were still fixing the problems! Thank you very much! You are the man!!

chinarut:
@vanderleivieirainacio have you tried to build from source with this patch in place to verify it works with Goku Makaimura Kai?

I ask because I had the same exact issue with Ultimate Ghosts & Goblins (**ULUS10105**) and my workaround is to roll back to v1.6.3. I verified I do see this game_id in the patch:

# Ultimate Ghosts & Goblins
 ULJM05147 = true
 **ULUS10105** = true
 ULES00419 = true
 NPJH50235 = true

However, I do **not** see any reference to your game_id: ULMJ05265 so i thought I’d ask before v1.11.0 rolls out!

[...]

PR #13720, opened 2020-11-27
Enable color reinterpret for Ultimate Ghosts & Goblins

hrydgard:
The game uses it by accident probably, but whatever - it's a clean fix.

Fixes, but only for "modern" APIs, #13717.

There's something else lurking too.


GPU/Common/GPUStateUtils.cpp
```diff
@@ -959,7 +959,7 @@ void ApplyStencilReplaceAndLogicOpIgnoreBlend(ReplaceAlphaType replaceAlphaWithS
 bool IsColorWriteMaskComplex(bool allowFramebufferRead) {
 	// Restrict to Outrun temporarily (by uglily reusing the ReinterpretFramebuffers flag)
```
sum2012, line 960: Should we need change this commit ?

hrydgard: Yeah, I should have removed that comment.

Issue #13718, opened 2020-11-26
[Suggestion][Solved to Latin Font issue] I will try make Font File. [ltn8~15.pgf]


nassau-tk:
### What happens?
I will try make ltn8~15.pgf.

- [ltn8.pgf] Qix++ will OK.
- [ltn12.pgf] Lunar has a line spaces problem.
- [ltn9/11/13/15?] SnakeWarriors Training has position & line spaces problem.

I have many problem yet.
Most important problem is "Line Spaces".

I tested changing "Win Ascent & Win Descent" on TTF.
But, I can't getting good result.

If anyone have a informations or advices.
Please, Give me the hints!!

[QIX++ (ltn8.pgf)]
![image](https://user-images.githubusercontent.com/48179091/100331505-978c5380-3013-11eb-8934-d1f94683d309.png)
![image](https://user-images.githubusercontent.com/48179091/100331573-abd05080-3013-11eb-8b05-093d248f8118.png)
![image](https://user-images.githubusercontent.com/48179091/100331595-b559b880-3013-11eb-958c-e8efdc8235b8.png)
![image](https://user-images.githubusercontent.com/48179091/100331626-bee32080-3013-11eb-98f6-ed1ebd0b78e6.png)
![image](https://user-images.githubusercontent.com/48179091/100331685-d28e8700-3013-11eb-9a60-52bacf26b320.png)
![image](https://user-images.githubusercontent.com/48179091/100333061-7cbade80-3015-11eb-865a-2df84dca9b10.png)

[Lunar] (Real PSP ltn12.pgf)
![image](https://user-images.githubusercontent.com/48179091/100330128-e6d18480-3011-11eb-9386-b5f290613bf1.png)
![image](https://user-images.githubusercontent.com/48179091/100330163-f355dd00-3011-11eb-8c0a-af1fabb48f92.png)

[Lunar] ("Maked ltn8.pgf" rename to ltn12.pgf)
![image](https://user-images.githubusercontent.com/48179091/100330439-4c257580-3012-11eb-9337-47da95f60f0d.png)
![image](https://user-images.githubusercontent.com/48179091/100330496-5f384580-3012-11eb-818f-2d5b9cd8bc5e.png)

I think "Warning screen" will looks good size.
But,The  line spaces to large that In game conversation window.

[SnakeWarriors] (test by maked ltn8.pgf "sans-serif".)
[...]

PR #13721, opened 2020-11-28
Change ltn x font

nassau-tk:
"Small latin fonts" are modified.
(From ltn8.pgf to ltn15.pgf)
See #13718 .


hrydgard:
good progress! Let's try them.

PR #13733, opened 2020-12-01
Apple M1 - initial work

hrydgard:
image

This gets it going, if you do something approximating the following (it's early days!)

* Switch to this branch, update submodules

* Install XCode.

* Install Homebrew using this method: https://github.com/mikelxc/Workarounds-for-ARM-mac  - do follow the advice to put it under /opt/.

* Install cmake: `brew install -s cmake`

* Install SDL2: `brew install -s SDL2`

* `export LIBRARY_PATH=/opt/homebrew/lib`

* `./b.sh --debug` (or whatever)

* Set Settings/Tools/Developer Options/CPU Core to IR Interpreter

* Go nuts

This PR contains some hacks that need to be gotten rid of - not even close to be ready for merge.

JIT works now!

Just need to get the cacheline size dynamically:

https://developer.apple.com/documentation/apple_silicon/addressing_architectural_differences_in_your_macos_code


jcchikikomori:
I am already curious if Apple Silicon has a better performance compared to Intel Macs
Also, sad for iPhone users

hrydgard:
It is very very fast. Much faster than my Intel mac, that one is 6 years old though. Don't see what's sad for iphone users.

Anyway, I picked over all the safe changes here to master. Not much left here now.

PR #13734, opened 2020-12-03
Why not iOS?

Halo-Michael:
It actually works on iOS


hrydgard:
Mostly because I haven't yet been able to test it on iOS, but this seems fine.

PR #13742, opened 2020-12-06
cmake: Add an install target.

orbea:
This adds an `install` target and a desktop file for both the SDL and Qt5 frontends.

This is very useful for installing ppsspp system wide as is done with distros and required for anyone that wants to install ppsspp into a non-standard location.


hrydgard:
thanks!

PR #13751, opened 2020-12-08
Initial support for vulkan on iOS. In addition, MoltenVK is bound to the framework.

Halo-Michael:
Initial support for vulkan on iOS.
MoltenVK is bound to the framework.
Todo:
1. Directly dependence the framework without using dlopen.
Disadvantage: May need to remove support for armv7, Vulkan may never support armv7 devices.
2. Need to wait for MoltenVK to support the mac with arm64 chips.
Cheat: We can modify the dylib made for iOS devices to support the mac with arm64 chips.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2097,8 +2097,8 @@ if(LIBRETRO)
 endif()
 
 if(TargetBin)
-	if(IOS OR APPLE)
-		if(APPLE AND NOT IOS)
+	if(APPLE)
+		if(NOT IOS)
 			set(ICON_PATH_ABS ${CMAKE_CURRENT_SOURCE_DIR}/icons/ppsspp.icns)
 			set(MACOSX_BUNDLE_ICON_FILE ppsspp.icns)
 			set_source_files_properties(${ICON_PATH_ABS} PROPERTIES MACOSX_PACKAGE_LOCATION "Resources")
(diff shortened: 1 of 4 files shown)


Halo-Michael:
Tips:
The original libMoltenVK.dylib for iOS file has not been signed. You need to use the developer certificate or ldid signature (for jailbroken devices) before use. The libMoltenVK.dylib provided here has been signed with ldid for use out of the box.

Halo-Michael:
The submission here is experimental and not guaranteed to work.

Halo-Michael:
Although it has been switched to Vulkan in the settings, obvious that the Vulkan engine is not be used. But after all it is just the first step.

Panderner:
@hrydgard this means for removing 32-bit support for iOS devices?

hrydgard:
Possibly.

Though it's probably time for that anyway...

kage2051:
> @hrydgard this means for removing 32-bit support for iOS devices?

Sadly it's a yes. The iPhone 5/5c (which was the last 32-bit iOS device, and it was released 8 years ago) isn't capable of running most of the games without stuttering/crashing anyway. In a world where 3GB of RAM is the recommended minimum, 1GB is not enough.

hrydgard:
Let's try this.

hrydgard:
Hm, I read this agian little fast there, we should probably indeed have waited until the new Vulkan SDK that will include proper support for ARM macs in the MoltenVK dylibs.. Oh well, can be fixed.

Issue #13756, opened 2020-12-08
[SaveData] All save data date & filesize is not appear correctly.


nassau-tk:
### What happens?
All savedata date is not overwrite.
Those data is saving correctly.
Maybe,just appearance of date?
~~Because,The old savedata was changed to the date.~~
~~The data was not saving after the date.~~

It will happen Windows & Andoroid.

### What *should* happen?
Appear the date correctly.

### What hardware, operating system, and PPSSPP version?  On desktop, GPU matters for graphical issues.
PPSSPP Ver.：1.10.3-1308
OS：Windows10（64-bit）Japanese
CPU：Ryzen7 3800X（8-core）
GPU：RADEON RX5700XT
Resolution：2560×1440
RAM：32GB

PPSSPP Ver.：1.10.3-1308
OS：Android 10.0
DeviceName：Samusung Galaxy S20 / SC-51A
CPU：Snapdragon865（Octa-core）
GPU：Adreno 650
Resolution：3200 x 1440 (Max)
RAM：12GB
ROM：128GB
ExternalStrage：microSDXC 512GB

[These game were saved at 2020/12/09 but it was not change date.]
![Screenshot_20201209-073949_PPSSPP](https://user-images.githubusercontent.com/48179091/101550448-b109a400-39f2-11eb-996e-dcc25ded11a8.jpg)

![Screenshot_20201209-073940_PPSSPP](https://user-images.githubusercontent.com/48179091/101550479-b961df00-39f2-11eb-8e61-dfe4d36a9101.jpg)

![Screenshot_20201209-074018_PPSSPP](https://user-images.githubusercontent.com/48179091/101550512-c67ece00-39f2-11eb-8f6e-f4709f69ff01.jpg)

But,I don't have memory that this game is saved the near the day.
Strange...
![Screenshot_20201209-073909_PPSSPP](https://user-images.githubusercontent.com/48179091/101550536-d1d1f980-39f2-11eb-8a6f-e58c73f6c798.jpg)

nassau-tk:
It will happen on "Screen of PSP's Save".
Because,It will not happen on original save screen like below.

![image](https://user-images.githubusercontent.com/48179091/101563060-ef5f8d00-3a0b-11eb-94b5-563b7b815904.png)

nassau-tk:
I tested 2 things.

[Patern 1]
I will create a new save data and check the date.
**Result : It made right date.But,If overwrite the saved data then date is not renew.**

[Patern2]
[...]

PR #13764, opened 2020-12-13
Savedata:Fixes incorrect date of savedata. 

shenweip:
Should help #13756.


unknownbrackets (approved):
It makes sense to me that it'd use the directory information rather than files in it.  I guess we could create tests to verify this further.

-[Unknown]

hrydgard:
Yeah, looks good.

Issue #13255, opened 2020-08-06
[iOS 14] PPSSPP crashes with JIT


xsm1997:
### What happens?
When connected to a debugger, things go normal.
When not connected to a debugger, PPSSPP crashes on all my games (tested 4) with JIT enabled. I am sure it is related to JIT because if I change to IR interpreter, things go normal but slow.

### What *should* happen?
Games run normally.

### What hardware, operating system, and PPSSPP version?  On desktop, GPU matters for graphical issues.
iPad Pro 9.7 (1st generation), iPad OS 14 Public Beta 4, with no Apple developer program subscription (certificates expire after 8 days).
PPSSPP v1.10.3 self-compiled.

I think maybe Apple added some new security restrictions related to JIT. And maybe it is a bug to iOS or a temporary situation. When Apple finally releases iOS 14, I will test again.

hrydgard:
Yeah, we've had reports of this before. I think the method we're using to fool it to believe it's running in the debugger is no longer working due to security fixes in iOS 14, and there's likely nothing that can be done about it without a jailbreak.

i30817:
People have to face that buying apple is a terrible malinvestment.

devinprater:
For some, like people with disabilities, it’s the only choice that allows one to use a phone, or tablet, productively. We just have to do the best we can with AltStore and other such tools. And with the IR Interpreter on my iPhone X R, even on iOS 14, it runs smoothly, unless the phone is under more-than-normal stress.
Devin Prater
d.prater@me.com
Https://devinprater.github.io

hrydgard:
I'm gonna close this because I don't think there's anything we can do about the issue. Apple's gonna do what Apple wants to do, today they're blocking the JIT. I'm glad the iOS port is usable at least for people who have no choice.

EhsanWaheed:
They’ve added jit support for iOS 14.2 and side loading via alt store also supports jit

xsm1997:
@EhsanWaheed Which commit did them actually add JIT support for iOS 14.2？

EhsanWaheed:
Hopefully but I think PPSSPP needs an update to use the iOS JIT as it crashes at the moment

EhsanWaheed:
@hrydgard What do you reckon?

xsm1997:
[...]

Issue #12575, opened 2020-01-13
Could the new PPSSPP for iOS be distributed on AltStore?


lazerl0rd:
AltStore (https://altstore.io) is an alternative app store for non-jailbroken iOS devices, and can distribute and resign applications onto Apple devices. This would be a great way to recieve updates for those of us who aren't jailbroken.

mariolopjr:
Hi @hrydgard @unknownbrackets I'd like to help with AltStore and ipa/deb building for iOS.

You can see repo here: https://github.com/mariolopjr/ppsspp-builder, based on some previous scripts and newer changes in PPSSPP, I've got GitHub CI working building complaint ipa. I can then work on getting these on AltStore. However, my repo will be tracking the tagged releases (I will explore building for each commit).

I hope this can help folks with non-jailbroken iOS devices. Thoughts?

hrydgard:
Sounds good, since I'm not actively supporting iOS myself, any such effort is appreciated. Happy to link to things or merge changes you need.

mariolopjr:
Thank you! I will be doing a PR for the git-version cmake to (hopefully) make it more compatible with ppsspp in a submodule. I've also reached out to AltStore.

I am not sure if you happen to know, but I'd also like to issue some fixes for the notch (I have an iPhone X I can test with). I also want to investigate setting the IR interpreter as the default setting the first time PPSSPP is run on an iOS device. Any pointers as to where to look for these? I am crawling the files (ofc), but am still unfamiliar with PPSSPP's structure. Thanks!!

I am also hoping to get a "preview" version out with MoltenVK dylibs linked for the experimental MoltenVK/Vulkan support on iOS (and macOS when I have the time!). I don't mind helping with those as well!

hrydgard:
As for the notch, you just need to return the right values in System_GetPropertyFloat - though it's already hooked up I see. Check out viewSafeAreaInsetsDidChange in ViewController.mm.

Setting the core to IRInterpreter by default can be done in Core/Config.cpp.

MoltenVK for iOS would be awesome.

mariolopjr:
My apologies, I thought I saw an iOS notch issue/bug (it may be I was looking at closed issues when exploring MoltenVK support).

Thanks for the pointers, lots of good stuff to explore!

devinprater:
[...]

PR #13766, opened 2020-12-13
Try to simplify iOS cmake build

hrydgard:
Removes support for 32-bit iOS devices and fixes compatibility issues with M1 macs and the latest XCode.

Adds  `./b.sh --ios-xcode` .

Unfortunately still struggling with getting the app to build.

It links and everything (after configuring auto signing), now I'm stuck in postbuild scripts:

image

Notes to self: look into #12575 , #13255 has good info from @jeeeyul .

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,10 +25,12 @@ endif()
 # Detect CPU from CMAKE configuration. Toolchains should set this up
 if(CMAKE_SYSTEM_PROCESSOR)
 	if(${CMAKE_SYSTEM_PROCESSOR} MATCHES "^aarch64")
-		set(ARM64 ON)
+		message("AARCH64 is a go")
+		set(ARM64_DEVICE ON)
 	elseif(${CMAKE_SYSTEM_PROCESSOR} MATCHES "^arm64")
+		message("ARM64 is a go")
 		# M1 Mac
-		set(ARM64 ON)
+		set(ARM64_DEVICE ON)
 	elseif(${CMAKE_SYSTEM_PROCESSOR} MATCHES "^arm")
 		message("ARM_DEVICE is a go")
 		set(ARM_DEVICE ON)
@@ -91,7 +93,7 @@ if(ANDROID OR WIN32 OR (UNIX AND NOT ARM_NO_VULKAN))
 endif()
 
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake/Modules)
-if(NOT (ARM64 AND MACOSX))
+if(NOT ((ARM64 AND MACOSX) OR IOS))
 	list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake/sdl)
 endif()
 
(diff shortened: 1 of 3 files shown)


hrydgard:
Hm, don't see how this could have worked.

Added some debug output:
```
	if(CMAKE_GENERATOR STREQUAL "Xcode")
		set(APP_DIR_NAME "${TARGET_BUILD_DIR}/${FULL_PRODUCT_NAME}")
	else()
		set(APP_DIR_NAME "$<TARGET_FILE_DIR:PPSSPP>")
	endif()
	add_custom_command(TARGET PPSSPP POST_BUILD
		COMMAND echo APP_DIR_NAME: ${APP_DIR_NAME}
		COMMAND pwd
		COMMAND mkdir -p ${APP_DIR_NAME}
		COMMAND tar -c -C ${CMAKE_CURRENT_BINARY_DIR} --exclude .DS_Store --exclude .git assets *.png | tar -x -C \"${APP_DIR_NAME}\"
		COMMAND /bin/bash "${CMAKE_SOURCE_DIR}/ios/macbundle.sh" \"${APP_DIR_NAME}\"
	)
```
and it turns out that `APP_DIR_NAME` is equal to / , which obviously doesn't work.

Looking better when I try to use the other option `"$<TARGET_FILE_DIR:PPSSPP>"` ... to be continued

hrydgard:
Finally, got it to work. Now I'm able to build PPSSPP for iOS on my M1 and run on an iPhone. 

@jeeeyul wondering if you have any feedback on my changes here?

Halo-Michael:
If run directly from xcode to the device, libMoltenVK.dylib may not be signed.

hrydgard:
Hm, that's right... 

MoltenVK's instructions seem to advocate using a XCFramework, that means static linking, or? https://github.com/KhronosGroup/MoltenVK/blob/master/Docs/MoltenVK_Runtime_UserGuide.md

Halo-Michael:
> Hm, that's right...
> 
> MoltenVK's instructions seem to advocate using a XCFramework, that means static linking, or?
> That would avoid any problems of signing the separate dylib... https://github.com/KhronosGroup/MoltenVK/blob/master/Docs/MoltenVK_Runtime_UserGuide.md

The file in xcframework does not look like an executable file, I can’t even sign it with ldid
2BB2AE52E97DA9854939F68DDD715E53

Halo-Michael:
So I'm just using dynamic library file.

Panderner:
> Removes support for 32-bit iOS devices

Here's a reason: Iphone 5/5C can't run most games ran at full speed even 2d ones.

kingko200:
> Finally, got it to work. Now I'm able to build PPSSPP for iOS on my M1 and run on an iPhone.
> 
> MoltenVK doesn't show up though (no Vulkan option). I think I'll need to do some work for that to happen.
> 
> @jeeeyul wondering if you have any feedback on my changes here?

Would you please share how did you make it work?

hrydgard:
Closing in favor of #13936. 

This does some unnecessary changes, better keep it small to risk various buildbots less.

PR #13771, opened 2020-12-15
Update libMoltenVK.dylib to support M1 chips

Halo-Michael:
EDIT(hrydgard): Also adds new iOS MoltenVK. From new Vulkan SDK


hrydgard:
Hm, I wonder why the iOS MoltenVK dylib is so huge? does it also contain x64 bits for simulator on old macs?

What do we do about signing? Should we instead use the static library approach?

Halo-Michael:
Programs built for iOS will always contain a lot of blank blocks. You can try to write a helloworld.c and compile it for iOS and macOS. Basically the size of the program built for iOS is about 56kb, and the size of the program built for macOS is about 1kb.
However, since iOS 10.3, 64-bit iOS devices have used apfs partitions as file partitions. Thanks to the characteristics of apfs, the blank space does not actually occupy the physical block of the hard disk at all, and can basically be understood as a kind of Similar to compression characteristics. Thanks to this feature, we actually don't need to care too much about these blank blocks in binaries.

Halo-Michael:
Regarding codesign, the temporary method is to use a jailbroken device for testing. I have signed libMoltenVK.dylib for iOS with ldid, so if you use a jailbroken device to test it, it should be available directly.
If there is no jailbroken device, you can also find the cached PPSSPP.app location after running ppsspp into the device through xcode run for the first time, use the codesign command to sign the libMoltenVK.dylib file, and then re-run ppsspp into the device. Xcode tries to skip the compilation process, read the ppsspp program in the cache and reinstall it. At this time, you should be able to see the vulkan option in ppsspp.
However, these are temporary solutions. The correct way is to add libMoltenVK.dylib as a third-party library that needs to be linked into the xcode project, and then xcode will automatically sign it. This may require that the ppsspp main program is directly linked to the libMoltenVK.dylib file, and the CMakeFile file needs to be properly and correctly modified. Unfortunately, my current knowledge of CMake is extremely limited, and I don't know how to make such a modification. Another method is to statically compile libMoltenVK.a directly into the ppsspp main program, so that there is no need to sign. This also requires a lot of modifications to the source code.
I suggest that libMoltenVK.a is statically compiled into the main ppsspp program on macOS, which may save a lot of work time.

My knowledge of English is also extremely limited, so the language support comes from Google Translate. If you have any sentences that are difficult to understand, please tell me, and I will try my best to change the language and elaborate again.

Halo-Michael:
About how to use codesign command:
1. Use `security find-identity -v -p codesigning` to get all your certificates
2. Well, I did not find a more reliable introduction website, and I can’t describe it easily, so I use pictures to show
(}YGQ`QOZWYYC02F7NC@RR4
%F1RGA97W(@JV83HX5I6{33
Dylib does not need entitlements

DamnBoi706:
So now is vulkan working properly in the latest build now?

DamnBoi706:
imma try a couple different of games and iOS devices see if it would make any difference (with the latest build)

Halo-Michael:
> So now is vulkan working properly in the latest build now?

Not really on iOS

DamnBoi706:
so it is not functional right now?

DamnBoi706:
Or some part of it does work I don't really understand that?

Panderner:
@Halo-Michael why are you closing this?

PR #13776, opened 2020-12-16
Add stack value in scempeg and scepsmf from jpcsp

sum2012:
Might help video corruption issue
Jpcsp do not have stack value  in psmfplayer so that I need to check the value later


hrydgard:
OK, let's try this.

PR #13780, opened 2020-12-17
Fixed "jpn0/kr0/ltn0,2,4,6

nassau-tk:
It will solved issue #13767 .
Latin's " i " will appear on BLAZBLUE CS2.


nassau-tk:
@unknownbrackets 
@hrydgard 

Thanks for your help!

PR #13791, opened 2020-12-19
Linux: Properly avoid committing address space for large mmap().

achurch:
A shared or writable mapping is charged to the process's memory commitment regardless of MAP_NORESERVE, so the mmap() call to find a 4G base can still fail depending on memory usage and overcommit settings.


hrydgard:
Oh, good catch!

PR #13799, opened 2020-12-20
Add sceKernelFreeHeapMemory and sceKernelAllocHeapMemoryWithOptionFunction

sum2012:
Find in WWE Smackdown vs. RAW 2006
But don't help the game

### DIFF
--- a/Core/HLE/sceKernelHeap.cpp
+++ b/Core/HLE/sceKernelHeap.cpp
@@ -126,9 +126,8 @@ static u32 SysMemForKernel_536AD5E1()
 	return 0;
 }
 
-//should have option variable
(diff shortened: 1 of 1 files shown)


Core/HLE/sceKernelHeap.cpp
```diff
@@ -110,6 +126,21 @@ static u32 SysMemForKernel_536AD5E1()
 	return 0;
 }
 
+//should have option variable
```
hrydgard, line 129: hm, why not declare the extra option parameter then? even if it's not used and only logged.

sum2012:
@hrydgard done,only this function jpcsp don't have it.

PR #13802, opened 2020-12-22
Mpeg: Add warm up staff

sum2012:
fix #13527

original log: https://gist.github.com/sum2012/ec213b7141b0fa08f6ae1a206faa92c7
modify log: https://gist.github.com/sum2012/1737d874a299a230a4ead995ff9eb556
Unlucky I cannot do JPCSPTrace of the game.Nothing record.
I do not sure which one need call ERROR_MPEG_NO_DATA

PR #13806, opened 2020-12-23
Add an option to disable video decoding threads.

shenweip:
See https://github.com/hrydgard/ppsspp/issues/13759#issuecomment-750292447. Use an adjustable visual option to balance performance and correctness. Fixes possible video blur.

A fixed example(LocoRoco2)
Number of video decoing threads was set to auto:

Number of video decoing threads was set to 4:

### DIFF
--- a/Core/Config.cpp
+++ b/Core/Config.cpp
@@ -521,7 +521,7 @@ static ConfigSetting cpuSettings[] = {
 	ReportedConfigSetting("SeparateSASThread", &g_Config.bSeparateSASThread, &DefaultSasThread, true, true),
 	ReportedConfigSetting("SeparateIOThread", &g_Config.bSeparateIOThread, true, true, true),
 	ReportedConfigSetting("IOTimingMethod", &g_Config.iIOTimingMethod, IOTIMING_FAST, true, true),
-	ReportedConfigSetting("VideoDecodingThreads", &g_Config.iVideoDecodingThreads, 0, true, false),
+	ReportedConfigSetting("DisableVideoDecodingThread", &g_Config.bDisableVideoDecodingThreads, false, true, false),
 	ConfigSetting("FastMemoryAccess", &g_Config.bFastMemory, true, true, true),
 	ReportedConfigSetting("FuncReplacements", &g_Config.bFuncReplacements, true, true, true),
 	ConfigSetting("HideSlowWarnings", &g_Config.bHideSlowWarnings, false, true, false),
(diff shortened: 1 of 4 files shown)


hrydgard:
This is weird! Correctness of the decoding shouldn't change with the number of threads... Something is very wrong.

shenweip:
Use of ffmpeg threads will increase decoding delay by one frame per thread. If the video data at the beginning is not enough, will cause the all the video data to be read by the loop `av_read_frame()` . When the end of the data is reached, the integrity of the last frame cannot be guaranteed.

hrydgard:
Seems like an artificial way to add that delay. What about #13802 ? Doesn't it have a similar effect?

shenweip:
#13802 doesn't fix LocoRoco2, but if the game can response to ERROR_MPEG_NO_DATA and add more data to ringbuffer, I think it will have effect.

hrydgard:
Okay. Well either way I think we should try to avoid using the number of threads to hack around issues like this - feels unsafe. It has to be possible to tweak the buffering somehow so it works without it...

shenweip:
Does it look like safer now?

hrydgard:
I don't want to add an option for this, and I think there are ways we can avoid it.

I'll help look into this in a while but won't have time for a few days.

shenweip:
Oh, well, I'll keep on paying attention to this too.

sum2012:
LocoRoco 2 use psmfplayer so that don't effect my patch

unknownbrackets:
Did #14199 affect this?

As I noted for Valkyrie Profile, we really shouldn't be submitting data to FFmpeg that we don't yet have.  Unfortunately, the parsing API makes it a bit tricky and we may have to improve our own parsing so we know when frames begin and end.

-[Unknown]

sum2012:
Yes #14199 fix LocoRoco 2 
(Finally I get back password :) )

unknownbrackets:
Glad you got your access working again.  Is there anything this change (disabling threads) does now that the low latency change didn't?  Maybe not having threads was automatically changing latency.

-[Unknown]

Gamemulatorer:
This pr maybe could help YGO tag force 6 special cards is to blurry.
Screenshot_2021-09-03-11-50-12-685_org ppsspp ppsspp

unknownbrackets:
Are the cards each videos?  That seems strange, although maybe the play animations?  I don't think the PSP firmware's Mpeg library could handle that many concurrent videos really, and it'd be slow to init for each one, so the stills at least seem like they have to just be images.

I thought previous reports were that this game used sceJpeg, which wouldn't be affected by this pull.

-[Unknown]

hrydgard:
Totally forgot about this one.

I'm leaning towards not having an option here but just forcing it on always. Don't think the lack of multithreaded decoding is likely to be a bottleneck on almost any hardware - the devices that are so old that they would struggle decoding this video also are generally not multicore.

hrydgard:
Replacing this PR with #16799, closing.

PR #13809, opened 2020-12-24
[Adhoc] - Fixed an issue where sometimes a player unable to join a host on Warriors Orochi 2

anr2me:
Fixed an issue where sometimes a player unable to join a host on Warriors Orochi 2 (should also fix similar issue on other games that use PDP/UDP on Windows)


ANR2ME:
Since this is just a small changes, i'll just merge it.

PR #13869, opened 2021-01-04
Make combo button more generic

iota97:
This does a few thing:

- Make `EmuScreen::pspKey` public, this method is used to make the EmuScreen process key (both PSP and VIRTKEY), having it public should help reduce code duplication in touch control (i.e: unthrottle mode, analog rotation, etc could be have a way simpler logic).
- ~~As there is not really a way to pick virtual button status (unless I missed it?) combo button now just have a on off status for toggle, which is a bit worse tbh, I want to find out a solution for this.~~
- ~~Add quite a lot of virtual button to combo button (I didn't test them all yet tho')~~

Was I would like to discuss is:
- ~~Would it be worth to start a more generic touch UI with freely bindable key from this?~~
- ~~Should we make possible to set few of our button icon as a option and have only generic button at that point (except analog sticks, DPad and face button that have a separate logic at least).~~
- ~~With this I got to 29 button, might need a way to save uint64_t in the config at some point.~~

Any suggestion in more that welcome :)

Edit:
Growed quite a bit, I'm pretty happy with the result yay :)


iota97:
Changed the UI logic quite a bit: instead of having Combo button configuration screen one can set the combo button in the visibility UI (I changed some menu name to make it have sense ofc) kinda like the custom right analog. I think it make more sense for the user to setup it up when he "add the button" making it visible.

Also we can now set shape and icon of the combo button, should be fairy simple to add more in the atlas if needed now :)

Here some screen of the result (I find it cleaner this way, but just my opinion):

hrydgard:
The need for this certainly points to the need for a more generic touch input system ... We could even have custom stuff like just a left/right slider for the analog stick for racing games etc. That's a bigger project though...

It does seem like a sensible approach to a quicker fix. The new UI is a bit unreadable when it's that wide since the value of each setting is so far away from the text, but that's mostly the fault of the UI system...  

Anyway, I'll think more about this once 1.11 is out. Marking this for 1.12.

iota97:
Yeah better to wait for 1.12, I'm not eager to make change to input system close to a release eheh.

I build an APK for my testing, it have this PR, the touch gesture one and the tilt control one as well, if someone want to try I'll leave it [here](https://drive.google.com/file/d/1gpq-b5_MvYDNtV9lE8ysEP-kmnYFCxGF/view?usp=sharing) as feedback never hurt :3

Spicy-Kimchi:
Look great, now this is just my selfish request:

Could you add A B C D E icons pretty please? |ω・`)

iota97:
Adding new icon should be really straightforward once you have them in the UI atlas, seem like building the atlas tools on Linux (or any CMake platform) got really broken in the /ext/native code clean up tho'.

I can find a way but it will take some time I guess :)

iota97:
> Could you add A B C D E icons pretty please? |ω・`)

Done, [link](https://drive.google.com/file/d/1gpq-b5_MvYDNtV9lE8ysEP-kmnYFCxGF/view?usp=sharing) updated with the new version :)

hrydgard:
Hm, sorry I kinda forgot about this one. I'm open to merging this, but the rebase with all the collisions is probably not going to be fun :(

iota97:
Hope I didn't miss anything with all the control refractoring since this, but it seem to work fine. 

I have quite some problem with the atlas file as I can't really build the tool on linux (when all the file in /ext/ got moved that CMake stopped working at all, probably quite a pain to fix with next to none user base). I left the upstream atlas and included all the file to build the update one, could you please just run the atlas build script on your machine?

hrydgard:
Cool, thanks.

I'll do the atlas build tomorrow. (Really should get rid of it for regular UI images though... it's not really worth it for non-text rendering anymore, but that's another discussion....)

hrydgard:
Done and merged in https://github.com/hrydgard/ppsspp/commit/57b4cd394a69029dcb8943e1bad33a4d149dcf99. (Github doesn't pick it up because I rebased).

Gamemulatorer:
Combo button doesn't work now?
where is combo button now located 😕
Screenshot_2021-07-12-02-21-11-010_org ppsspp ppsspp

iota97:
"Customize Touch Controls" -> "Customize" -> tap the custom button you want to set. Basically where you enable button visibility. I find way less back and forward to set thing up like this, and having the old screen with 10 button was not really the best.

Gamemulatorer:
@iota97 tnx

PR #13875, opened 2021-01-05
Compat.ini Cleanup and Add games to RequireBufferedRendering

Panderner:
This PR is for removing unused options for games that does not affected and adding missing IDs for games

Removing unused options for games that doesn't affected
Luxor games doesn't need to add ForceMax60FPS it loads faster even I overclocked the CPU to 1000mhz 
Disabling RequireBlockTransfer for games added from BlockTransferAllowCreateFB because BlockTransferAllowCreateFB always enables Simulate block transfer effects even we disabled Simulate block transfer effects.

We're add games to RequireBufferedRendering that cause black screen when buffered rendering when it's not turned on


hrydgard:
Yeah, these changes look good.

PR #13878, opened 2021-01-05
Save on game start

### LINKED ISSUE #13710: About Scroll Memory on Game List - problem after task kill
### What happens?
@iota97 
@hrydgard 

Report!

If closed PPSSPP by task kill on Android then PPSSPP select the root directry of storage at next boot.

When in game & menu screen both.

### What *should* happen?
Is this a normally?
If you can fix this then PPSSPP should memory the scroll point in this case too.

iota97:
Should fix #13710, tho' I can't reproduce on my device sadly. 

If saving everything is too much let me know and I'll look on a way to save only 1 setting in the ini :)


hrydgard:
Eh, should be fine...

PR #13885, opened 2021-01-07
Android: Add button to easily switch to SD card directory.

hrydgard:
image

(just realized the Home button is redundant on the screenshot, will fix that).

This uses a variety of tricks to figure out where the additional SD card is. If one is available, show a button. 

Not tested on a lot of devices yet. There's also not yet a selector in case there are multiple SD cards (not sure how many devices support that....)

This probably won't work on Android 12 and later (see #11997 ) , but until then and on older devices it will be useful.

Fixes (or at least tries to) #10199, at least for some devices. Additionally adds graphics for an upcoming fullscreen toggle button.

Might help #13827 ?

Tested on:

* Pocophone F1
* Huawei Honor 7S
* ASUS Zenfone 5


hrydgard:
This is at a useful state now and doesn't crash on old phones anymore. I'm gonna merge.

PR #13894, opened 2021-01-08
[Adhoc Matching] Fix Lord of Arcana

anr2me:
- Fix possible of crash issue due to Out of userMemory when AdhocMatching Events piling up.
- Fix piling up AdhocMatching Events on Lord of Arcana

PS: There is an issue of invalid address causing the Client side to be stuck at Loading if the Host is entering the mission door first, so the Host need to enter last, may be some data didn't get synced when the host is entering first.

**Test builds:** (based on v1.10.3-1518-gd6b4e1f8a)
Win32&64: https://www.dropbox.com/s/b2mxcsw3w6gb9wq/PPSSPP_1.10.3-testbuild_Win32x64.zip?dl=0
Android(ARMv7): https://www.dropbox.com/s/329379pz4i7g0eo/PPSSPP_1.10.3-testbuild_ARMv7.apk?dl=0

### DIFF
--- a/Core/HLE/proAdhoc.cpp
+++ b/Core/HLE/proAdhoc.cpp
@@ -1203,18 +1203,24 @@ void notifyAdhocctlHandlers(u32 flag, u32 error) {
 // Note: Must not lock peerlock within this function to prevent race-condition with other thread whos owning peerlock and trying to lock context->eventlock owned by this thread
 void notifyMatchingHandler(SceNetAdhocMatchingContext * context, ThreadMessage * msg, void * opt, u32_le &bufAddr, u32_le &bufLen, u32_le * args) {
 	// Don't share buffer address space with other mipscall in the queue since mipscalls aren't immediately executed
-	MatchingArgs argsNew;
+	MatchingArgs argsNew = { 0 };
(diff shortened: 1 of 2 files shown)


Lemoncak3:
Can you look into it, i tried it today we can see each other at lobby and start a quest. But after that i can play it just fine but other players only got black screen.

ANR2ME:
That black screen is probably the "invalid address" i mentioned on my first post, this game is using a lot of AdhocMatching events rapidly, and i suspects that some issue might occur if an event's callback occurred at the wrong time due to long queue (ie. the game was expecting a certain events but a different one is being executed). 
So i'm planning to change the way it handles the events by not exiting the syscall yet until the event related to it is fully executed, at least on some of AdhocMatching syscall.

As i remembered there is a game who doesn't work well if the callback got executed too soon, since the game need to prepare something first (ie. the game is initializing a flag not long after calling AdhocMatching syscall where this flag will be used on the callback handler, otherwise the callback handler will ignore the event, thus getting an issue), i don't remember which game it was, hopefully i wrote the name of the game in one of my comment in the source code to avoid breaking that game with the new changes.

PR #13895, opened 2021-01-08
Use the same logic of game setting for main menu full screen

iota97:
Tested on Linux SDL and Qt


hrydgard:
Oh yeah, that's better. Forgot we had that message :)

PR #13898, opened 2021-01-09
Add Browse... button to Android game browser

hrydgard:
Should allow you to pick the SD card, working around #13827 until proper Storage Access Framework support is added.

Does some highly unsupported path mangling to convert a storage uri to a local folder...

### DIFF
--- a/Common/Net/URL.cpp
+++ b/Common/Net/URL.cpp
@@ -126,7 +126,7 @@ std::string UriDecode(const std::string & sSrc)
 	const unsigned char * const SRC_END = pSrc + SRC_LEN;
 	const unsigned char * const SRC_LAST_DEC = SRC_END - 2;   // last decodable '%' 
 
-	char * const pStart = new char[SRC_LEN];
+	char * const pStart = new char[SRC_LEN];  // Output will be shorter.
 	char * pEnd = pStart;
 
 	while (pSrc < SRC_LAST_DEC)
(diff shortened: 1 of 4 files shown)


unknownbrackets (commented):
How well does this local-only flag work?  Does it restrict well?

The "right" way ought to be this: https://developer.android.com/reference/android/content/ContentResolver.html#openAssetFileDescriptor(android.net.Uri,%20java.lang.String,%20android.os.CancellationSignal)

I'd imagine something like (psuedo):

```c++
// Java land - with null checks...
parcel = openAssetFileDescriptor(uri, "r").getParcelFileDescriptor();
int desc = parcel.getFd();

// C++ land
std::string old = getcwd();
fchdir(desc);
std::string path = getcwd();
chdir(old);

// Back to Java
parcel.close();

// In C++ land: use path.
```

Haven't tested it.  But that's what I'd hope for.  Maybe with the local flag that could make #11997 pretty okay?

Edit: The above is, by the way, a technique that some have used to escape sandboxes which you have to be careful about when forking a privileged process to a less privileged one.  So there might be restrictions (to mitigate security bugs) on it for Android, I don't know.

-[Unknown]

hrydgard:
Yes, something like that could work. Apparently we can even preserve the permission to the folder (although not across some events...).

The local restriction - when choosing folders, it seems to already be limited to local paths. Only when you choose a single or multiple files do you get to select stuff from Google Drive. 

So it seems to be maybe more workable than we thought...

PR #13905, opened 2021-01-10
FixPGF for Euro Characters.(Total Balance emphasis)

nassau-tk:
Fix for Euro Characters.


hrydgard:
Hm, some of the files shrunk suspiciously much in size?

nassau-tk:
Yes.
Because,l checked included characters per pgf by pgf-tool.
And,I deleted the not included characters from my pgf.
Maybe,Latin is most down.

If you think  some problem then you should not merge it. :smile:

hrydgard:
No, I just wanted to understand why they shrunk :)   That sounds reasonable, let's try it.

nassau-tk:
[My Note]

If I want to get characters positioning balance included Euro characters then a few games will appear the characters on bottom position more than the previous one.

Example) BOXER's Road2 The Real

The cause is because value of jpn0 is being used for value of Ascender & Descender on ltn0~7.
This is because the height of EURO character is abnormally high.
So, A margin in the upper part of the character spreads with that.

Result : Line spaces too much wide in a few games.

I thought about that "What is High priority? What's should fixed?" on my PGF.
And, I decide that "appear the special characters for German French Spanish Italian uniformly position".
If improvement method of an overall balance is found, it will fixed once again.

Euro special characters are really "troublesome".

nassau-tk:
Ooooops!!

I misstook understood correctly for Original PGF.

The orginal PGF is not perfect for all games.
Maybe, I will make re-fix & PR soon.

PR #13909, opened 2021-01-11
PGF Re-Fixed Euro Characters

nassau-tk:
I'm so sorry.
I misstook understood correctly for Original PGF.

The original PGF was not perfect for all games.
This is my answer by the my study.
So, I re-fixed my PGF files.

These PGF will become so good appearance for Euro Characters.
(And, some other little fix included.)


hrydgard:
No need to be sorry - we all learn!

nassau-tk:
Thank you!! :smile:

PR #13911, opened 2021-01-12
Enable BlockTransferAllowCreateFB for Gradius Collection

Panderner:
This PR increases performance for Gradius Collection to all games (except Gradius IV).

If I disable Simulate block transfer effects for all games except Gradius IV the video freezes.

Unfortunately all games are extremely lag for lower-end devices however Gradius IV works fine for 1.5 GHz quad core Mali-400MP2 device.


hrydgard:
The game looks exactly the same with this change, and you've played more than a few levels? If so I'm happy to merge.

Panderner:
Oh yes there's a difference that BlockTransferAllowCreateFB disables linear filtering for upscaled resolutions for Gradius 1, Gradius 2 and Gradius Gaiden.

BlockTransferAllowCreateFB off: 
Screenshot_2021-01-17-18-42-17-889_org ppsspp ppsspp

BlockTransferAllowCreateFB on: 
Screenshot_2021-01-17-18-41-10-685_org ppsspp ppsspp

hrydgard:
Oh, hm, that's interesting. Might be a bug, not sure? Either way, if that's the only difference, I think we're good.

Panderner:
This PR is now merged can we revert #13929?

unknownbrackets:
If it's drawing at a higher resolution and aniso is on, those results are expected right?

It does require block transfer, doesn't it?

-[Unknown]

Panderner:
> If it's drawing at a higher resolution and aniso is on, those results are expected right?
> 
> It does require block transfer, doesn't it?
> 
> -[Unknown]

It only used for Gradius 1, Gradius 2 and Gradius Gaiden. It does not require Block transfer for Gradius 3 and 4. and now Ys Seven is no longer require to turn on block transfer effects due to BlockTransferAllowCreateFB is enabled

PR #13920, opened 2021-01-15
Fix GCC9 Crash

liberodark:


hrydgard:
That's interesting. Does the crash also go away if you turn optimization down to -O2? 

I don't really know why we use -O3 anyway, it's not likely to be a large boost.

liberodark:
This problem is on some SBCs.
Personally I haven't used -O2 but I think one of my colleagues there already did.

orbea:
> I don't really know why we use -O3 anyway, it's not likely to be a large boost.

Using `-O3` can expose experimental features in compilers, the safer default is `-O2`.

CMakeLists.txt
```diff
@@ -281,11 +281,11 @@ if(NOT MSVC)
 
 	set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -g -D_DEBUG")
 	set(CMAKE_CXX_FLAGS_MINSIZEREL "${CMAKE_CXX_FLAGS_MINSIZEREL} -Os -D_NDEBUG")
-	set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -O3 -D_NDEBUG")
+	set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -O3 -fno-tree-slp-vectorize -D_NDEBUG")
```
orbea, line 284: This is the wrong place for compiler flags and it should be tested during configure if the compiler actually supports it.

I also suspect this doesn't fix the root issue which is probably undefined behavior in ppsspp or a compiler bug.

orbea (changes requested):

orbea:
@liberodark I would suggest you spend some time narrowing this down so a better fix or more specific workaround can be found. 

First set `-fno-tree-slp-vectorize` for every file and then remove the flag for half off them, then repeat until its narrowed down to specific file(s). Next disable it with a gcc pragma until it narrowed down to specific code which may be in the included headers.

For example here was an old workaround for a now fixed gcc7 bug.

https://github.com/libretro/bsnes-mercury/pull/50/files

And when it was fixed in gcc.

https://github.com/libretro/bsnes-mercury/pull/71/files

rtissera:
We had a ppsspp crash with O3 GCC 9.3.0 on some ARM devices (RK3326 at least) on PPSSPP for batocera.linux 

I bissected all individual flags between O2 to O3 to narrow down the problematic optimization at a global level.

hrydgard:
Well, now that you have narrowed down the optimization, it would be even more useful if you could narrow it down to the file. It could be a bug in PPSSPP, just as well as it could be an optimizer bug.

But I think I'm just gonna bump it down to -O2.

rtissera:
Ok let me some time to find down in our git history the affected platform
and games to reproduce and get a better scope of the issue.

Le dim. 17 janv. 2021 à 10:42, Henrik Rydgård <notifications@github.com> a
écrit :

unknownbrackets:
Had any luck narrowing this down to specific files or groups thereof?

-[Unknown]

hrydgard:
Is this still an issue?

Can you get a backtrace of the crash?

hrydgard:
No feedback in quite a while, closing.

PR #13922, opened 2021-01-16
Make sure we don't try to set a negative viewport size.

hrydgard:
Don't have a repro for #13921, but should fix it for now. 

Needs more investigation later, but it might be just be a precision issue in all those clips etc.


Panderner:
Unfortunately it getting worse for Gradius Collection and it started up flickering loading screen.

Issue #13910, opened 2021-01-11
Rockman Dash 1 and 2 graphics broken since v1.10.3-1000-g349fc7071


Panderner:
This issue has graphics glitches for Rockman Dash games for Vulkan.

Rockman Dash 1:
![Screenshot_2021-01-12-07-40-42-992_org ppsspp ppsspp](https://user-images.githubusercontent.com/37503397/104251492-d0d92f80-54aa-11eb-8ab9-bd653dd6bc03.jpg)
![Screenshot_2021-01-12-07-47-05-463_org ppsspp ppsspp](https://user-images.githubusercontent.com/37503397/104251511-dafb2e00-54aa-11eb-8199-36cd3be02946.jpg)
![Screenshot_2021-01-12-07-47-14-876_org ppsspp ppsspp](https://user-images.githubusercontent.com/37503397/104251526-e2bad280-54aa-11eb-96c4-0821300b7a78.jpg)

Rockman Dash 2:
![Screenshot_2021-01-12-07-40-54-143_org ppsspp ppsspp](https://user-images.githubusercontent.com/37503397/104251566-f82ffc80-54aa-11eb-843d-65d5b1d06dfa.jpg)
![Screenshot_2021-01-12-07-41-13-345_org ppsspp ppsspp](https://user-images.githubusercontent.com/37503397/104251578-ffefa100-54aa-11eb-969a-f41821bb5a4f.jpg)
![Screenshot_2021-01-12-07-47-31-385_org ppsspp ppsspp](https://user-images.githubusercontent.com/37503397/104251586-0716af00-54ab-11eb-85f3-144a08697363.jpg)

Last build work: v1.10.3-987-g327eb21e0
First bad: v1.10.3-1000-g349fc7071

ANR2ME:
I think you need to mention your device model too, just in case it's specific to certain GPU brand/type.

Panderner:
I'm using POCO M3 and here's a GE Dumps: 
[Rockman Dash 1.zip](https://github.com/hrydgard/ppsspp/files/5799279/Rockman.Dash.1.zip)
[Rockman Dash 2.zip](https://github.com/hrydgard/ppsspp/files/5799280/Rockman.Dash.2.zip)

Panderner:
This only happens in Vulkan but it does not happen in OpenGL

ANR2ME:
[...]

PR #13928, opened 2021-01-17
Adds two new tests to GPU driver test screen: Adreno shader logic test and flat shading

hrydgard:
The adreno test tests for a serious shader compiler bug mentioned in #13910.
Basically it miscompiles ``if (color.r < 0.002 && color.g < 0.002 && color.b < 0.002) discard;``.

Access through Settings/Tools/Developer Tools/GPU Driver Test/Shader.

Bad (adreno test):

image

Good (adreno test):

image

Very clear repro on Adreno 630, Pocophone F1.

The flat shading test is an untested attempt at a repro of #13451 (corrupt rendering of flat shading on iOS 14.x).

(will test that soon).

### DIFF
--- a/Common/GPU/OpenGL/GLFeatures.cpp
+++ b/Common/GPU/OpenGL/GLFeatures.cpp
@@ -594,6 +594,8 @@ std::string ApplyGLSLPrelude(const std::string &source, uint32_t stage) {
 	if (!gl_extensions.IsGLES && gl_extensions.IsCoreContext) {
 		// We need to add a corresponding #version.  Apple drivers fail without an exact match.
 		version = StringFromFormat("#version %d\n", gl_extensions.GLSLVersion());
+	} else if (gl_extensions.IsGLES && gl_extensions.GLES3) {
+		version = StringFromFormat("#version %d es\n", gl_extensions.GLSLVersion());
 	}
 	if (stage == GL_FRAGMENT_SHADER) {
 		temp = version + glsl_fragment_prelude + source;
--- a/Common/Render/DrawBuffer.cpp
+++ b/Common/Render/DrawBuffer.cpp
@@ -236,6 +236,18 @@ void DrawBuffer::DrawImageStretch(ImageID atlas_image, float x1, float y1, float
 	V(x1,	y2, color, image->u1, image->v2);
 }
 
+void DrawBuffer::DrawImageStretchVGradient(ImageID atlas_image, float x1, float y1, float x2, float y2, Color color1, Color color2) {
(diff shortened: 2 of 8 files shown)


unknownbrackets:
Is the flat shading bug that it doesn't draw at all, that it uses the wrong provoking vertex, or that it ignores it and uses both colors?

-[Unknown]

hrydgard:
When it breaks, it's worse than any of those - instead of drawing a textured polygon it just fills with a single seemingly unrelated color covering the whole polygon.

I can't seem to get it to break under controlled conditions in the test though, but it breaks horribly in a few games and all the breakage goes away when I remove the flat modifier.

unknownbrackets:
I wonder if it's related to another feature, such as an alpha test (since that's common) or lighting or something.

One time created an *extremely* hacky tool to "edit" framedumps and trim things off to narrow down a bug like this.  Maybe exploring that further could be helpful...

-[Unknown]

hrydgard:
Nope, the weirdest thing is that drawcalls this happens to are super simple, no extra state. One example is the intro video in FF: Crisis Core. No alpha blend, no depth or stencil, no discard. So my little test quad should be a very close replication, I even rearranged the test shader to look almost the same... 

(Replicating it this way, if I can, is better for the driver bug report...)

hrydgard:
Even though it currently fails to reproduce the flat shading error, I'm gonna merge this in anyway, just for the Adreno test.

unknownbrackets:
Maybe it has to do with a previous draw or later draw corrupting things in combination with the flat shading.

If ALL draws are forced to flat shading, does it still happen?

-[Unknown]

hrydgard:
Hm, that's actually a good question...

xsm1997:
image
image
A screenshot from MI10T (Android 11, MIUI12.5, Snapdragon 865) and another screenshot from iPad mini 5 (iPadOS 14.3, Apple A12).
There is no difference between the two devices on Flat shaded tex as far as I have seen, but the iPad mini 5 is affected by the flat bug.
It seems that the test can not reproduce the bug.

hrydgard:
Yes, I have not succeeded in reproducing the iOS flat bug in controlled conditions, it only reproduces during gameplay.

unknownbrackets:
It reproduces in a dump right?  We could add a subdomain (maybe for #11888 anyway) with frame dumps and register the app to open them.  Would that be a viable reproduction?

-[Unknown]

hrydgard:
Yeah, I might try to do something like that. Yes, they would strongly prefer a reduction, but this might be a viable way.

Issue #11055, opened 2018-05-20
Earth Saver Plus - Inseki Bakuha Daisakusen (Japan) (PSN) text problem


benderscruffy:
PPSSPP v1.5.4-993-gab65c5210 Windows 64 bit
during the game the text is way to big to fit in the screen
i am using my own fonts off my own psp
here is a screen from jpcsp
![jpcsp screen](https://user-images.githubusercontent.com/2696404/40277158-2cd37e70-5c4d-11e8-862b-c13a4b1f871b.PNG)
here it is from ppsspp
![ppsspp screen](https://user-images.githubusercontent.com/2696404/40277160-35fa5384-5c4d-11e8-9253-e7b14ac8a1d7.PNG)
[ppsspplog.zip](https://github.com/hrydgard/ppsspp/files/2020180/ppsspplog.zip)

benderscruffy:
i am using the fonts from the official firmware

benderscruffy:
v1.7.1-210-g9bda7cb64
still happens

benderscruffy:
one other thing i noticed is during gameplay there is black lines 
happens in both DX11 and opengl
![earth1](https://user-images.githubusercontent.com/2696404/49129832-3fc78e80-f30c-11e8-9d0a-1aa3e73b6d45.png)
[DUMP.ZIP](https://github.com/hrydgard/ppsspp/files/2622928/DUMP.ZIP)

benderscruffy:
Snake Warriors - Training (USA) (minis) (PSN) has the same problem
![snake](https://user-images.githubusercontent.com/2696404/71535716-e7b1f000-2942-11ea-880f-af851b53c0dc.PNG)

unknownbrackets:
> Loading module sceFont_Library with version 0101, devkit 06020010

> sceFontNewLib(088a8660, 09ffdea0)
> sceFontGetNumFontList(0993a7ac, 09ffdea0)
> sceFontGetFontList(0993a7ac, 09ffdea4, 18)
> sceFontFindOptimumFont(0993a7ac, 09ffebc4, 09ffdea0)
> sceFontOpen(993a7ac, 0, 0, 9ffdea0)
> sceFontGetFontInfo(993a7f8, 9ffec6c)
> sceFontClose(993a7f8)
> sceFontDoneLib(0993a7ac)

> sceFontNewLib(088a8660, 09ffd6d0)
> sceFontGetNumFontList(09c8718c, 09ffd6d0)
[...]

PR #13948, opened 2021-01-19
PGF Fixed Bold & Italic property and camouflage the Font name.

nassau-tk:
See #11055 .

My PGF was not setting Property of Bold & Italic.
- Because, I didn't know that these property and font name was use in program.
- Of course, I'm understanding that is not absolutry.

And, I camouflaged to font name for "Original PGF font name".
jpn0.pgf : FTT-NewRodin Pro DB
kr0.pgf   : AsiaKNHH-SONY-uni
ltn0/2/4/6/8/10/12/14.pgf  :  FTT-NewRodin Pro Latin
ltn1/3/5/7/9/11/13/15.pgf  :  FTT-Matisse Pro Latin


nassau-tk:
>Some checks were not successful
15 successful, 1 failing, and 1 skipped checks

Is it OK?

hrydgard:
Yes, seems to be some temporary problem with the Qt build.

PR #13951, opened 2021-01-20
Try to improve sceMp3Decode

sum2012:
From Hanayaka Nari Wa ga Ichizoku  JPCSPTrace log https://gist.github.com/sum2012/e2d513c55342c40f5e12

But don't help the game #5213

### DIFF
--- a/Core/HLE/sceMp3.cpp
+++ b/Core/HLE/sceMp3.cpp
@@ -157,8 +157,6 @@ void __Mp3DoState(PointerWrap &p) {
 }
 
 static int sceMp3Decode(u32 mp3, u32 outPcmPtr) {
-	DEBUG_LOG(ME, "sceMp3Decode(%08x,%08x)", mp3, outPcmPtr);
-
 	AuCtx *ctx = getMp3Ctx(mp3);
 	if (!ctx) {
 		ERROR_LOG(ME, "%s: bad mp3 handle %08x", __FUNCTION__, mp3);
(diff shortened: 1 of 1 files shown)


Core/HLE/sceMp3.cpp
```diff
@@ -167,10 +165,12 @@ static int sceMp3Decode(u32 mp3, u32 outPcmPtr) {
 		
 	int pcmBytes = ctx->AuDecode(outPcmPtr);
 	if (pcmBytes > 0) {
+		DEBUG_LOG(ME, "%d = sceMp3Decode(%08x,%08x)", pcmBytes, mp3, outPcmPtr);
 		// decode data successfully, delay thread
 		return hleDelayResult(pcmBytes, "mp3 decode", mp3DecodeDelay);
 	}
-	return pcmBytes;
+	ERROR_LOG(ME, "%d = sceMp3Decode(%08x,%08x)", 0x807f00fd, mp3, outPcmPtr);
+	return ERROR_AVCODEC_INVALID_DATA;
```
unknownbrackets, line 173: The only negative case right now would be invalid address:
https://github.com/hrydgard/ppsspp/blob/b76e906ed0960a4a352ad9eb291cf0558637193b/Core/HW/SimpleAudioDec.cpp#L332

This probably won't change anything.  If we're seeing the JpcspTrace return an error it may mean we need better validation.

-[Unknown]

sum2012: JPCSP also don't have have this error code in sceMp3Decode.
I should wait it first.

ANR2ME: According to this http://lukasz.dk/mirror/forums.ps2dev.org/viewtopic108b.html?t=10798
Error 0x807F00FD might be related to ID3 Tag
image

Gamemulatorer:
How about this now?

unknownbrackets:
I think the issue was fixed by #13981, and the code currently incorporates a similar change.  Thanks for helping getting this fixed - I'm gonna close this PR.

-[Unknown]

PR #13965, opened 2021-01-24
Fix Greek characters & Roman numbers & all balance on jpn0.pgf

nassau-tk:
「Greek」
image

「Roman Numbers」
image


nassau-tk:
I don't know that use Greek & Roman numbers in what games.

But, Some time use both characters in Japan.
The fix is basic style same as Latin fonts.

Except: Russian (monospaced font)

PR #13967, opened 2021-01-24
[adhoc] Fix The Warriors

anr2me:
Apparently there are games (ie. The Warriors) who tried to do PtpSend immediately after PtpOpen, without trying to do PtpConnect first.


hrydgard:
Do we have an issue about this, to tag for closing? 

Looks okay, although it seems very odd to call send without connect.

ANR2ME:
Yeah, i still need to test the behavior of PtpOpen to see whether it's possible to send without connecting first. 
Currently this commit is the only way to make The Warriors to work properly, and shouldn't have any impact on other games.

ANR2ME:
After testing it on a small homebrew, looks like the error code returned from PtpSend was wrong, it should be "disconnected" instead of "not connected"

ANR2ME:
Okay i'm done

hrydgard:
Oh, that's making more sense indeed.

PR #13983, opened 2021-01-27
Remove duplicates from ThreadManForKernel

Florin9doi:


sum2012:
The game I test maybe give 2 difficult of upper case and lower of 0X28B6489C ?

sum2012:
The log is from
https://github.com/hrydgard/ppsspp/issues/13205
18:22:889 hle\scekernelmodule.cpp:751 W[LOADER]: Unknown syscall in known module 'ThreadManForKernel': 0x28b6489c

sum2012:
I will test again when new build produce

Florin9doi:
#13205 was closed by #13217

hrydgard:
hex numbers don't have upper and lower case, only their representation in c++ source code :P  still the same number.

sum2012:
I test again that game.No more that log now.

PR #13992, opened 2021-01-28
Fix Russian characters ltn0~7.pgf

nassau-tk:
Fix Greek characters ltn0~7.pgf


nassau-tk:
Oops!

Greek ==> Russian is right!

PR #13994, opened 2021-01-29
FindFFmpeg: Fix a few issues

glebm:
1. postproc now looks for `postprocess.h` (there is no `postproc.h` header).
2. pkg-config fallback condition now works (`find_path/library` set the variable to `${var}-NOTFOUND` but it was checking for an empty string).


hrydgard:
Looks good, thanks!

Issue #13667, opened 2020-11-16
Dynasty Warriors Multi Raid 2 (真·三国无双：联合作战2)Online questions


xiaolang0668:
![真三国无双2](https://user-images.githubusercontent.com/74450832/99263854-ad9f5480-285a-11eb-8e33-6d5951ad3c34.jpg)

Test dynastyy warriors multi raid 2 game, only 2 people successfully enter the battlefield, more than three people enter the battlefield directly offline, hope the gods repair.

xiaolang0668:
![aaa](https://user-images.githubusercontent.com/74450832/99265987-4505a700-285d-11eb-931f-ea3fd58bcb2a.jpg)
![bb](https://user-images.githubusercontent.com/74450832/99266124-6797c000-285d-11eb-857c-2a9b98f5951d.jpg)


> 
> Test dynastyy warriors multi raid 2 game, only 2 people successfully enter the battlefield, more than three people enter the battlefield directly offline, hope the gods repair.

MojoJojoDojo:
Huh what is this issue about lol?
You cant play the game with over 2 players?

xiaolang0668:
> Huh what is this issue about lol?
> You cant play the game with over 2 players?

Three players have an invisible task, leave one, can not see the task of the player, can not see, more than three players, after the start of the task into the field, not all offline, only two players online, the rest are offline.

xiaolang0668:
![20201117011843](https://user-images.githubusercontent.com/74450832/99285825-05e25080-2873-11eb-9617-5c878af95ef6.jpg)
![20201117011708](https://user-images.githubusercontent.com/74450832/99285840-0bd83180-2873-11eb-89e3-3428d6265476.jpg)


![dxl0](https://user-images.githubusercontent.com/74450832/99286562-00393a80-2874-11eb-9bef-8528ecd6742c.gif)

ANR2ME:
@xiaolang0668 have you tried to Enable "Forced First Connect" in Networking settings?

xiaolang0668:
> @xiaolang0668 have you tried to Enable "Forced First Connect" in Networking settings? on all players

Whether the test switch is opened or not is dropped.

xiaolang0668:
![22](https://user-images.githubusercontent.com/74450832/99538478-fb959300-29e7-11eb-9117-261f61e26f87.png)
Four people can not receive tasks in normal synchronization, and there will be problems when they enter the game.

[...]

PR #13997, opened 2021-01-29
[Adhoc] Fix connection issue on Dynasty Warriors (Shin Sangoku Musou) games

anr2me:
Fix connection issue on Dynasty Warriors (Shin Sangoku Musou) games when playing with more than 2 players. https://github.com/hrydgard/ppsspp/issues/13667

Not sure whether this is the correct behaviour of `sceNetAdhocPtpOpen` or not, may need to test a small homebrew to simulate 3 players.

### DIFF
--- a/Core/HLE/proAdhoc.cpp
+++ b/Core/HLE/proAdhoc.cpp
@@ -128,16 +128,19 @@ bool isPDPPortInUse(uint16_t port) {
 	return false;
 }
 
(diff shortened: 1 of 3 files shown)


ANR2ME:
Apparently it's allowed to bind the same [local] port using PtpOpen as long the destination mac address or destination port are different that existing one.

PR #14001, opened 2021-01-29
CMake: Add USE_UBSAN

glebm:
1. Adds an option to use the undefined behaviour sanitizer.
2. Cleans up the ASAN configuration.

UBSAN is kept debug-only, similar to ASAN, though there is no real reason to do this as they both work just fine in release modes.


glebm:
Does not link with `-DUSE_UBSAN=ON`:

```
/usr/bin/ld: lib/libCommon.a(ShaderTranslation.cpp.o):(.data.rel+0x18): undefined reference to `typeinfo for glslang::TShader'
```

glebm:
Figured it out

glebm:
UBSAN found 2 UBs on startup: #14002 #14003

glebm:
Most of the UB found by UBSAN immediately are alignment issues.

While x86(_64) is quite lenient towards these (but not in SIMD!), ARM CPUs do not like them at all

hrydgard:
Actually it's not true that it's insta-crash on ARMv7, those generally have no problems with scalar load/store with misalignment. ARMv5 and earlier is another story though.

Still, of course, good to fix these.

glebm:
Yeah that's what I meant with "before ARMv7". MIPS also crashes (e.g. jz4770).

This is not entirely benign even with CPUs that support unaligned access, because none of them support unaligned SIMD instructions. With `-O3`, the compiler can auto-vectorize code, which then crashes.

hrydgard:
Oh sorry I misread your post. But ARMv6 also handles scalar misalignment fine.

Modern x86 and ARM64 actually both support misaligned SIMD load/stores, although with limitations. Only in the last few generations have x86 allowed unaligned SIMD memory operands.

PR #14028, opened 2021-01-30
Only force the log level if it set via an argv

glebm:
Follow-up to #14019

@unknownbrackets


hrydgard:
Oops again. I need to review more carefully.

unknownbrackets (approved):

PR #14060, opened 2021-02-04
[Adhoc] Fix possible lock issue and Updated PdpStat

anr2me:
- Fix possible lock issue during AdhocMatchingStart
- Updated PdpStat to prevent rcv_sb_cc from exceeding the buffer size arg (since we use larger buffer size to prevent micro stutters or disconnection issue due to too many dropped packets with small buffer size).
TODO: May need to improve it to be able to calculate the correct size if there are multiple datagram messages

### DIFF
--- a/Core/HLE/proAdhoc.cpp
+++ b/Core/HLE/proAdhoc.cpp
@@ -1896,13 +1896,35 @@ uint16_t getLocalPort(int sock) {
 	return ntohs(localAddr.sin_port);
 }
 
-u_long getAvailToRecv(int sock) {
+u_long getAvailToRecv(int sock, int udpBufferSize) {
 	u_long n = 0; // Typical MTU size is 1500
+	int err = -1;
 #if defined(_WIN32) // May not be available on all platform
-	ioctlsocket(sock, FIONREAD, &n);
+	err = ioctlsocket(sock, FIONREAD, &n);
 #else
-	ioctl(sock, FIONREAD, &n);
+	err = ioctl(sock, FIONREAD, &n);
 #endif
+	if (err < 0)
+		return 0;
+
+	if (udpBufferSize > 0 && n > 0) {
+		// Cap number of bytes of full DGRAM message(s?) up to buffer size
+		static int lastUdpBufSize = 0;
+		static char* buf = NULL;
+		if (udpBufferSize > lastUdpBufSize) {
+			// Reusing temp buffer to prevent causing too many fragmentation due to repeated alloc -> free (was getting out of memory issue)
+			char *tmp = (char*)realloc(buf, udpBufferSize);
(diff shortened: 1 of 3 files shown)


Core/HLE/proAdhoc.cpp
```diff
@@ -1896,13 +1896,35 @@ uint16_t getLocalPort(int sock) {
...
 #endif
+	if (err < 0)
+		return 0;
+
+	if (udpBufferSize > 0 && n > 0) {
+		// Cap number of bytes of full DGRAM message(s?) up to buffer size
+		static int lastUdpBufSize = 0;
+		static char* buf = NULL;
+		if (udpBufferSize > lastUdpBufSize) {
+			// Reusing temp buffer to prevent causing too many fragmentation due to repeated alloc -> free (was getting out of memory issue)
```
hrydgard, line 1915: if you were really running out of memory, you would have been leaking. I don't think fragmentation here would cause such serious issues..

hrydgard:
What games have issues that this will fix? Just curious.

ANR2ME:
Currently there are no game that have issue with current socket stats, just trying to get the stat data correctly.

There is one game i know that relies on the number of byte available to receive and affects the FPS, which is Warriors Orochi 2.
This game will wait until there are not enough data to be received before rendering new frame, otherwise it will keeps receiving in a loop to drains the buffer, so if the buffer is too large and filled up it could cause the FPS & VPS to drops to 0/0.

My previous solution to prevent the FPS/VPS from dropping too much was by reducing the buffer size, but apparently VPN doesn't worked well with small buffer size (may be due to higher latency it need to be able to buffer more data).
Capping the number of bytes up to the buffer size arg set by the game is just an extra prevention of FPS drops on games that relies on the `rcv_sb_cc` value like Warriors Orochi 2.

Core/HLE/proAdhoc.cpp
ANR2ME: It might not be fragmentation, i only gets out of memory when debugging it through Visual Studio 2019, may be it's VS bug on memory allocation?
It's just a simple:
```c++
char *buf = (char*)malloc(size);
err = recvfrom(...);
free(buf);
```
So i can't really tell where the leaks came from, and i certainly don't want to get issue while debugging.

hrydgard: That's really odd, something like that shouldn't run of RAM.

However, if realloc fixes it, let's go with that, I guess..

ANR2ME: Hmm... i think we don't really need to provide a large buffer since we only need to retrieved the message size, and it seems with `MSG_TRUNC` we can provide small buffer and it will still return the actual message size (one message)

PR #14063, opened 2021-02-05
[Adhoc] Implement blocking AdhocMatchingStop

anr2me:
Implement blocking AdhocMatchingStop to prevent possible data loss and PSP memory leaks due to unprocessed pending Matching Event's callbacks.

PS: Currently this PR is making Lord of Arcana worse than already is compared to v1.10.3-1731-g5babc1af2 :( 
Will need to be tested whether it's causing regression or improving other games.

Test builds:
------------
Win32/x64: https://www.dropbox.com/s/b2mxcsw3w6gb9wq/PPSSPP_1.10.3-testbuild_Win32x64.zip?dl=0
Android(ARMv7): https://www.dropbox.com/s/329379pz4i7g0eo/PPSSPP_1.10.3-testbuild_ARMv7.apk?dl=0


ANR2ME:
i'll close this for now since it causes regression on some games.

PR #14089, opened 2021-02-09
GL: Call CreateDeviceObjects *after* updating render_.

hrydgard:
Also remove a redundant call to fetch render_ in the FramebufferManager constructor, it's already passed in.

Should fix another crash in https://github.com/hrydgard/ppsspp/issues/14082#issuecomment-775457717 .


hrydgard:
Argh, Qt build broken again due to servers or whatever...

PR #14126, opened 2021-02-12
Fix many marks of jpn0.pgf.

nassau-tk:
I fixed some major symbols on jpn0.pgf.
Circle,square,triangle,numbers in the circle,etc...and more symbols.


nassau-tk:
Sorry, @hrydgard 
I misstook to upload file.
I will Re-upload from other branch.

PR #14170, opened 2021-02-17
Fix "Enable chat" in Linux Qt build

TotalCaesar659:
Same as https://github.com/hrydgard/ppsspp/pull/14165 but in Linux Qt build.


unknownbrackets (approved):

clort81:
I opened chat by accident and now the "Chat 0" at bottom of screen won't go away.

How can I make it go away? :(

PR #14235, opened 2021-02-28
[Adhoc] Always enable TCPNoDelay to improve response time

anr2me:
Removing TCPNoDelay option from Network Settings, and always enable it to improve response time.


hrydgard:
Removing unnecessary options is always nice :)

PR #14261, opened 2021-03-04
Allow polygon upscaling in Boku no Natsuyasumi series 

### LINKED ISSUE #14259: Boku no natsuyasumi (UCJS10038) rendering problem
### What happens?

The game isn't being rendered right, rendered resolution is set to x5 but the game has still the same resolution.

### What *should* happen?

The game should improve render right the 3D models to up the x1 original

### Rendering backend used (Direct3D 9, Direct3D 11, Vulkan, OpenGL)

No matter which one you choose, they all work the same 

### LINKS

https://forums.ppsspp.org/showthread.php?tid=5634
https://puu.sh/dPW5F/1a0f170ed6.png

LunaMoo:
There is some minor block transfer glitch with those games(tested only 1st), but it also happens without IntraVRAMBlockTransferAllowCreateFB hack, so I'll open a new issue for it if it wasn't already existing. This fixes #14259 at least as much as it can be "fixed" since most of the game is still made of pre-rendered backgrounds in a PSOne resolution stretched to PSP which can't be pretty.

### DIFF
--- a/assets/compat.ini
+++ b/assets/compat.ini
@@ -673,6 +673,10 @@ ULJM06218 = true
 UCAS40346 = true
 UCAS40347 = true
 
+# Boku no Natsuyasumi: Mushi Mushi Hakase to Teppen-yama no Himitsu!!
+UCJS10038 = true
+UCJS18013 = true
+
 # Note! This whole flag is disabled temporarily by appending "Disabled" to its name). See 7914
 [YugiohSaveFixDisabled]
 # The cause of Yu-gi-oh series 's bad save (cannot save) are load "save status" and use cwcheat,


hrydgard:
If the IntraVRAMBlockTransferAllowCreateFB works, prefer that since it's more restrictive.

LunaMoo:
Ok, going to check, I don't really know the difference between them, also will test other 2 games if they really have same problem, I only remembered they looked same on PSP.

LunaMoo:
It works same with the IntraVRAMBlockTransferAllowCreateFB, also other 2 games didn't need this to begin with, even through they look alike.

hrydgard:
What these do is to if we detect that the game wants to transfer a block of pixels to some memory that we have not yet detected to be a "framebuffer", we create one and do the copy on the GPU, instead of reading the image back to memory using the CPU. 

The "IntraVRAM" one only allows this when both ends of the copy are in VRAM, which might help prevent accidentally capture instances when the game actually wants to later read the image using the CPU (in such cases, games usually block transfer directly to main RAM instead of VRAM).

hrydgard:
Seems you managed to lose most of the changes_

LunaMoo:
Yeah as I said, I tested other 2 games and they don't need it. They also don't suffer from the small issue I mentioned, maybe both of those things are also affected/solved by "Hook_bokunonatsuyasumi4_download_frame" which they are using, would frame download work instead of this IntraVRAMBlockTransferAllowCreateFB hack? I could check for it, through, last time I used known functions, it wasn't reliably saving hashes.

hrydgard:
Nah I think those things are separate.

PR #14291, opened 2021-03-14
Fix character's Side Space ltn0.pgf/ltn4.pgf

nassau-tk:
I found the characters side space problem.
Like a GranTurismo(English).


hrydgard:
Got before/after screenshots? Just curious what changed.

nassau-tk:
Yes sure!

[ltn0 Now]
image

[ltn0 Fix]
image

Side bearings is too much troublesome setting.
If I setting narrow then characters will overwrap like above.
But, If I setting over wide then characters goes out of screen or text window.
I feel that this case is too much overwrap.
So, I fix side bearings more wide.
(It is overwrap a little yet. I do that it intentional this.)

hrydgard:
Looks great!

PR #14313, opened 2021-03-23
Add a setting for choosing background animation in PPSSPP's menus

hrydgard:
Off and Floating Symbols are the only choices so far, might add more in the future.

Was requested as a way to disable the symbols, which I agree can be a bit distracting. Figured might as well set it up as a multi-choice from the start.

Will need a small translation update, since I reorganized System settings slightly ("UI Language" heading is changed to "UI")

image

### DIFF
--- a/Core/Config.cpp
+++ b/Core/Config.cpp
@@ -494,6 +494,8 @@ static ConfigSetting generalSettings[] = {
 #endif
 	ConfigSetting("InternalScreenRotation", &g_Config.iInternalScreenRotation, ROTATION_LOCKED_HORIZONTAL),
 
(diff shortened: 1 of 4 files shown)


LunaMoo:
@hrydgard would it be possible to support external shaders as UI animation? Like a post process, but on top of the background while under text etc.. There's tons of animation examples in https://www.shadertoy.com/ and those for the most part can be easily translated to PPSSPP inputs.

hrydgard:
Yes, that would be possible. Feel free to implement ;)

LunaMoo:
Not sure if I'll be able to, but will definitely try once I get some free weekend. I tried to do it once in the distant past, but just replaced the shader that draws everything with a post process animation and it went on top of everything...^_^;

PR #14330, opened 2021-03-31
Set ent_top/ent_size/stub_top/stub_size of struct NativeModule

Linblow:
Members `ent_top`, `ent_size`, `stub_top` and `stub_size` of the internal `struct NativeModule` were not set in function `__KernelLoadELFFromPtr`. Previously, they were only initialized to zero.

The following PSP kernel functions return a `SceModule *` corresponding to the internal `struct NativeModule`:
`sceKernelFindModuleByAddress` 
`sceKernelFindModuleByName` 
`sceKernelFindModuleByUID` 

The aforementioned members may be needed by the caller of the functions listed above.


hrydgard:
Good find! Seems likely to be right.

PR #14332, opened 2021-04-01
Feature: reducehash configure different value for different dump dimensions

Banh-Canh:
Hello !

Basically what I said here some time ago : https://forums.ppsspp.org/showthread.php?tid=27097, I was wondering if it could interest anyone else here ?
I added the following settings to the textures.ini by imitating how the [hashranges] worked.
```
[options]
version = 1
hash = xxh64
ignoreAddress = true
reduceHash = true
reduceHashGlobalValue = 0.5
;; if reduceHash = true, then it makes ppsspp assume only the top half is important (by default) 
;; I tried to make it configurable.

[reducehashrange]
;; if reduceHash = true, then it allows to configure reduceHashSize that is specific to the specified dimensions
;; syntax : w,h = reducehashvalue
128,256 = 0.3125
;; resulting dump will be named like : 128x256_addresscluthash.png
```

It is not going to be used a lot but it worked for me (with Grownlanser WoT) and I was able to greatly reduce the amount of duplicates textures dumps (by using the above settings) without causing issues (so far, after multiple playthrough).
It did not mess with other games I use with reduceHash=true (Persona 2 IS) or with reduceHash=false (Persona 3).

### DIFF
--- a/Core/TextureReplacer.cpp
+++ b/Core/TextureReplacer.cpp
@@ -77,10 +77,12 @@ bool TextureReplacer::LoadIni() {
 	aliases_.clear();
 	hashranges_.clear();
 	filtering_.clear();
+	reducehashranges_.clear();
 
 	allowVideo_ = false;
 	ignoreAddress_ = false;
 	reduceHash_ = false;
+	reduceHashGlobalValue = 0.5;
 	// Prevents dumping the mipmaps.
 	ignoreMipmap_ = false;
 
@@ -192,6 +194,14 @@ bool TextureReplacer::LoadIniValues(IniFile &ini, bool isOverride) {
 		}
 	}
 
+	if (ini.HasSection("reducehashranges")) {
+		auto reducehashranges = ini.GetOrCreateSection("reducehashranges")->ToMap();
+		// Format: w,h = reducehashvalues 
+		for (const auto& item : reducehashranges) {
+			ParseReduceHashRange(item.first, item.second);
+		}
+	}
+
 	return true;
 }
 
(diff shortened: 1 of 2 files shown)


LunaMoo:
Reduce hash was initially a float, however Henrik did not liked that, so dunno if this will get a pass.

hrydgard:
So what value did you use that had a huge impact? I don't really see how smaller values than 0.5 would help, except in really odd edge cases.

Banh-Canh:
A value of 0.3125, stricly for dump with 128x256 as dimensions, drastically reduced the amount of duplicates for Growlanser WoT.
The rest of the textures are still ok with the default 0.5 value.

Almost every dumped textures of this dimensions (126x256), only have relevant data in the first third, resulting in thousands of duplicates.

I agree though, that it might be one of those odd edge cases.
Here are the kind of textures that have issues with the default value.

Sample 1:
00000000c66b4cff790017ff

Sample 2:
00000000c66b4cff347404b7

hrydgard:
The problem of course is that the more you reduce it, the more you risk problems with other textures clashing in the top-left part but different elsewhere. Though I suppose that restricting it to certain dimensions will reduce the likelyhood of problems a lot.

I think I'm fine with this, now that there are clear examples of the problem.

efonte:
I think it would be better not to add the dimensions to the name (128x256_addresscluthash.png) so that they are uniform with the rest of the images.

Also it would be good to add a parameter to customize the file name of all the images for example:

Example default value: `new/0000000001489f3548166ef0.png`
```ini
imgName = {address}{clut}{data}
```

Example 1: `new/00000000_01489f35_48166ef0_32_32.png`
```ini
imgName = {address}_{clut}_{data}_{width}_{heigh}
```
Example 2: `new/32_32/00000000_01489f35_48166ef0.png`
```ini
imgName = {width}_{heigh}/{address}_{clut}_{data}
```

Issue #8094, opened 2015-10-23
MGS Peace Walker recruit function


Antonio1994:
I was a long time off from ppsspp and git hub. Now that I'm back I see a lot off progress on this game and AFAIK only the recruit option is not working. I play on a PC and tryed connecting via WiFi and router but it doesn't work. Does anyone have a clue what could cause it? Sorry if its double posted, haven't found the same issue here ;),

thedax:
I could be wrong, but I think that function tries to connect to an actual router to "download" new recruits. Such functionality isn't implemented in PPSSPP.

Antonio1994:
Hmmm...haven't thought about that...I'll check the log if anything shows
up...I'm stuck at lvl90 with the combat unit so I thought off speeding
things up :-)

subota, 24. listopada 2015. korisnik thedax <notifications@github.com>
napisao je:

asl97:
@Antonio1994 if you talking about the ap recruit, i think it depend on the amount of wireless access points and their signal strength. it doesn't "download" anything so it doesn't require internet access.

in any case, iirc, those recruit that you get from the ap recruit are pretty weak,
it's better to target those extra (hard) missions for recruits

Antonio1994:
I have S ranked every mission and submission. I dont need recruits, I'm just saying this function doesn't work properly :)

LunaMoo:
If anyone cares, I got to mess around with this game and made this work:
https://gist.github.com/LunaMoo/ec340ddf0f6bc7c7277e7b9603061a8e

 That code is hackish, full of guesses, definitely not for merging.
In short the 3 functions that have to be implemented for this to work are:
```
sceNetApctlScanSSID2
sceNetApctlGetBSSDescIDList2
sceNetApctlGetBSSDescEntry2
```
 But I also added a nasty hack which checks for and replaces one opcode for something I couldn't figure out.

 It's also feeding the game with completely random stuff instead of allowing some kind of UI selection or emulating it properly ~ which actually isn't all that bad as with static data feed from stationary device like a pc the game would heavily limit this feature. Still could be improved as the random data should follow some rules, I filled everything with same number of random data which is completely wrong, could in fact break some other game or affect the randomness of recruits if it for example depend on lenght of data or takes only ascii characters for example.

 Use at your own risk / someone will have to investigate it further, don't have any psp I could use anymore and I don't think anyone ever documented those syscalls at least I couldn't find anything on them.:]

benderscruffy:
[...]

PR #14345, opened 2021-04-04
[APctl] Fix stuck issue when scanning AP to Recruit on MGS:PW

anr2me:
This should fix recruiting on MGS:PW using faked AP https://github.com/hrydgard/ppsspp/issues/8094


Core/HLE/sceNet.cpp
```diff
@@ -1219,21 +1226,57 @@ static int sceNetApctlGetBSSDescEntryUser(int entryId, int infoId, u32 resultAdd
...
+
 	switch (infoId) {
 	case PSP_NET_APCTL_DESC_IBSS: // IBSS, 6 bytes
-		Memory::WriteStruct(resultAddr, &netApctlInfo.bssid);
+		if (entryId == 0)
+			Memory::WriteStruct(resultAddr, &netApctlInfo.bssid);
+		else {
+			// Generate a BSSID/MAC address
+			char dummyMAC[ETHER_ADDR_LEN];
+			memset(&dummyMAC, entryId, sizeof(dummyMAC));
```
CookiePLMonster, line 1240: Nitpick, but `char dummyMAC[ETHER_ADDR_LEN] {};` will yield the same result without an explicit memset.

Core/HLE/sceNet.cpp
```diff
@@ -1219,21 +1226,57 @@ static int sceNetApctlGetBSSDescEntryUser(int entryId, int infoId, u32 resultAdd
...
+		}
 		break;
 	case PSP_NET_APCTL_DESC_SIGNAL_STRENGTH:
 		// Return 1 byte
-		Memory::WriteStruct(resultAddr, &netApctlInfo.strength);
+		if (entryId == 0)
+			Memory::WriteStruct(resultAddr, &netApctlInfo.strength);
+		else {
+			// Randomize signal strength between 1%~99% since games like MGS:PW are using signal strength to determine the strength of the recruit
+			Memory::Write_U8(rand()*99 + 1, resultAddr);
```
hrydgard, line 1278: rand()*99+1 will not do what you think. rand() returns very large integer numbers, so the only thing limiting here is that you just write a byte.

Probably not worth messing with a better random generator here, so the "right" way to do this with rand() should be something like:

```(int)(((float)rand()/(float)RAND_MAX)) * 99.0 + 1.0)```

Comment (ANR2ME) [reply Core/HLE/sceNet.cpp:1278]: Ah you i right, i was taking the code example from here https://www.cplusplus.com/reference/cstdlib/rand/
And it should be rand() % 99 instead of rand()*99 >.< i must have been sleepy

File: Core/HLE/sceNet.cpp (hunk 1 above)
Comment (ANR2ME) [reply Core/HLE/sceNet.cpp:1240]: i see, i thought initializing it with {} have the same effect as {0} which fill it with 0 and without it became uninitialized, well we don't really need to null-terminating a mac address anyway since it's not a string, so i should probably remove it.

Comment (hrydgard) [reply Core/HLE/sceNet.cpp]: I think CookiePLMonster thought you were zero-initializing it, in which case it would have worked.

You can remove the & sign though, leaving `memset(dummyMAC, entryId, sizeof(dummyMAC));`

File: Core/HLE/sceNet.cpp (hunk 2 above)
Comment (hrydgard) [reply Core/HLE/sceNet.cpp]: yeah, that'd also be acceptable for this use case.

File: Core/HLE/sceNet.cpp [hunk 3]
```diff
@@ -1219,21 +1226,56 @@ static int sceNetApctlGetBSSDescEntryUser(int entryId, int infoId, u32 resultAdd
...
+	dummySSID[6] += static_cast<char>(entryId);
+
 	switch (infoId) {
 	case PSP_NET_APCTL_DESC_IBSS: // IBSS, 6 bytes
-		Memory::WriteStruct(resultAddr, &netApctlInfo.bssid);
+		if (entryId == 0)
+			Memory::WriteStruct(resultAddr, &netApctlInfo.bssid);
+		else {
+			// Generate a BSSID/MAC address
+			char dummyMAC[ETHER_ADDR_LEN]{ static_cast<char>(entryId) };
```
hrydgard, line 1239: be aware that this will only set the first entry to entryId, the rest will be 0. that might be fine though. if you want all the bytes to be set to entryId, switch back to memset.

ANR2ME: eh.. it's not the same with memset? ugh okay, i'll change it back

Core/HLE/sceNet.cpp
ANR2ME: ah right, if i wanted to use `&` i need to explicitly pointing it to the first char like `&dummy[0]`

PR #14370, opened 2021-04-11
Enable save encryption in libretro build

stuken:
Since the libretro build doesn't read the ppsspp.ini file the default true value for EncryptSave is not picked up and all save files are stored plainly.
Uncrypted saves fail to load in Toca Race Driver 2 and 3 and potenitally others.

This should bring libretro in line with the default behaviour of the main build.


unknownbrackets (approved):
Sigh, annoying that it's all separate.  There's probably more settings that differ in non-obvious ways.

-[Unknown]

PR #14390, opened 2021-04-17
Update our miniupnp fork to match upstream.

hrydgard:
There's no difference from upstream anymore, since https://github.com/miniupnp/miniupnp/pull/486 was merged.

Should take care of #14386. 

Could probably just switch to using upstream directly, but this'll do just as well.


hrydgard:
Got lazy and didn't test the build properly. Fixing.

PR #14411, opened 2021-04-25
FileUtil cleanup

hrydgard:
Get rid of some redundancy, before I go nuts adding Android storage framework support to it. I think it's actually the right level to do that at, even though it's not gonna be super pretty.

Also moves some stuff into the namespace to avoid clashes, and then renaming the functions to look more consistent.


hrydgard:
Heh, thanks for finishing it up :)

PR #14418, opened 2021-04-28
[Adhoc] Replacing sceKernelDelayThread with ScheduleEvent on fake HLE to wait for event availability

anr2me:
I'm not sure which one is better, i just had this worry that calling `sceKernelDelayThread` too often could potentially wake up other threads at a bad/critical time.
But if something like that is not possible to happen, i prefer using the simple one (ie. `sceKernelDelayThread`) than the complicated Scheduled Event.

### DIFF
--- a/Core/HLE/sceNet.cpp
+++ b/Core/HLE/sceNet.cpp
@@ -171,7 +171,7 @@ static void __ApctlPollEvent(u64 userdata, int cyclesLate) {
 
 	u32 waitVal = __KernelGetWaitValue(threadID, error);
 	if (error == 0) {
-		if (netApctlInited && apctlEvents.empty()) {
+		if (netApctlInited && apctlEvents.empty() && npAuthEvents.empty()) {
 			// Try again in another 10ms until adhocctl terminated.
 			CoreTiming::ScheduleEvent(usToCycles(10000) - cyclesLate, apctlPollEvent, userdata);
 			return;
--- a/Core/HLE/sceNet.cpp
+++ b/Core/HLE/sceNet.cpp
@@ -208,6 +208,7 @@ void __NetApctlInit() {
 	netApctlInited = false;
 	netApctlState = PSP_NET_APCTL_STATE_DISCONNECTED;
 	apctlStateEvent = CoreTiming::RegisterEvent("__ApctlState", __ApctlState);
+	apctlPollEvent = CoreTiming::RegisterEvent("__ApctlPollEvent", __ApctlPollEvent);
 	apctlHandlers.clear();
 	apctlEvents.clear();
 	memset(&netApctlInfo, 0, sizeof(netApctlInfo));
--- a/Core/HLE/sceNet.cpp
+++ b/Core/HLE/sceNet.cpp
@@ -74,6 +74,7 @@ u32 apctlThreadHackAddr = 0;
 u32_le apctlThreadCode[3];
 SceUID apctlThreadID = 0;
 int apctlStateEvent = -1;
+int apctlPollEvent = -1;
 int actionAfterApctlMipsCall;
 std::recursive_mutex apctlEvtMtx;
 std::deque<ApctlArgs> apctlEvents;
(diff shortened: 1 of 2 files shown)


Core/HLE/sceNet.cpp
```diff
@@ -313,6 +352,13 @@ void __NetDoState(PointerWrap &p) {
 		apctlStateEvent = -1;
 	}
 	CoreTiming::RestoreRegisterEvent(apctlStateEvent, "__ApctlState", __ApctlState);
+	if (s >= 6) {
+		Do(p, apctlPollEvent);
+	}
+	else {
+		apctlPollEvent = -1;
+	}
+	CoreTiming::RestoreRegisterEvent(apctlPollEvent, "__ApctlPollEvent", __ApctlPollEvent);
```
ANR2ME, line 361: Currently, i'm having crash issue due to this newly added `RestoreRegisterEvent` when loading old SaveState on Def Jam Fight for NY
image
image
Which causes an issue during `__UsbMicDoState`

May be we should create newly registered events after savestate fully loaded? 
So it won't insert newly registered event id in the middle and messed up id(s) of existing event(s) that might be active when the savestate were created, thus causing the id pointing to the wrong event function.

unknownbrackets: It should allocate a new one at the end - the problem is that it does need to set the variable to the ID.

The problem here is that you're calling RestoreRegisterEvent when you never called RegisterEvent in the first place.  That's not valid and ends up causing this error.

-[Unknown]

unknownbrackets:
If the thread is meant to stall until some other event happens, using events instead of sleep does make sense.

-[Unknown]

Core/HLE/sceNet.cpp
ANR2ME: I copy that `RestoreRegisterEvent` from other DoState, the event itself will be created when Apctl being initialized, but this game doesn't even use Apctl (usually used for infrastructure) because it doesn't support infrastructure, thus never gets registered.

unknownbrackets: It has to be initialized in init now and then everything works.

-[Unknown]

ANR2ME: The `__NetDoState` didn't have any issue, but because there is a new register event in `__NetDoState`, the following DoState have a possibility to crashed (in this case the `__UsbMicDoState` getting the error), probably because the id got shifted

ANR2ME:
Yes, it's just to poll and check whether there is pending event that need to be processed (ie. `if (adhocctlEvents.empty())`)

What about the overhead, which one should be lighter? 
Because there are games (ie. DBZ SB2) that have Adhocctl initialized at boot time thus these loop will be running all the time (including during single player)

ANR2ME:
Hmm.. i wanted to do the same for AdhocMatching events, but AdhocMatching ended broken :(

Looks like it's a bad idea to stall the thread while have a mipscall/callback running, the callback ended being stalled when there are no more pending event to be processed, and the game ended gave up waiting for information update from their callback handler.

unknownbrackets:
Hm, __NetApctlCallbacks is run on an internal HLE thread right?

Callbacks are different from mipscalls.  The simplest thing might be to set a flag if you ran `hleEnqueueCall(it->second.entryPoint, 3, args.data);`, since presumably you want those to run right away - not be delayed.  Would you just delay only when none were enqueued?

-[Unknown]

ANR2ME:
The mipscall supposed to be called on a separate thread than the threads that called the sceNetApctl* syscalls, and there is about 200ms to 1seconds timestamp difference from the sceNetApctl* call to the callback execution(depends on what kind of sceNetApctl* syscall), probably waiting for WLAN state (ie. when scanning, joining, etc.) which could takes time.

Also, at the time the callbacks gets called the ApctlState will need to be changed first so the game's callback handler gets the new state on the handler execution, so i tried to simulate BeforeAction using scheduledEvent and blocking the thread that supposed to execute the callback to change the state then resume the thread on the event (which will also resume the callback execution), since we only have AfterAction on callback.

PR #14539, opened 2021-06-18
Fix some Japanese characters position.

nassau-tk:
Fix like below.
I fixed Bracket mark & many Kanji's potision.

image


unknownbrackets (approved):

hrydgard:
Nice!

PR #14545, opened 2021-06-19
An attempt at cleaning up the checks in ThreadUtil

hrydgard:
Had reported issues on some platforms.

### DIFF
--- a/Common/Thread/ThreadUtil.cpp
+++ b/Common/Thread/ThreadUtil.cpp
@@ -1,13 +1,19 @@
 #include "ppsspp_config.h"
 
-#ifdef _WIN32
+#if PPSSPP_PLATFORM(WINDOWS)
+
 #include <windows.h>
+
 #ifdef __MINGW32__
 #include <excpt.h>
 #endif
+
 #define TLS_SUPPORTED
+
 #elif defined(__ANDROID__)
+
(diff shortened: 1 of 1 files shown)


Sanaki:
No change on x86_64 Linux, but changing `return (int)gettid();` on line 152 to `return (int)syscall(__NR_gettid);` did resolve it for me. That said, I harvested that solution from elsewhere, I don't know enough about the implications of it to feel comfortable submitting a PR myself in this case.

hrydgard:
Baffling to me, as far as I understand, defining _GNU_SOURCE should make gettid() available.

I'll change it to the syscall.

FridgeRacer:
> No change on x86_64 Linux, but changing `return (int)gettid();` on line 152 to `return (int)syscall(__NR_gettid);` did resolve it for me. That said, I harvested that solution from elsewhere, I don't know enough about the implications of it to feel comfortable submitting a PR myself in this case.

Thanks! Cloned and tried to build PPSSPPSDL on my Nintendo Switch (Ubuntu 18.04) today but had the same error, this allowed me to progress past it.

hrydgard:
I've merged the fix. Though it's still weird.

Sanaki:
No argument there. Thanks.

PR #14550, opened 2021-06-21
[Adhoc] Making sure not to get random port when original port is not zero.

anr2me:
This probably doesn't fix anything, just a precaution of a possibility for `original_port + port_offset` to be zero after wrapping it to 16-bit and ended getting random port instead of static port.


ANR2ME:
Since this is not something major, i'll merge this.

PR #14567, opened 2021-06-27
UWP: Add a missing bounds check from  TextDrawerWin32::DrawStringBitmap

CookiePLMonster:
Fixes a crash on switching tabs due to an out of bounds access. While it's not a clean fix, it improves feature parity between Win32 and UWP implementations.


unknownbrackets (approved):

hrydgard:
Post-approved!

PR #14569, opened 2021-06-27
Refactor and unify analog input settings

### LINKED ISSUE #11552: Controller stick range

gyf304:
Previously PPSSPP analog stick mapping is done by each backend separately, with the XInput implementation being the more feature-complete one. The other ones usually have fewer features or are somewhat broken (e.g. lacking circle -> square mapping causing characters unable to run diagonally).

This PR attempts to fix that by unifying all analog stick mapping implementations to a common one.

This adds dead zone, inverse dead zone, sensitivity, and circle -> square mapping functionalities to all platforms using the `NativeAxis` abstraction.

The InverseMode option is also removed, as the "Off", "X", "Y", "X + Y" options do not quite make sense to me: "Off" can be achieved using the default values for "sensitivity" and "inverse dead zone", applying axis mapping to only one of the axes also seems not very useful.

The defaults are also set to values that represent 1-1 input mapping. Circle-to-square mapping is also enabled by default since most controllers use a [2-norm (Euclidian norm)](https://en.wikipedia.org/wiki/Norm_(mathematics)#Euclidean_norm) space. (PSP uses the [max-norm](https://en.wikipedia.org/wiki/Norm_(mathematics)#Maximum_norm_(special_case_of:_infinity_norm,_uniform_norm,_or_supremum_norm)))

Resolves https://github.com/hrydgard/ppsspp/issues/11552

### DIFF
--- a/Core/Config.cpp
+++ b/Core/Config.cpp
@@ -962,18 +962,11 @@ static ConfigSetting controlSettings[] = {
 	ConfigSetting("AnalogRotationCWKeyX", "AnalogRotationKeyCWY", "AnalogRotationKeyCWScale", "ShowAnalogRotationCWKey", &g_Config.touchAnalogRotationCWKey, defaultTouchPosHide, true, true),
 	ConfigSetting("AnalogRotationCCWKeyX", "AnalogRotationKeyCCWY", "AnalogRotationKeyCCWScale", "ShowAnalogRotationCCWKey", &g_Config.touchAnalogRotationCCWKey, defaultTouchPosHide, true, true),
 
-#ifdef _WIN32
-	ConfigSetting("DInputAnalogDeadzone", &g_Config.fDInputAnalogDeadzone, 0.1f, true, true),
-	ConfigSetting("DInputAnalogInverseMode", &g_Config.iDInputAnalogInverseMode, 0, true, true),
-	ConfigSetting("DInputAnalogInverseDeadzone", &g_Config.fDInputAnalogInverseDeadzone, 0.0f, true, true),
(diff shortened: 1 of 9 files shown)


hrydgard:
Oh nice! Looks very close to what I was planning to do at some point, hehe.

Been a busy weekend, I'll take a closer look tomorrow or so.

gyf304:
Any thoughts on this? The only potential issue I see here is that it wouldn't distinguish input between controllers, I don't know how much of an issue this will be though.
I have a long weekend this week (July 4th). I can spend a little bit more time reworking this if required.

hrydgard:
Yeah, it's not like the existing code does that either...

Overall this looks fine, I'm just gonna give it a go myself and then probably merge. 

Ideally these settings should be on a screen that displays joystick data pre and post mapping, but that's for later.

hrydgard:
Currently getting a crash on startup, looks like a race condition where ScreenManager doesn't exist yet but already getting an event. Seems like something I should fix though, heh.

image




I'm gonna try to fix it later.

hrydgard:
Merged it together with a fix. Github didn't pick that up, so closing.

Let's see how it goes :)

gyf304:
Thanks!

hrydgard:
Okay so there's a reported issue with this, #14586. Maybe the default settings are not great, or there's some other bug - seems like a simple deadzone thing though from the description.

unknownbrackets:
I'd say the circle setting should be hidden on some platforms where the values are known to be circular, always.

-[Unknown]

PR #14589, opened 2021-07-08
Fix black screen on multiple libretro content loads.

### LINKED ISSUE #14572: [Libretro] Black screen on successive launches (all renderers), then crash on close content (D3D11)
### What happens?
1st launch works perfectly fine, then if you close content and launch the game again you'll get a black screen with all renderers (audio can still be heard) and if you're using D3D11 and close content after the 2nd launch, RA crashes.

Logs: https://pastebin.com/aNcaWBW4
Crash log (D3D11): https://pastebin.com/vEbhLsFY

### What *should* happen?
Should be able to close content, launch again, close content, launch again, etc. without black screen and crash.

Windows 10 / i5-4670K / GTX 970

This is with the Libretro core, it worked before, I tried to bisect but had some crashes on launch so I had to skip some commits, but it started happening between 3be5c7bd9aed60ce2c24ccf318b56e845cb3fe59 (good) and 2d6a73792b9505e882d02a175d090820414c8df0 (bad).

#### Which game or games?
Doesn't matter.

#### Checklist
 - [x] Test in the [latest git build](https://buildbot.orphis.net/ppsspp/) in case it's already fixed.
 - [x] [Search for other reports](https://github.com/hrydgard/ppsspp/search?q=my+issue&type=issues) of the same issue.

stuken:
Moves ThreadPool teardown to retro_unload_game. Gives ThreadPool threads some breathing room to terminate before game load starts spamming it with tasks.

Should fix #14572 outside of pathological start stop behaviour.

### DIFF
--- a/Common/Thread/ThreadManager.cpp
+++ b/Common/Thread/ThreadManager.cpp
@@ -98,10 +98,6 @@ static void WorkerThreadFunc(GlobalThreadContext *global, ThreadContext *thread)
 }
 
 void ThreadManager::Init(int numRealCores, int numLogicalCoresPerCpu) {
-	if (!global_->threads_.empty()) {
-		Teardown();
-	}
-
 	numComputeThreads_ = std::min(numRealCores * numLogicalCoresPerCpu, MAX_CORES_TO_USE);
 	int numThreads = numComputeThreads_ + EXTRA_THREADS;
 	numThreads_ = numThreads;
--- a/Common/Thread/ThreadManager.cpp
+++ b/Common/Thread/ThreadManager.cpp
@@ -98,6 +98,10 @@ static void WorkerThreadFunc(GlobalThreadContext *global, ThreadContext *thread)
 }
 
 void ThreadManager::Init(int numRealCores, int numLogicalCoresPerCpu) {
+	if (!global_->threads_.empty()) {
+		Teardown();
+	}
+
 	numComputeThreads_ = std::min(numRealCores * numLogicalCoresPerCpu, MAX_CORES_TO_USE);
 	int numThreads = numComputeThreads_ + EXTRA_THREADS;
 	numThreads_ = numThreads;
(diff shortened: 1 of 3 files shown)


Common/Thread/ThreadManager.cpp
```diff
@@ -98,10 +98,6 @@ static void WorkerThreadFunc(GlobalThreadContext *global, ThreadContext *thread)
 }
 
 void ThreadManager::Init(int numRealCores, int numLogicalCoresPerCpu) {
-	if (!global_->threads_.empty()) {
-		Teardown();
-	}
-
```
hrydgard, line 104: Does this part actually hurt? Seems that just calling Teardown from retro_unload_game should be enough, without deleting these three lines. I don't remember exactly why I added these though, heh.

stuken: It seemed redundant in terms of the lifecycle of the thread manager in the main build. As far as i can reason init is only ever called once and any full scale cleanup is handled by exiting the application. As a library it does hang around way longer and the cleanup was needed so decoupling the init and teardown behaviours when we have appropriate lifecycle callbacks seemed logical to me. That teardown call was the source of the issue in the first place so if for whatever reason the new retro_unload_game based teardown call fails we would fall back to the whatever race condition that got us here in the first place.

That said I have not been able to figure out why nothing was rendering. On Nvidia and AMD machines it was 100% reproducable yet on an Intel Iris Xe there were zero issues. Plus I didnt notice any rendering related tasks hitting that thread pool so im at a loss as to what was going on. I'm in a here be dragons part of the map so if you think it needs to come back im not going to argue :)

hrydgard: I would prefer that the change is not in this PR unless it's really needed. I'm gonna clean it up myself and test the other platforms separately though.

stuken: Pushed the revert now.

hrydgard: Thanks!

PR #14590, opened 2021-07-08
Disable analog stick deduplication

gyf304:
As discussed in #14586


gyf304:
(Squash and merge probably?)

PR #14592, opened 2021-07-08
Break out EmuScreen's key/axis inputs to ControlMapper.cpp/h

hrydgard:
Additionally, break out the analog stick value remapping to the same file.

This will make more sense soon when I change where the analog mapping is called from, but want this refactor in separately for testing.

In short, #14569 moved the analog stick value remapping to a common location right at the bottom of the callstack, but I think I want the value mapping to happen after inputs have been mapped and processed, right before sending it to the virtual PSP. Also, I later want to be able to process the mappings directly from within a new controls testing screen, where you'll also be able to modify the analog parameters to check that you can cover the whole space properly.

### DIFF
--- a/UWP/CoreUWP/CoreUWP.vcxproj.filters
+++ b/UWP/CoreUWP/CoreUWP.vcxproj.filters
@@ -1021,9 +1021,6 @@
     <ClCompile Include="..\..\ext\libzip\zip_pkware.c">
       <Filter>Ext\libzip</Filter>
     </ClCompile>
-    <ClCompile Include="..\..\ext\libzip\zip_mkstempm.c">
-      <Filter>Ext\libzip</Filter>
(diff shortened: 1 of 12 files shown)


UWP/CoreUWP/CoreUWP.vcxproj.filters
```diff
@@ -1021,9 +1021,6 @@
     <ClCompile Include="..\..\ext\libzip\zip_pkware.c">
       <Filter>Ext\libzip</Filter>
     </ClCompile>
-    <ClCompile Include="..\..\ext\libzip\zip_mkstempm.c">
-      <Filter>Ext\libzip</Filter>
-    </ClCompile>
```
hrydgard, line 1026: not sure what happened here, heh.

Issue #14571, opened 2021-06-28
Surf's Up multiplayer bug/issue(?)


Carlson1:
This is a report for a potential bug/issue regarding the game Surf's Up on PPSSPP.

### What happens?
My friend and I have been trying to play Surf's Up on multiplayer, however, we couldn't see each other whenever one of us hosted. We have tried various methods of multiplayer games, however, none worked.
![sketch-1624886437658](https://user-images.githubusercontent.com/86608310/123644579-cb3b9000-d825-11eb-8e4c-f76b723156cb.png)

### What *should* happen?
What SHOULD originally happen is for when a player is hosting, the other player(s) should be able to see the host's username, and have the option to join the host's game.
![sketch-1624886218857](https://user-images.githubusercontent.com/86608310/123645225-70eeff00-d826-11eb-9af7-f42828d71aae.png)

### What are you using?
Phone.
#### What hardware / device and operating system?
Android 8.0.0.
#### What graphics card (GPU) or mobile phone model?
Samsung Galaxy A6+.
#### What PPSSPP version (standalone/official), and did it work before?
Latest official version released on the Google Play Store, and it worked before.
#### Which game or games?
ULUS10262

#### Checklist
 - [X] Test in the [latest git build](https://buildbot.orphis.net/ppsspp/) in case it's already fixed.
 - [X] [Search for other reports](https://github.com/hrydgard/ppsspp/search?q=my+issue&type=issues) of the same issue.
 - [X] Try resetting settings or older versions and include if the issue is related.
 - [X] Include logs or screenshots of issue.
 - Note 1: We tested with the proper multiplayer settings needed to play a game on multiplayer.
 - Note 2: I checked everything related to the internet router and connection.

I hope this isn't too much trouble... I'm still not sure wheter this is only a client-side issue or not. We will continue trying to play on Surf's Up multiplayer. Hope you have a good day! :)

ANR2ME:
Hmm.. "Latest official version released on the Google Play Store, and it worked before." we need the version number where it worked the last time in order to find out which changes breaks it.

Could you find out which version is the last working one by testing the builds in this list https://buildbot.orphis.net/ppsspp/index.php?m=fulllist
You only need to tests those (from the top) that have `ANR2ME` in the Commit Log and can be downloaded, there shouldn't be many of them.

itsXPCT:
Hi,me again.I'm friend he is trying to play with.
We tried everything,latest build too,stable from Google Play.
We did following

[...]

Issue #14543, opened 2021-06-19
[Prototype]Isle of Minno breaks Ad-Hoc connection after minigame loading screen - 1.11.3


itsXPCT:
**[Prototype] - game is not released officially,but as leak of game company which means it's unreleased**
Hi,me again.Found this interesting prototype of unreleased game and thought i could test multiplayer on it,
while i did i found bug,so i wanted to report it

### What happens?
PC Host:Game basically loads in menu just fine,everything goes fine,host and join device act fine,until you proceed to game,but after loading screen it just displays connection lost

Android device:Game loads in menu until Android freezes at Loading Screen,and PC displays "Connection Lost"
### What *should* happen?
Game should work and join same on both devices

### What are you using?
Android and PC
#### What hardware / device and operating system?
Honor 9S - Android 8.1.0
PC - Windows 8.0

#### What graphics card (GPU) or mobile phone model?
Honor 9S - PowerVR GE8320
PC - Intel (R) HD Graphics

#### What PPSSPP version (standalone/official), and did it work before?
1.11.3 stable
1.11.3 g906b from https://buildbot.orphis.net

#### Which game or games?

Isle Of Minno
UCET00372
Europe

#### Checklist

 - [x] Test in the [latest git build](https://buildbot.orphis.net/ppsspp/) in case it's already fixed.
 - [x] [Search for other reports](https://github.com/hrydgard/ppsspp/search?q=my+issue&type=issues) of the same issue.
 - [x] Try resetting settings or older versions and include if the issue is related.
 - [ ] Include logs or screenshots of issue.

ANR2ME:
I tested this game on localhost and found 2 issue:
![image](https://user-images.githubusercontent.com/7974720/122651070-a88fe400-d160-11eb-9f9d-d05227c77ee0.png)
1). After joining and the mission started, this game will move group from a public group "MINNO" to a private group, but this private group is using non-alphanumeric characters which is considered as invalid group name based on the AdhocServer code from aemu project.
Not sure if non-alphanumeric characters are allowed or not on real PSP, does anyone have tried the multiplayer feature of this game on PSP vs PSP?

2). This game is trying to bind the same socket on PTP(TCP)'s `PtpListen` and PDP(UDP)'s `PdpCreate`, which think it's allowed on most platform, but not sure why on my Win10 it's getting Socket Error with errno = 0 like this... which is weird.
[...]

PR #14599, opened 2021-07-10
Change the default port offset

ghost:
I doing these changes because I am seeing a that a lot of issues come by using the previous default port offset of 0.

**Reasons for the change:**

1. The default port offset of 0 is not a good one for the majority of PPSSPP's users (Android users).
Because Android (and maybe IOS) users can't use ports that are below 1024 and thats what most of what the PSP games use for communicating (Except few games that also got infrastructure mode to them and Monster Hunter).
Seems like the port that is used on infrastructure in games for voice chat is used for the adhoc part as well
2. Most of the issues here and on the forums and discord server and even on github stem from the default port offset being 0.
3. Users that dont have adminstrator rights also got issues with ports below 1024.
4. Port offset 10000 is the port offset (or higher)  recommended to use by @ANR2ME and others and it also appears on the github wiki page I created
5. Port offset 0 is often used to play with real PSP hardware and with JPCSP but no one uses it for that purpose and PPSSPP networking settings and the Github wiki page still tell you to use port offset 0 if you wanna have compatibility with either of them.
6. Very few games do not work with a port offset of 0 (no idea why but possibly a port conflict).

Originally I also wanted to remove the "UPNP use original port offset" option as I can't seem to get it to work correctly but I decided to keep it anyway for now.

### DIFF
--- a/UI/GameSettingsScreen.cpp
+++ b/UI/GameSettingsScreen.cpp
@@ -833,7 +833,7 @@ void GameSettingsScreen::CreateViews() {
 	qc5->SetEnabledPtr(&g_Config.bEnableQuickChat);
 #endif
 
-	networkingSettings->Add(new ItemHeader(n->T("Misc", "Misc (default = compatibility)")));
+	networkingSettings->Add(new ItemHeader(n->T("Misc"));
 	networkingSettings->Add(new PopupSliderChoice(&g_Config.iPortOffset, 0, 60000, n->T("Port offset", "Port offset (0 = PSP compatibility)"), 100, screenManager()));
 	networkingSettings->Add(new PopupSliderChoice(&g_Config.iMinTimeout, 0, 15000, n->T("Minimum Timeout", "Minimum Timeout (override in ms, 0 = default)"), 50, screenManager()));
 	networkingSettings->Add(new CheckBox(&g_Config.bForcedFirstConnect, n->T("Forced First Connect", "Forced First Connect (faster Connect)")));
--- a/UI/GameSettingsScreen.cpp
+++ b/UI/GameSettingsScreen.cpp
@@ -833,7 +833,7 @@ void GameSettingsScreen::CreateViews() {
 	qc5->SetEnabledPtr(&g_Config.bEnableQuickChat);
 #endif
 
-	networkingSettings->Add(new ItemHeader(n->T("Misc"));
+	networkingSettings->Add(new ItemHeader(n->T("Misc", "Misc (default = compatibility)")));
 	networkingSettings->Add(new PopupSliderChoice(&g_Config.iPortOffset, 0, 60000, n->T("Port offset", "Port offset (0 = PSP compatibility)"), 100, screenManager()));
 	networkingSettings->Add(new PopupSliderChoice(&g_Config.iMinTimeout, 0, 15000, n->T("Minimum Timeout", "Minimum Timeout (override in ms, 0 = default)"), 50, screenManager()));
 	networkingSettings->Add(new CheckBox(&g_Config.bForcedFirstConnect, n->T("Forced First Connect", "Forced First Connect (faster Connect)")));
--- a/Core/Config.cpp
+++ b/Core/Config.cpp
@@ -984,7 +984,7 @@ static ConfigSetting networkSettings[] = {
 	ConfigSetting("EnableWlan", &g_Config.bEnableWlan, false, true, true),
 	ConfigSetting("EnableAdhocServer", &g_Config.bEnableAdhocServer, false, true, true),
 	ConfigSetting("proAdhocServer", &g_Config.proAdhocServer, "socom.cc", true, true),
-	ConfigSetting("PortOffset", &g_Config.iPortOffset, 0, true, true),
+	ConfigSetting("PortOffset", &g_Config.iPortOffset, 10000, true, true),
 	ConfigSetting("MinTimeout", &g_Config.iMinTimeout, 0, true, true),
 	ConfigSetting("TCPNoDelay", &g_Config.bTCPNoDelay, true, true, true),
 	ConfigSetting("ForcedFirstConnect", &g_Config.bForcedFirstConnect, false, true, true),


hrydgard:
Given the above, I think I agree that 10000 is the right choice for a default. I assume you agree @ANR2ME, since you recommend this port already?

MojoJojoDojo:
Looking at these issues as the places he suggested it about it:
https://github.com/hrydgard/ppsspp/issues/14543#issuecomment-864443422
https://github.com/hrydgard/ppsspp/issues/14571#issuecomment-870900836
That or higher but you still have less chance to encounter an error or an issue that way for Android users.

hrydgard:
It would just be a shame if we changed it and we discover that exactly 10000 + (some port a popular game uses) is equal to some reserved port or something. But even that I guess shouldn't matter much, except if some firewall does weird stuff etc...

MojoJojoDojo:
> It would just be a shame if we changed it and we discover that exactly 10000 + (some port a popular game uses) is equal to some reserved port or something. But even that I guess shouldn't matter much, except if some firewall does weird stuff etc...

Googling "port offset 10000" would also give you results that use it for other programs/apps and systems so its seems like there is little to no port reservation there.
But up to you to decide.

ANR2ME:
As i remembered someone said that 10k doesn't work for him for a certain game (i forgot which issue), may be we should use a higher port?
According to this _common ports_ the range goes up to 43k http://tcp-udp-ports.com/unofficial-ports.php
May be we should use 50k+ since ports above 50k usually used for client-side which is allocated randomly.

UI/GameSettingsScreen.cpp
```diff
@@ -833,7 +833,7 @@ void GameSettingsScreen::CreateViews() {
 	qc5->SetEnabledPtr(&g_Config.bEnableQuickChat);
 #endif
 
-	networkingSettings->Add(new ItemHeader(n->T("Misc", "Misc (default = compatibility)")));
```
ANR2ME, line 836: Instead of changing the Section Header, you should change the one specific to Port Offset (ie. "Port offset (0 = PSP compatibility)" ), because there are more options in Misc that most common people shouldn't touch when not having any issue.

MojoJojoDojo: oops sry.

MojoJojoDojo:
> As i remembered someone said that 10k doesn't work for him for a certain game (i forgot which issue), may be we should use a higher port?
> According to this _unofficial common ports_ the range goes up to 43k http://tcp-udp-ports.com/unofficial-ports.php and there are also some _officially registered ports_ up to 48k http://tcp-udp-ports.com/ports48000-49000.htm
> May be we should use 50k+ since ports above 50k usually used for client-side which is allocated randomly.

Should I change it to 50k? I thought its too high and maybe games use 10000 or higher or a port but I dont know (I think monster hunter does but not sure).

UI/GameSettingsScreen.cpp
ANR2ME: Nevermind, i got confuse too, by saying "default = PSP compatibility" doesn't right either when default port offset is non-zero ... i wonder what kind of wording we should use...

ANR2ME:
Yeah, i also think 50k might be too high, but i wondered why 10k didn't worked for that guy, 10k probably safe enough

UI/GameSettingsScreen.cpp
ANR2ME: May be we should just replace it with "Misc (workaround)" ? since these options are workaround anyway LOL

MojoJojoDojo: > Nevermind, i got confuse too, by saying "default = PSP compatibility" doesn't right either when default port offset is non-zero ... i wonder what kind of wording we should use...

btw where is that?
Like I am trying to find it in the code but its not in the GameSettingsScreen.cpp (which it should be?) lol.

ANR2ME: Sorry, i mean the "Misc (default = compatibility)" one

hrydgard:
Alright, let's go with 10000 and see how it goes. Hope this reduces the amount of questions.

Linblow:
Here is how the PTP/PDP adhoc ports are used in the original PSP adhoc librairies:

- Port `0` is unused (a random user port is chosen, see below)
- Ports **`[1; 32767]`** or `[1; 0x7fff]` in hex are for user (ie. games that use adhoc)
Total number of adhoc ports available for user: `32767`
- Ports `[32768; 65534]` or `[0x8000; 0xfffe]` are reserved for system (games cannot use them)
- Port `65535` or `0xffff` is unused and reserved

Note that:
- Port `0x7fff` the highest PTP/PDP adhoc port for user
- Port `0xffff ` is the highest port number usable for BSD sockets
- Ports `[0; 1024]` are usually important and reserved  by the underlying system PPSSPP is running on

Since the emulated adhoc librairies use internal BSD sockets for communication, the port number chosen randomly, or manually from the user (game), **should be changed** to be at least > 1024. In practise though, lots of ports over 1024 are still important (eg. port 1194 for openvpn).

To be safe, the port offset should be within range `[10000; 32768]`. 
PPSSPP **should** thrown an error, or use the default new port offset (`10000`), when that range is not satisfied.

unknownbrackets:
IMHO, once we're changing the default port offset we definitely can't be compatible with real PSPs, right?

In that case, it seems like the best solution to this problem is something like #7268, whether using UDP or QUIC or whatever.  I'm not saying that's the easiest, though.

Many port numbers might be used by various services/software running on client devices:
https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.txt

Running on devices that do more than just run games makes it tricky, though if PSP games can't use 0x8000+, that certainly gives some breathing room...

-[Unknown]

hrydgard:
Well people will still have the option to set it back to 0 to be compatible with real PSPs. Although, we absolutely should make that clearer in the UI! It does make sense to make the (today) likely more common use case the default.

MojoJojoDojo:
TBH I tried doing PSP to PPSSPP communication lately and a lot of games got issues connecting because a lot of the adhoc code in PPSSPP got changed... (I only got 1-2 games to work consistently maybe). Also the compatbility of the pro online plugin is super low compared to PPSSPP.
Someone will need to update the code in the pro online plugin or something I guess (I heard the code is very convulated in general tho).

Issue #14602, opened 2021-07-10
Vulkan: PPSSPP uses RGB565 texture format that's not mandatory-supported


hrydgard:
Turns out that VK_FORMAT_R5G6B5_UNORM_PACK16 is mandatory but VK_FORMAT_B5G6R5_UNORM_PACK16, which we were using, is not (though widely supported).

https://www.khronos.org/registry/vulkan/specs/1.2/html/chap33.html#features-required-format-support

The Intel driver on Linux reports this format as not supported, as mentioned in my first attempt to fix this, #14588 . Unfortunately, that change led to #14601 so had to revert.

hrydgard:
Since #14603 safely works around this, I'll push this off until later.

hrydgard:
Not planning to do this, the existing workaround is fine.

PR #14603, opened 2021-07-10
Vulkan: Turn off all 16-bit formats if B5G6R5 format is not available.

hrydgard:
Works around #14602 for now.

Also minor code cleanup.


hrydgard:
note to self: tested and confirmed on intel linux.

PR #14619, opened 2021-07-17
Enable scoped storage enforcement on Android 11+.

hrydgard:
Targets Android SDK 30, so scoped storage gets turned on for real on Android 11+ devices, see #11997.

~~This has a number of UX issues and bugs we need to work through, but at least games start up. ~~ Upgrades are handled smoothly by keeping existing storage access until you uninstall. After a reinstall, you'll need to re-select your old PSP directory manually in settings :(

We need to merge this soon, since after August we can no longer release new PPSSPP updates on Google Play without doing this. Most people should not be badly affected once this is in, just a new procedure to learn to give PPSSPP permissions to the folders you want to access on first startup.

We absolutely need to merge this AND release a new PPSSPP version with it before whenever Android 12 releases in September, October or November, and the old ways will no longer work for new installs.

This will take care of #11997 once done. Again, do not merge yet!

## Sub-issues discovered and fixed:

- /PSP as content URI:
  - [x] Save states not saving
  - [x] Save state thumbnails not showing
  - [x] In-game saves not working (game tested: Wipeout Pulse)
- Other
  - [x] File listing performance
  - [x] The code for determining free storage space is broken
  - [x] File move not implemented
  - [x] File append not working, probably some other edge cases
  - [x] Add support for returning error codes
  - [ ] Actually catch exceptions and provide proper error codes for things like disk full
  - [x] Make things work as expected when the memstick folder is named PSP 
- Planned features
  - [x] On first run, ask the user where they want their PSP folder. 
    - [x] Get it working
    - [ ] Make it pretty and make sure the user understands the choice
  - [x] Ability to auto-migrate your data when changing your PSP folder
  - [x] Threaded file move with progress bar
  - [ ] Set correct MIME types? Worth the trouble?

## Setup screen
If we find no folder configured and scoped storage enabled (first boot, or reinstall on Android11+) we will pop up a screen that gives the user two options, along with some advice:

* Choose or create a folder (browse button)
  - Choose if you had an old PPSSPP version and thus an old PSP folder
  - Allows sharing data between PPSSPP Regular and PPSSPP Gold

* Default PSP folder location
  - Will be removed if you uninstall PPSSPP

Not sure what else to say. There seem to be little reason to use the default folder location really, now that performance is pretty good in both cases...

### DIFF
--- a/Common/File/DiskFree.cpp
+++ b/Common/File/DiskFree.cpp
@@ -41,9 +41,11 @@ bool free_disk_space(const Path &path, int64_t &space) {
 	int res = statvfs(path.c_str(), &diskstat);
 
 	if (res == 0) {
+		// Not sure why we're excluding Android here anyway...
 #ifndef __ANDROID__
 		if (diskstat.f_flag & ST_RDONLY) {
-			space = -1;
+			// No space to write.
+			space = 0;
 			return true;
 		}
 #endif
--- a/Common/File/DiskFree.cpp
+++ b/Common/File/DiskFree.cpp
@@ -23,7 +23,7 @@
 #include "Common/File/AndroidStorage.h"
 #include "Common/Data/Encoding/Utf8.h"
 
-bool free_disk_space(const Path &path, uint64_t &space) {
+bool free_disk_space(const Path &path, int64_t &space) {
 #ifdef _WIN32
 	ULARGE_INTEGER free;
 	if (GetDiskFreeSpaceExW(path.ToWString().c_str(), &free, nullptr, nullptr)) {
@@ -43,7 +43,7 @@ bool free_disk_space(const Path &path, uint64_t &space) {
 	if (res == 0) {
 #ifndef __ANDROID__
 		if (diskstat.f_flag & ST_RDONLY) {
-			space = 0;
+			space = -1;
 			return true;
 		}
 #endif
--- a/UI/MemStickScreen.cpp
+++ b/UI/MemStickScreen.cpp
@@ -25,23 +25,26 @@
 #include "Common/UI/ViewGroup.h"
 
 #include "Common/StringUtils.h"
-#include "Common/File/FileUtil.h"
-#include "Common/File/Path.h"
 #include "Common/System/System.h"
 #include "Common/System/NativeApp.h"
-#include "Common/Data/Text/I18n.h"
 #include "Common/System/Display.h"
+#include "Common/Data/Text/I18n.h"
+
+#include "Common/File/AndroidStorage.h"
+#include "Common/File/FileUtil.h"
+#include "Common/File/Path.h"
 
 #include "Core/Util/GameManager.h"
 #include "Core/System.h"
+#include "Core/Config.h"
 
 #include "UI/MemStickScreen.h"
 #include "UI/MainScreen.h"
 #include "UI/MiscScreens.h"
 
-#include "Core/Config.h"
-
-MemStickScreen::~MemStickScreen() { }
+MemStickScreen::~MemStickScreen() {
+	pendingMemStickFolder_ = g_Config.memStickDirectory;
+}
 
 void MemStickScreen::CreateViews() {
 	using namespace UI;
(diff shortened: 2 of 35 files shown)


hrydgard:
Making progress here, in-game saves are working. 

Getting quite terrible stutter around save dialogs etc, hopefully the planned optimization will be good enough...

hrydgard:
Now that most of this is merged, cut it back down to just the enabling commit.

hrydgard:
With those two commits, performance is now pretty close to acceptable, actually.

hrydgard:
image

First cut at a startup experience. Ugh. Gonna need some design work..

hrydgard:
It was suggested by unknown that maybe it's better to directly default to the internal directory and add a warning somewhere, allowing the user to choose a permanent location. 

I couldn't quite figure out how to design that though, so for the initial merge (possibly not the official release) I'm going with a popup screen offering the choice, as above. For moves after the install, there'll be a second screen where you confirm the details of the move, and that will have a progress bar etc if you are moving data from a previous location.

Such a popup screen is not really much worse than the old obnoxious ask-for-permissions screen/button, I figure, anyway.

Most things seem to be working okay now, so I'm actually gonna look into merging this fairly soon so it gets some exercise. Of course, this can easily screw up a user's data, so best be careful and test properly before merging...

hrydgard:
Shrunk this again by pulling out a bunch of stuff into #14642 , should help any bisect.

Now the entire flow seems to be working okay, including moving the directory around after the fact. You have the option to move the files or not, but can't control overwrite in case of collisions yet - not sure how desirable that is, but I guess we could warn if you're about to overwrite any newer files...

hrydgard:
This is essentially feature complete, but needs a bunch of polish, see the missing checkboxes above. Even so, I think this should actually go in soon almost as-is so it gets as much testing as possible. It really isn't much slower anymore and seems to work quite alright. Of course, before I press merge I'm gonna subject this to a lot more manual testing, too.

Anyway, marking as ready for review, for the so inclined.

android/AndroidManifest.xml
```diff
@@ -45,7 +45,9 @@
         android:logo="@drawable/ic_banner"
         android:isGame="true"
         android:banner="@drawable/tv_banner"
-        android:requestLegacyExternalStorage="true">
+        android:requestLegacyExternalStorage="true"
+		android:preserveLegacyExternalStorage="true"
+		>
```
unknownbrackets, line 50: nit: Indentation different.

-[Unknown]

Common/File/DiskFree.cpp
```diff
@@ -43,7 +43,7 @@ bool free_disk_space(const Path &path, uint64_t &space) {
 	if (res == 0) {
 #ifndef __ANDROID__
 		if (diskstat.f_flag & ST_RDONLY) {
-			space = 0;
+			space = -1;
 			return true;
```
unknownbrackets, line 47: It returns true though, isn't it better to interpret that as 0 bytes free?  We'll cast this to unsigned and try to write.

Otherwise it should at least return false?  But this isn't really an error.  The free space for a game to create new save data is 0 bytes.  It's a known value, so I'd consider it successful.

-[Unknown]

UI/NativeApp.cpp
```diff
@@ -339,12 +342,13 @@ static void PostLoadConfig() {
...
+#if PPSSPP_PLATFORM(ANDROID)
+	CreateDirectoriesAndroid();
+#endif
 }
 
-static bool CreateDirectoriesAndroid() {
-	// On Android, create a PSP directory tree in the external_dir,
-	// to hopefully reduce confusion a bit.
+bool CreateDirectoriesAndroid() {
+	// TODO: Really not sure why this code is Android-exclusive, except the ".nomedia" part.
```
unknownbrackets, line 351: You added this for permissions iirc because it needed to happen after permissions were granted.

-[Unknown]

UI/MemStickScreen.h
```diff
@@ -30,19 +30,52 @@
 // Currently only useful for Android.
 class MemStickScreen : public UIDialogScreenWithBackground {
 public:
-	MemStickScreen() {}
-    virtual ~MemStickScreen();
+	MemStickScreen(bool initialSetup);
+	~MemStickScreen() {}
 
 	std::string tag() const override { return "game"; }
+	void CreateViews() override;
```
unknownbrackets, line 37: Hm, what's the reason for this to be public?  Only RecreateViews() should get called right?

-[Unknown]

UI/MemStickScreen.cpp
```diff
@@ -25,23 +25,83 @@
...
+
+	// Doesn't already exist, create.
+	// Should this ever happen?
+	if (newMemstickFolder.Type() == PathType::NATIVE) {
+		if (!File::Exists(newMemstickFolder)) {
+			File::CreateFullPath(newMemstickFolder);
+		}
+		if (!File::WriteDataToFile(true, "1", 1, testWriteFile)) {
+			// settingInfo_->Show(sy->T("ChangingMemstickPathInvalid", "That path couldn't be used to save Memory Stick files."), nullptr);
+			// TODO: Display an error!
```
unknownbrackets, line 67: Good todos are good.

-[Unknown]

UI/MemStickScreen.cpp
```diff
@@ -25,23 +25,83 @@
...
+			return UI::EVENT_DONE;
+		}
+		File::Delete(testWriteFile);
+	} else {
+		// TODO: Do the same but with scoped storage? Not really necessary, right? If it came from a browse
+		// for folder, we can assume it exists and is writable, barring wacky race conditions like the user
+		// being connected by USB and deleting it.
```
unknownbrackets, line 74: I wonder in what exciting ways we would crash if they did that, then started PPSSPP up again later without the USB device connected...

-[Unknown]

UI/MemStickScreen.cpp
```diff
@@ -25,23 +25,83 @@
...
+	g_Config.memStickDirectory = newMemstickFolder;
+	g_Config.SetSearchPath(GetSysDirectory(DIRECTORY_SYSTEM));
+	g_Config.UpdateIniLocation();
+
+	return true;
+}
+
+static std::string FormatSpaceString(int64_t space) {
+	if (space >= 0) {
+		// TODO: Smarter display (MB, GB as appropriate). Don't we have one of these somewhere?
```
unknownbrackets, line 94: Probably should be in Common:
https://github.com/hrydgard/ppsspp/blob/0c40e918c92b897f745abee0d09cf033a1572337/Windows/W32Util/Misc.cpp#L56

But also:
https://github.com/hrydgard/ppsspp/blob/87dcf60ed037d5f2168d5f0033dda002b098e177/Core/Dialog/SavedataParam.cpp#L1050

-[Unknown]

UI/MemStickScreen.cpp
```diff
@@ -51,45 +111,50 @@ void MemStickScreen::CreateViews() {
...
-	settingInfo_ = new SettingInfoMessage(ALIGN_CENTER | FLAG_WRAP_TEXT, new AnchorLayoutParams(dp_xres - leftSide - 40.0f, WRAP_CONTENT, leftSide, dp_yres - 80.0f - 40.0f, NONE, NONE));
-	settingInfo_->SetBottomCutoff(dp_yres - 200.0f);
+	leftColumn->Add(new TextView(iz->T("MemoryStickDescription", "Choose PSP data storage (Memory Stick)"), ALIGN_LEFT, false));
+
+	leftColumn->Add(new Choice(iz->T("Create or Choose a PSP folder")))->OnClick.Handle(this, &MemStickScreen::OnBrowse);
+	leftColumn->Add(new TextView(iz->T("ChooseFolderDesc", "* Data will stay even if you uninstall PPSSPP.\n* Data can be shared with PPSSPP Gold\n* Easy USB access"), ALIGN_LEFT, false));
 
-	root_->Add(settingInfo_);
+	leftColumn->Add(new Choice(iz->T("Use App Private Directory")))->OnClick.Handle(this, &MemStickScreen::OnUseInternalStorage);
+	leftColumn->Add(new TextView(iz->T("InternalStorageDesc", "* Warning! Data will be deleted if you uninstall PPSSPP!\n* Data cannot be shared with PPSSPP Gold\n* USB access through Android/data/org.ppsspp.ppsspp/files"), ALIGN_LEFT, false));
```
unknownbrackets, line 134: How well does the ini parser handle embedded newlines?

-[Unknown]

UI/MemStickScreen.cpp
hrydgard: We would fail the check in NativeApp.cpp "Couldn't read directory '%s' specified by memstick_dir.txt", and it would ask the user for a new directory.

UI/MemStickScreen.cpp
hrydgard: It does handle it, if I remember correctly. but better check....

UI/MemStickScreen.h
hrydgard: No reason, fixing.

UI/NativeApp.cpp
hrydgard: Right..

UI/MemStickScreen.cpp
hrydgard: Right.. I'll just move NiceSizeFormat out to Common for now, could merge them later but they're slightly different (rounding, decimal...)

UI/MemStickScreen.cpp
hrydgard: yeah, at least on the right side, heh.

hrydgard:
Just tested this branch on an Android 4.0 device out of curiosity, works just fine. So we can keep our extreme compatibility range no problem (not surprised though but good to check).

Common/File/DiskFree.cpp
```diff
@@ -41,9 +41,11 @@ bool free_disk_space(const Path &path, int64_t &space) {
 	int res = statvfs(path.c_str(), &diskstat);
 
 	if (res == 0) {
+		// Not sure why we're excluding Android here anyway...
```
unknownbrackets, line 44: https://github.com/hrydgard/ppsspp/commit/f687613d6875a0cb95eb35ea671e71bcc6a36aa6

IIRC, the flag field wasn't there back then, at least...

-[Unknown]

hrydgard:
This is working well now, and I will merge it later this week after some more testing.

hrydgard:
Failed test results so far:

* Galaxy S8 (Android 9, API 28): Misdetecting scoped storage, broken folder browser and wacky content URIs. Should have been using the old ways.

hrydgard:
Tested the flow again on some devices and made a few minor UX improvements.

This will be merged tomorrow.

hrydgard:
YOLO

UI/MemStickScreen.cpp
```diff
@@ -51,49 +112,91 @@ void MemStickScreen::CreateViews() {
...
+	// For legacy Android systems, so you can switch back to the old ways if you move to SD or something.
+	// TODO: Gonna need a scroll view.
 #if PPSSPP_PLATFORM(ANDROID)
-	if (Android_IsContentUri(path.ToString())) {
-		freeSpaceAtMemStick = Android_GetFreeSpaceByContentUri(path.ToString());
-	} else {
-		freeSpaceAtMemStick = Android_GetFreeSpaceByFilePath(path.ToString());
+	if (!System_GetPropertyBool(SYSPROP_ANDROID_SCOPED_STORAGE)) {
+		leftColumn->Add(new Choice(iz->T("Use PSP folder at root of storage")))->OnClick.Handle(this, &MemStickScreen::OnUseStorageRoot);
+		leftColumn->Add(new TextView(iz->T("ChooseFolderDesc", "* Data will stay even if you uninstall PPSSPP.\n* Data can be shared with PPSSPP Gold\n* Easy USB access"), ALIGN_LEFT, false));
```
iota97, line 144: `Data can be shared with PPSSPP Gold` doesn't work really well for Gold version itself, maybe something like: "Data can be shared between normal and Gold version of PPSSPP".

The path `Android/data/org.ppsspp.ppsspp/files` displayed below should be different for the gold version too to avoid confusion.

hrydgard: Oh yeah, I noticed this a few days ago and then completely forgot to fix it. Thanks!

PR #14627, opened 2021-07-19
Path and I/O fixes preparing for Android scoped storage

hrydgard:
This is most of what's currently in #14619 , but without enabling the enforcement. Want to get this in before, to check that it has no adverse effects.

Prep for #11997


hrydgard:
I'm gonna go ahead and merge this, hoping for the best :)  Doesn't seem to break anything that I can see.

PR #14659, opened 2021-07-29
Update InstallZipScreen.cpp

hsigamerppsspp:


hrydgard:
What?

(Better SD card support is coming in the upcoming version anyway).

PR #14669, opened 2021-08-01
sceNetAdhoc: fix format specifier for debug log

SeanMcG:
With the correct printf specifier, there is no need to cast to int.


hrydgard:
%zu is from C99 but still support for it on all platforms isn't that old. On the other hand, by now, it really should work...

PR #14670, opened 2021-08-01
fix overlinking of OpenGL libraries

SeanMcG:
In CMake, FindOpenGL recommends linking to one of ~~GLX~~libGL (CMP0072 LEGACY) or GLX / libOpenGL (CMP0072 GLVND), not both. See https://cmake.org/cmake/help/latest/module/FindOpenGL.html.

We should prefer the libOpenGL where possible and fall back to GLX if not.

As always, Apple is a special snowflake and it links to OpenGL.framework, so skip the library selection logic in that case.

PR #14677, opened 2021-08-02
cmake: use vendor-neutral GL implementation when available

SeanMcG:
Otherwise, fall back to legacy GL

This should fix the libretro regression caused by #14670


hrydgard:
thanks!

PR #14679, opened 2021-08-03
Savestate load undo

iota97:
Create a savestate (shared between all game) before loading the selected one that can be used to undo the load.

About the UI, well I really dunno tbh... (Edit: remembered we had rewind button so placed it next to it)
Screenshot_2021-08-03_11-12-09

### DIFF
--- a/Core/SaveState.cpp
+++ b/Core/SaveState.cpp
@@ -412,18 +412,19 @@ namespace SaveState
 		return filename.GetFilename() + " " + sy->T("(broken)");
 	}
 
-	Path GenerateSaveSlotFilename(const Path &gameFilename, int slot, const char *extension)
-	{
+	std::string GenerateFullDiskId(const Path &gameFilename) {
 		std::string discId = g_paramSFO.GetValueString("DISC_ID");
 		std::string discVer = g_paramSFO.GetValueString("DISC_VERSION");
-		std::string fullDiscId;
 		if (discId.empty()) {
 			discId = g_paramSFO.GenerateFakeID();
 			discVer = "1.00";
 		}
-		fullDiscId = StringFromFormat("%s_%s", discId.c_str(), discVer.c_str());
+		return StringFromFormat("%s_%s", discId.c_str(), discVer.c_str());
+	}
 
-		std::string filename = StringFromFormat("%s_%d.%s", fullDiscId.c_str(), slot, extension);
+	Path GenerateSaveSlotFilename(const Path &gameFilename, int slot, const char *extension)
+	{
+		std::string filename = StringFromFormat("%s_%d.%s", GenerateFullDiskId(gameFilename).c_str(), slot, extension);
 		return GetSysDirectory(DIRECTORY_SAVESTATE) / filename;
 	}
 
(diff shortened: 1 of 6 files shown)


hrydgard:
Right, given that it's a global option, unlike the slot-specific undo-save, it doesn't really belong in the popups you get by clicking the image.

Maybe you can do a horizontal LinearLayout and put the buttons next to each other instead? Concerned about vertical space on small phone screens.

Undo latest save might also be convenient here.. Hm.

iota97:
Using horizontal layout it use the same vertical space as before with rewind I guess, may still need to scroll on some small device tho'.

Screenshot_2021-08-03_21-17-12

hrydgard:
Definitely an improvement!

I'll actually review the code tomorrow, but it's a great feature to have. 

Also I would like to redesign the screen a bit to make it harder to accidentally misclick load/save, but that's another story, and quite difficult without losing convenience. There'll still be value in this either way.

Core/SaveState.cpp
```diff
@@ -412,18 +412,19 @@ namespace SaveState
 		return filename.GetFilename() + " " + sy->T("(broken)");
 	}
 
-	Path GenerateSaveSlotFilename(const Path &gameFilename, int slot, const char *extension)
-	{
+	std::string GenerateFullDiskId(const Path &gameFilename) {
```
unknownbrackets, line 415: nit: This is called disc everywhere else.  Personally, I usually use "disk" for storage, and disc for CDs and UMDs.  But either way, I think we should be consistent here.

-[Unknown]

Core/SaveState.cpp
```diff
@@ -471,6 +460,63 @@ namespace SaveState
...
+						callback(status, message, data);
+					}
+				};
+
+				if (!backup.empty()) {
+					Save(backup.WithExtraExtension(".tmp"), -2, saveCallback, cbUserData);
+				} else {
+					auto sy = GetI18NCategory("System");
+					if (callback)
+						callback(Status::FAILURE, sy->T("Failed to save state for load undo. Error in the file system."), cbUserData);
```
unknownbrackets, line 487: Hm, if I'm trying to load a state and saving fails, maybe it's my disk being full... I feel like I'd still want to load state in this situation, and be frustrated if the emulator wouldn't let me load just because I can't save.

-[Unknown]

Core/SaveState.cpp
```diff
@@ -471,6 +460,63 @@ namespace SaveState
...
+		if (g_Config.sStateLoadUndoGame != GenerateFullDiskId(gameFilename)) {
+			auto sy = GetI18NCategory("System");
+			if (callback)
+				callback(Status::FAILURE, sy->T("Error: load undo state is from a different game"), cbUserData);
+			return false;
+		}
+
+		Path fn = GetSysDirectory(DIRECTORY_SAVESTATE) / LOAD_UNDO_NAME;
+		if (!fn.empty()) {
+			Load(fn, -2, callback, cbUserData); // Slot number is visual only, -2 will display special message (kinda have to find a better way)
```
unknownbrackets, line 510: I think -2 is okay if we make it a `constexpr` at the top of the file somewhere after includes (I would not expose it in the .h.)

-[Unknown]

unknownbrackets (approved):

PR #14683, opened 2021-08-03
Add ULES00581 to compat.ini for MTX EU JIT hack

Linblow:
Forgot to add ULES00581 to compat.ini for MTX Mototrax EU JIT hack
Also converted CRLF to LF in compat.ini


hrydgard:
Yeah, maybe we should probably bite the bullet on compat.ini to make it consistent, even though it does mess up the per-line history...

Although, does it really matter? Is there a huge advantage from having CR-only line endings in this file?

Linblow:
As you said, it's only for consistency. I personally don't mind about CRLF

hrydgard:
Alright, I merged the first of your commits to master. Closing the CRLF part for now.

PR #14690, opened 2021-08-06
Improve translatability of memstick screen

hrydgard:
@iota97 I think this solves the issue from 83b6139

### DIFF
--- a/UI/MemStickScreen.cpp
+++ b/UI/MemStickScreen.cpp
@@ -155,9 +155,9 @@ void MemStickScreen::CreateViews() {
 	leftColumn->Add(new Choice(iz->T("Use App Private Directory")))->OnClick.Handle(this, &MemStickScreen::OnUseInternalStorage);
 	// Consider https://www.compart.com/en/unicode/U+26A0 (unicode warning sign?)? or a graphic?
 	leftColumn->Add(new TextView(iz->T("DataWillBeLostOnUninstall", "Warning! Data will be lost when you uninstall PPSSPP!")))->SetBullet(true);
-	leftColumn->Add(new TextView(iz->T("DataCannotBeShared", "Data cannot be shared between PPSSPP regular / Gold!")))->SetBullet(true);
+	leftColumn->Add(new TextView(iz->T("DataCannotBeShared", "Data CANNOT be shared between PPSSPP regular/Gold!")))->SetBullet(true);
 #if GOLD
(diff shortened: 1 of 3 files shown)


UI/MemStickScreen.cpp
```diff
@@ -134,28 +134,33 @@ void MemStickScreen::CreateViews() {
...
+	leftColumn->Add(new TextView(iz->T("DataCanBeShared", "Data can be shared between PPSSPP regular/Gold.")))->SetBullet(true);
+	leftColumn->Add(new TextView(iz->T("EasyUSBAccess", "Easy USB access")))->SetBullet(true);
 
 	leftColumn->Add(new Choice(iz->T("Use App Private Directory")))->OnClick.Handle(this, &MemStickScreen::OnUseInternalStorage);
-	leftColumn->Add(new TextView(iz->T("InternalStorageDesc", "* Warning! Data will be deleted if you uninstall PPSSPP!\n* Data cannot be shared between PPSSPP regular/Gold\n* USB access through "
-#ifdef GOLD
-	"Android/data/org.ppsspp.ppssppgold/files"
+	// Consider https://www.compart.com/en/unicode/U+26A0 (unicode warning sign?)? or a graphic?
+	leftColumn->Add(new TextView(iz->T("DataWillBeLostOnUninstall", "Warning! Data will be lost when you uninstall PPSSPP!")))->SetBullet(true);
+	leftColumn->Add(new TextView(iz->T("DataCannotBeShared", "Data cannot be shared between PPSSPP regular / Gold!")))->SetBullet(true);
```
iota97, line 158: nit: `regular / Gold` have space before and after the slash (in the other string is `regular/Gold`). 

Also you may want `cannot` in all caps for emphasis

UI/MemStickScreen.cpp
```diff
@@ -134,28 +134,33 @@ void MemStickScreen::CreateViews() {
...
+#if GOLD
+	leftColumn->Add(new TextView(iz->T("USBAccessThrough", "USB access through Android/data/org.ppsspp.ppssppgold/files")))->SetBullet(true);
```
iota97, line 160: Mhhh I don't think this fix the problem `USBAccessThrough` is the key of both, this mean when a translator add a translation he have no way to set the 2 different string in the ini (unless I'm missing something about how the ini work?). Better use `GoldUSBAccessThrough` for the gold version.

hrydgard: Fixed

UI/MemStickScreen.cpp
hrydgard: Fixed

PR #14696, opened 2021-08-07
Allow to disable right analog diagonal

### LINKED ISSUE #14691: An option to disable diagonal on right analog (touch control)
### What should happen

using right analog as camera is very inacurate on touch screen because there is always accidental diagonal input. being able to disable diagonal is necessary.

### Who would this benefit

touch screen users who map the right analog to camera controls

### Platform (if relevant)

Android

any game that use dpads/action buttons as camera

none

### Checklist

iota97:
Fixes #14691.


hrydgard:
Kind of a weirdly specific feature, but does make some sense.

PR #14706, opened 2021-08-08
Fix Translation lang ja_JP.ini

nassau-tk:
Translation for Japanese.
But, Sorry I cant check on Android device.
Probably, No problem.

PR #14709, opened 2021-08-09
Scoped storage translation

iota97:
Add the keys in en_US for reference, fix ja_JP from #14706 and make progress string translatable

### DIFF
--- a/UI/MemStickScreen.cpp
+++ b/UI/MemStickScreen.cpp
@@ -368,6 +368,7 @@ UI::EventReturn ConfirmMemstickMoveScreen::OnMoveDataClick(UI::EventParams &para
 
 void ConfirmMemstickMoveScreen::update() {
 	UIDialogScreenWithBackground::update();
+	auto iz = GetI18NCategory("MemStick");
 
 	if (moveDataTask_) {
 		if (progressView_) {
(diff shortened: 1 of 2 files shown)


nassau-tk:
Is it means that if merged this PR then translate Japanese automatically?

hrydgard:
I think it's slightly early to translate these maybe? I expect to polish this stuff up a bit more, which might include changing the text and order of it quite a bit..

iota97:
I agree is too early, but since nassau-tk translated some already I see value in having them in (mostly because I think is better to make as clear as possible what can happen to your data). 

Maybe should just not update the en_US one for now tho'.

hrydgard:
Yeah, fair enough.

nassau-tk:
When Japanese translation wasn't indicated, Japanese would feel weak at the knees in fear.
lmao
It means...
The Japanese who don't try to understand English. I think they give up operating.

hrydgard:
Let's get it in, but be ready that there might be some changes to these.

nassau-tk:
Of course sir.
Then I will try translate quicklly.

nassau-tk:
OKey.
Translate is working.
But, A part is over character.(2 characters.)
Probably, 99.9% Japanaese can understanding it.
We'll assume that it's OK for now.

hrydgard:
Oh, hehe. I'll leave as much space as possible, maybe make the text shrink or wrap if it's not enough space. Gotta be ready for long translatations here I think ...

Issue #14570, opened 2021-06-28
Use the AltKit library to automatically enable JIT for jailed iOS users running AltServer


lonkle:
**Note**: I spoke with someone on Discord ([unknown]) and they told me to get ahold of someone named @hrydgard about this to look into. It should be pretty straightforward, but I'll detail as much as I can in this ticket.

### What should happen?
Jailed user's should be able to use the JIT compiler on iOS using using AltKit (https://github.com/rileytestut/AltKit)
whenever a user is running AltServer (which can enable JIT for a running app as long as it's on the same Wi-Fi network, or plugged in)

### Who would benefit from it?
Every Jailed iOS User

### What's it specific to (remove if not)?
iOS

### Does anything else have a similar feature?
I just worked with the Play! developers who also use CMake as part of their pipeline and they got this working for their jailed iOS 13.3 - 15.0+ users here: https://github.com/jpd002/Play-/issues/1078

That same developer (jpd002) made a PR for AltKit to integrate their changes to our package for all CMake projects (like PPSSPP).

### Additional Notes / Guidelines
You're free to use AltKit to enable AltJIT whenever you want, but I thought I'd provide my notes to you guys from my experience in order to help out. Guidelines:

1. Use the function (shown in the Github readme linked) to ask to enable on altJIT on launch
2. If it fails at launch you really have nothing to lose by requesting JIT at random intervals - given the user's conditions need to be right for the function to respond back to your app with "JIT enabled successful" - you could just use a software interpreter in the meantime and auto switch to JIT at the next available chance.

*Optional*: If you don't want to use the "enable AltJIT" function at random intervals until JIT gets successful enabled then I'd suggest requesting it once at app launch, and then on each instance the iPhone connects to a WiFi network or to USB - both of which can connect an iPhone AltServer which is the software that automatically (externally and automatically via this function) enables JIT for jailed users.

hrydgard:
Sounds good, and I would be happy to merge a pull request with this support, but can't prioritize this right now.

lonkle:
No prioritization needed from you, I'm just opening this issue for anyone on your team (thanks for labeling it). Tbh, we haven't even merged in jpd002's CMake compatibility update PR quite yet anyway (tho any CMake dev can just use his PR directly). 

Anyway, this whole ticket simply outlines what you need to do to get your jailed iOS users (sideloading is becoming bigger and bigger) JIT whenever they need it. Good luck and wonderful project! Keep it up. 🎉

[...]

PR #14733, opened 2021-08-16
[WIP] iOS AltKit support for JIT

JoeMatt:
refs #14570 

## Overview

- Add AltKit CMake (currently a custom branch until https://github.com/rileytestut/AltKit/pull/1 is merged)
- Update Info.plist for bonjour usage string
- Add basic code to connect to AltKit
- Conditionally import AltKit in AppDelegate.mm

Build instructions I'm using;
```sh
cmake -DCMAKE_TOOLCHAIN_FILE=./cmake/Toolchains/ios.cmake -DIOS_PLATFORM=OS -DCMAKE_IOS_SDK_ROOT=/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk -H. -Bbuild.ios -GXcode
```

## References

https://github.com/jpd002/Play-/issues/1078
https://github.com/jpd002/AltKit/tree/cmake_poc

## Todo

- [ ] Activate/Disable JIT in PPSSPP on success/failure callbacks
- [ ] Fix import AltKit (how to do in CMake?)
- [x] Fix CMake error  `CMake Error at ext/AltKit/CMakeLists.txt:3 (project):
  Language 'C' is currently being enabled.  Recursive call not allowed.`
  
  
## Discussion

I wanted to test JIT in AltKit for another project, and PPSSPP seemed like a good project to test since I don't have JIT code active in my other project.

I'm a CMake noob, I couldn't figure out the AddLanguage error, so I'm posting this WIP PR to see if anyone else wants to take a stab or give some feedback/help.

Thanks

ping @lonkle

### DIFF
--- a/cmake/Toolchains/ios.cmake
+++ b/cmake/Toolchains/ios.cmake
@@ -12,7 +12,7 @@
 # PPSSPP platform flags
 set(MOBILE_DEVICE ON)
 set(USING_GLES2 ON)
-set(IPHONEOS_DEPLOYMENT_TARGET 6.0)
+set(IPHONEOS_DEPLOYMENT_TARGET 9.0)
 add_definitions(
   -DGL_ETC1_RGB8_OES=0
   -U__STRICT_ANSI__
@@ -96,3 +96,7 @@ set(CMAKE_SYSTEM_FRAMEWORK_PATH
 set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)
 set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)
 
+add_compile_options($<$<NOT:$<COMPILE_LANGUAGE:Swift>>:-fPIC>)
+# Causes error
+# project(Common LANGUAGES C Swift)
+add_subdirectory(ext/AltKit)
--- a/cmake/Toolchains/ios.cmake
+++ b/cmake/Toolchains/ios.cmake
@@ -17,7 +17,8 @@ add_definitions(
   -DGL_ETC1_RGB8_OES=0
 )
 
-add_definitions($<$<NOT:$<COMPILE_LANGUAGE:Swift>>:-U__STRICT_ANSI__ >)
+# TODO: Not sure how to do this, need to conditionally flag or else Swift test fails
+#add_definitions($<$<NOT:$<COMPILE_LANGUAGE:Swift>>:-U__STRICT_ANSI__ >)
 
 set(OPENGL_LIBRARIES "-framework OpenGLES")
 set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -mios-version-min=${IPHONEOS_DEPLOYMENT_TARGET}")
--- a/cmake/Toolchains/ios.cmake
+++ b/cmake/Toolchains/ios.cmake
@@ -15,8 +15,9 @@ set(USING_GLES2 ON)
 set(IPHONEOS_DEPLOYMENT_TARGET 9.0)
 add_definitions(
   -DGL_ETC1_RGB8_OES=0
-  -U__STRICT_ANSI__
 )
+
+add_definitions($<$<NOT:$<COMPILE_LANGUAGE:Swift>>:-U__STRICT_ANSI__ >)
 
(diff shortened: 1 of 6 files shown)


hrydgard:
Yeah this might be useful for some iOS users. Enabled CI for your account, I'm open to merging this.

lonkle:
Wanted to add a note on this PR after working with @JoeMatt today to really follow this entire conversation here: https://github.com/jpd002/Play-/issues/1078 - and the PoC branch's commits of jpd002's Play! app which is the best example of integrating AltKit into a Cmake project: https://github.com/jpd002/AltKit/tree/cmake_poc - we'll make an official Cmake build guide down the road but these are early days.

cmake/Toolchains/ios.cmake
```diff
@@ -12,12 +12,14 @@
 # PPSSPP platform flags
 set(MOBILE_DEVICE ON)
 set(USING_GLES2 ON)
-set(IPHONEOS_DEPLOYMENT_TARGET 6.0)
+set(IPHONEOS_DEPLOYMENT_TARGET 9.0)
```
unknownbrackets, line 15: Is this the minimum supported version or the maximum?  Just asking to understand the impact.

-[Unknown]

unknownbrackets (commented):
Currently, `./b.sh --ios` uses Make, which apparently isn't supported for Swift:

> Swift language not supported by "Unix Makefiles" generator

Generally, `./b.sh` is the recommended way to build and also what our CI checks use to make sure other changes don't break iOS, so we'd need to get that working (not sure if using xcodebuild or etc.?)

-[Unknown]

ios/PPSSPP-Info.plist
```diff
@@ -202,5 +202,11 @@
 			</dict>
 		</dict>
 	</array>
+	<key>NSBonjourServices</key>
```
ds8k, line 205: I think this needs a dummy `ALTDeviceID` plist key in order to enable JIT automatically

https://discord.com/channels/625766896230334465/626109028824514570/882747590679076895

cmake/Toolchains/ios.cmake
ds8k: It's minimum supported version

ios/PPSSPP-Info.plist
JoeMatt: Updated

cmake/Toolchains/ios.cmake
JoeMatt: Min, plus I don't think any device runnion ios 6-8 could even run ppsspp.

JoeMatt:
> Currently, `./b.sh --ios` uses Make, which apparently isn't supported for Swift:
> 
> > Swift language not supported by "Unix Makefiles" generator
> 
> Generally, `./b.sh` is the recommended way to build and also what our CI checks use to make sure other changes don't break iOS, so we'd need to get that working (not sure if using xcodebuild or etc.?)
> 
> -[Unknown]

I think cmake would be the easier way to go. Would you prefer if I update the CI scripts to work with cmake?

unknownbrackets:
Well, `b.sh` just runs CMake for you with the right arguments, but it's going to use Unix Makefiles.  It might be okay if it uses `-G Xcode`, as long as you modify `b.sh` to do that.  It'd be best if this could continue working with CMake without forcing a specific generator.

But yes, either way it would be CMake.  We don't include any non-CMake build files for iOS, as far as I know.

-[Unknown]

JoeMatt:
> Well, `b.sh` just runs CMake for you with the right arguments, but it's going to use Unix Makefiles. It might be okay if it uses `-G Xcode`, as long as you modify `b.sh` to do that. It'd be best if this could continue working with CMake without forcing a specific generator.
> 
> But yes, either way it would be CMake. We don't include any non-CMake build files for iOS, as far as I know.
> 
> -[Unknown]

Noted thanks.

In the mean time, I think I'm pretty much stuck anyway right now.
I have the taret building the swift libs and linking them, but the header includes aren't working since its making static libs not .frameworks and I have no idea how to add the required headers for `#import` to work in `AppDelegate.m`

PR #14735, opened 2021-08-16
(Libretro) Switch back to earlier base_width/base_height values

inactive123:
We decided on fixing this issue at the frontend level instead of the core side. @hrydgard Our apologies for this secondary PR, we'd appreciate if it could be merged quickly.


unknownbrackets (approved):
If this is for the surface, matching the internal resolution is important.  If this is for the display, it can of course be any size...

-[Unknown]

PR #14744, opened 2021-08-19
Improve touch control layout screen

iota97:
Helps #14743 a bit.

1. Allow touch control to be only half inside the screen, our bound are kinda large so I see people wanting to place them even further (half inside should be enough to drag it back, really worst case scenario one can reset but shouldn't be needed).

2. Dragging now select the best match based on the click distance from the various button center, this should help when you get 2 button overlapping and it keep selecting the first that was added to the screen (was kinda annoying tbh).

3. Here sadly thing get tricky: action button was always added (it just draw nothing if is not visible) so it get in the way. I make so it get added only if there is at least 1 action button visible ~~but this still going to need some changes to make it right~~.

Edit: Made it handle action button as it should, tested all the possible combination (rendering an overlay of the drag area) and all seem fine.

### DIFF
--- a/UI/TouchControlLayoutScreen.cpp
+++ b/UI/TouchControlLayoutScreen.cpp
@@ -50,6 +50,12 @@ public:
 		return false;
 	};
 
+	void Draw(UIContext &dc) override {
+		dc.PushScissor(screenBounds_);
+		MultiTouchButton::Draw(dc);
+		dc.PopScissor();
+	}
+
 	virtual void SavePosition() {
 		x_ = (bounds_.centerX() - screenBounds_.x) / screenBounds_.w;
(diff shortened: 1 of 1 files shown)


hrydgard:
Code looks good. I'm starting to think though that maybe we should turn the four action buttons into separate entities instead of having them joined together in a clump. Would also simplify this code.

I'm open to merging this as-is as an intermediate step though. What do you think @iota97  ?

iota97:
I kinda prefer them together, it make setting them up easier for most use cases (and if one want to split them for some different layout can still use the combo button that can have the same icon now as well).

hrydgard:
Hm, yeah, that's true, didn't think of that. At least that's a solution we can point to if someone wants to split them.

PR #14747, opened 2021-08-20
Make touch layout position accurate

### LINKED ISSUE #14743: Touch Control customization screen is super buggy

iota97:
This make screen rateo and scale correct. Fixes #14743.

I prefer this rather than add a drag drop menu or such for the left column button as I think that will be really hard to get the UX right (like if I place a button next to the edge and I want to drag it back should I open the menu or no?)

Result:
Screenshot_2021-08-20_12-44-41

This is just a test to show it's accurate (ofc the column width is normal in the PR):
Screenshot_2021-08-20_12-43-51
Screenshot_2021-08-20_12-43-36

### DIFF
--- a/UI/TouchControlLayoutScreen.cpp
+++ b/UI/TouchControlLayoutScreen.cpp
@@ -32,6 +32,8 @@
 #include "UI/TouchControlLayoutScreen.h"
 #include "UI/TouchControlVisibilityScreen.h"
 
+static float layoutAreaScale = 1.0f;
+
 static u32 GetButtonColor() {
 	return g_Config.iTouchButtonStyle != 0 ? 0xFFFFFF : 0xc0b080;
 }
@@ -51,9 +53,11 @@ public:
 	};
 
 	void Draw(UIContext &dc) override {
+		scale_ = theScale_*layoutAreaScale; // Scale down just for rendering
 		dc.PushScissor(screenBounds_);
 		MultiTouchButton::Draw(dc);
 		dc.PopScissor();
+		scale_ = theScale_/layoutAreaScale; // is this is needed?
 	}
 
 	virtual void SavePosition() {
@@ -83,10 +87,8 @@ protected:
 		return std::max(0.5f, opacity);
 	}
 	const Bounds &screenBounds_;
-
-private:
-	float &x_, &y_;
 	float &theScale_;
+	float &x_, &y_;
 };
 
 class PSPActionButtons : public DragDropButton {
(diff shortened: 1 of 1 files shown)


hrydgard:
Hm, the arrow keys look weirdly large in the picture where you prove that it's accurate?

iota97:
Oh yeah, its just little spacing and big scale, was to check if the line overlap in the same way. Was just to test with scale and spacing as well, not a real layout I would use for sure.

UI/TouchControlLayoutScreen.cpp
```diff
@@ -612,7 +619,8 @@ void TouchControlLayoutScreen::CreateViews() {
 	tabHolder->SetTag("TouchControlLayout");
 	root_->Add(tabHolder);
 
-	layoutView_ = root_->Add(new ControlLayoutView(new AnchorLayoutParams(leftColumnWidth + 10, 0.0f, 0.0f, 0.0f, false)));
+	root_->Add(new ItemHeader("", new AnchorLayoutParams(leftColumnWidth + 10, bounds.h*(1.0-layoutAreaScale)-40, NONE, NONE, false)));
```
unknownbrackets, line 622: Hm, it's a bit weird to use an ItemHeader, might be better to use a spacer or something.

-[Unknown]

iota97: Mhh, I was wondering to use it to add a tip like: "You may edit visibility and add additional controls pressing the Customize button on the left" as that screen might get missed.

Tho' I was not really sure how to write it down and then I forgot :P

Gamemulatorer:
Can you fix the word "customize" is not correct align on my android 11 with 6.5 inches screen.
Screenshot_2021-09-12-17-10-58-282_org ppsspp ppsspp

PR #14754, opened 2021-08-22
Fix store attempt

iota97:
`Path::empty()` was not working properly, I wonder if this was causing #14752.


hrydgard:
Really? path_ is an std::string, it should already default to "", right?

iota97:
Oh, yeah. I must be confusing with some other thing, my bad. Closing.

PR #14768, opened 2021-08-28
Fix the math in cardboard VR mode for wide aspect ratios

### LINKED ISSUE #12982: Google Cardboard settings on a bigger phone

hrydgard:
Avoids issues with very wide aspect ratios. Also very minor cleanups.

Basically, splits the full screen in half properly before centering, instead of chopping both width and height in half and reusing the normal centering, which doesn't work when the screen is wider than 16:9.

Fixes #12982

### DIFF
--- a/GPU/Common/PresentationCommon.cpp
+++ b/GPU/Common/PresentationCommon.cpp
@@ -151,10 +151,11 @@ void PresentationCommon::GetCardboardSettings(CardboardSettings *cardboardSettin
 		cardboardSettings->enabled = false;
 		return;
 	}
+
 	// Calculate Cardboard Settings
 	float cardboardScreenScale = g_Config.iCardboardScreenSize / 100.0f;
 	float cardboardScreenWidth = pixelWidth_ / 2.0f * cardboardScreenScale;
-	float cardboardScreenHeight = pixelHeight_ / 2.0f * cardboardScreenScale;
+	float cardboardScreenHeight = pixelHeight_ * cardboardScreenScale;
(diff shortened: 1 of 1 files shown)


Dominiquedb:
https://github.com/hrydgard/ppsspp/issues/12982#issuecomment-907685703

I have not given up on the feature, never will, I love psp games in vr glasses, and having it working perfectly in ppsspp is the dream.

This update is good, but not quite there yet to get the optimal experience. Same settings in v1.11.3 and v1.11.3-1203-g2735b09eb, vs RetroArch with shutter-to-side-by-side shader.
More tweaking possibilities for the interpupillary distance would still improve upon the experience, the option to actually move the two screens into each other for instance would be great.

As you can see in the screenshots, the center of the screen remains on the same location, but by changing the eye separation (X- shift) closer together a bigger image is possible, taking full advantage of the phone's display.

Screenshot_20210829-124906405
Screenshot_20210829-125403821-01
Screenshot_20210829-125000109-01
Screenshot_20210829-131822401-01

hrydgard:
Yeah, okay. Now that the math is right, just extending the range of the sliders should get us closer to that, and maybe additionally blending them into each other... That would take a bit of work though, so as a first step I'm just gonna extend the slider range a bit, shortly.

PR #14777, opened 2021-08-29
Handle OpenGL vertex shader generation failures better.

hrydgard:
Suspect that this is happening with the HW tesselation crash in KOF '96.

Should help #14774 . Might still glitch but hopefully won't crash.


hrydgard:
Just gonna merge this, shouldn't be able to break anything that's currently working.

PR #14778, opened 2021-08-29
Better binding popup

iota97:
Just to note when is waiting for the key press for the bind the relative row will "blink" (sadly I have terrible timing with screenshot so it that weird color in the pic eheh).

Screenshot_2021-08-29_23-42-02

This will be better with some change in the main control binding list I think, but not sure how to design that (suggestion are welcome as always).

~~PS: I have removed the "refocus row" code path for now as its broken in master, will open an issues about that and another focus related bug :)~~

### DIFF
--- a/UI/ControlMappingScreen.cpp
+++ b/UI/ControlMappingScreen.cpp
@@ -45,6 +45,9 @@
 #include "UI/ControlMappingScreen.h"
 #include "UI/GameSettingsScreen.h"
 
+constexpr int NEW_KEY_ADD = -1;
+constexpr int NEW_MOUSE_ADD = -2;
+
 class SingleControlMapper : public UI::LinearLayout {
 public:
 	SingleControlMapper(int pspKey, std::string keyName, ScreenManager *scrm, UI::LinearLayoutParams *layoutParams = nullptr);
@@ -60,12 +63,14 @@ private:
 
 	enum Action {
 		NONE,
-		REPLACEONE,
-		REPLACEALL,
-		ADD,
+		REPLACE_ONE,
+		REPLACE_ALL,
+		ADD_KEY,
+		ADD_MOUSE,
 	};
 
-	UI::Choice *addButton_ = nullptr;
+	UI::Choice *addKeyButton_ = nullptr;
+	UI::Choice *addMouseButton_ = nullptr;
 	UI::Choice *replaceAllButton_ = nullptr;
 	std::vector<UI::View *> rows_;
 	Action action_ = NONE;
@@ -115,20 +120,31 @@ void SingleControlMapper::Refresh() {
 	}
 	root->Add(replaceAllButton_)->OnClick.Add([=](EventParams &e) {
 		auto mc = GetI18NCategory("MappableControls");
-		action_ = REPLACEALL;
-		scrm_->push(new KeyMappingDialog(pspKey_, std::bind(&SingleControlMapper::MappedCallback, this), mc));
+		action_ = REPLACE_ALL;
+		scrm_->push(new KeyMappingDialog(pspKey_, 0, std::bind(&SingleControlMapper::MappedCallback, this), mc));
 
 		return UI::EVENT_DONE;
 	});
 
-	addButton_ = root->Add(new Choice(" + ", new LayoutParams(WRAP_CONTENT, itemH)));
-	addButton_->OnClick.Add([=](EventParams &e) {
+	addKeyButton_ = root->Add(new Choice(" + ", new LayoutParams(WRAP_CONTENT, itemH)));
+	addKeyButton_->OnClick.Add([=](EventParams &e) {
 		auto mc = GetI18NCategory("MappableControls");
-		action_ = ADD;
-		scrm_->push(new KeyMappingDialog(pspKey_, std::bind(&SingleControlMapper::MappedCallback, this), mc));
+		action_ = ADD_KEY;
+		scrm_->push(new KeyMappingDialog(pspKey_, NEW_KEY_ADD, std::bind(&SingleControlMapper::MappedCallback, this), mc));
 
 		return UI::EVENT_DONE;
 	});
+
+	if (g_Config.bMouseControl) {
+		addMouseButton_ = root->Add(new Choice("M", new LayoutParams(WRAP_CONTENT, itemH)));
+		addMouseButton_->OnClick.Add([=](EventParams &e) {
+			auto mc = GetI18NCategory("MappableControls");
+			action_ = ADD_MOUSE;
+			scrm_->push(new KeyMappingDialog(pspKey_, NEW_MOUSE_ADD, std::bind(&SingleControlMapper::MappedCallback, this), mc));
+
+			return UI::EVENT_DONE;
+		});
+	}
 
 	LinearLayout *rightColumn = root->Add(new LinearLayout(ORIENT_VERTICAL, new LinearLayoutParams(rightColumnWidth, WRAP_CONTENT)));
 	rightColumn->SetSpacing(2.0f);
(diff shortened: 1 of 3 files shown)


unknownbrackets:
Hm, so this changes it so that the PSP view now uses a different dialog, only for it, to bind.  And now, whether you click the PSP button (i.e. D-Pad Left) or the mapped button (i.e. kbd.j), it shows the same screen and lets you replace or delete here.

I think the cancel button should probably be Cancel and centered (to match how they look in other pop ups.)

A bit worried about the pop up being too "busy."  If we do this, it might pair better with the PSP view, although I think "Rebind All" at least would need to go through a simpler flow (so it can still visually highlight and dodge the button that's now being prompted.)

-[Unknown]

UI/ControlMappingScreen.cpp
```diff
@@ -349,30 +289,148 @@ bool KeyMappingNewKeyDialog::key(const KeyInput &key) {
...
+}
+
+void KeyMappingDialog::CreatePopupContents(UI::ViewGroup *parent) {
 	using namespace UI;
 
 	auto km = GetI18NCategory("KeyMapping");
+	auto mc = GetI18NCategory("MappableControls");
+	auto di = GetI18NCategory("Dialog");
+
+	ScrollView *scroll = new ScrollView(ORIENT_VERTICAL, new LinearLayoutParams(FILL_PARENT, 45, 1.0f));
```
unknownbrackets, line 351: Specifying a height in layout params for a ScrollView currently doesn't work, but #14779 will fix it.  That said, I don't know if you meant to use 45 - that's pretty short?

-[Unknown]

UI/ControlMappingScreen.cpp
```diff
@@ -349,30 +289,148 @@ bool KeyMappingNewKeyDialog::key(const KeyInput &key) {
...
+	LinearLayout *items = new LinearLayoutList(ORIENT_VERTICAL);
+	items->SetSpacing(5.0f);
+
+	static const std::string hint = "Hint: you may map the same key to different PSP button to be able to press multiple one with a single key!";
+	parent->Add(new TextView(km->T(hint), FLAG_DYNAMIC_ASCII | FLAG_WRAP_TEXT, true, new LinearLayoutParams(Margins(10,0))));
```
unknownbrackets, line 356: This isn't dynamic ASCII, it can be translated.  That said, I'm worried this might be a bit confusing... not sure if there's a better way.

If we keep it, trying to think of a shorter message... maybe "Hint: Mapping the same key to multiple PSP buttons will press both at once"?

-[Unknown]

UI/ControlMappingScreen.cpp
```diff
@@ -349,30 +289,148 @@ bool KeyMappingNewKeyDialog::key(const KeyInput &key) {
...
+			this->selectedIndex_ = 0;
+			m->SetBinding(true);
+			g_Config.bMapMouse = true;
+			return UI::EVENT_CONTINUE;
+		});
+	}
+
+	items->Add(new Spacer(10.0));
+	Choice *b = items->Add(new Choice(di->T("Back"), new LinearLayoutParams(FILL_PARENT, 45)));
+	b->OnClick.Handle<UIScreen>(this, &UIScreen::OnBack);
```
unknownbrackets, line 406: I'd suggest OnCancel and naming it cancel.  DR_BACK should mainly be used for screens rather than pop ups.

-[Unknown]

UI/ControlMappingScreen.h
```diff
@@ -78,27 +78,26 @@ class KeyMappingNewKeyDialog : public PopupScreen {
...
+	virtual bool axis(const AxisInput &axis) override;
 
 protected:
 	void CreatePopupContents(UI::ViewGroup *parent) override;
 
-	bool FillVertical() const override { return false; }
-	bool ShowButtons() const override { return true; }
-	void OnCompleted(DialogResult result) override {}
+	virtual bool FillVertical() const override { return false; }
+	virtual bool ShowButtons() const override { return true; }
```
unknownbrackets, line 95: nit: This is the default anyway?  Also, you don't actually show these buttons because the button1_ text is always "".  So this override isn't needed.

I suspect you copied this from elsewhere, but we're currently preferring overrides to be without the `virtual` (since it's already an `override`, it's duplicative to say `virtual` too.)

-[Unknown]

UI/ControlMappingScreen.h
```diff
@@ -78,27 +78,26 @@ class KeyMappingNewKeyDialog : public PopupScreen {
...
+class BindingChoice;
+
+class KeyMappingDialog : public PopupScreen {
 public:
-	explicit KeyMappingNewMouseKeyDialog(int btn, bool replace, std::function<void(KeyDef)> callback, std::shared_ptr<I18NCategory> i18n)
-		: PopupScreen(i18n->T("Map Mouse"), "", ""), callback_(callback), mapped_(false) {
-		pspBtn_ = btn;
-	}
+	explicit KeyMappingDialog(int btn, std::shared_ptr<I18NCategory> i18n)
+		: PopupScreen(std::string(i18n->T("Mapping for"))+" "+i18n->T(KeyMap::GetPspButtonName(btn))), pspBtn_(btn) {}
```
unknownbrackets, line 86: nit: I know this is used elsewhere, but I personally greatly prefer to use for translation:
```c++
std::string label = ReplaceAll(km->T("Currently mapped for %1:"), "%1", mc->T(pspButtonName));
```

Rather than:
```c++
std::string label = km->T("Currently mapped for") + " " + mc->T(pspButtonName) + ":";
```

The reason?  Word order and punctuation can be very tricky to get right the second way.  Another language may prefer to phrase this as "[BUTTON]のマッピング：", which as you can see is in the opposite order and has different punctuation spacing.

-[Unknown]

UI/ControlMappingScreen.h
iota97: Yeah was copy pasted form the previous mapping popup views (fixed them as well).

UI/ControlMappingScreen.cpp
iota97: Mhhhh, after #14779 how can I make the popup actually wrap content? It get full screen height even with few item (same goes now for the grid setting in the main screen, but probably less visible giving it have quite some button in it).

Anyway gotta call if for the night, I'll think about it tomorrow :)

unknownbrackets: Fixed the WRAP_CONTENT issue in #14781.  Better for that to be fixed properly so we're not hacking a workaround...

-[Unknown]

iota97:
Rebased on master with the focus fixes and so, also made it select the action based on where you press in the control mapping screen (e.g. if you press the " + " the popup will open waiting for the new binding already).

hrydgard:
Good stuff, but I kinda liked the look of Unknown's version:

image

I'll look more at this later today.

hrydgard:
It works but on the main mapping screen it's kinda redundant, makes more sense on top of Unknown's PSP shadow screen. There's also bug where if you click a binding to replace it, and then click Add, two things are suddenly blinking...

unknownbrackets:
I think there are pros and cons to both ways, but here's the change I'd started:

https://github.com/hrydgard/ppsspp/compare/master...unknownbrackets:ui-ctrl-mapping

It turns the replace checkbox into a selection:
Image showing View, Add, and Replace selections next to vector image of PSP

Selecting Bind All in "View" mode then asks if you want Add or Replace:
Dialog "Add or replace all PSP bindings?"

And if you click a button first, you see its mapped buttons along with Add/Replace (if in "View" mode, the default):
Dpad Up mapping showing 3 buttons mapped

If you select Add or Replace on the left first, it skips these extra steps and goes directly to the current mapping, which just shows a dialog waiting for you to press a button.

A downside of the interface I just described is that there isn't a clear way to remove one, although the items in the dialog could be made clickable.  That said, I think it's also less dense/busy.

-[Unknown]

iota97:
I think the PSP shadow UI should be used for quick setup, so it should be better a less dense UI. 

I would go for unknown one, just a couple of thing I noticed:
- the button list for analog make little sense, probably could use an item header or something to separate the 4 axis.
- nit: in `ShowKeyMappingDialog` the method `ShowButtons` is market both virtual and override while the method `FillVertical` is the default so not really needed (I'm not the only one copy pasting that part I see :P)
- the PSP body have like 50% alpha, but the screen image does not so it kinda look off with different background (pic rel)

Screenshot_2021-09-08_07-05-09

PR #14782, opened 2021-08-30
Support virtual button in touch gesture

iota97:
Follow up to #13107. 

It won't allow mapping all the virtual button as I would make little sense like to bind fast-forward while swiping (as is not toggle is useless even on double tap).


UI/GamepadEmu.h
```diff
@@ -289,25 +313,44 @@ namespace CustomKey {
...
+	struct key {
+		const char* n; // UI name
+		uint32_t c; // Key code
+	};
+	static const key keyList[] = {
+		{ "Square", CTRL_SQUARE },
+		{ "Triangle", CTRL_TRIANGLE },
+		{ "Circle", CTRL_CIRCLE },
+		{ "Cross", CTRL_CROSS },
+		{ "Up", CTRL_UP },
```
hrydgard, line 327: Hm, not sure exactly how, but wish we could reuse the strings that exist elsewhere for the same purpose in the main control mapper.

iota97: We had a method already to get the name in a `std::string`, tho' I added  a new one to get the `char *` as the string get translated when you click on the choice and the c_str() pointer might be invalid by then.

Made the same for combo button while was at it :)

Saramagrean:
sorry, what's about error button?

Screenshot_20210901-212621

hrydgard:
I'm sure that was not intended @iota97  :)

PR #14786, opened 2021-09-01
Step-Trace Logger for Windows Version

SylvieWrath:
Adds a toggleable manual trace-logger that activates on Manual Breaks, Manual Gos, Step Into, Step Over, and Step Out. Outputs to text files in \Trace Logs\ directory. 
image
Uses a checkbox on the disassembly debugger to toggle whether it's enabled.


LunaMoo:
Looking at the changes(ie. readme edit and some other unrelated/formatting changes) it seems you wanted to keep that in your personal fork and not as a PR?

Windows/Debugger/CtrlDisAsmView.cpp
```diff
@@ -1352,14 +1355,176 @@ void CtrlDisAsmView::disassembleToFile() {
...
 void CtrlDisAsmView::getOpcodeText(u32 address, char* dest, int bufsize)
 {
 	DisassemblyLineInfo line;
 	address = manager.getStartAddress(address);
-	manager.getLine(address,displaySymbols,line);
-	snprintf(dest, bufsize, "%s  %s",line.name.c_str(),line.params.c_str());
+	manager.getLine(address, displaySymbols, line);
+	
+	// Begin trace logger mods
+	std::string instructionName = std::string(line.name.c_str());
```
iota97, line 1368: `line.name` is a `std::string` already, no reason to get a `const char *`, construct a `std::string` out of it (elided by the compiler likely) and use it to construct `instructionName`. 

`c_str()` is useful mostly when you are interacting with C style method and function (e.g. printf) that require `const char *`.

TL;DR `std::string instructionName = line.name;` should do.

Windows/Debugger/CtrlDisAsmView.cpp
```diff
@@ -1352,14 +1355,176 @@ void CtrlDisAsmView::disassembleToFile() {
...
+	instructionName += "\t"; // Plus an extra tab character for increased readability
+
+	//DEBUG_LOG(CPU, parameters.c_str());
+
+	// Setting the paremeters "column" to have 32 characters uniformly.
+	while (parameters.length() < 32) {
+		parameters += " ";
+	}
+	// removed the space between %s and %s since i pre-formatted them. 
+	snprintf(dest, bufsize, "%s%s", instructionName.c_str(), parameters.c_str());
```
iota97, line 1389: `snprintf` and similar is made to format string, I suggest to search on google if there is syntax to format your string as you want with it directly first (there is a lot in it, too much to be worth learning rather than searching when you need tho').

In this case: https://stackoverflow.com/questions/293438/left-pad-printf-with-spaces

Windows/Debugger/CtrlDisAsmView.cpp
```diff
@@ -1352,14 +1355,176 @@ void CtrlDisAsmView::disassembleToFile() {
...
+
+	return;
+	// end modifications
 }
 
+void CtrlDisAsmView::enableTraceLogger()
```
iota97, line 1395: nit: you may want a method `SetEnabled(bool)` rather than 2 here (not that it really matter tho').

Windows/Debugger/CtrlDisAsmView.cpp
```diff
@@ -1377,3 +1542,50 @@ u32 CtrlDisAsmView::getInstructionSizeAt(u32 address)
 	u32 next  = manager.getNthNextAddress(start,1);
 	return next-address;
 }
+
+int regStringToMIPSGPReg(const char* name) {
```
iota97, line 1546: Here you may want to create a new struct that have both the int and the string.

```
struct regStringPair {
    int reg;
    const char *name;
};
```

And then declare and array with all the pair:

```
const regStringPair regNames[] = {
    {MIPS_REG_V0, "vo"},
    {MIPS_REG_V1, "v1"},
    ....
};
```

Then just loop the array and if you have match return what you want.

Windows/Debugger/CtrlDisAsmView.cpp
```diff
@@ -1377,3 +1542,50 @@ u32 CtrlDisAsmView::getInstructionSizeAt(u32 address)
...
+		if (Name == "ra") return MIPS_REG_RA;
+		if (Name == "hi") return MIPS_REG_HI;
+		if (Name == "lo") return MIPS_REG_LO;
+	}
+	else if (Name == "zero") {
+		return MIPS_REG_ZERO;
+	}
+	else return MIPS_REG_INVALID;
+	// extra return just to make the compiler happy.
+	return MIPS_REG_INVALID;
```
iota97, line 1590: It's not just to make the compiler happy, let's say someone add a new name to GPR_NAMES and forget to add it in this function, what will happen? I might sound silly now, but if you edit your code after some year is kinda easy for thing like this to happen.

I would remove the `else return MIPS_REG_INVALID;` like tho' as that is not needed and just left the last return if we didn't catch any condition before.

Windows/Debugger/CtrlDisAsmView.cpp
```diff
@@ -1377,3 +1542,50 @@ u32 CtrlDisAsmView::getInstructionSizeAt(u32 address)
 	u32 next  = manager.getNthNextAddress(start,1);
 	return next-address;
 }
+
+int regStringToMIPSGPReg(const char* name) {
+	std::string Name = name;
```
iota97, line 1547: you may just change the parameter type to `std::string` (or better` const std::string &`), if you pass a `const char*`
to a method asking for a `std::string` it will be implicit converted anyway.

Windows/Debugger/Debugger_Disasm.cpp
```diff
@@ -32,11 +32,37 @@
...
+
+
+
 // How long (max) to wait for Core to pause before clearing temp breakpoints.
 static const int TEMP_BREAKPOINT_WAIT_MS = 100;
 
 static FAR WNDPROC DefGotoEditProc;
 
+// Begin Trace Logger Changes
+int check = CreateDirectoryA("Trace Logs\\", NULL);
```
iota97, line 48: Why is this in global scope? Am I missing something?

Windows/Debugger/Debugger_Disasm.cpp
```diff
@@ -349,14 +394,16 @@ void CDisasm::runToLine()
 
 BOOL CDisasm::DlgProc(UINT message, WPARAM wParam, LPARAM lParam)
 {
+
 	//if (!m_hDlg) return FALSE;
-	switch(message)
+	switch (message)
```
iota97, line 399: All this indentation change add a lot of noise to the diff (you may want to keep the old style).

Windows/Debugger/Debugger_Disasm.cpp
SylvieWrath: It made sense to me to keep this in the global scope as long as the things below it are also in the global scope. The idea being I wanna be able to access file reading/writing at any point but the file it's writing to is determined by the date/time the debugger window is open, so whatever the "current" file is makes sense to keep globally, to me.

Windows/Debugger/Debugger_Disasm.cpp
SylvieWrath: why ARE definhitions for functions here highlighted in blue and showing weird-ass line coordinates?? This is something i don't quite get.

Windows/Debugger/CtrlDisAsmView.cpp
SylvieWrath: I think this was literally some of the first code i wrote so yeah i can clean this up.

Windows/Debugger/CtrlDisAsmView.cpp
SylvieWrath: I think i will try this. And maybe move it outside mips.h, too.

Windows/Debugger/CtrlDisAsmView.cpp
SylvieWrath: ah thanks, i was having trouble figuring out how to directly pass an std::string as an argument to a function.

Windows/Debugger/CtrlDisAsmView.cpp
SylvieWrath: Keeping this as is for now just because i prefer separate functions for now.

Windows/Debugger/CtrlDisAsmView.cpp
SylvieWrath: Im confused what you are suggesting here? (though i did fix my code according to your suggestion upthread)

Windows/Debugger/Debugger_Disasm.cpp
SylvieWrath: Ah nevermind, i misunderstood what you were talking about and fixed it, once i upload it that should be fixed.

Windows/Debugger/CtrlDisAsmView.cpp
```diff
@@ -1352,12 +1400,175 @@ void CtrlDisAsmView::disassembleToFile() {
 	}
 }
 
+
+// The following member has been modified to accomodate "nicer-looking" output (according to my needs). Assumes maximum length of an instruction name is 9 or 10 chars.
+// What's the maximum length the params string could reasonably be?
```
iota97, line 1405: I still think this should not be modified, you may add a new function like `getOpcodeTextFormatted` or something so it doesn't affect other part of the code where this is used.

Also I think all this code can be done like this (untested tho'):

```
DisassemblyLineInfo line;
address = manager.getStartAddress(address);
manager.getLine(address,displaySymbols,line);
snprintf(dest, bufsize, "%10s\t%32s", line.name.c_str(), trimString(line.params).c_str());
```

Windows/Debugger/CtrlDisAsmView.cpp
iota97: Yeah, way better.

Just a few thing:
You may just change the parameters type from `const char *name` to `const std::string &name` and then in the loop you can check with `if (name == regNames[i].name)`.

After this no need to add `.c_str()` when you call this function.

When you check the return type of this check for MIPS_REG_INVAID rather than -1, it get more clear what you are doing like that.

I would change the loop with `i < ARRAY_SIZE(regNames)`, so you don't have to count them and even if you change the array it will still work.

Windows/Debugger/Debugger_Disasm.cpp
```diff
@@ -32,11 +32,33 @@
...
+#include <fstream>
+#include <string>
+#include <sstream>
 // How long (max) to wait for Core to pause before clearing temp breakpoints.
 static const int TEMP_BREAKPOINT_WAIT_MS = 100;
 
 static FAR WNDPROC DefGotoEditProc;
 
+// Begin Trace Logger Changes
+static int check = CreateDirectoryA("Trace Logs\\", NULL);
```
iota97, line 45: Most of this should be moved either as a class member to to local scope when not needed elsewhere

Also STEP_INTO etc could use a enum.

Windows/ppsspp.rc
```diff
@@ -187,7 +187,8 @@ BEGIN
     PUSHBUTTON      "Step Out",IDC_STEPOUT,228,0,40,14
     PUSHBUTTON      "Next HLE",IDC_STEPHLE,276,0,40,14
     PUSHBUTTON      "Breakpoint",IDC_MEMCHECK,317,0,45,14
-    EDITTEXT        IDC_THREADNAME,370,3,150,10,ES_AUTOHSCROLL | ES_READONLY | NOT WS_BORDER
+    EDITTEXT        IDC_THREADNAME,365,3,150,10,ES_AUTOHSCROLL | ES_READONLY | NOT WS_BORDER
```
iota97, line 190: Few like above there is `IDD_DISASM DIALOGEX 0, 0, 510, 415` that should have his width changed too now (at least I think, never used this kind of file and I'm not even on windows...).

Windows/Debugger/Debugger_Disasm.cpp
```diff
@@ -912,3 +993,46 @@ void CDisasm::UpdateDialog(bool _bComplete)
 	InvalidateRect (GetDlgItem(m_hDlg, IDC_DEBUGMEMVIEW), NULL, TRUE);
 	UpdateWindow (GetDlgItem(m_hDlg, IDC_DEBUGMEMVIEW));
 }
+
+// More trace logger changes
+void CDisasm::writeToTraceLogger(int writeCode) {
+	CtrlDisAsmView* ptr = CtrlDisAsmView::getFrom(GetDlgItem(m_hDlg, IDC_DISASMVIEW));
+	u32 currentPc = cpu->GetPC();
+	u32 newAddress = currentPc + ptr->getInstructionSizeAt(currentPc);
+	traceLogger.open(traceLogPath.c_str(), std::ios::app);
```
iota97, line 1002: I don't think you should keep opening and closing the file at every write, we have our internal class for handling path and file btw, but I guess as this is windows only might be fine using std one directly.

Windows/Debugger/CtrlDisAsmView.cpp
SylvieWrath: That's a good call actually. I'll mess with that later. As for your code, well i could just steal it wholesale and test but now i feel like i should have figured it out myself XD

Windows/Debugger/Debugger_Disasm.cpp
SylvieWrath: I actually need to review how to use enum.

Windows/Debugger/Debugger_Disasm.cpp
SylvieWrath: If you don't close it after every write then it won't update until the file until you close the disassembly window. Which seems like better functionality in my mind?

Windows/Debugger/CtrlDisAsmView.cpp
SylvieWrath: Temporarily unresolving because i didn't realize the ARRAY_SIZE function existed and i will try to incorporate that as soon as i next have spoons to look at this.

Windows/Debugger/Debugger_Disasm.cpp
iota97: You may call `traceLogger.flush();` to get the changes updated on disk without closing the file.

Windows/Debugger/CtrlDisAsmView.cpp
iota97: Just to note ARRAY_SIZE is not a function but rather a macro, it get expanded by preprocessor to something like:
`sizeof(array)/sizeof(array[0])` and doesn't work on dynamic array or pointer at all. Nothing to concern about here, just for future reference.

Windows/Debugger/Debugger_Disasm.cpp
SylvieWrath: What is the functional difference between opening/closing constantly and using the flush command and closing at the end, out of curiosity? Like what makes one more reliable or something than the other?

iota97: Well I expect open/close to be slower (as it have to do more operation), but mostly is about how easy is the code to read, I mean if you want to save the changes to disk that operation is called flush and is better to use it rather than making other to achieve the same result.

Just to note in general open/close will reset the tell position while flush won't (in this case you are always appending so doesn't matter here).

SylvieWrath: okay, i will fix this too. Thank you for the explanation!

hrydgard:
While the idea is OK, the committer isn't around anymore and there are issues with this implementation. Might resurrect this in a different form in a later time, but for now, closing.

PR #14793, opened 2021-09-05
Multiple key for single button

iota97:
Allows to bind 2 button combination.

### DIFF
--- a/Common/Input/InputState.h
+++ b/Common/Input/InputState.h
@@ -80,21 +80,42 @@ enum {
 // Represents a single bindable key
 class KeyDef {
 public:
-	KeyDef() : deviceId(0), keyCode(0) {}
-	KeyDef(int devId, int k) : deviceId(devId), keyCode(k) {}
+	KeyDef() : deviceId(0), keyCode(0), deviceId2(0), keyCode2(0), combo(false) {}
+	KeyDef(int devId, int k) : deviceId(devId), keyCode(k), deviceId2(0), keyCode2(0), combo(false) {}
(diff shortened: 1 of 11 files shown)


hrydgard:
I like the feature, but asking for one or two buttons gets really confusing for new users I would think, and the messages will flicker by quickly for normal bindings. I've never seen a binding interface like this before. 

I need to think about if I want this in for 1.12, and how we'll solve the UX issue. But maybe it can be a hidden feature for a while?

hrydgard:
Could you rebase this? I'd like to experiment with this again.

iota97:
Rebase had quite some conflict (more "hidden" than not). Seems fine from quick test but I have no controller or multiple device to more complex test.

No UI because it changed in the meanwhile and I don't really have the time to understand what happened (also we was not sure how to do it anyway). This mean u have to press to button at binding time to create the combo, but it's kinda hidden feature :)

hrydgard:
It might be okay to be a hidden feature, as long as it's obvious afterwards what you did :)

I'll play with it and see what UI changes might be necessary, if any.

hrydgard:
Sorry for taking so long with this, but I have finally played with it a bit more and I want it in as is :)

Wanna rebase it or should I? Sorry again for the need for another rebase...

iota97:
I don't have access to a PC with a compiler/PPSSPP source code untill next month.

The conflict seems easy enough but tried the Web conflict fix once and it messed with indentation so I don't trust it anymore :P

It's probably the best if you could it if you don't wanna wait :)

hrydgard:
Alright, I'll do it pretty soon then :)

hrydgard:
Closing in favor of #17168.

PR #14816, opened 2021-09-09
Fix wave on different dps

iota97:
Fixes missing line when `pixel_in_dps_x` was not 1, also add a check about vertex buffer overflow (on 4k I think it could overflow the buffer, better be safe than sorry anyway).

### DIFF
--- a/Common/Render/DrawBuffer.cpp
+++ b/Common/Render/DrawBuffer.cpp
@@ -14,11 +14,6 @@
 #include "Common/Log.h"
 #include "Common/StringUtils.h"
 
-enum {
-	// Enough?
-	MAX_VERTS = 65536,
-};
-
 DrawBuffer::DrawBuffer() : count_(0), atlas(0) {
 	verts_ = new Vertex[MAX_VERTS];
 	fontscalex = 1.0f;
--- a/Common/Render/DrawBuffer.h
+++ b/Common/Render/DrawBuffer.h
@@ -14,6 +14,11 @@
 struct Atlas;
 
 enum {
+	// Enough?
+	MAX_VERTS = 65536,
+};
+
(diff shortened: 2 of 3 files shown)


Gamemulatorer:
Can you also add like a super wallpaper on android?

ResurrectionRedmi9:
Reduce or REMOVE the bluish effect when recent game is set as the animation background.
Screenshot_2021-09-09-18-29-06-195_org ppsspp ppsspp

iota97:
About super wallpaper I'm not totally sure what you meant but if you want something that change based on how you navigate the menu is currently not possible (ofc it can be implemented, but would take some work ig).

About the blueish effect it's caused by Recent game only have a 20% opacity so you see the default UI background behind, I kinda agree it should not blend as it will look off more than not. Will change that too.

Gamemulatorer:
> About super wallpaper I'm not totally sure what you meant but if you want something that change based on how you navigate the menu is currently not possible (ofc it can be implemented, but would take some work ig).
>
This is super wallpaper animation looks like.
https://youtu.be/IpJanHEUaE4

unknownbrackets:
I don't really agree on the recent game thing - it was meant to be more subtle and less in your face.  I don't think making it show as 100% looks as nice, then you should just set a single image...

-[Unknown]

iota97:
That also true.

Mhhh I think some user could still want to cycle all the images without blending, could have both ig? (Like a "Recent Game - Solid Background" or something). Not sure what's the best here.

iota97:
Removed the recent screen change, is out of the scope of this PR. I we want to add another animation with full alpha it can be done in a second time anyway.

unknownbrackets (approved):

PR #14818, opened 2021-09-10
Focus based moving background

iota97:
Add another animation that scroll and zoom the current background image based on how you move on the menu (based on current focused view position, last scrollView moved scroll position and screenManager stack size).

There is a bug: if one view is focused and you scroll without moving the focus (e.g. scroll wheel) the 2 movement cancel one another, but working around that result in way more complex code and I think it's fine enough for an UI animation.

Preview: https://youtu.be/O8axvu6X5L0.

PS: name is kinda terrible, but I can't think of anything better...

### DIFF
--- a/Common/UI/Root.cpp
+++ b/Common/UI/Root.cpp
@@ -24,7 +24,6 @@ static std::mutex eventMutex_;
 static std::function<void(UISound)> soundCallback;
 static bool soundEnabled = true;
 
-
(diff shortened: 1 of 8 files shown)


Gamemulatorer:
Its also cool if the background is also moving in main menu while scrolling recent, games, homebrew & demos. But overall this is really cool thanks @iota97 :)

PR #14819, opened 2021-09-10
Show grid setting in recent game tab too, slider for grid icon size

iota97:
Move the setting button in the tab holder and show it in recent tab too (I kinda like it more with no spacing like this, but I don't really have a strong opinion on this).
Screenshot_2021-09-26_14-33-40

Also use a slider rather than 2 button (this add support for unlimited stacks of popup rather than just 1).


Gamemulatorer:
How about this now?

Issue #14490, opened 2021-05-30
[Feature Request] Add touch gliding


ghost:
### What should happen?
Just like this please :)
https://github.com/stenzek/duckstation/pull/1723

### Who would benefit from it?

Android user
#### Hardware, device, or operating system?
Android
#### Checklist

 - [ ] Check the [latest git build](https://buildbot.orphis.net/ppsspp/) in case it's already implemented.
 - [ ] [Search for other requests](https://github.com/hrydgard/ppsspp/search?q=my+issue&type=issues) of the same feature.

PR #14843, opened 2021-09-13
DPad drag fixes

### LINKED ISSUE #14429: Moving the finger from an empty area in the screen onto the DPAD doesn't work
For every key, it is possible to touch and hold on an empty area in the screen and move the finger onto a key and the key fires.

This is not the case with the DPAD keys. If you miss the key, you'll have to lift your finger and put it back again.

### LINKED ISSUE #14721: Prevent analogs from triggering other buttons (touch controls)
### What should happen

This is a very specific problem and very minor, but I often drag the analog button too far and it triggers the unthrottle button. Also happens when sliding the right analog up and down because there is very small space.

I could always rearrange the position but, I wonder wether it's possible to prevent the analogs from trigerring other buttons when dragged (just like how it doesn't trigger the dpads).

doesn't trigger dpads:
Screenshot_20210812-123005
but triggers other buttons:
Screenshot_20210812-123019
right analog:
Screenshot_20210812-123524

Thank you.

### Who would this benefit

Android

### Platform (if relevant)

Android

Any

None

### Checklist

iota97:
Fixes #14721 and fixes #14429.

### DIFF
--- a/UI/GamepadEmu.cpp
+++ b/UI/GamepadEmu.cpp
@@ -35,6 +35,7 @@
 #include "UI/GamepadEmu.h"
 
 static uint32_t usedPointerMask = 0;
+static uint32_t analogPointerMask = 0;
 
 static u32 GetButtonColor() {
 	return g_Config.iTouchButtonStyle != 0 ? 0xFFFFFF : 0xc0b080;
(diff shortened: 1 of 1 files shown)


ResurrectionRedmi9:
How about this one https://github.com/hrydgard/ppsspp/issues/14490

unknownbrackets (approved):

PR #14848, opened 2021-09-13
Fix DPAD size calculation

iota97:
Not really sure what logic was behind the previous code, but didn't really work when changing spacing and scale too much.

### DIFF
--- a/UI/GamepadEmu.cpp
+++ b/UI/GamepadEmu.cpp
@@ -239,8 +239,14 @@ PSPDpad::PSPDpad(ImageID arrowIndex, const char *key, ImageID arrowDownIndex, Im
 }
 
 void PSPDpad::GetContentDimensions(const UIContext &dc, float &w, float &h) const {
-	w = D_pad_Radius * spacing_ * 4;
-	h = D_pad_Radius * spacing_ * 4;
+	const AtlasImage *image = dc.Draw()->GetAtlas()->getImage(arrowIndex_);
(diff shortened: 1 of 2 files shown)


hrydgard:
This is the kind of thing that's a bit scary to change because people will "feel" like something is not like they're used to.

Is this very similar to the old calculations near the default size?

iota97:
I see the concern...

About ProcessTouch I changed the float value scaling them based on what the value was before and after the change on default setting indeed. The only difference here should be that before the size was circular rather than a square like any other button once it was selected.

About ContextDimension the size is just a little smaller on default (it now match the draw size).

Before even at default size the size for pressing it was a square a bit bigger than the displayed size, after the press you could drag away a bit further than where you had to press and was also circular, not really sure if is possible to match this without having it all broken when scale or spacing change. Also I think we should be consisted with other button as much as possible.

I think is close enough, but maybe we can wait after 1.12 so we can get feedback is someone is hyper sensitive to this change. It was broken like this for like 8 year I guess and none ever complained afaik, no need to rush this next to a release imho.

hrydgard:
I'll try it out later and see if I can notice any difference. But yeah, might just wait until after.

hrydgard:
Okay, tested it, and it works well enough for me and indeed really not that much different. Now that we're not just before a release, no reason not to get in.

hrydgard (approved):

PR #14868, opened 2021-09-16
Little fix to logo screen

iota97:
Basically `MemStickScreen` get created with `LogoScreen` below in the `ScreenManager` stack:
https://github.com/hrydgard/ppsspp/blob/e6a2da69c5dde977fecb468cc7906addfca1fb3e/UI/NativeApp.cpp#L803-L804

But `KEY_UP` get send to all layer:
https://github.com/hrydgard/ppsspp/blob/e6a2da69c5dde977fecb468cc7906addfca1fb3e/Common/UI/Screen.cpp#L97-L101

So `LogoScreen` was triggering the "skip code" and went to `MainScreen` on key up (hard to notice with a touch screen tho').

### DIFF
--- a/UI/MiscScreens.cpp
+++ b/UI/MiscScreens.cpp
@@ -707,7 +707,7 @@ void LogoScreen::sendMessage(const char *message, const char *value) {
 }
 
 bool LogoScreen::key(const KeyInput &key) {
-	if (key.deviceId != DEVICE_ID_MOUSE) {
+	if (key.deviceId != DEVICE_ID_MOUSE && (key.flags & KEY_DOWN)) {
 		Next();
 		return true;
 	}


UI/MiscScreens.cpp
```diff
@@ -707,7 +707,7 @@ void LogoScreen::sendMessage(const char *message, const char *value) {
 }
 
 bool LogoScreen::key(const KeyInput &key) {
-	if (key.deviceId != DEVICE_ID_MOUSE) {
+	if (key.deviceId != DEVICE_ID_MOUSE && key.flags & KEY_DOWN) {
```
hrydgard, line 710: This should be:

`if (key.deviceId != DEVICE_ID_MOUSE && (key.flags & KEY_DOWN))`

Yes you do need those extra parenthesis in C++...

iota97: Bitwise have precedence over logical: https://en.cppreference.com/w/cpp/language/operator_precedence. It should be fine like this no?

hrydgard: I don't know how you read the table, but logical have precedence over bitwise, not the other way around.

iota97: In the table the operator with the smaller number get evaluated before, if two operator have the same level of precedence they get evaluated based on the associativity column (i.e. left-to-right or right-to-left).

I double checked with the compiler as well:

```
printf("Default: %d, &&: %d, &: %d\n", true && 0x10 & 0x01, (true && 0x10) & 0x01, true && (0x10 & 0x01));
// Result: Default: 0, &&: 1, &: 0
```

Anyway this is for sure a source of confusion so better use parenthesis (you might be confusing with & and == that have this need mb?).

hrydgard: Oh duh, yeah, I saw the != and thought that would be involved in it, but you're right - in this case it'll work. 

However, I really do want parenthesis for this, because this kind of thing has been the source of many bugs.

PR #14869, opened 2021-09-16
Clean up Memstick screen a bit

iota97:
Show only the info for the current selected one, make text wrap when is too long and move the continue/back below to give more space on the side on small device (on long screen phone may be better on the side tho', not really sure).

Few previews (ignore the space on the left, just notch doing thing):
Screenshot_20210917-011707_PPSSPP
Screenshot_20210917-011717_PPSSPP
Screenshot_20210917-011723_PPSSPP
Screenshot_20210917-011753_PPSSPP

About implementation: I wanted to try to show the info right below the selected one but was really weird in the end, with this design we could use ChoiceStrip more maybe, but I gotta admit I'm bit worried about all the "choice number" code path there, probably nothing too scary but may be more bug prone (given where the choice count change a lot), not sure what's better...

### DIFF
--- a/UI/MemStickScreen.cpp
+++ b/UI/MemStickScreen.cpp
@@ -110,40 +110,40 @@ void MemStickScreen::CreateViews() {
 	auto di = GetI18NCategory("Dialog");
 	auto iz = GetI18NCategory("MemStick");
 
-	Margins actionMenuMargins(15, 15, 15, 0);
+	root_ = new LinearLayout(ORIENT_VERTICAL);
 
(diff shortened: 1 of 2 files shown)


hrydgard:
Thanks, I'm gonna take over this in the weekend and change it up a bit more. But the main concept of not totally overwhelming the user with text is probably a good one.

hrydgard:
Replacing this with https://github.com/hrydgard/ppsspp/pull/14875 , sorry :)

PR #14870, opened 2021-09-17
[AdhocMatching] An attempt to handle unknown/incorrect source port better on AdhocMatching

anr2me:
An attempt to handle unknown/incorrect source port better on AdhocMatching by keeping track of the new port (taken from source port) for further communication.

However, this can only work if the player was able to receive something on the correct port in order to get the source port, which mean the player who supposed to receive the data for the first time must be able to receive data on the normal/correct AdhocMatching port.

I still don't know why the source port can sometimes be different than the bound port, and the only clue i can find was the port might be remapped by router or ISP as mentioned here https://stackoverflow.com/questions/31772531/udp-source-port-different-on-client-getsocketname-and-server-recv

But, this issue seems to be happening on LAN multiplayer too which is strange.

PS: I'm not sure whether this will work better or not in the case of "incorrect port" being shown on screen, since i'm having difficulty trying to reproduce the issue on my end.

PPS: Players who had the "unknown/incorrect port" warning shown on their screen are not the one who had their port being remapped, since they're able to receive data on the "correct" port, the sender is the one who had their port being remapped.

Anyone who had this issue quite often can test these test builds:
Win32&64: https://www.dropbox.com/s/2t3mtdhb0f045cn/PPSSPP_1.11-testbuild_Win32x64.zip?dl=0
Android(ARMv7): https://www.dropbox.com/s/b41bm43mtn1gpnn/PPSSPP_1.11-testbuild_ARMv7.apk?dl=0 
Or download it from this PR's artifacts.

### DIFF
--- a/Core/HLE/proAdhoc.h
+++ b/Core/HLE/proAdhoc.h
@@ -221,6 +221,19 @@ typedef struct SceNetEtherAddr {
   uint8_t data[ETHER_ADDR_LEN];
 } PACK SceNetEtherAddr;
 
+inline bool operator<(const SceNetEtherAddr& lhs, const SceNetEtherAddr& rhs) {
+	uint64_t l = 0;
+	uint64_t r = 0;
+	const uint8_t* lp = lhs.data;
+	const uint8_t* rp = rhs.data;
+	for (int8_t i = 5; i >= 0; i--) {
+		int8_t sb = (CHAR_BIT * i);
+		l |= (uint64_t)*lp++ << sb;
+		r |= (uint64_t)*rp++ << sb;
+	}
+	return (l < r);
+}
+
 // Broadcast MAC
 extern uint8_t broadcastMAC[ETHER_ADDR_LEN];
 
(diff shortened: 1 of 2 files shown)


MojoJojoDojo:
This gonna affect GameMode games as well because they mostly use AdhocMatching in the lobby anyway.
I already got the message once on The Force Unleashed before so I know.

ANR2ME:
> 
> 
> This gonna affect GameMode games as well because they mostly use AdhocMatching in the lobby anyway.
> I already got the message once on The Force Unleashed before so I know.

GameMode use a different predefined port than AdhocMatching, so games that use both AdhocMatching and GameMode might have double issue, one on the AdhocMatching port and another one on GameMode port.

So i'm thinking to do the same (keeping track source port) on GameMode later (may be tonight or tomorrow), since it can also be used for detecting port remapping issue (since we know which port it supposed to came from).

hrydgard:
So should we merge this in the meantime, or wait?

ANR2ME:
> 
> 
> So should we merge this in the meantime, or wait?

May be wait for a few hours to get some feedback (if any testing this)

I'll create a separate PR for GameMode later.

MojoJojoDojo:
I think i see an improvement even on internet play.
Someone that triggered incorrect port no longer triggers it again.
Good job.

ANR2ME:
If you can't see other's room but when switching side others can see your room, it's probably because you're the one who had the port being remapped, thus unable to receive the room broadcast from other players, which is why you're not the one getting the warning message.

I guess it's safe to merge this.

MojoJojoDojo:
Nah I did get it on Linux its just didnt appear on Windows for some reason.
Its never me, people trigger it for me.

PR #14911, opened 2021-09-22
Update zh_CN.ini

Narugakuruga:
translate based on‘ context


hrydgard:
Yeah, context is tricky...

we have way too many strings, and we would really need some context explanations in the INI files..

Narugakuruga:
add commentary for each line would take a lot of time. Can probably note some important parts.
I thought the "rhombus" is a bottom of some kind of controller until I open PPSSPP🤣

Issue #14913, opened 2021-09-22
[Translate][Appearance] Some text box width is not enuogh space for Japanese translation.


nassau-tk:
### What should happen

Some text box is not enough space for Japanese.
It should be shrink the text automatically or change to more wide.

example) In Edit touch control layout...
![image](https://user-images.githubusercontent.com/48179091/134352248-9c701be9-4c2f-48dd-889a-08e15a579256.png)

example) In Control mapping
![image](https://user-images.githubusercontent.com/48179091/134353169-91269604-6732-4b02-8da9-95171a297f42.png)

example) In Gesture mapping
![image](https://user-images.githubusercontent.com/48179091/134356457-a354e08c-4153-4b66-877d-ddb0b89918f0.png)

Very unshapely.:weary:

### Who would this benefit

All Japanese　user.

### Games this would be useful in

All

### Checklist

- [X] Check the [latest git build](https://buildbot.orphis.net/ppsspp/) in case it's already implemented.
- [X] [Search for other requests](https://github.com/hrydgard/ppsspp/search?q=my+issue&type=issues) of the same feature.

nassau-tk:
Related #14522  .

PR #14914, opened 2021-09-22
Fix Translation to Japanese.

nassau-tk:
Please check with #14913 too.
Thank you!


hrydgard:
I tagged #14913 with 1.12 to not forget it, I will take care of it :) Thanks!

PR #14916, opened 2021-09-22
Update th_TH.ini

Saramagrean:
- more translations.
- fix sorting of MemStick categorie.

Note: string "Memory Stick folder" use in System setting, "Memory Stick Folder" use in Manually specify PSP folder.


hrydgard:
Do you think we should combine the two `Memory Stick folder` strings? I notice you translated them slightly differently?

Saramagrean:
Yeah, can combine, it same context. :)

PR #14918, opened 2021-09-22
Icon for combo

### LINKED ISSUE #14915: Custom button shape is not match on Arrow Up/Down/Right/Left.
### What should happen

Custom button shape is not match on Arrow Up/Down/Right/Left.

In Custom key setting.
If you choose Arrow Up in shape then appear "D-pad Down shape".
If you choose Arrow Down in shape then appear "D-pad Up shape".
If you choose Arrow Right in shape then appear "D-pad Left shape".
If you choose Arrow Left in shape then appear "D-pad Right shape".

It's not meaning "Icon".
It's meaning "Shape".

Designation in left, right, up or down should be reversed.

### Who would this benefit

All

### Platform (if relevant)

_No response_

All

_No response_

### Checklist

iota97:
For the series "A picture is worth a thousand words": let's use image preview rather than string, fixes #14915.

Screenshot_2021-09-22_23-09-35

I updated the en_US language file only, I think hrydgard have the right tool to sync all up (if not I'll updated them all myself with old good regex tho').


hrydgard:
Fantastic :) 

Yeah, I'll sync all the lang files later. This will be the command to do it, after removing them from en_US, once I've updated my tool shortly:

langtool --remove-unknown-lines

nassau-tk:
Awesome!!

PR #14921, opened 2021-09-23
[AdhocMatching] Fix some more race condition possibility

anr2me:
I found some more race condition issue while testing Gundam: Senjou No Kizuna Portable with 4 players

### DIFF
--- a/Core/HLE/sceNetAdhoc.cpp
+++ b/Core/HLE/sceNetAdhoc.cpp
@@ -6137,6 +6137,9 @@ void broadcastHelloMessage(SceNetAdhocMatchingContext * context)
 */
 void sendAcceptPacket(SceNetAdhocMatchingContext * context, SceNetEtherAddr * mac, int optlen, void * opt)
 {
+	// Lock the peer
+	std::lock_guard<std::recursive_mutex> peer_guard(peerlock);
+
 	// Find Peer
 	SceNetAdhocMatchingMemberInternal * peer = findPeer(context, mac);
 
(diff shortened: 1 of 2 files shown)


Core/HLE/sceNetAdhoc.cpp
```diff
@@ -1809,7 +1809,8 @@ static int sceNetAdhocPdpRecv(int id, void *addr, void * port, void *buf, void *
 						return WaitBlockingAdhocSocket(threadSocketId, PDP_RECV, id, buf, len, timeout, saddr, sport, "pdp recv");
 					}
 
-					return hleLogSuccessVerboseX(SCENET, ERROR_NET_ADHOC_WOULD_BLOCK, "would block");
+					VERBOSE_LOG(SCENET, "%08x=sceNetAdhocPdpRecv: would block", ERROR_NET_ADHOC_WOULD_BLOCK); // Temporary fix to avoid a crash on the Logs due to trying to Logs syscall's argument from another thread (ie. AdhocMatchingInput thread)
+					return ERROR_NET_ADHOC_WOULD_BLOCK; // hleLogSuccessVerboseX(SCENET, ERROR_NET_ADHOC_WOULD_BLOCK, "would block");
```
hrydgard, line 1813: I think it's kinda weird to only change one return point of the function, there are more that use hleLog. But I guess if it solves the problem for now...

ANR2ME: Because it's the only return code that floods the the log from AdhocMatching thread

hrydgard: Doesn't mean the others are completely safe though? But I guess I'm ok with this as a temporary fix.

ANR2ME: Yes, but we'll be changing the thread into PSPThread in the future so the logs won't be an issue anymore.

PR #14946, opened 2021-09-28
[Adhoc] Fix threading issue when exiting PPSSPP due to unjoined AdhocServer thread

anr2me:
I bumped this issue when AdhocServer failed to bind with socket error 10013 (probably blocked by firewall or the port reserved by hyper-v) during Initialization and exited the AdhocServer thread immediately, thus the condition to join it failed to meet upon Shutdown, and causing a popup window to appears on Debug build.


hrydgard:
I'm really not sure if this solves the problem completely or just papers over it, but I think I'm okay to merge nonetheless...

PR #14947, opened 2021-09-28
Updated translation

Felipefpl:
Updated translation to brazilian portuguese.

Edit: i noticed an entry in about box pointing to freenode, idk if the channel is active there but perhaps it needs to be changed. ;)


hrydgard:
Thanks!

Took the liberty of correcting the line endings while merging, to make the diff look good. Merged separately, so closing.

Issue #14557, opened 2021-06-24
Legend Of The Dragon freezes on Waiting for other players message - 1.11.3


itsXPCT:
Hi,me again,found bug and wanted to report it
### What happens?
Game freezes on that message,with sound still playing in background
I also looked at logs,and it spams sceNetAdHocDiscoverGetStatus
### What *should* happen?
Game should proceed fine without any issues

### What are you using?
Android
PC

**(i tried switching host \ client,nothing helped)**
#### What hardware / device and operating system?
Android - PowerVR GE8320
Windows - 8.0
#### What graphics card (GPU) or mobile phone model?
PowerVR GE8320
Intel (R) HD Graphics
#### What PPSSPP version (standalone/official), and did it work before?
1.11.3
#### Which game or games?

ULUS10250
Legend Of The Dragon
USA

#### Checklist

 - [x] Test in the [latest git build](https://buildbot.orphis.net/ppsspp/) in case it's already fixed.
 - [x] [Search for other reports](https://github.com/hrydgard/ppsspp/search?q=my+issue&type=issues) of the same issue.
 - [x] Try resetting settings or older versions and include if the issue is related.
 - [ ] Include logs or screenshots of issue.

ANR2ME:
We currently does have `sceNetAdHocDiscover*` HLE being implemented i think (i don't even know what it's supposed to be used for), so that's probably the reason it's waiting forever probably waiting for a state change or a callback.

I will try to make that function to return an error later, so the game won't stuck waiting forever.

ANR2ME:
Hmm.. `sceNetAdhocDiscover*` probably have a dialog which triggered by `sceNetAdhocDiscoverInitStart` and updated by `sceNetAdhocDiscoverUodate`, but this game probably hiding the dialog since it doesn't shows up when i tested it with JPCSP+prx files, while JPCSP HLE only faking the result after delaying it for 2000 ms.
One of the filed in the parameters struct seems to be group name, so this Discover thing probably tried to find someone else in that group.
`sceNetAdhocDiscoverInitStart netAdhocDiscoverParam=0x08A903F8(SceNetAdhocDiscoverParam unknown1=0x00000000, groupName='NkFree', unknown2=0x0000003C, result=0x00000000)`

Btw, anyone know any other games that use `sceNetAdhocDiscover` syscalls? i'm curious how the dialog looks like :D in case one of them didn't hide the dialog

ANR2ME:
[...]

PR #14950, opened 2021-09-28
[AdhocDiscover] An attempt to implement AdhocDiscover

anr2me:
An attempt to implement AdhocDiscover, doesn't fully working yet to be a success, but at least we can cancel the waiting process properly now, instead of stuck forever and unable to go back to previous menu.

~~Please don't merge this yet, since i only~~ tested this on 2 games Digivice Ver. Portable and Legend of the Dragon #14557
~~I need to test this on~~ other games that seems to use AdhocDiscover too, such as: 
- Dissidia 012
- Monster Hunter 2nd
- Got Eater 2

I believe there will be more games appeared in this list that use AdhocDiscover syscalls in the future https://report.ppsspp.org/logs/kinds/

PS: These are based on JPCSP HLE (which didn't work there, and still stuck) along with some information from LLE to improve it (JPCSP LLE didn't work either, but can be canceled properly), also added some information/code based on decompiled pspnet_adhoc_discover.prx

### DIFF
--- a/Core/HLE/sceNetAdhoc.cpp
+++ b/Core/HLE/sceNetAdhoc.cpp
@@ -66,6 +66,13 @@ bool netAdhocInited;
 bool netAdhocctlInited;
 bool networkInited = false;
 
+#define DISCOVER_DURATION_US	2000000 // 2 seconds is probably the normal time it takes for PSP to connect to a group (ie. similar to NetconfigDialog time)
+u64 netAdhocDiscoverStartTime = 0;
+s32 netAdhocDiscoverStatus = NET_ADHOC_DISCOVER_STATUS_NONE;
+bool netAdhocDiscoverIsStopping = false;
+SceNetAdhocDiscoverParam* netAdhocDiscoverParam = nullptr;
+u32 netAdhocDiscoverBufAddr = 0;
+
 bool netAdhocGameModeEntered = false;
 int netAdhocEnterGameModeTimeout = 15000000; // 15 sec as default timeout, to wait for all players to join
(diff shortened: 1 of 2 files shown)


ANR2ME:
Okay i tested some games, apparently this AdhocDiscover thing is usually used for card/avatar exchange/trade/export/import in sleep mode.
- God Eater 2 (Trade Avatar in Sleep Mode): I can cancel the process properly and goes back to main menu
- Monster Hunter 2nd G: I have no clue how to trigger the AdhocDiscover, may be some kind of trading?
- Dissidia 012 (Communications->Send Friend Card): Looks like i can only cancel it by using power button? not sure how to do that with emulators tho :D something related to sleep mode i guess
- 
image
image

I guess this PR should be pretty safe, mostly can be canceled properly, other than on games that required to turn off sleep mode by using power button, since emulators doesn't have things like that (right?)

MojoJojoDojo:
So it basically like StreetPass on 3DS ? I guess it wasnt very used lol.

ANR2ME:
Well AdhocDiscover seems like only used to find someone else using AdhocDiscover on the same game too, after it found someone it will probably use the regular adhoc functions to communicate further.

ANR2ME:
@hrydgard can i have this PR merged for now? 
This should be able to prevent some games from freezing forever, and continuing it later when i had the time to check the decompiled prx that have so many switch case.

hrydgard:
Oh, I mixed this up with another PR I think. Happy to merge if you're confident in it.

Issue #14554, opened 2021-06-22
Rurouni Kenshin 2 multiplayer doesn't work android to android.


MajinDipu:
### What happens?
The game's full name is Rurouni Kenshin Meiji Kenkaku Romantan Kensai. It's a japanese only game. So the games multiplayer works fine PC to PC or Android to PC crossplay. But whenever I try to play Android to Android it doesn’t work. Both of us can see each other and accept challenge. Look

![Screenshot_20210622_235743_org ppsspp ppsspp](https://user-images.githubusercontent.com/75547787/122998327-b5b20a80-d3ce-11eb-8151-4274516fd459.jpg)
But when I try to move forward to the character select screen the host sees loading for like 30 sec.
![Screenshot_20210622_235754_org ppsspp ppsspp](https://user-images.githubusercontent.com/75547787/122998595-104b6680-d3cf-11eb-957c-fad6cb347ae9.jpg)
Then the host screen shows this msg in japanese after like 30 secs.
![Screenshot_20210622_235820_org ppsspp ppsspp](https://user-images.githubusercontent.com/75547787/122998681-29ecae00-d3cf-11eb-9583-eda63774aec5.jpg)
The client's screen is interesting because instead of a 30 seconds loading screen it instantly shows an error message like this:
![Screenshot_20210623_000240_org ppsspp ppsspp](https://user-images.githubusercontent.com/75547787/122998785-4c7ec700-d3cf-11eb-80fd-6ba6c81c75df.jpg)

Since the game is in Japanese here's how to play multiplayer:- Select the 5th option (going down) in the main menu. You will see 4 options there. The top one is for hosting and the 2nd one is for joining. Challenge with O and accept with O.

### What *should* happen?
It just should work! The error should not be there! And the host should not have 30 seconds loading screen. In crossplay between PC-Android, the lobby progresses to character select screen instantly.

#### What hardware / device and operating system?
2 Androids.
Used localhost on single PC and PC-Android crossplay before then it worked. It just Won't work in mobile-mobile situation. 
#### What graphics card (GPU) or mobile phone model?
Huawei Y9 and Huawei Y11 Prime. I also tried with a foreign dude. He was on android but I dunno what was his device.
#### What PPSSPP version (standalone/official), and did it work before?
v1.11.3-832
It always worked in android to PC crossplay and PC to PC but never worked mobile to mobile in previous versions. Didn’t even work with the stable versions.
#### Which game or games?
NPJH50658

#### Checklist
[...]

PR #14969, opened 2021-10-05
[Adhoc] Workaround for a more cross-platform behavior on PtpConnect (but inaccurate)

anr2me:
This is a workaround to solves different behavior between Windows and non-Windows as mentioned at https://github.com/hrydgard/ppsspp/issues/14554#issuecomment-934432824

**On Windows:** a non-blocking `connect` will keeps returning `EWOULDBLOCK` when the other side is not `listen`ing yet
**On Linux/Android(probably Unix/BSD too):** have similar behavior to PSP, which return `EWOULDBLOCK`/`EAGAIN` for a short while and then became `ECONNREFUSED` for the rest of `connect` attempts on the same socket when no body is listening.

Since there is no way to implement `ERROR_NET_ADHOC_CONNECTION_REFUSED` properly using non-blocking socket on Windows, this PR will do the opposite, which is making non-Windows platform to behave the same with Windows by using `ERROR_NET_ADHOC_WOULD_BLOCK` instead (which might be an inaccurate error code)

This should have a side effect of making some games to work when they're not supposed to work due to timings issue, because `ERROR_NET_ADHOC_WOULD_BLOCK` will trigger the game to try again, while `ERROR_NET_ADHOC_CONNECTION_REFUSED` will cause the game to stop trying.

PS: i haven't tested this yet since i only have 1 android device, so we'll need to wait for feedback first.
Android test build: https://www.dropbox.com/s/qtqpn3iv4uc4sae/PPSSPP_1.11_kenshin2test.apk?dl=0

### DIFF
--- a/Core/HLE/sceNetAdhoc.cpp
+++ b/Core/HLE/sceNetAdhoc.cpp
@@ -3615,9 +3615,13 @@ int NetAdhocPtp_Connect(int id, int timeout, int flag, bool allowForcedConnect) 
 								return hleLogDebug(SCENET, ERROR_NET_ADHOC_WOULD_BLOCK, "would block");
 							}
 						}
-						// No connection could be made because the target device actively refused it.
+						// No connection could be made because the target device actively refused it (on Windows/Linux/Android), or no one listening on the remote address (on Linux/Android).
 						else if (errorcode == ECONNREFUSED) {
-							return hleLogError(SCENET, ERROR_NET_ADHOC_CONNECTION_REFUSED, "connection refused");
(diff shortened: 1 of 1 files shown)


hrydgard:
What's the status of this, want it in for 1.12?  (Even if it misses 1.12, it can still catch a ride on the inevitable 1.12.1....)

MajinDipu:
prolly shouldn’t comment here but what does that mean? not merging anytime soon?

MojoJojoDojo:
There are no Android artifacts so someone needs to build it.

ANR2ME:
Okay, i'll make a testbuild for android

MajinDipu:
OK I am excited to test this but I am gonna need some moments here

hrydgard:
Okay, so, merge? @ANR2ME

ANR2ME:
Okay, we can merge this now

PR #14970, opened 2021-10-05
Update script to be more flexible

iota97:
Before the script was looking for issues id without square brackets around to avoid adding duplicate.

Was kinda fine for the first time we updated the readme but it's confusing now as you would need to edit the readme with no square bracket and then run the script that add them and the link.

Let's just add bracket if they are missing and then add the link that are missing.


hrydgard:
Oh cool, was wondering why the script didn't seem to work right...

PR #14971, opened 2021-10-05
Update script to remove old link

iota97:
This help a lot when moving to history. Running the script will now update both readme and history to have all and only the needed link.

I run the script already since now even if we remove or edit something it will get removed from the link list.


hrydgard:
cool!

PR #14972, opened 2021-10-05
Vulkan: Fix texture scaling shader+ Auto Max Quality texture filtering

hrydgard:
This was broken on Mali (and probably other GPUs). Need to be careful with the barriers when blitting stuff that was just compute-shader-generated.

Also, simply gets rid of the Auto level of texture scaling. It's a performance trap (usually a terrible choice).


unknownbrackets (approved):
Added some fixes to config UI on various platforms.

FWIW seemed like Qt's setting for texture scaling was just broken previously.

-[Unknown]

hrydgard:
Thanks for the cleanups!

Gamemulatorer:
Auto upscale level still present on opengl and it cause this glitch if enable.
Screenshot_2021-10-06-19-51-55
Screenshot_2021-10-06-19-50-47

hrydgard:
That seems unlikely, in the latest builds, the setting string "Auto" isn't even in the list?

Gamemulatorer:
Ah yes I forget to update but issue still persistent
using the latest git build.

**Watch here**

https://user-images.githubusercontent.com/37603562/136206063-cf06509b-515d-4de3-bd60-69e88f0267c7.mp4

Or maybe because of this ancient phone mali-450 gpu?

Gamemulatorer:
> Hm, not sure... I'll look into it.

Nevermind I cannot reproduce on my redmi note 9.

hrydgard:
Hm. I haven't made Auto Max Quality default yet, so not so critical to fix. I don't want it to do bad stuff on old hardware, but leaving it as-is for now.

unknownbrackets:
The black boxes might just be that it ran out of RAM and could not allocate a texture.

-[Unknown]

hrydgard:
I think it's more likely that we set up a combination of mip parameters that it didn't like. Seen that kind of thing before with OpenGL ES...

PR #14994, opened 2021-10-08
More crash fixes

hrydgard:
Just some more Android-crash-report driven fixes/tweaks.


unknownbrackets (approved):
Well, we're also not checking the data offset + data length here.  We could do something with strnlen to limit the strings.

-[Unknown]

hrydgard:
True. but at least it'll hopefully catch most forms of corruption as is, you have to be quite unlucky to get a crash in the paramsfo parsing with corrupt data now.

PR #15012, opened 2021-10-13
Fix macOS 10.8~10.15 run compatibility

vit9696:
REF: https://github.com/Homebrew/homebrew-core/issues/84737#issuecomment-941788983

Screenshot 2021-10-13 at 03 46 13


Gcenx:
I believe you can remove the -mmacosx-version-min= usage within the cmake file

vit9696:
Very likely, but I believe it is there for compatibility with cmake 2.x, so I think it is better to leave it.

vit9696:
Given that this is a regression to a formerly guaranteed macOS minimum target I request this to be included in v2.12.3. CC @hrydgard 

Also, can we update https://ppsspp.org/downloads.html to point to https://github.com/hrydgard/ppsspp/releases/latest for macOS?

PR #15048, opened 2021-10-23
Add option to control drag analog behaviour

### LINKED ISSUE #15047: I need the analog stick dragging feature back
Since the 1.12 update the capability of dragging the analog stick into another button has been deleted because it was considered useless but... I used this A LOT so it would be nice if at least they added an on/off option or something.

iota97:
Fixes #15047. Kinda a weird use cause tho'...


hrydgard:
Let's not do this, heh. see comment in the issue.

PR #15086, opened 2021-11-02
Update id_ID.ini

PutuDesca:
Fixed Indonesia translate


Gamemulatorer:
Stop spamming

PutuDesca:
I don't spam, but every time I publish a translation update it always fails when I check the uwp build 

do you know how to solve it?

PR #15090, opened 2021-11-02
New update translation id_ID.ini

PutuDesca:


PutuDesca:
do i need to make a balinese translation for ppsspp?

PR #15092, opened 2021-11-03
Last update id_ID.ini

PutuDesca:


hrydgard:
Do you really want to change all these words to start with uppercase in the middle of sentences? It's a pretty common convention in English but rarely seen in other languages.

hrydgard:
I mean, if it is commonly done in Indonesian languages, then it's fine. But just wondering.

PutuDesca:
I canceled doing this project because the spelling of the language is not right so I think for this translation it looks a little strange

PutuDesca:
So the results are not quite right because the grammar of Indonesian writing is very much so a lot of vocabulary that I made may not be right

PR #15101, opened 2021-11-06
Update CMakeLists.txt & zh_CN.ini

Halo-Michael:
Bundle libMoltenVK.dylib for iOS & correct the wrong translation.


hrydgard:
Looks fine to me. I think we still don't have the code that actually sets up and uses MoltenVK for iOS though..

PR #15102, opened 2021-11-06
Well, Settings.bundle is missing for long time :\

Halo-Michael:
Well, xcode build have "little" performance with gcc build. In fact, xcode build has not been properly maintained for a long time. This should bring the Settings.bundle back to the gcc build.


Halo-Michael:
Actually I noticed bring "ext/vulkan/iOS/Frameworks" back to add_executable may be more friendly to xcode build.

Halo-Michael:
> Actually I noticed bring "ext/vulkan/iOS/Frameworks" back to add_executable may be more friendly to xcode build.

Not really

PR #15104, opened 2021-11-06
Update id_ID.ini

PutuDesca:
Update id_ID.ini
As Well As Translation Fixes
#PutuDesca & eerdfz

Always spirited......


PutuDesca:
For the id_ID translation, it's enough to get here first because I think it's good enough

PR #15108, opened 2021-11-06
[Do not direct merge/WIP] bring something back

Halo-Michael:
PPSSPP's support for iOS is really out of time and broken. Also, cmake is difficult to generate an ideal xcode project. So this pull request I removed generate xcode project to bring back:
1. Settings.bundle, which is missing for long time, if I just revert https://github.com/hrydgard/ppsspp/commit/e68462b55f4308495aa6ad43810430318e57b654 it still doesn't come back.
2. Launch Screen, direct use ibtool to compile `Launch Screen.storyboard` to replace the boring black screen.
https://user-images.githubusercontent.com/45062120/140621790-3229154d-3164-49c7-b647-d5745d5f0671.mov
I don’t know if there is a better way to still generate a suitable xcode project, so please evaluate before deciding whether to merge.


vit9696:
The idea sounds pretty nice to my eyes, will need to test this myself a bit later. By the way, can we also generate an ipa file and automatically upload it to artifacts and releases? This would ease the ability to sideload PPSSPP on non-jailbroken devices.

Halo-Michael:
> The idea sounds pretty nice to my eyes, will need to test this myself a bit later. By the way, can we also generate an ipa file and automatically upload it to artifacts and releases? This would ease the ability to sideload PPSSPP on non-jailbroken devices.

Here is my build scripts: https://github.com/Halo-Michael/ppsspp-builder
And Here is my builds: https://github.com/Halo-Michael/halo-michael.github.io/tree/master/ipas

Halo-Michael:
> > The idea sounds pretty nice to my eyes, will need to test this myself a bit later. By the way, can we also generate an ipa file and automatically upload it to artifacts and releases? This would ease the ability to sideload PPSSPP on non-jailbroken devices.
> 
> Here is my build scripts: https://github.com/Halo-Michael/ppsspp-builder And Here is my builds: https://github.com/Halo-Michael/halo-michael.github.io/tree/master/ipas

Basically the last build works fine for me.

Halo-Michael:
If you can't jailbreak, sideload with [https://altstore.io/](AltStore) it is a very fashionable choice, and its team actively maintains this project.

vit9696:
I use Developer ID to sign applications on my devices, though prefer not to jailbreak the primary ones indeed. As for the scripts, they look pretty neat. Could you integrate them in the CI, so that deb/ipa files appear on the release page, just like the [macOS releases](https://github.com/hrydgard/ppsspp/releases), and on commit artifacts page, the [same way](https://github.com/hrydgard/ppsspp/runs/4131666832)?

Halo-Michael:
> If you can't jailbreak, sideload with [https://altstore.io/](AltStore) it is a very fashionable choice, and its team actively maintains this project.

Also, [AltKit](https://github.com/rileytestut/AltKit) is a sub project, allowed every iOS version enable JIT without jailbreak!
Here is a demo about how UTM teams just bring it into UTM: https://github.com/utmapp/UTM/commit/32205149606fc7589e32a7193dcefcda3881c06d

vit9696:
Cannot AltKit server be placed in another application/extension or even in a separate thread?

Halo-Michael:
> Cannot AltKit server be placed in another application/extension or even in a separate thread?

AltServer is an macOS/Windows application. That means it needs a computer, but Wi-Fi connect is enough.

Halo-Michael:
> I use Developer ID to sign applications on my devices, though prefer not to jailbreak the primary ones indeed. As for the scripts, they look pretty neat. Could you integrate them in the CI, so that deb/ipa files appear on the release page, just like the [macOS releases](https://github.com/hrydgard/ppsspp/releases), and on commit artifacts page, the [same way](https://github.com/hrydgard/ppsspp/runs/4131666832)?

I'm not really know how to write github ci script :\

vit9696:
Hmm, but do not you just call a shell script from there? I can help you with the asset upload part as long as the ipa/deb files are generated.

vit9696:
> > Cannot AltKit server be placed in another application/extension or even in a separate thread?
> 
> AltServer is an macOS/Windows application. That means it needs a computer, but Wi-Fi connect is enough.

You can use [this](https://github.com/osy/Jitterbug#entitlements) instead of AltServer. For free accounts you need 2 Apple devices (e.g. an iPhone and an iPad) and for a proper Developer ID you can run the app with just a single device.

hrydgard:
So is this ready and working as expected now? I don't really know enough about the iOS ecosystem to know what this will fix and break...

Is there a risk it'll break some iOS buildbots like the one by @W-MS ?

W-MS:
My buildbot works fine, it should be okay. I totally agree with @Halo-Michael and @vit9696. We do already using github actions for testing purposes, it could be an effortless stable buildbot. We could try it these days.

hrydgard:
OK, great. Since the PR is tagged "[Do not direct merge/WIP]", I'll wait for Halo-Michael's comment too.

Halo-Michael:
This obviously destroys xcode compilation, but it is good for direct compilation.
We should also change things like ios/README.md to ensure that there are no outdated parts.

hrydgard:
Right.. Xcode compilation is useful for debugging though, any way to keep both? maybe hide xcode compilation under a cmake flag so it's not the default?

Halo-Michael:
But Xcode compilation is already break, If we want keep it we need to fix it first anyway

hrydgard:
It does work last I tried it (though you have to do some manual messing around assigning a "team" and changing the package name in the UI)

PR #15115, opened 2021-11-09
Close

PutuDesca:
Cancel


hrydgard:
Let's do both changes at the same time: add the ini and update langregion. So when you're done, you can just push your .ini file to this branch.

PutuDesca:
Ok

PR #15125, opened 2021-11-11
Cancel

PutuDesca:
Hindi Language PPSSPP


hrydgard:
Hm, as far as I know, "Indian" is not a language, India has 20 official languages (and countless more), and `in` is not a langauage code: https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes .

I believe this is Hindi though, which has the code "hi".

It's important to get the language code right, so that we can default to the correct language on devices set up with these languages.

PutuDesca:
Ok, actually this is hindi

Issue #15137, opened 2021-11-13
PPSSPP profiling notes


hrydgard:
I'm messing around with a trial of Superluminal, which is a really powerful, although commercial, sampling profiler, that just like Very Sleepy works without defining scopes manually (though you can get better output if you do) but has a nice timeline view.

Taking down notes here of anything unexpected I find. Not sure how much the numbers are affected by running in the profiler though - I'm thinking, for Windows-message-dominated stuff like the dialog creation, maybe quite a bit?

### Findings

First, pre-creating the debug dialogs is surprisingly slow, could probably simply create these on demand if we care about saving a bit of startup time on Windows:

![image](https://user-images.githubusercontent.com/130929/141653223-55117150-393b-40ff-923c-0c30dd482ac9.png)

Startup of the emu thread has some stuff, but is really not that slow (PPSSPP UI):

![image](https://user-images.githubusercontent.com/130929/141653320-6842bfae-ac57-4d92-a107-f161e8837446.png)

CleanRecent takes 200ms for me on Windows and probably more on mobile. It can probably be made faster or run in the background quite easily.

Related: #15135

hrydgard:
Profiling during actual gameplay was kinda disappointing because it can't handle JIT code in the stack. Might be possible to circumvent with the integration API though.

hrydgard:
There are some surprisingly large stalls due to vkAllocateMemory, that happen occasionally. An easy repro for me on Windows with NV is the "Suicide Barbie" homebrew demo.

![image](https://user-images.githubusercontent.com/130929/142779428-ae6612c1-9cc6-46b9-8d5b-951651bb2040.png)

In this case, we could easily do vkAllocateMemory on a thread, and only "finish" the "defragmentation" of the push buffer when done. Or allocate a larger buffer up front. Still, it's really weirdly slow here...

hrydgard:
Actually from experience with other Vulkan allocators like [VMA](https://gpuopen.com/vulkan-memory-allocator/), I know that we are not doing things the way drivers expect these days. We should be making fewer much larger allocations, and suballocating more aggressively as needed, instead of every pushbuffer managing its own allocation. 

In fact I'm considering just dropping our custom allocation code and just bringing in VMA, it's very battle tested now and is used by a large majority of commercial games that use Vulkan, if my information is correct.

hrydgard:
#15162 fixes the Suicide Barbie stalls by integrating VMA...

PR #15138, opened 2021-11-13
Windows debugger: Load the dialogs on demand.

hrydgard:
I think this is especially good for the GeDebugger dialog since we now can avoid initializing that extra GL context unless you open the dialog.

This actually saves as much as a second of startup time on Windows. Maybe doesn't seem important, but when developing and iterating on stuff, the less you have to wait the better.

Additionally, optimizes CleanRecent for an extra 150ms savings (in my testing).

See #15137

### DIFF
--- a/Core/Config.cpp
+++ b/Core/Config.cpp
@@ -29,6 +29,7 @@
 #include "Common/Net/URL.h"
 
 #include "Common/Log.h"
+#include "Common/TimeUtil.h"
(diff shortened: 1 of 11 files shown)


hrydgard:
Just gonna go ahead and merge. There's a small risk it might regress some secondary path for opening these dialogs, but if so, should be easy to fix.

PR #15144, opened 2021-11-16
Turn Off Sound When Fast Forwarding

turbo420:
I am not 100% sure this is right way to do this but we will see

### DIFF
--- a/Core/Config.cpp
+++ b/Core/Config.cpp
@@ -919,6 +919,7 @@ static ConfigSetting graphicsSettings[] = {
 
 static ConfigSetting soundSettings[] = {
 	ConfigSetting("Enable", &g_Config.bEnableSound, true, true, true),
+	ConfigSetting("FastForwardSound", &g_Config.bFastForwardSound, true, true, true),
 	ConfigSetting("AudioBackend", &g_Config.iAudioBackend, 0, true, true),
 	ConfigSetting("ExtraAudioBuffering", &g_Config.bExtraAudioBuffering, false, true, false),
 	ConfigSetting("GlobalVolume", &g_Config.iGlobalVolume, VOLUME_FULL, true, true),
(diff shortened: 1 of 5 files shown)


Gamemulatorer:
What's the difference of this in **alternate speed volume** mute?
Screenshot_2021-11-17-07-08-09-588_org ppsspp ppsspp

LunaMoo:
None, I suppose @turbo420 simply didn't knew we have that feature already.

hrydgard:
It is not exactly intuitive, but hopefully googling for a solution will end up here or in similar issues though.

unknownbrackets:
Maybe we could change the name slightly, but I personally like it applying:
 * Even when speed is decreased to 50%.
 * Only as a decrease in volume, not total silence.
 * Whether toggled alternate speed is used or fast forward is used.

And I think it'd be silly/more confusing to have multiple settings (one for fast forward, another for alternate speed.)

@turbo420 thanks for opening a pull request and contributing.  As some feedback for next time, I'd suggest implementing `g_Config.bFastForwardSoundOff` as a function or similar, and calculating based on `g_Config.bFastForwardSound && PSP_CoreParameter().fastForward`.  Not in this case, because I think we have this feature already... but I know I always appreciate feedback on my code.

-[Unknown]

turbo420:
I didn't know that alternate speed volume did that lol. Maybe a slight name change will help. 
I did learn a lot about ppsspp code though.

Gamemulatorer:
Maybe change alternate speed volume to like this?
Screenshot_2021-11-17-10-21-11-471_com github stenzek duckstation

unknownbrackets:
So you think we should remove the feature to decrease the volume when using a speed other than fast forward, such as 50% or 200%?

-[Unknown]

Issue #15173, opened 2021-11-27
Need help removing rolling bar effect from crt shader by KillaMaaki


ryzenvideous:
The rolling bar effect for me is a little visually obstructive and I would like to find a way to remove it. On the issue #12546 I came across, a user helped by guiding on how to remove by removing lines of code from the fsh file and even presented the full code with the concerned lines removed. 

What I did is I created a text file, copied and pasted the code, saved it as crt.fsh and created a shaders folder within the psp folder to move it there but it seems that the rolling bar effect still remains unchanged for some reason. Could someone please help me out with this?

Also, I'm using PPSSPP for Android.

iota97:
On android you can't edit shader file as you don't have access to them but you may just add a new one.

You will need a `.ini` file that describe the shader with a `.fsh` and a `.vsh` in the shaders folder like this: [shaders.zip](https://github.com/hrydgard/ppsspp/files/7611469/shaders.zip)

After this you should see a new shader in the list named `CRT Scanlines - not moving`.

ryzenvideous:
> On android you can't edit shader file as you don't have access to them but you may just add a new one.
> 
> You will need a `.ini` file that describe the shader with a `.fsh` and a `.vsh` in the shaders folder like this: [shaders.zip](https://github.com/hrydgard/ppsspp/files/7611469/shaders.zip)
> 
> After this you should see a new shader in the list named `CRT Scanlines - not moving`.

Dude, you're a legend 😄

I tried other ways like finding the ini which includes the lines for the crt shader from Mr Hrydgard's own list of files, downloaded the concerned fsh and vsh and changed the code and text accordingly but it somehow couldn't work, but the files you provided do!

Once again you're a legend, and playing ULUS10176, or any other PSP titles will be even more enjoyable! The shader looks almost like how the original PSP screen looked! Or at least that's how I recall it.. regardless it just looks great!

Happy Thanksgiving to you, Sir.

Issue #12546, opened 2020-01-03
Could provide a crt-shader without shaking scanline?


Oend:
CRT shader looks pretty good, but when I don't move, I can see scanlines are shaking or move or something (not sure how to say it) on static screen, this breaks the enjoying for beautiful static image.

Or could you tell me how to add crt-shader from RetroArch? so I can add it by myself.

very appreciative for help :) :) :)

Oend:
https://forums.ppsspp.org/showthread.php?tid=18926
ok, so I copy crt-pi.glsl from RetroArch and rename it fsh, it works but looks very different with ppsspp-crt, ppsspp-crt looks much better.....much much better.....

Oend:
Just need to remove that shaking it will be perfect......

Oend:
ok I'm wrong, that's not crt-pi that I add, that's scanlines.fsh from ppsspp-self
For crt-pi from RetroArch, I edit ppsspp-shader-ini, add it

[CRT-PI]
Name=CRT-PI
Author=NO
Fragment=crt-pi.fsh
Vertex=fxaa.vsh

After ppsspp load it, nothing changes, looks this way doesn't work

iota97:
There are 2 CTR shader in PPSSPP, only one (ctr.fsh) have any effect that change over time, if you want to remove those effect:

Open the .fsh file with a text editor such as notepad or notepad++
Now look at the line that mention u_time.x:

Remove u_time.x * 0.5 will remove the like shaking over time, or you may just want to change the 0.5 to other value to change the speed of the effect instead:

`  int vPos = int( ( v_texcoord0.y + u_time.x * 0.5 ) * 272.0 );`

The other line controll the "black shadow" passing behind:
```
float rollbar = sin( ( v_texcoord0.y + u_time.x ) * 4.0 );
gl_FragColor.rgba = c + (rollbar * 0.02);
```

Here as well you may just multiply u_time.x to change it's speed, to remove the effect instead just remove the `+ (rollbar * 0.02)` from the next one.

Here the code without any effect changing overtime:
```
// Retro (CRT) shader, created to use in PPSSPP.

#ifdef GL_ES
precision mediump float;
precision mediump int;
#endif

uniform sampler2D sampler0;
varying vec2 v_texcoord0;
uniform vec4 u_time;

void main()
{
    // scanlines
    int vPos = int( ( v_texcoord0.y ) * 272.0 );
    float line_intensity = mod( float(vPos), 2.0 );
    
[...]

PR #15174, opened 2021-11-27
Make CRT animation speed configurable

iota97:
As it was requested twice in #12546 and #15173 I think we can just make use of setting for this.

### DIFF
--- a/assets/shaders/crt.fsh
+++ b/assets/shaders/crt.fsh
@@ -8,11 +8,12 @@ precision mediump int;
 uniform sampler2D sampler0;
 varying vec2 v_texcoord0;
 uniform vec4 u_time;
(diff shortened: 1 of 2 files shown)


ryzenvideous:
I would like to add that the moving scanlines aren't all that much of a problem for me, it was the repeating float bar which was a little bit annoying. That said I still prefer the static scanlines version because at the speed they go there is a little bit of glitching going on occasionally in the manifestation of white translucent stripes - are they a byproduct of the speed of the scanlines moving not in sync with the fps or intended as another effect in the shader itself?

Anyway, I second this idea.

ryzenvideous (approved):

ryzenvideous (approved):

PR #15185, opened 2021-11-30
CMake: Allow use system miniUPnPc and ( on macOS ) system glew

CarterLi:
Most Linux distros and macOS homebrew provide miniUPnPc package. Using external libraries allows them to be upgraded independently and keep them up to date.

GLEW: It was disabled on macOS due to an issue building with bundled ffmpeg. Since ppsspp works fine with the latest ffmpeg, this limitation should be reduced.


abouvier:
I think the include directives in `Core/Util/PortManager.h` should be updated to allow building with system miniupnpc.
Like `#include <miniupnpc/miniwget.h>` instead of `#include "ext/miniupnp/miniupnpc/miniwget.h"`.

ANR2ME:
> I think the include directives in `Core/Util/PortManager.h` should be updated to allow building with system miniupnpc. Like `#include <miniupnpc/miniwget.h>` instead of `#include "ext/miniupnp/miniupnpc/miniwget.h"`.

Does using system miniupnpc mean we're not using static library? because currently PortManager.h have `#define MINIUPNP_STATICLIB` being defined

PR #15186, opened 2021-11-30
ext: Update vulkan headers & binaries

CarterLi:
Bundled vulkan headers & binaries were pretty outdated. This PR upgrade them to the latest version 1.2.198


hrydgard:
Don't think there's been much interesting new functionality in recent extensions, but yeah, good to keep up.

PR #15187, opened 2021-11-30
cmake: Include a default version for git-version.cmake

orbea:
@unknownbrackets Stated:

> Just a note that some people patch this file (git-version.cpp) to prevent it from using git to lookup the version number.

https://github.com/hrydgard/ppsspp/pull/15184#issuecomment-982718132

So it might alleviate a lot of the need to patch `git-version.cpp` by having a default version in case the `git` executable or `.git` directory is missing. It would also have a more helpful version than `unknown` for any users with such a build. The downside is that it would be one more place to update the version string every release. I don't have a strong opinion on this, but I think it would be probably a good idea.

### DIFF
--- a/git-version.cmake
+++ b/git-version.cmake
@@ -1,5 +1,5 @@
 set(GIT_VERSION_FILE "${OUTPUT_DIR}/git-version.cpp")
-set(GIT_VERSION "unknown")
+set(GIT_VERSION "1.12.3")
 set(GIT_VERSION_UPDATE "1")
 
 find_package(Git)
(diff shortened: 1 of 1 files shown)


hrydgard:
Hmmm.. Would really prefer not to have one more thing to manually update :(

orbea:
That is a fair concern and I'm fine if you decline. I just wanted to suggest the idea.

Currently including this PR here are the places that have the version string.
```
$ grep --exclude-dir=.git -r '1\.12\.3'
README.md:What's new in 1.12.3
android/AndroidManifest.xml:    android:versionName="1.12.3.0">
git-version.cmake:set(GIT_VERSION "1.12.3")
```
It could be possible to wrap them all up in a single script that updates them with sed, for example here is a naive attempt where only `README.md` would need to be manually updated.
```
#!/bin/sh

set -eu

old_version="$(git describe --abbrev=0)"
new_version="${1:?}"

for file in $(grep --exclude-dir=.git -rwl "${old_version#v}"); do
  [ "$file" != README.md ] || continue
  sed -i "s/${old_version#v}/$new_version/" "$file"
done
```

hrydgard:
Yeah I suppose it's not too bad. I'll consider it.

orbea:
I'll close this due to lack of activity. Feel free to still merge if you want it.

PR #15210, opened 2021-12-07
(libretro) Force enable g_Config.bRenderDuplicateFrames + minor audio buffer tweaks

jdgleaver:
This is a small follow-up to #15202. As recommended by hrydgard, this PR force enables `g_Config.bRenderDuplicateFrames` for the libretro core, to improve the likelihood that one actual 'frame' will be emulated for each call of `retro_run()`. This greatly improves frame pacing for 30 fps and 20 fps content.

Since enabling `g_Config.bRenderDuplicateFrames` also means that we write fewer audio samples per call of `retro_run()` when running sub-60 fps content, it is no longer necessary for the core to override the default frontend latency setting, so the `RETRO_ENVIRONMENT_SET_MINIMUM_AUDIO_LATENCY` callback has been removed.

Finally, the hardcoded limit on the number of frames that can be sent per use of the audio batch callback has been lifted - this is now determined at run time based on the response of the frontend.

### DIFF
--- a/libretro/libretro.cpp
+++ b/libretro/libretro.cpp
@@ -63,10 +63,6 @@
 // to calculating the average for the last 180
 // frames, or 3 seconds of runtime...
 #define AUDIO_FRAMES_MOVING_AVG_ALPHA (1.0f / 180.0f)
-// Workaround for a RetroArch audio driver
-// limitation: a maximum of 1024 frames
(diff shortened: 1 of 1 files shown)


hrydgard:
Unknown is right that it might be good to also look at elapsed emulator time, but indeed, this is gonna improve things.

PR #15222, opened 2021-12-10
Fix shutdown assert / potential crash

hrydgard:
Split into two commits, one with the fix, the other with the new asserts and stuff I added to find it..


unknownbrackets (approved):
Ah good, I saw this VMA assert and assumed an ordering issue but didn't get to it with the use-after-free fix.

-[Unknown]

PR #15223, opened 2021-12-10
Split off the fonts from the atlas

### LINKED ISSUE #14876: Should consider splitting the UI atlas into font and "other"

hrydgard:
Fixes #14876 

Uses the new mechanism added by unknown in #15021 , but slightly differently than intended there:

Platforms that can use OS fonts don't need font_atlas.zim for normal text, but they do need ascii font for debug, so instead use a new separate asciifont_atlas.zim on those platforms. This is better than including the debug font in ui_atlas.zim since that one is now smaller and lighter to update as needed, while asciifont_atlas.zim can stay static forever, probably.

We now no longer need to ship the full font on Android, shrinking the distribution by about 1MB and slightly speeding up app load time.

The same could be done for ppge_atlas.zim , not sure if worth it.

Also, removes the silly chinese.txt and korean.txt, and instead "learns characters" directly from the language files.


Common/UI/Context.cpp
```diff
@@ -40,9 +40,18 @@ void UIContext::BeginFrame() {
...
+#else
 			fontTexture_ = CreateTextureFromFile(draw_, "font_atlas.zim", ImageFileType::ZIM, false);
-			if (!fontTexture_)
-				WARN_LOG(SYSTEM, "Failed to load font_atlas.zim");
+#endif
+			if (!fontTexture_) {
+				// Load the smaller ascii font only, like on Android. For debug ui etc.
+				fontTexture_ = CreateTextureFromFile(draw_, "asciifont_atlas.zim", ImageFileType::ZIM, false);
+				if (!fontTexture_) {
+					WARN_LOG(SYSTEM, "Failed to load font_atlas.zim");
```
iota97, line 52: Should be: "Failed to load asciifont_atlas.zim"

unknownbrackets (approved):
Went ahead and removed android/assets/ entirely, because it didn't have the duplicate zims and was breaking cmake.

-[Unknown]

PR #15292, opened 2022-01-09
Updated pt-br translation

Felipefpl:
Updated the translation with the minor changes made to some strings in the english file plus some minor fixes to other ones.


assets/lang/pt_BR.ini
```diff
@@ -1,1218 +1,1219 @@
...
+Both = Ambos
+Buffer graphics commands (faster, input lag) = Buffer dos comandos dos gráficos (mais rápido, atraso na entrada dos dados)
+Buffered Rendering = Renderização com buffer
+BufferedRenderingRequired = Aviso: Este jogo requer que o "modo de renderização" esteja definido "com buffer".
+Camera = Câmera
+Camera Device = Dispositivo de câmera
+Cardboard Screen Size = Tamanho da tela	(em % do visor)
+Cardboard Screen X Shift = Deslocamento do X (em % do espaço vazio)
+Cardboard Screen Y Shift = Deslocamento do Y (em % do espaço vazio)
+Cardboard VR Settings = Configurações do Google VR Cardboard
```
iota97: VR Cardboard have little to do with Google I think

assets/lang/pt_BR.ini
```diff
@@ -1,1218 +1,1219 @@
...
+Settings = Configurações
+Share Games (Server) = Compartilhar jogos (servidor)
+Share on PPSSPP startup = Compartilhar na inicialização do PPSSPP
+Stop Sharing = Parar com o compartilhamento
+Stopping.. = Parando...
+
+[Reporting]
+Bad = Ruim
+FeedbackCRCCalculating = CRC do Disco: Calculando...
+FeedbackCRCValue = CRC do Disco: [VALOR]
```
iota97: `[VALUE]` should be left as it  was as it's used to mark where the actual value should be written in the string

iota97:
Also I'm not sure why diff are so messed up, maybe related to line ending or something :/

hrydgard:
Hm, yeah - can you re-save with the opposite kind of line endings to get a clean diff? :)

assets/lang/pt_BR.ini
Felipefpl: That's strange, the word google is in the language file of latest unofficial build in the site.

assets/lang/pt_BR.ini
Felipefpl: Thank you for the tips, it would be good if there was a warning in the language file about that, there is no way that simple translators like me would know about this. ;)

Felipefpl:
> Hm, yeah - can you re-save with the opposite kind of line endings to get a clean diff? :)

Line endings? I dont know anything about this, all i did was translate the file, i didnt change anything in its structure, i can only deduce something is wrong with the unofficial build since it came like this.

iota97:
Probably the text editor you used saved the file using Windows line ending encoding (CRLF) while the original file had Unix line ending (LF). You may use a tool like dos2unix to convert back to Unix line ending.

If you click on `Files changed` tab in this PR you'll see how it show every line as changed, if the line ending is correct it should just show changed line.

Felipefpl:
> Probably the text editor you used saved the file using Windows line ending encoding (CRLF) while the original file had Unix line ending (LF). You may use a tool like dos2unix to convert back to Unix line ending.
> 
> If you click on `Files changed` tab in this PR you'll see how it show every line as changed, if the line ending is correct it should just show changed line.

I see, i searched dos2unix but i'm not a fan of command line stuff, but thanks to your guidance about those line endings i found a way to fix it on windows here:

http://www.sql313.com/index.php/43-main-blogs/maincat-dba/62-using-notepad-to-change-end-of-line-characters

Thank you for your help.

hrydgard:
Thank you, that's much better.

Felipefpl:
> Thank you, that's much better.

You are welcome, and thank YOU for your/everyone's work on ppsspp.

PR #15299, opened 2022-01-13
[AdhocGameMode] Fix Bomberman Panic Bomber multiplayer

anr2me:
- Fix invalid arg issue on sceNetAdhocPtpAccept
- Fix argument issue on sceNetAdhocGameModeUpdateReplica
- Fix many issue due to more than one GameMode Schedulers can be scheduled
- Fix buffer overflow issue on GameMode buffer

This should fix Bomberman Panic Bomber https://github.com/hrydgard/ppsspp/issues/14641#issuecomment-1010016225

### DIFF
--- a/Core/HLE/sceNetAdhoc.cpp
+++ b/Core/HLE/sceNetAdhoc.cpp
@@ -151,11 +151,12 @@ void __NetAdhocShutdown() {
 }
 
 bool IsGameModeActive() {
-	return netAdhocGameModeEntered;
+	return netAdhocGameModeEntered && masterGameModeArea.data;
 }
 
 static void __GameModeNotify(u64 userdata, int cyclesLate) {
 	SceUID threadID = userdata >> 32;
+	u32 error;
 
 	if (IsGameModeActive()) {
 		// Need to make sure all replicas have been created before we start syncing data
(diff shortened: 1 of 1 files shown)


ANR2ME:
Btw, while testing this game and idling in lobby, i bumped into `CoreTiming` issue.
Not sure whether this is related to AdhocMatching threads or not (ie. similar to MGS:PW), but at the time it breaks, both AdhocMatching threads were sleeping.

Using `Dynarec(JIT)`
image

Using `IRInterpreter`
image

I can't reproduce it using `Interpreter`, probably because it runs very slow on Debug build.

MojoJojoDojo:
Does it help with The Force Unleashed as well?

ANR2ME:
> Does it help with The Force Unleashed as well?

nope, it still shows missing player names, it's probably an issue occurred before GameMode initiated

hrydgard:
Coretiming issue is slightly scary, and likely unrelated..

Change looks good, merging.

unknownbrackets:
CoreTiming is not re-entrant, if CoreTiming::Advance(), CoreTiming::ForceCheck(), or anything in run loop until for the CPU is called from two threads simultaneously, or if jit runs on two threads somehow, bad things will happen.

`WaitBlockingAdhocSocket()` seems like it can trigger this, or at least it schedules an event, and does so very unsafely on a separate thread than the CPU...

-[Unknown]

ANR2ME:
The one on another thread never use blocking behavior as i remembered, thus never use scheduled events.

PR #4074, opened 2013-10-08
add scePauth support

tpunix:
test with Monster Hunter Portable 3rd.
How it work:
When first times run, it will write encrypted data to ms0:/PAUTH.  This data can be decrypt on real PSP. The second times run, it will load the decrypted data and continue.

The tools: http://pan.baidu.com/s/1gigSO

PR #15373, opened 2022-02-04
Delete pauth_tool, useless since pauth keys known

ghost:
pauth_tool was added because of #4074
pauth_tool can be removed because of #12559


hrydgard:
Yeah, in case it's needed for for any reason, it can be dug out of git history.

PR #15394, opened 2022-02-11
Allow custom UI themes

iota97:
This allow to add other color themes to `themes` folder, might add a few to assets later if I manage to make something decent :)

Silly example:
Screenshot_2022-02-11_12-31-28

I made one of the worst looking HTML page ever to help creating theme, uploaded the source [here](https://github.com/iota97/Misc/blob/master/Studies/Graphics/PSP_Themes.html). Will improve it as well.


hrydgard:
I've actually been thinking about an easier way to do this, which would be simply a hue and a saturation slider, then we could rotate the colors in the vertex shader when drawing the UI very cheaply. Would also be a lot less code (though slightly less flexible since you then couldn't set the focused and unfocused colors entirely separately, etc).

iota97:
That would be quite a fancy thing, I didn't thought of it. Will take a look into it :)

hrydgard:
Actually I can take care of that, kind of wanted that feature for a while. And the sum total of the changes will probably actually be larger, come to think of it. Let's see :)

hrydgard:
Ah nice. I can take it from there.

hrydgard:
Hm, some good points there, yeah. Really like the slider though :P  And some extra vertex shader work in the UI should be no big deal and hopefully not much cause for breaks...

hrydgard:
I made an alternate branch that gets it working on all backends, and has live slider update:

https://github.com/hrydgard/ppsspp/commit/77b4874b6e71fa1bdee09d046bf63607782b8b65

But yeah, it won't support switching bright and dark, and that kind of thing. In a way, I'm not sure that's a bad thing... Kinda keeps the "PPSSPP look" intact while still allowing some customization. 

Not sure what is best.

UI/UI.vcxproj
```diff
@@ -101,6 +102,7 @@
     <ClInclude Include="TouchControlLayoutScreen.h" />
     <ClInclude Include="TouchControlVisibilityScreen.h" />
     <ClInclude Include="InstallZipScreen.h" />
+    <ClCompile Include="Theme.h" />
```
unknownbrackets, line 105: *ClInclude

-[Unknown]

UI/Theme.cpp
```diff
@@ -0,0 +1,238 @@
...
+				section.Get("PopupStyleFg", &info.uPopupStyleFg, 0xFFFFFFFF);
+				section.Get("PopupStyleBg", &info.uPopupStyleBg, 0xFF303030);
+
+				appendTheme(info);
+			}
+		}
+	}
+}
+
+static UI::Style MakeStyle (uint32_t fg, uint32_t bg) {
```
unknownbrackets, line 159: nit: We usually don't put a space before the `(` in arguments.

-[Unknown]

UI/Theme.cpp
```diff
@@ -0,0 +1,238 @@
...
+#include "Common/File/VFS/VFS.h"
+#include "Common/Data/Format/IniFile.h"
+
+#include "Core/Config.h"
+
+#include "Common/UI/View.h"
+#include "Common/UI/Context.h"
+#include "Core/System.h"
+
+struct themeInfo {
```
unknownbrackets, line 34: nit: Prefer ThemeInfo to stay consistent with other struct/class names.

-[Unknown]

UI/Theme.cpp
```diff
@@ -0,0 +1,238 @@
...
+	std::string name;
```
unknownbrackets, line 35: I feel like it'd be really nice (maybe not in this pull) to also allow overriding the main zim (if not the font) per theme.  Since we read all the metadata from there, this would allow customizing icons, etc.

It might also be nice to have a default ini in the repo or somewhere that has the default theme (but also like you have, still work without it.)

-[Unknown]

iota97: ZIM override would allow for custom touch controls too that got requested quite a few times.

Mhhhh, I got the feeling this the way to go rather than hue shift (kinda fancier with the slider but that's it).

hrydgard: Now I kinda feel .. why not both? You pick a theme, and then you can shift it if you don't like its colors.. Maybe it's overkill.

iota97: Such use cases could be covered by a web app to create or edit themes. We could also change the lightness value without changing the image redering as it happens on the vertex stage.
We could get the web interface make the ZIM file in the future too or thing like this to make it easier for the user.

But we could have both in the end, never disliked overkills in software to begin with :P

hrydgard: These shaders are at least quite a bit less complicated than some of the stuff we generate sometimes, so I think they'll be fine. I will test them on Mali-400 though :)

iota97:
About this do we really need all those settings for the theme?

- Item and button aren't really consistent around the code base, we use button in a couple of place only with no much semantic difference from choice AFAICT.
- No idea where highlighted is used, can't seem to find it anywhere.
- InfoStyleBg doesn't seem to be used anywhere.

hrydgard:
Yeah, let's try to minimize the number of them. Do delete any unused ones and feel free to merge as appropriate.

UI/Theme.cpp
OOKRONOO: PPSSPP

hrydgard:
OK, I'm happy with this. Ready to go in?

hrydgard (approved):

iota97:
Yeah, I wanted to make at least another theme at first but I'm terrible at that :)

hrydgard:
Hm, I'm thinking about the theme names .. doesn't look like we've hooked them up for translation?

iota97:
> Hm, I'm thinking about the theme names .. doesn't look like we've hooked them up for translation?

This should do it: f4b5c16e1b084c895773f22accf8b3cd01b99755, I just haven't synced the lang file.

hrydgard:
Oh that's right!

Gamemulatorer:
What Happen?
Can't see the new theme colors?
Screenshot_2022-02-17-19-04-05-239_org ppsspp ppsspp

iota97:
There isn't any new theme yet, but you may add them in /PSP/themes/whatever.ini

Gamemulatorer:
> There isn't any new theme yet, but you may add them in /PSP/themes/whatever.ini

Like this?

PR #15396, opened 2022-02-11
Add UI Tint/Saturation settings

hrydgard:
Does the color tinting in the vertex shader.

Makes the background grayscale, and sets a color, so it'll follow the tint properly.

Lighter alternative to full custom themes as in #15394, although I'm kinda torn which way to go.

iota97 started work on the same too at my suggestion. This uses a simpler hsv conversion, not sure which one looks best though. Adds support for all the backends too.


iota97:
Just a few thing to improve the UI:
- Remove darkening background (kinda off with live update)
- Allow to increase saturation as well

Pushed it here: https://github.com/iota97/ppsspp/tree/vertex-hue-ui

Time to make a RGB animation now ;)

hrydgard:
Pulled your change, nice, makes sense.

hrydgard:
Still a bug left with uniform bufer sizes, fixing.

iota97:
Pushed at the [same branch as before](https://github.com/iota97/ppsspp/tree/vertex-hue-ui) a fix for the live update that wasn't quite complete :P

Gamemulatorer:
how about this now?

hrydgard:
Confirmed that this does indeed work just fine on Mali-400.

PR #15406, opened 2022-02-17
Control profiles

### LINKED ISSUE #13303: Feture request  changeable controler profiles
The ability to have up to 5 profiles per game and be able to swap between them using a button
So you can have 2 types of mapible button
While x is held use profile y
Or
When x is pressed swap to profile y

Kind of like steam

### LINKED ISSUE #14404: Suggestion about the controls.
Hello, this is a amazing project, congrats.
I wanted suggest this feature:

Save the controls settings like they had a "profile".
For example: If I wanted this:

Screenshot_20210424-105215_PPSSPP

I could save a profile in "default" section of the controls settings.

Honestly, the control factor could be one reason why some people don't want play the emulator, but I found this extremely confortable way to play the games, with these positions and sizes.

Thank you :D

iota97:
Implement control profiles.

Use cases:
- One may want to share the same controls preset in a few game and have other different setting per game, this make it easier.
- Run time change as different game area/situation may map better to different controls layout (mostly on touch controls).
- More binding on the controller: you may bind a key to change the profile and then have all key bind to action as fast forward, screen shot and so on in that profile. (This also change touch control so you can add a touch control to tell in which state you are).

Look like when I implemented right analog setting I didn't notice the section separation, this will reset those settings (and gesture one) :/

If the key mapping have no entry save will create no section, so load will trigger this:
https://github.com/hrydgard/ppsspp/blob/9bc6b965566b3d805275876def74397f18bfc103/Core/KeyMap.cpp#L676

This will make default key map to be loaded instead of empty binding, not sure if there was any reason for that check to be there so I'm asking first if I could just remove it. Or I should make save create the section always?

Fixes #14404 and fixes #13303.

### DIFF
--- a/Core/Config.cpp
+++ b/Core/Config.cpp
@@ -547,21 +547,6 @@ static const ConfigSetting generalSettings[] = {
 	ConfigSetting("GridView1", &g_Config.bGridView1, true),
 	ConfigSetting("GridView2", &g_Config.bGridView2, true),
 	ConfigSetting("GridView3", &g_Config.bGridView3, false),
-	ConfigSetting("RightAnalogUp", &g_Config.iRightAnalogUp, 0, true, true),
-	ConfigSetting("RightAnalogDown", &g_Config.iRightAnalogDown, 0, true, true),
-	ConfigSetting("RightAnalogLeft", &g_Config.iRightAnalogLeft, 0, true, true),
-	ConfigSetting("RightAnalogRight", &g_Config.iRightAnalogRight, 0, true, true),
-	ConfigSetting("RightAnalogPress", &g_Config.iRightAnalogPress, 0, true, true),
-	ConfigSetting("RightAnalogCustom", &g_Config.bRightAnalogCustom, false, true, true),
-	ConfigSetting("RightAnalogDisableDiagonal", &g_Config.bRightAnalogDisableDiagonal, false, true, true),
-	ConfigSetting("SwipeUp", &g_Config.iSwipeUp, 0, true, true),
-	ConfigSetting("SwipeDown", &g_Config.iSwipeDown, 0, true, true),
-	ConfigSetting("SwipeLeft", &g_Config.iSwipeLeft, 0, true, true),
-	ConfigSetting("SwipeRight", &g_Config.iSwipeRight, 0, true, true),
-	ConfigSetting("SwipeSensitivity", &g_Config.fSwipeSensitivity, 1.0f, true, true),
-	ConfigSetting("SwipeSmoothing", &g_Config.fSwipeSmoothing, 0.3f, true, true),
-	ConfigSetting("DoubleTapGesture", &g_Config.iDoubleTapGesture, 0, true, true),
-	ConfigSetting("GestureControlEnabled", &g_Config.bGestureControlEnabled, false, true, true),
 
 	// "default" means let emulator decide, "" means disable.
 	ConfigSetting("ReportingHost", &g_Config.sReportHost, "default"),
@@ -606,6 +591,7 @@ static const ConfigSetting generalSettings[] = {
 	ConfigSetting("EnablePlugins", &g_Config.bLoadPlugins, true, true, true),
 
 	ReportedConfigSetting("IgnoreCompatSettings", &g_Config.sIgnoreCompatSettings, "", true, true),
+	ConfigSetting("SettingsVersion", &g_Config.uSettingsVersion, 0u, true, true), // Per game for game configs
 };
 
 static bool DefaultSasThread() {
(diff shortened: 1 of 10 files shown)


unknownbrackets:
So this essentially implements hot-swappable controller profiles.  Initially I was imagining you'd assign custom game configs to specific profiles, but this seems more versatile.

The intent of that line was to allow for defaults on first startup, i.e. to avoid these lines:
[ControlMapping"](https://github.com/hrydgard/ppsspp/blob/9bc6b965566b3d805275876def74397f18bfc103/Core/KeyMap.cpp#L685-L686)

Otherwise, if I have no inis yet and start up, I would get no defaults.  That said, I don't know of anything that would break if we allowed saving empty sections (though I'm not terribly sure how useful an empty controller profile is...?)

Personally, I feel like it'd be nicer to put all controller profiles in the same ini, just with different section names.  Instead of `[ControlMapping]` it could be `[ControlProfile1]` etc.  I feel like this could just go in controls.ini?

We could also "port" the settings in the separate section on first startup with those settings missing from the new place.  Otherwise there will definitely be people confused or thinking we removed features.

-[Unknown]

Core/Config.cpp
```diff
@@ -1581,6 +1595,24 @@ bool Config::Save(const char *saveReason) {
 		Section *control = iniFile.GetOrCreateSection("Control");
 		control->Delete("DPadRadius");
 
+		// Migrate from wrong section...
```
unknownbrackets, line 1598: I think this is okay, but I'll note that it's ideal to check some ini value at a default (like check if DoubleTapGesture is -1 and set to 0 the actual default, or something) around the defaulting, and not delete.

The reason is that sometimes we ask users to test against old versions to validate a bug, and it's convenient when their old settings just work and are kept.  The current way would result in this both resetting to defaults in the old versions, and then subsequently resetting to defaults in the new versions too (because they'd exist again.)

I know DPadRadius is doing this, though...

-[Unknown]

iota97: I don't really get what you mean with the default check.

If we want to allow old versions there are 2 options I can think:
- Have both settings in the ini and update both, that's kinda ugly
- Leave the settings where they are and just change the profile save/load logic to get them as well

If we want to allow only old version settings if the user used an old version already:
- Update both if the keys are already present

Not sure what's best or if there are other ways.

unknownbrackets: Sorry for the delay, I've had a complex week and missed this notification.

I meant something like:

```c++
ConfigSetting("DoubleTapGesture", &g_Config.iDoubleTapGesture, -1, true, true),

...

if (g_Config.iDoubleTapGesture == -1) {
	// The -1 value is invalid, we use it entirely to know that the control settings are defaults.
	g_Config.iDoubleTapGesture = 0;

	loadOldControlSettings();
}
```

That way, if `DoubleTapGesture = 0` (or any other value) already exists, it won't port the old settings over.  If it doesn't exist yet, it will copy them over (and very briefly use -1 during the loading process, but other code should only see 0 and other valid values.)

Checking this way would deterministically tell us if it's time to port the settings, but it would also allow us to keep the old settings without deleting them.  No need to update the old ones.

Alternatively, we could do effectively the same with a "SettingsVersion" key under `[General]`, which might be easier to understand for people reading the code and future changes.

-[Unknown]

iota97: Oh, now I get it.

Went for the "SettingsVersion" way because we have bool to port as well and can't set them to invalid value :)

unknownbrackets (approved):

hrydgard:
will take a real look soon! sorry for the delay in reviewing.

Core/Config.cpp
```diff
@@ -601,6 +586,7 @@ static ConfigSetting generalSettings[] = {
 	ConfigSetting("EnablePlugins", &g_Config.bLoadPlugins, true, true, true),
 
 	ReportedConfigSetting("IgnoreCompatSettings", &g_Config.sIgnoreCompatSettings, "", true, true),
+	ConfigSetting("SettingsVersion", &g_Config.uSettingsVersion, 0u, true, true), // Per game for game configs
```
hrydgard, line 589: We traditionally haven't had a version number for settings. I hope the addition of this will not lead to a proliferation of changes, because testing all kinds of settings upgrades will be very laborious...

hrydgard (commented):
Sorry for the long delay in this one, I really don't know what I want to do with control profiles, and since this has some potential backwards compatibility issues, I'm gonna bump this to after 1.13. 

Having multiple control profiles has a lot of value of course so we're gonna get this in later in some form.

Core/Config.cpp
iota97: We could have a bool specific for this port like `bRightAnalogSettingPorted` that would reduce the temptations to add new settings version.

Anyway I see your concern with this PR, if/when you decide what's best for the project I'll change and rebase accordingly :)

iota97:
I had some spare time at last, changed the logic to leave all setting where they are avoiding possible compatibility issues. There is simply a list of setting that get saved in the profile that are from general.

Core/Config.cpp
hrydgard: I'm taking back my comment here, SettingsVersion is the better option, otherwise if we make more corrections in the future, we'll just end up with more specific bools and the combinations get harder to understand.

hrydgard:
This now has some conflicts due to the way too long review process, sorry :(

Could you rebase those out? Then I'll play around with this myself to get a feel for it. I do want this feature in in some form.

iota97:
Squashed the commit for my own sanity before rebasing, I have a local backup of the other versions if needed.

PS: any reason we are manually writing down VIRTKEYS enum values?

hrydgard:
What do you mean by manually writing down virtkey values? Giving them the string names?

iota97:
I mean we write `VIRTKEY_CONTROLLER_PROFILE_1 = 0x40000027`. We could omit the value and it will auto increase anyway, this seems just more affected by typo and git conflict to me. Not sure if there is any pro in this ;)

unknownbrackets:
Since those values are saved externally, it's kind of a good thing that it causes conflicts when two people use the same number.  I don't think it's bad to be explicit in "exposed" values.  I think it makes sense to let auto-numbering happen when you don't care if i.e. two compiler versions auto-number the same way.

-[Unknown]

Gamemulatorer:
This fix #16960?

hrydgard:
Tried this again. As before, i'm concerned with the UX.

While it's not too hard to understand what's going on:

image

it's gonna be hard to remember which is which. It's also not clear what happens when you have a game config - are all setups saved in the game specific config, or global? Also, it's unclear from the UI if this applies to bindings, touch screen UI, or both.

I'm just afraid it's too confusing.. But like I said, I do like the functionality...

iota97:
We could add a string on top explaining what get saved. More strings to traslate tho'.

About global vs game specific: this should be global to allow to "port" configs from one game to another.

We could possibly add a simple text input to add a little description to help remember.

I find this more useful than not, even if a bit confusing should be easy enough to understand if a few tries :)

hrydgard:
The confusion over what belongs to game and what is global does really go beyond this particular thing.. I've been thinking for a long time for how to highlight what's what in the UI, and how to make it understandable... I have a few concepts in mind, but to make it managable, need to make as much as possible of the settings menus dynamically generated from tables (like if we add a little color flash to settings that are being modified for the specific game, don't want to manually add that to all of them...).

NABN00B:
I've posted a couple of ideas on Discord that could be considered for the GUI of this feature:
- replace the list layout with a more visual grid layout
- provide a way to customize the name of control profiles
- give each profile some sort of visual representation, eg a texture or icon

MS Paint mockups:
- Main layout:
control_profiles3
- The popup menu of the Manage button:
control_profiles3_manage

PR #15408, opened 2022-02-18
Allow to set default background color in a theme

iota97:
I think it make sense :)


hrydgard:
Yeah, makes sense indeed.

PR #15409, opened 2022-02-19
Upgrade to Vulkan 1.3 headers (and remove the .hpp files)

hrydgard:
Good to keep up. Will wanna try the VK_ARM_rasterization_order_attachment_access extension once it becomes available on devices (basically enables framebuffer fetch in Vulkan, with a bit trickier setup than in GL which will also take some work).


unknownbrackets (approved):
Ah, my old pet peeve of pointlessly updating copyright years in all files.

vulkan_funcs.hpp is still there, was that meant to be removed too?

-[Unknown]

PR #15418, opened 2022-02-24
Show game crashed title

iota97:
Just add a title telling that the game crashed. Making it only look like a BSOD might be confusing for phone user (also Windows 10 have a quite different one nowaday)

Screenshot_2022-02-24_12-24-21


hrydgard:
Yeah, good stuff.

PR #15450, opened 2022-03-19
[Build] Generate iOS IPA

anr2me:
This is moved from https://github.com/hrydgard/ppsspp/pull/15446 since i need to test another action on my master branch

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -230,7 +230,7 @@ jobs:
         sudo apt-get install libsdl2-dev libgl1-mesa-dev libglu1-mesa-dev
 
     - name: Create macOS git-version.cpp for tagged release
-      if: startsWith(github.ref, 'refs/tags/') && runner.os == 'macOS' && matrix.extra == 'test'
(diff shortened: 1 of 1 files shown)


hrydgard:
Hm, couple of tests still showing yellow, not sure what's up with that. Otherwise, is this ready to go?

ANR2ME:
As mentioned by @unknownbrackets before:
```
Changing the matrix on the test changes the required check names. If we merge this, we need to change the configuration of the branch protection.

-[Unknown]
```

The test stuck probably because this PR changed the artifact naming to include the id (ie. macos, clang) thus no longer recognized

hrydgard:
Hm, do we still need this? There's some conflict.

anr2me:
We already have the manual generate IPA, so this is not needed anymore.

This PR only used to automatically packed and released the IPA build to https://github.com/hrydgard/ppsspp/releases since the files used to generate the IPA already existed all this time (seems to be built together with the MacOS build) but not exposed to artifacts/release.

PR #15464, opened 2022-04-01
Allow to change atlas per theme

iota97:
In the theme ini: `UIAtlas = ui_atlas_name` (without extension, code will add `.meta` and `.zim` as appropriate).

Font atlas can't be changed as I don't think it's worth changing it.


unknownbrackets (approved):
Seems reasonable to me.  I hope the extra VFS calls aren't overly slow on Android, sigh.

-[Unknown]

hrydgard:
This broke UWP/Xbox apparently, see #15496

hrydgard:
Checked in a fix. It doesn't seem to find the ini file, though.

PR #15481, opened 2022-04-14
Fix some NEON code that had bad compile-time checks

hrydgard:
Some of it didn't compile.. Hopefully it actually works too.


unknownbrackets (approved):
Ah, good, sorry I didn't notice these before.

-[Unknown]

PR #15501, opened 2022-04-24
Make the existing ReinterpretFramebuffers/ShaderColorBitmask path work for Split/Second

### LINKED ISSUE #13957: Split-Second Missing HUD for 3rd person view

hrydgard:
Fixes #13957.

It took writing and debugging #15500 for me to understand what the issue with the old path was..

Much simpler alternative to #15500, or we could merge both but disable Split/Second for this one. Needs some benchmarks I guess...

### DIFF
--- a/GPU/Common/DrawEngineCommon.cpp
+++ b/GPU/Common/DrawEngineCommon.cpp
@@ -501,7 +501,7 @@ bool DrawEngineCommon::ApplyFramebufferRead(bool *fboTexNeedsBind) {
 		lastFrameBlit = gpuStats.numFlips;
 	}
 	++blitsThisFrame;
-	if (blitsThisFrame > MAX_REASONABLE_BLITS_PER_FRAME * 2) {
+	if (blitsThisFrame > MAX_REASONABLE_BLITS_PER_FRAME * 4) {
 		WARN_LOG_ONCE(blendingBlit2, G3D, "Skipping additional blits needed for obscure blending: %d per frame, blend %d/%d/%d", blitsThisFrame, gstate.getBlendFuncA(), gstate.getBlendFuncB(), gstate.getBlendEq());
 		return false;
 	}
--- a/GPU/Common/GPUStateUtils.cpp
+++ b/GPU/Common/GPUStateUtils.cpp
@@ -1033,7 +1033,7 @@ void ConvertMaskState(GenericMaskState &maskState, bool allowFramebufferRead) {
 }
 
 // Called even if AlphaBlendEnable == false - it also deals with stencil-related blend state.
-void ConvertBlendState(GenericBlendState &blendState, bool allowFramebufferRead) {
+void ConvertBlendState(GenericBlendState &blendState, bool allowFramebufferRead, bool forceReplaceBlend) {
 	// Blending is a bit complex to emulate.  This is due to several reasons:
 	//
 	//  * Doubled blend modes (src, dst, inversed) aren't supported in OpenGL.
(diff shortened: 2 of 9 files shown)


hrydgard:
I think we should definitely merge this one actually, and only then consider merging #15500.

GPU/Common/DrawEngineCommon.cpp
```diff
@@ -501,7 +501,7 @@ bool DrawEngineCommon::ApplyFramebufferRead(bool *fboTexNeedsBind) {
 		lastFrameBlit = gpuStats.numFlips;
 	}
 	++blitsThisFrame;
-	if (blitsThisFrame > MAX_REASONABLE_BLITS_PER_FRAME * 2) {
+	if (blitsThisFrame > MAX_REASONABLE_BLITS_PER_FRAME * 4) {
```
unknownbrackets, line 504: Should we just increase MAX_REASONABLE_BLITS_PER_FRAME?  It seems weird to multiply the constant, when this is its only use...

-[Unknown]

unknownbrackets (approved):

Gamemulatorer:
Performance on vulkan become slow unlike to opengl it's the same 🤔

### vulkan
Screenshot_2022-04-25-15-20-59-527_org ppsspp ppsspp

### opengl
Screenshot_2022-04-25-15-22-55-074_org ppsspp ppsspp

hrydgard:
@Gamemulatorer Check how performance is now :)

GPU/Common/DrawEngineCommon.cpp
hrydgard: I'm just gonna remove the whole thing, see #15515

Gamemulatorer:
> @Gamemulatorer Check how performance is now :)

Ok will I try :)

Gamemulatorer:
What's happening :(
Screenshot_2022-05-01-21-27-48-128_org ppsspp ppsspp

Gamemulatorer:
Both happen with ogl and vk

Panderner:
@Gamemulatorer latest build?

hrydgard:
@Gamemulatorer  Maybe your game ID is missing from compat.ini ? What is it?

unknownbrackets:
Is that black thing new since that recent merge, though?  Even if the game ID was missing, it would seem bad if the change made it worse in that way.

-[Unknown]

hrydgard:
I can't figure out how the change would cause that, works here on several devices. Maybe you have an older compat.ini that neither enables the old nor the new workaround, though in that case, it shouldn't have worked for you previously either...

Panderner:
Works fine for me on PPSSPP v1.12.3-1138-g36bc7a3e6
Screenshot (30)

But it has bugs for software rendering for this game:
Screenshot (29)

Gamemulatorer:
Nevermind I forget that I have a compat.ini file in system folder.
But it's ok now using Vulkan.
Screenshot_2022-05-02-18-18-35-988_org ppsspp ppsspp

hrydgard:
Great, thanks for reporting!

I have plans for some other changes that should speed the game up further, but they're a bit invasive and will need to wait for after the upcoming release.

PR #15507, opened 2022-04-27
Partially fixes SenseMe app

anr2me:
This will only works for offline music at `PSP/MUSIC`, while the SenseMe channels will need HTTP/HTTPS support to be fully working.

image
image

PS: For some reason the FPS can occasionally dropped below 60, causing the sound to be distorted :(
Edit: i didn't aware i was using IR interpreter, thus getting the occasional FPS drops

### DIFF
--- a/Core/HLE/sceHprm.cpp
+++ b/Core/HLE/sceHprm.cpp
@@ -44,6 +44,16 @@ static u32 sceHprmIsRemoteExist() {
 	return 0;
 }
 
+static u32 sceHprmRegisterCallback() {
+	DEBUG_LOG(HLE, "sceHprmRegisterCallback()");
+	return 0;
+}
+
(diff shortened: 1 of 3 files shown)


Core/HLE/sceHprm.cpp
```diff
@@ -44,6 +44,16 @@ static u32 sceHprmIsRemoteExist() {
 	return 0;
 }
 
+static u32 sceHprmRegisterCallback() {
+	DEBUG_LOG(HLE, "sceHprmRegisterCallback()");
```
hrydgard, line 48: These are unlikely to be called repeatedly, I think we should log these two (and maybe other ones here too) as:

ERROR_LOG(HLE, "UNIMPL %s", __FUNCTION__);

PR #15511, opened 2022-04-30
Detect ISO files from file contents if extensions are wrong.

hrydgard:
On older Android versions (that still support Android Storage Framework), filenames are not always preserved in Content URIs, they can look like `.../msf:342` and things like that. 

So let's identify a few more file types by content if filename identification fails.


unknownbrackets (approved):

hrydgard:
Got a report of a case that this will not handle though, that is one strange path:

image

PR #15516, opened 2022-05-01
Support different image format for atlas

iota97:
Provide the PNG as well in the assets to make it easier to use.

Make the default logic less messy, I don't have a UWP machine to test tho' :/


hrydgard:
Hm, what's the purpose of this? The atlas needs to be generated by the atlas generator anyway, you're not supposed to manually edit it...

iota97:
The idea was to make it easier to edit the atlas without having to compile or use the atlas tools.

Many people are able to edit a PNG, but having to mess with compilers and scripts might turn off someone.

If you think it's better to always regenerate the atlas feel free to close this ofc :)

hrydgard:
The issue is that such direct edits will become a maintenance nightmare when the atlases change... things can get shuffled around quite a bit. 

Have been previously leaning towards building the atlas at runtime from small pngs for non-font atlases, although the reduced I/O performance on modern Android kind of probably made that a bad idea :(

iota97:
I see :/

PR #15520, opened 2022-05-02
libkirk: Set Segher Boessenkool's original copyright notice

Linblow:
Segher Boessenkool (@segher) is the original author of libkirk's `ec.c` and `bn.c` files, which are originally for the Wii K-233.
Those files were used, and modified for KIRK.
After contacting him, I learnt his original copyright notice was somehow altered (probably years ago). 
This small PR restores his original copyright notice.


hrydgard:
Not sure how that happened, but yeah let's get it fixed.

Issue #15461, opened 2022-03-28
PPSSPP libretro core for RetroArch does not expose WLAN switch


Junohea:
### What should happen

The PPSSPP libretro core for RetroArch should have an exposed option to enable the PSP's "WLAN switch" in the same manner the standalone version provides.

The standalone version of PPSSPP has an option to toggle the "WLAN switch" state
Settings > Networking > "Enable networking/WLAN"
![image](https://user-images.githubusercontent.com/13029766/160466066-bd2ab8c3-d77e-4aa7-8cb2-22551ab904ea.png)

Games such as Phantasy Star Portable 2 require this value and will show it in-game (PPSSPP standalone with networking enabled)
![image](https://user-images.githubusercontent.com/13029766/160466182-db1655e5-158f-4645-832b-4ba0e5449a4c.png)

...and will not allow you to proceed if they do not detect the WLAN switch as enabled 
(shown: PPSSPP libretro core in RetroArch, all available netplay or networking options produce the following)
![image](https://user-images.githubusercontent.com/13029766/160470936-a635c291-164f-4794-b33d-90c9ab8a6ee1.png)
![image](https://user-images.githubusercontent.com/13029766/160467542-4395cda4-cd4e-43be-82c2-2a5a9c548b97.png)

- [x] This issue has shown up previously but was closed at the time
https://github.com/hrydgard/ppsspp/issues/2201
- [x] Tested latest nightly PPSSPP core in RetroArch. Also tested a few different older builds of the PPSSPP core.

### Who would this benefit

Anyone that would like to use RetroArch to play PSP games, specifically for games that perform checks against the state of the WLAN switch

### Platform (if relevant)

libretro / Retroarch

### Games this would be useful in

ULUS-10529 - Phantasy Star Portable 2, NPJH-50332 - Phantasy Star Portable 2 Infinity

### Checklist

- [X] Check the [latest git build](https://buildbot.orphis.net/ppsspp/) in case it's already implemented.
- [X] [Search for other requests](https://github.com/hrydgard/ppsspp/search?q=my+issue&type=issues) of the same feature.

johannes21j:
when i open the game it is normal when i play it kick me out then i delete Textures configuration setting then it stop kick me out please help

BunnyCinnamon:
I don't know the difference between libreto core and standalone, but I would assume networking is something only the standalone launcher can do?

ANR2ME:
[...]

PR #15538, opened 2022-05-13
Expose networking options in libretro core

amverweij:
This was inspired by #15461 and the fact that I wanted to play Soulcalibur Broken Destiny on my LAN.

I have been able to play a network game with the Libretro core on my Pi against PPSSPPQt standalone on my desktop. So there is added value. Network games between PPSSPP Libretro cores work as well.

<strike>I have not been able to play network games with the Libretro core on my Pi against another Libretro core on my desktop. The systems see each other, however when you challenge the other player, the challenge never arrives and nothing happens.</strike>

Some of these options might not make any sense for the Libretro core, however I can't judge that. I'm open to suggestions for improving this PR.

The idea of how to enter the IP address was borrowed from the Gambatte Libretro Core. Also I looked at Flycast for inspiration about hiding and showing menu items.

### DIFF
--- a/libretro/libretro.cpp
+++ b/libretro/libretro.cpp
@@ -689,7 +689,7 @@ static void check_variables(CoreParameter &coreParam)
    ppsspp_enable_builtin_pro_ad_hoc_server.Update(&g_Config.bEnableAdhocServer);
 
    ppsspp_chat_button_position.Update(&g_Config.iChatButtonPosition);
-   ppsspp_chat_screen_position.Update(&g_Config.iChatButtonPosition);
+   ppsspp_chat_screen_position.Update(&g_Config.iChatScreenPosition);
    ppsspp_upnp_use_original_port.Update(&g_Config.bUPnPUseOriginalPort);
    ppsspp_port_offset.Update(&g_Config.iPortOffset);
    ppsspp_minimum_timeout.Update(&g_Config.iMinTimeout);
--- a/libretro/libretro.cpp
+++ b/libretro/libretro.cpp
@@ -428,7 +428,6 @@ static RetroOption<std::string> ppsspp_change_mac_address[] = {
     {"ppsspp_change_mac_address12", "MAC address Pt 12: --:--:--:--:--:-X", MAC_INITIALIZER_LIST}
 };
 static RetroOption<int> ppsspp_wlan_channel("ppsspp_wlan_channel", "WLAN channel", {{"Auto", 0}, {"1", 1}, {"6", 6}, {"11", 11}} );
-static RetroOption<bool> ppsspp_discord_presence("ppsspp_discord_presence", "Send Discord \"Rich Presence\" information", true);
 static RetroOption<bool> ppsspp_enable_builtin_pro_ad_hoc_server("ppsspp_enable_builtin_pro_ad_hoc_server", "Enable built-in PRO ad hoc server", false);
 static RetroOption<std::string> ppsspp_change_pro_ad_hoc_server_address("ppsspp_change_pro_ad_hoc_server_address", "Change PRO ad hoc server IP address (localhost = multiple instances)", {
     {"socom.cc", "socom.cc"},
@@ -540,7 +539,6 @@ void retro_set_environment(retro_environment_t cb)
    for (int i = 0; i < 12; ++i)
       vars.push_back(ppsspp_change_mac_address[i].GetOptions());
    vars.push_back(ppsspp_wlan_channel.GetOptions());
-   vars.push_back(ppsspp_discord_presence.GetOptions());
    vars.push_back(ppsspp_enable_builtin_pro_ad_hoc_server.GetOptions());
    vars.push_back(ppsspp_change_pro_ad_hoc_server_address.GetOptions());
    for (int i = 0; i < 12; ++i)
@@ -696,7 +694,6 @@ static void check_variables(CoreParameter &coreParam)
 
    ppsspp_enable_wlan.Update(&g_Config.bEnableWlan);
    ppsspp_wlan_channel.Update(&g_Config.iWlanAdhocChannel);
-   ppsspp_discord_presence.Update(&g_Config.bDiscordPresence);
    ppsspp_enable_builtin_pro_ad_hoc_server.Update(&g_Config.bEnableAdhocServer);
 
    ppsspp_upnp_use_original_port.Update(&g_Config.bUPnPUseOriginalPort);
@@ -805,6 +802,7 @@ void retro_init(void)
    g_Config.flash0Directory = retro_base_dir / "flash0";
    g_Config.internalDataDirectory = retro_base_dir;
    g_Config.bEnableNetworkChat = false;
+   g_Config.bDiscordPresence = false;
 
    VFSRegister("", new DirectoryAssetReader(retro_base_dir));
 
--- a/libretro/libretro.cpp
+++ b/libretro/libretro.cpp
@@ -452,28 +452,6 @@ static RetroOption<int> ppsspp_pro_ad_hoc_ipv4[] = {
 };
 static RetroOption<bool> ppsspp_enable_upnp("ppsspp_enable_upnp", "Enable UPnP (need a few seconds to detect)", false);
 static RetroOption<bool> ppsspp_upnp_use_original_port("ppsspp_upnp_use_original_port", "UPnP use original port (enabled = PSP compatibility)", true);
-static RetroOption<bool> ppsspp_enable_network_chat("ppsspp_enable_network_chat", "Enable network chat", true);
-static RetroOption<int> ppsspp_chat_button_position("ppsspp_chat_button_position", "Chat button position", {
-    {"Bottom left", BOTTOM_LEFT},
-    {"Bottom center", BOTTOM_CENTER},
-    {"Bottom right", BOTOM_RIGHT},
-    {"Top left", TOP_LEFT},
-    {"Top center", TOP_CENTER},
-    {"Top right", TOP_RIGHT},
-    {"Center left", CENTER_LEFT},
-    {"Center right", CENTER_RIGHT},
-    {"None", 8}
-});
-static RetroOption<int> ppsspp_chat_screen_position("ppsspp_chat_screen_position", "Chat screen position", {
-    {"Bottom left", BOTTOM_LEFT},
-    {"Bottom center", BOTTOM_CENTER},
-    {"Bottom right", BOTOM_RIGHT},
-    {"Top left", TOP_LEFT},
-    {"Top center", TOP_CENTER},
-    {"Top right", TOP_RIGHT},
-    {"Center left", CENTER_LEFT},
-    {"Center right", CENTER_RIGHT}
-});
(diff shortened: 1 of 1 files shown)


libretro/libretro.cpp
```diff
@@ -379,9 +386,112 @@ static RetroOption<bool> ppsspp_retain_changed_textures("ppsspp_retain_changed_t
...
+    {"Bottom left", BOTTOM_LEFT},
+    {"Bottom center", BOTTOM_CENTER},
+    {"Bottom right", BOTOM_RIGHT},
+    {"Top left", TOP_LEFT},
+    {"Top center", TOP_CENTER},
+    {"Top right", TOP_RIGHT},
+    {"Center left", CENTER_LEFT},
+    {"Center right", CENTER_RIGHT}
+});
+static RetroOption<int> ppsspp_port_offset("ppsspp_port_offset", "Port offset (0 = PSP compatibility)", {{"0", 0}, {"5000", 5000}, {"10000", 10000}, {"15000", 15000}});
```
ANR2ME, line 437: Is this limiting port offset to 0, 5000, 10000, 15000 ? 
If so, this will make libretro unable to play with standalone version that use free value (just like Minimum timeout setting)

MojoJojoDojo:
I suggest enabling built in proadhoc server all the time like i suggested doing that for the standalone PPSSPP.
IMO there is no reason it should be toggable.

stuken:
I played around with this a bit a while back and did notice some instability when the built in adhoc pro server is enabled.

There is also some desync issues between libretro and the main builds.

libretro/libretro.cpp
amverweij: Yes indeed.

My first version had a drop down with 65535 possible choices. That took quite long to open, even on a fast desktop. And it was quite unpractical to scroll all the way down to 10000 (which is the value suggested in the AdHoc wiki). I ended up opening the options file with a text editor after ssh-ing into my Pi.

The Libretro core targets devices without a keyboard. As far as I can see, either we supply a discrete list of sane choices, or we use a single drop down control for each digit of the port offset  (similar to the IP address).

I was hoping we could get away with the discrete list of sane choices in this case.

amverweij: Would it be acceptable if I extend the list to 0, 5000-5020, 10000-10020, 15000-15020?

amverweij:
I would prefer the defaults to be the same as for the standalone version, unless there is a good reason to deviate for the Libretro core. Thank you @MojoJojoDojo for spotting the difference for UPnP. Fixed.

libretro/libretro.cpp
```diff
@@ -559,6 +682,20 @@ static void check_variables(CoreParameter &coreParam)
...
    bool isFastForwarding = environ_cb(RETRO_ENVIRONMENT_GET_FASTFORWARDING, &isFastForwarding);
    coreParam.fastForward = isFastForwarding;
+
+   ppsspp_enable_wlan.Update(&g_Config.bEnableWlan);
+   ppsspp_wlan_channel.Update(&g_Config.iWlanAdhocChannel);
+   ppsspp_discord_presence.Update(&g_Config.bDiscordPresence);
+   ppsspp_enable_builtin_pro_ad_hoc_server.Update(&g_Config.bEnableAdhocServer);
+
+   ppsspp_chat_button_position.Update(&g_Config.iChatButtonPosition);
+   ppsspp_chat_screen_position.Update(&g_Config.iChatButtonPosition);
```
iota97, line 692: Should be `iChatScreenPosition`.

stuken:
The psp mac address is hardcoded in the libretro config at https://github.com/hrydgard/ppsspp/blob/287dd66f20f2a03acd316a6f5d5b98ff2eb6ad26/libretro/libretro.cpp#L614. This probably needs to be unique per player. You should look at persisting the mac out to the config file. There is a randomize mac method in the codebase that can be used to populate it.

Also I dont think the chat stuff will work on libretro. IIRC it is handled in the ui rendering code which is totally bypassed in that build.

libretro/libretro.cpp
amverweij: Good catch. I'm not sure whether the chat options make a lot of sense for devices without keyboard. However reading what other people type could still be good.

Furthermore the chat button would be part of PPSSPP's on-screen keyboard? Can that be used from the Libretro core?

libretro/libretro.cpp
ANR2ME: Why not making it like the Minimum Timeout? it seems to use Min, Max, and Step, right?

If libretro doesn't support using Min,Max,Step, may be @MojoJojoDojo can tell us which port offset are commonly used by players? i heard some ppl also use 65000?

ANR2ME:
Just being curious, does the OpenPSID will be unique too? 
we currently generate a unique OpenPSID using the MAC address, because some games (ie. Patapon 3) need a unique OpenPSID for each players, otherwise there will be indexing issue (ie. all players will be moving the 1st Player's character instead of their own character)

ANR2ME:
> I played around with this a bit a while back and did notice some instability when the built in adhoc pro server is enabled.
> 
> There is also some desync issues between libretro and the main builds.

Even on standalone, multiplayer still have inconsistency issue between platforms (mostly Linux/Android vs Windows/MacOS) and between architecture (x86 vs ARM) on some games (mostly japanese games)

libretro/libretro.cpp
MojoJojoDojo: Well the default is 10000 so most people use that and it works fine.

ANR2ME: did you said that some games worked better with 65k long ago?

MojoJojoDojo: No other people said that 60k is a good choice for Gran Turismo but besides that I dont think i managed to get it to work well...
Previously I mostly used offset 1500 for playing though.

amverweij:
I removed the chat options, since these indeed do not make sense. The chat overlay does not show when using the Libretro core.

How about the discord presence? RetroArch supports this natively [(see here)](https://www.retroarch.com/?page=discord). I suppose it makes more sense to let the front end deal with this, meaning we should also disable it in the PPSSPP Libretro core.

libretro/libretro.cpp
amverweij: > Why not making it like the Minimum Timeout? which seems to use Min, Max, and Step, right?
> 
> If libretro doesn't support using Min,Max,Step, may be @MojoJojoDojo can tell us which port offset are commonly used by players? i heard some ppl also use 65000?

Using Min, Max, and Step to generate a drop down with entries between 0 and 65535 is definitely possible, technically. But it does not yield a usable input widget. If in doubt, just change the constructor of the ppsspp_port_offset RetroOption to
`
static RetroOption<int> ppsspp_port_offset("ppsspp_port_offset", "Port offset (0 = PSP compatibility)", 0, 65535, 1);
`
and see for yourself. It would be much better if we can settle for a sane and limited set of choices. 

As long as people who want to play a network game together can agree on a common port to use, we should be good?

libretro/libretro.cpp
amverweij: The chat functionality has been removed for the Libretro core. It doesn't make sense in the Libretro core since the chat overlay is not displayed by the front end.

amverweij:
I guess picking a random MAC address solved my issues with playing with Libretro core versus Libretro core. Very nice!

libretro/libretro.cpp
ANR2ME: the step doesn't need to be 1, it can be 500 or 1k or even larger steps

libretro/libretro.cpp
```diff
@@ -379,9 +412,98 @@ static RetroOption<bool> ppsspp_retain_changed_textures("ppsspp_retain_changed_t
...
+   {"ppsspp_pro_ad_hoc_server_address08", "PRO ad hoc server IP address Pt  8: ---.---.-x-.--- ", 0, 10, 1},
+   {"ppsspp_pro_ad_hoc_server_address09", "PRO ad hoc server IP address Pt  9: ---.---.--x.--- ", 0, 10, 1},
+   {"ppsspp_pro_ad_hoc_server_address10", "PRO ad hoc server IP address Pt 10: ---.---.---.x-- ", 0, 10, 1},
+   {"ppsspp_pro_ad_hoc_server_address11", "PRO ad hoc server IP address Pt 11: ---.---.---.-x- ", 0, 10, 1},
+   {"ppsspp_pro_ad_hoc_server_address12", "PRO ad hoc server IP address Pt 12: ---.---.---.--x ", 0, 10, 1}
+};
+static RetroOption<bool> ppsspp_enable_upnp("ppsspp_enable_upnp", "Enable UPnP (need a few seconds to detect)", false);
+static RetroOption<bool> ppsspp_upnp_use_original_port("ppsspp_upnp_use_original_port", "UPnP use original port (enabled = PSP compatibility)", true);
+static RetroOption<int> ppsspp_port_offset("ppsspp_port_offset", "Port offset (0 = PSP compatibility)", {{"0", 0}, {"5000", 5000}, {"10000", 10000}, {"15000", 15000}});
+static RetroOption<int> ppsspp_minimum_timeout("ppsspp_minimum timeout", "Minimum timeout (override in ms, 0 = default))", 0, 15000, 100);
```
ANR2ME, line 456: If this going to became a drop down of 150 items, it's probably too much, you can use larger steps like 500

libretro/libretro.cpp
amverweij: Thank you for the explanation, 1k steps would certainly yield a sane and discrete set of options.

libretro/libretro.cpp
amverweij: Agreed, will make 5000 the max

ANR2ME:
> I guess picking a random MAC address solved my issues with playing with Libretro core versus Libretro core. Very nice!

because adhoc games identify players by mac address, but some games use OpenPSID too, which is why both the mac and openpsid need to be unique for each player.

As a side note, some games are locking the save data to mac address, trying to load this save data on a different mac address will make the game to shows corrupted message/dialog

amverweij:
If people know their desired mac address, they can make their save states work by setting it manually in the Libretro core options.

amverweij:
(Disclaimer: I'm in no way an authority on how PPSSPP works. The following is what I could understand by setting some breakpoints and reading the relevant bits of code.)

It seems initializing the OpenPSID works out alright. In retro_load_game.cpp, we first call check_variables (at libretro.cpp:813) and then proceed with calling PSP_InitStart (at libretro.cpp:816).

The global variable g_Config.sMACAddress will be set from the call to check_variables. The OpenPSID is initialized by __OpenPSIDInit, which is called from __KernelInit in the ExecLoader thread. This thread is started from CPU_Init, which is called from PSP_InitStart (at System.cpp:432).

It follows that when we initialize the OpenPSID from g_Config.sMACAddress, this is done using the MAC address that is stored in the Libretro core settings file. One would need to restart the Libretro core in order to have the OpenPSID in line with the current MAC address after manually changing it in the settings, though.

hrydgard:
Looks fine, everyone on board to get it in?

ANR2ME:
> Looks fine, everyone on board to get it in?

I'm good with this

amverweij:
Thank you all for your constructive feedback!

PR #15541, opened 2022-05-17
(libretro) Add option to detect and notifiy frontend of internal frame rate changes (60 <-> 30 <-> 20 fps, etc.) + restore 'Duplicate Frames in 30 Hz Games' core option

jdgleaver:
This PR adds a new `Detect Frame Rate Changes (Notify Frontend)` option to the libretro core. When enabled, the core will detect frame rate changes that are a factor of the nominal display refresh rate (e.g. 60 <-> 30 <-> 20 fps) and notify the frontend, such that the frontend can call `retro_run()` with proper timing instead of relying on over-saturation of the audio buffer to control game speed.

When used in conjunction with RetroArch's new automatic `VSync Swap Interval` feature (https://github.com/libretro/RetroArch/pull/13927), this enables perfect frame pacing for 30 fps and 20 fps games on a 60 Hz display (and for 30 fps content on a 120 Hz display).

Since we now have a way to achieve correct pacing without force-enabling `g_Config.bRenderDuplicateFrames`, the `Duplicate Frames in 30 Hz Games` core option has been restored (and is disabled by default, as was originally the case).

Caveats:

- The `Detect Frame Rate Changes (Notify Frontend)` option is only available when frame skipping and duplicate frames are disabled
- This feature works best when the core is running at full speed - i.e. if the host platform cannot run at 100% speed, then frame rate transitions may be delayed or missed (but if the core is running slow then good frame pacing is impossible anyway, and the option is essentially irrelevant in that case...)

### DIFF
--- a/libretro/libretro.cpp
+++ b/libretro/libretro.cpp
@@ -34,6 +34,8 @@
 #include "Core/Host.h"
 #include "Core/MemMap.h"
 #include "Core/System.h"
+#include "Core/CoreTiming.h"
+#include "Core/HW/Display.h"
 
 #include "GPU/GPUState.h"
 #include "GPU/GPUInterface.h"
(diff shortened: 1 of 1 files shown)


hrydgard:
Interesting method. You might be able to get more accurate results by hooking into the (rather convoluted) logic in sceDisplay.cpp, but if this works for you, I don't mind it, as it's isolated to libretro.cpp :)

hrydgard:
Was this tested on Android? See #15561

PR #15456, opened 2022-03-24
[Adhoc] Fix inconsistency issue between Windows and non-Windows

anr2me:
This should:
- Fix disconnection issue on some games that use `sceNetAdhocPtpFlush` (mostly occurred on Linux) when the remote peer already closed the connection (ie. R-Type Command https://github.com/hrydgard/ppsspp/issues/14641#issuecomment-1059791525)
- _Might fix_ Unknown Source Port issue that occurred randomly on some games that use PDP communication (ie. MGS:PW). Test build confirmed by UltimaGamerX at discord to work better than official PPSSPP (MGS:PW co-op didn't work for him using official ppsspp). _But may need to be tested by more players who had Unknown Source Port issue to make sure, since i couldn't reproduce that issue anymore._
- _Might fix_ `sceNetAdhocPtpConnect` issue on Linux when the remote host is not listening yet on some games (ie. Taiko no Tatsujin Portable DX https://github.com/hrydgard/ppsspp/issues/14641#issuecomment-1067286995). _Need someone to test this on non-Windows._

**Test build:**
Android: https://sam.nl.tab.digital/s/EmmKP45q2PC7Sbi
iOS: https://sam.nl.tab.digital/s/c3Qst63Qw65qTE8

PR #15542, opened 2022-05-18
[Adhoc] Fix more inconsistency between Windows and non-Windows

anr2me:
- Fix some minor issue from https://github.com/hrydgard/ppsspp/pull/15456 (mostly logging issue)
- Fix some possible crashes issue.
- Fix Taiko no Tatsujin Portable DX on Linux https://github.com/hrydgard/ppsspp/issues/14641#issuecomment-1067286995 due to `select` can flags an unconnected socket as writable and readable on Linux.
- Tales of VS on Windows https://github.com/hrydgard/ppsspp/issues/14641#issuecomment-1107655077 due to `getpeername` can return a success when the other side haven't `accept`ed the connection yet on Windows.
- Detects connectivity to the AdhocServer better/more accurate.

PS: Tales of VS have additional issue where joining can fails on latency higher than 67ms, seems to be related to AdhocMatching, i'll create a separate PR for AdhocMatching-related issues.


hrydgard:
Thanks, looks good!

PR #15544, opened 2022-05-18
Updated brazilian portuguese translation

Felipefpl:
Updated the translation based on latest changes to en.ini file plus some small fixes.


assets/lang/pt_BR.ini
```diff
@@ -937,7 +937,7 @@ Stopping.. = Parando...
 [Reporting]
 Bad = Ruim
 FeedbackCRCCalculating = CRC do Disco: Calculando...
-FeedbackCRCValue = CRC do Disco: %1
+FeedbackCRCValue = CRC do Disco: [VALUE]
```
hrydgard, line 940: This should be %1 to fill in the right value.

Felipefpl: Thanks for the tip, didnt notice that when making the changes.

PR #15555, opened 2022-05-26
Fix themes as well

iota97:
Same as #15554.


unknownbrackets (approved):
Ah, sorry, I should've remembered this one as well.

-[Unknown]

PR #15571, opened 2022-06-07
Update zh_CN.ini

xujibbs:


xujibbs:
combine to another update

PR #15581, opened 2022-06-09
Add additional rescaps for UWP port

tunip3:
added runFullTrust, broadFileSystemAccess and  expandedResources rescaps
The first two expands what devices the port can access whilst running on Xbox.
expandedResources improves performance on Xbox by giving the app access to more system resources.


hrydgard:
Oh cool, didn't know about these. Hope adding them won't limit the use in any scenario... Let's merge and see how it goes.

PR #15588, opened 2022-06-11
Fix-ja_jp.ini

nassau-tk:
Fixed some UI text.

But, I couldn't change some text below.
image

### DIFF
--- a/assets/lang/ja_JP.ini
+++ b/assets/lang/ja_JP.ini
@@ -28,8 +28,8 @@ AnalogLimiter Tip = アナログリミッタボタンが押されたとき
 Auto = 自動
 Auto-centering analog stick = アナログパッドを自動でセンタリングする
 Auto-hide buttons after seconds = 指定した秒数後に自動でボタンを隠す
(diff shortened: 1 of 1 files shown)


hrydgard:
That's weird, it should have worked to change those. I'll look into it soon.

hrydgard:
It does seem to work for me, at least in recent builds.

nassau-tk:
And!! 
Show Memory Stick folder
too. Thanks!

nassau-tk:
> It does seem to work for me, at least in recent builds.
> 
> image

OK!!

hrydgard:
Hm, show memory stick folder doesn't seem to have been added to the ini files, but should work to translate if you do.

nassau-tk:
Oh.... I miss take it about MS folder text.
Ok. I will fix later.

And, Color Tint & Saturation is not change on my 1.12.3-1215.
I will check next merged build.

hrydgard:
Try 1.12.3-1264...

Issue #15526, opened 2022-05-08
Tales of Phantasia Cross Edition: missing some layers in vulkan on MTK device


leoxxx:
### Game or games this happens in

ULJS00293 Tales of Phantasia Narikiri Dungeon X 

### What area of the game

All

### What happens

Missing some layers.

![2022-05-08-15-36-08](https://user-images.githubusercontent.com/5364709/167287635-58a2072e-dfdb-4566-a32f-3b6c50a1189e.png)
![2022-05-08-15-36-21](https://user-images.githubusercontent.com/5364709/167287626-744dd9fa-312d-46a8-837d-b2b4078ef324.png)

Tales of Phantasia Cross Edition is inclueded in Tales of Phantasia Narikiri Dungeon X.
Error in vulkan, but O.K. in opengl on my MTK device. 
It's O.K. in vulkan on my Snapdragon(855) device and PC.
Tales of Phantasia Narikiri Dungeon X is O.K. in vulkan.

### What should happen

![1](https://user-images.githubusercontent.com/5364709/167288142-1996f517-557a-4d69-92c8-c32d3c9c3ae6.png)
![2](https://user-images.githubusercontent.com/5364709/167288147-2cb9eee0-ed9d-4f2e-ade4-4d7d59b0ae66.png)

### Platform

Android

### Mobile phone model or graphics card

Helio X20 (MT6797+Mali-T880 MP4) Android Tablet

### PPSSPP version affected

V1.12.3

### Graphics backend (3D API)

Vulkan

### Checklist

- [X] Test in the [latest git build](https://buildbot.orphis.net/ppsspp/) in case it's already fixed.
- [X] [Search for other reports](https://github.com/hrydgard/ppsspp/search?q=my+issue&type=issues) of the same issue.
- [ ] Try resetting settings or older versions and include if the issue is related.
- [X] Try changing graphics settings to determine if one causes the glitch (especially speed hacks or enhancements/replacements.)
- [X] Include logs or screenshots of issue.

Gamemulatorer:
Might be related to https://github.com/hrydgard/ppsspp/issues/12890

leoxxx:
Maybe.
Tales of Phantasia Full Voice Edition
Tales of Phantasia Cross Edition
They are not same.

Gamemulatorer:
How about the latest build
https://buildbot.orphis.net/ppsspp/

leoxxx:
g045709dbc still error.

[...]

Issue #15016, opened 2021-10-14
List of games with graphics rendering issues using mali gpu with vulkan backend


Gamemulatorer:
### What should happen

Some games have graphics glitch on mali-gpu using **vulkan** backend but I found a workaround to fix them.

Map Graphics Issue Affected Games Using **Vulkan** Only.
Midnight Club L.A Remix
![Screenshot_2021-10-14-08-49-18-082_org ppsspp ppsspp](https://user-images.githubusercontent.com/37603562/137234149-6d2960cd-0d1c-470c-a450-c8af7303e3d6.jpg)

Midnight Club 3 Dub Edition
![Screenshot_2021-10-14-08-48-14-272_org ppsspp ppsspp](https://user-images.githubusercontent.com/37603562/137234188-1d19ec0c-1775-4fdb-8e4e-eda4dcb3aa96.jpg)

Gensou Suikoden
![IMG_20211014_091325](https://user-images.githubusercontent.com/37603562/137234237-81f2a1b4-1e97-47c7-a43e-40ef98747e70.jpg)

Surf's Up™
![Screenshot_2021-10-14-08-44-21-508_org ppsspp ppsspp](https://user-images.githubusercontent.com/37603562/137234262-f15dafae-8e0d-4ab4-a0bb-da6265f8b3b5.jpg)

**The Simpsons Game** Also Have Graphics Glitch Using Vulkan.
![Screenshot_2021-10-14-08-38-56-700_org ppsspp ppsspp](https://user-images.githubusercontent.com/37603562/137234423-975db19f-fd4f-4c86-9e4d-1e031cc8b0d0.jpg)

***WorkAround*** I found to help those games is by adding them to compat.ini **PixelDepthRounding**
Midnight Club L.A Remix
![Screenshot_2021-10-14-08-48-53-196_org ppsspp ppsspp](https://user-images.githubusercontent.com/37603562/137234820-37949ce2-d476-477d-9356-899e0d8a37b0.jpg)

Midnight Club 3 Dub Edition
![Screenshot_2021-10-14-08-46-08-214_org ppsspp ppsspp](https://user-images.githubusercontent.com/37603562/137234850-e9f53d0d-38f7-4fb7-af90-f9d7ed1cf433.jpg)

Gensou Suikoden
![IMG_20211014_091402](https://user-images.githubusercontent.com/37603562/137234872-3b2063ed-8a7a-40f9-b2d7-c9b12d041a48.jpg)

Surf's Up™
[...]

PR #15611, opened 2022-06-18
Workaround old Mali driver bug with discard + stencil

hrydgard:
Seems to be similar to the old Adreno bug, though happening in very slightly different situations - I tried to narrow the GPU test to isolate it but failed :(

Either way, this workaround works for #13833 (but not #15526). I'm selectively enabling it only for known affected games out of the usual performance regression paranoia and the looming next release, might refine later.

Don't have 100% confidence in the driver version range, but it's something at least.

This also adds a simple driver bug display in system info.


unknownbrackets (approved):

ghost (commented):
Can you remove tales of phantasia because it doesn't fix the issue instead added the following games please.
* Megamind
* Badass Rumble 
* The Simpsons Game
* Surf Up
See https://github.com/hrydgard/ppsspp/issues/15016

hrydgard:
Thanks, I tagged #15016 to take a look at it before release.

PR #15613, opened 2022-06-19
Allow to repeat a "single" button

iota97:
Allow to simulate button smashing on a single custom button.


hrydgard (approved):
Looks good, just missing adding new translation strings in the inis :)

Though maybe should do that as separate commits anyway for reviewability.

PR #15614, opened 2022-06-19
Avoid including signal.h from CommonFuncs.h

hrydgard:
Trying to work-around #15609 .

(Could be argued that CommonFuncs.h should be split up, too...)

### DIFF
--- a/Common/CommonFuncs.h
+++ b/Common/CommonFuncs.h
@@ -36,6 +36,9 @@
 #define Crash() {asm ("bkpt #0");}
 #elif PPSSPP_ARCH(ARM64)
 #define Crash() {asm ("brk #0");}
+#else
+#include <signal.h>
+#define Crash() {kill(getpid(), SIGINT);}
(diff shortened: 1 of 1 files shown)


Common/CommonFuncs.h
```diff
@@ -28,11 +30,12 @@
 #include <unistd.h>
 #include <errno.h>
 
-#if defined(_M_IX86) || defined(_M_X86)
+#if PPSSPP_ARCH(X86) || PPSSPP_ARCH(AMD64)
 #define Crash() {asm ("int $3");}
-#else
-#include <signal.h>
-#define Crash() {kill(getpid(), SIGINT);}
```
iota97, line 35: I think that leaving this as a default fallback in an `#else` at the end might help ports to other platform such as POWER or RISC-V.

hrydgard: Yeah, that's a fair point.

PR #15619, opened 2022-06-20
Update lang file

iota97:
Add missing key and sync the files


Saramagrean:
Please add string for "Animation speed" (in shader setting). 🙏

iota97:
Fixed animation speed translation, the key was `Animation speed (0 = disable)` and that `=` was quite a problem. Now is just `Animation speed (0 -> disable)`.

PR #15620, opened 2022-06-21
Add numpad enter as a confirm

iota97:
Treat Numpad Enter the same way as Enter. Am I the only one using it?


hrydgard:
Apparently :)

hrydgard (approved):

PR #15627, opened 2022-06-24
ffmpeg: fix compile bug

zack-huangzihan:
Err is generated when USE_FFMPEG is set to OFF, The Camera
module relies on FFMPEG

Signed-off-by: ZiHan Huang <zack.huang@rock-chips.com>


hrydgard:
thanks!

PR #15633, opened 2022-06-28
Update zh_CN.ini

xujibbs:


leoxxx:
中文和英文中间加空格根本没必要。英文与英文中间有空格是因为单词。

hrydgard:
@leoxxx please feel free to fix it :)

hrydgard:
We should simply follow what's common practice in Chinese mobile software I think - but as I'm not familiar with it, I don't feel qualified to make the changes :)

PR #15694, opened 2022-07-15
Update zh_CN.ini

xujibbs:
amend some translation and love space


LunaMoo:
We literally had an issue about using space between Chinese and Romanian letters being incorrect and as soon as a fix was merged we get this PR to add spaces back...🤔

ANR2ME:
According to this https://github.com/ruanyf/document-style-guide/blob/master/docs/text.md the space should be half-width

hrydgard:
Yeah, I'm confused now @leoxxx , which is it? :)

unknownbrackets:
I'm guessing this might be a question of whether "life, the universe, and everything" should have two commas or one.

-[Unknown]

leoxxx:
@hrydgard 
https://zh.wikipedia.org/wiki/星海遊俠系列
`银河游侠系列（日语：スターオーシャン，英语：Star Ocean，香港和台湾译作“银河游侠”，又译“星之海洋”）是由tri-Ace开发、史克威尔艾尼克斯（原艾尼克斯）发售的角色扮演游戏系列。 `
So, I said "it' a habbit".

hrydgard:
Oh no. Well, I don't really feel qualified to make this decision, as my Chinese knowledge extends to about ten characters and a few words...

leoxxx:
But I want to point out some about translations.
https://www.merriam-webster.com/dictionary/artifact
a defect in an image (such as a digital photograph) that appears as a result of the technology and methods used to create and process the image
https://zh.wikipedia.org/wiki/壓縮失真
artifacts=失真， not"伪像".
https://github.com/hrydgard/ppsspp/wiki/如何使用PPSSPP多人联机游戏 , it's a decode url,  add spaces will be broken the url.

leoxxx:
@hrydgard 
You just choose which you like. We will follow which you choose.

Halo-Michael:
In any case, this is clearly unreasonable:
A59FF2E1-91F4-4386-A0DD-322A5788DBBD

Narugakuruga:
> But I want to point out some about translations' issue. https://www.merriam-webster.com/dictionary/artifact
> 
> > c: a defect in an image (such as a digital photograph) that appears as a result of the technology and methods used to create and process the image
> > … can produce a very good picture, but there will be some loss of detail and some color artifacts such as adjacent colors bleeding into each other.
> > — Consumer Reports
> 
> [https://zh.wikipedia.org/wiki/壓縮失真](https://zh.wikipedia.org/wiki/%E5%A3%93%E7%B8%AE%E5%A4%B1%E7%9C%9F) artifacts="失真", not"伪像". [https://github.com/hrydgard/ppsspp/wiki/如何使用PPSSPP多人联机游戏](https://github.com/hrydgard/ppsspp/wiki/%E5%A6%82%E4%BD%95%E4%BD%BF%E7%94%A8PPSSPP%E5%A4%9A%E4%BA%BA%E8%81%94%E6%9C%BA%E6%B8%B8%E6%88%8F) It's a decode url, add spaces will be broken the url.

“伪像”更能表达清楚一点，不用太纠结本本上的定义。PPSSPP提高分辨率的时候有的游戏的特效会变成条带状，可以理解成fake image

"伪像" tells more clearly. Don't entangle in the definition of the book too much. When PPSSPP increases the resolution, the graphic effects of some games will become stripes, which can be understood as fake image

Narugakuruga (approved):

leoxxx:
@Narugakuruga
既然是中国人，那我就不用英语了。
我表示不赞同。就和你说一个词吧。frame size，这个词是直译是帧大小，但实际指的是画面大小，帧指的就是一个画面。你要是说“帧大小”可能会被误会成“帧率”。伪像这词要给大众看，你觉得会有多少人知道是啥意思？

ANR2ME:
> @Narugakuruga 既然是中国人，那我就不用英语了。 我表示不赞同。就和你说一个词吧。frame size，这个词是直译是帧大小，但实际指的是画面大小，帧指的就是一个画面。你要是说“帧大小”可能会被误会成“帧率”。伪像这词要给大众看，你觉得会有多少人知道是啥意思？

Google translation:
> Since I am Chinese, I don't need English.
>
> I disagree. Just one word with you.  Frame size, the word is literally the frame size, but it actually refers to the size of the picture, and the frame refers to a picture. If you say "frame size" you may be mistaken for "frame rate". The word "artifact" is to be seen by the public, how many people do you think will know what it means?

If would be nice if you use English when posting at github, even when you don't need English, because the devs might not understood what you guys are talking about :) even auto-translation can translate it in a weird wording

PS: i think what he/she meant by artifacts/bleeding color is when using small resolution or small sized font some pixels may bleed nearby pixels and making it hard to read when the spacing is too tight/small, but this kind of issue should only happen when the frame is being downscaled, upscaling with bilinear/bicubic might also bleed a little bit (the default scaling/stretching from internal resolution to window size is bilinear i think)

unknownbrackets:
Just in case it helps, any phrasing works.  What it's meaning by:

> Lower resolution for effects (reduces artifacts)

Is, in longer form:

> Detect buffers used for rendering effects, such as shadows, motion blur, shiny lighting, lens flare, or similar.  Render these buffers using a reduced resolution. (Doing this may reduce ugly markings on the screen from these effects, which occur because they don't get blurred enough at high resolution.)

That's obviously longer than would fit or make sense.  That's the meaning that was shortened in English, so any words that accomplish that same meaning should be good words in another language.

-[Unknown]

leoxxx:
> > @Narugakuruga 既然是中国人，那我就不用英语了。 我表示不赞同。就和你说一个词吧。frame size，这个词是直译是帧大小，但实际指的是画面大小，帧指的就是一个画面。你要是说“帧大小”可能会被误会成“帧率”。伪像这词要给大众看，你觉得会有多少人知道是啥意思？
> 
> Google translation:
> 
> > Since I am Chinese, I don't need English.
> > I disagree. Just one word with you.  Frame size, the word is literally the frame size, but it actually refers to the size of the picture, and the frame refers to a picture. If you say "frame size" you may be mistaken for "frame rate". The word "artifact" is to be seen by the public, how many people do you think will know what it means?
> 
> It would be nice if you use English when posting at github, even when you don't need English, because the devs might not understood what you guys are talking about :) even auto-translation can translate it in a weird wording
> 
> PS: i think what you/he/she meant by artifacts/bleeding color is when using small resolution or small sized font some pixels may bleed nearby pixels and making it hard to read when the spacing is too tight/small, but this kind of issue should only happen when the frame is being downscaled, upscaling with bilinear/bicubic might also bleed a little bit (the default scaling/stretching from internal resolution to window size is bilinear i think)
> 
> PPS: i don't even know who said what anymore when you guys posting in chinese wording >.<

I am sorry, I can't describe this comment completly. Some words need use net translation.
The general idea: I disagree about Narugakuruga's comment. E.g. literal translation, frame size=帧大小.It's a profession professional terminology. Infact it's a graph or screen size(画面大小). If it's "帧大小"，it will be mistaked "frame size" (帧率) for Chinese. So, I think "伪像" won't be konwn for Chinese.
I hope it will let you don't mistake.

leoxxx:
> Just in case it helps, any phrasing works. What it's meaning by:
> 
> > Lower resolution for effects (reduces artifacts)
> 
> Is, in longer form:
> 
> > Detect buffers used for rendering effects, such as shadows, motion blur, shiny lighting, lens flare, or similar.  Render these buffers using a reduced resolution. (Doing this may reduce ugly markings on the screen from these effects, which occur because they don't get blurred enough at high resolution.)
> 
> That's obviously longer than would fit or make sense. That's the meaning that was shortened in English, so any words that accomplish that same meaning should be good words in another language.
> 
> (I'm noting this because it's actually a different kind of "artifacts" than colors bleeding into each other.)
> 
> -[Unknown]

You agree with Narugakuruga,is that so?
I can't understand you meaning. If you can, please use Chinese for it.

unknownbrackets:
Unfortunately, my Chinese is bad enough that it would only add confusion.  I don't agree or disagree with either here, just clarifying the English meaning of artifacts here.

For example, in some games you'll see an "after image" of characters with this setting disabled.  On the PSP, games rendered at a lower resolution so this after image was a blurry shadow.  But with 4x PSP render resolution in PPSSPP, you just see the character multiple times.

This setting makes it so it blurs those properly, so it looks like it did on the PSP.

I think Dissidia is an example of a game where this happens.

-[Unknown]

leoxxx:
I see.
https://www.merriam-webster.com/dictionary/artifact
Then, I think C is the similar meaning, is it?

unknownbrackets:
Yes, this one:

> c : a defect in an image (such as a digital photograph) that appears as a result of the technology and methods used to create and process the image

-[Unknown]

leoxxx:
https://wangwei1237.gitbook.io/digital_video_concepts/shi-pin-zhi-liang-du-liang/4_1_0_compressionlossartifactsandvisualquality/4_1_2_commonartifacts
https://www.wiki8.com/weiying_155172/
https://wenku.baidu.com/view/48084e090366f5335a8102d276a20029bd6463a7.html
artifact: 伪像；失真；伪影；假像
"失真" doesn't refers only to color error in Chinese. It's meanings a error image. All meanings are right. 
I think “图像错误” is a straightaway meaning for Chinese.

PR #15695, opened 2022-07-16
Update zh_CN.ini

xujibbs:
amend Line 793 to ensure a proper wiki page and translation update


iota97:
The URL was pointing to the Wiki page in Chinese, that was intentional. Do you have any problem with that link?

xujibbs:
my fault, I can't connect it this morning and I simply think it is wrong.

PR #15706, opened 2022-07-19
(UWP) Small Building Tweaks.

ghost:
This prevents two files from being included in UWP appx/msix packages generated via Visual Studio.

The first is a CMake txt file which has no use in a finished build, and the second is a redundant asset file that is never used by the final build.

These are very minimal changes and lead to zero differences in both building and using the app. But a cleaner finalized package is still worthwhile. :)

### DIFF
--- a/UWP/CommonUWP/CommonUWP.vcxproj
+++ b/UWP/CommonUWP/CommonUWP.vcxproj
@@ -632,6 +632,7 @@
   </ItemGroup>
   <ItemGroup>
     <Text Include="..\..\ext\libpng17\CMakeLists.txt" />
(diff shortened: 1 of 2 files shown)


hrydgard (approved):

GABO1423:
> Hm, getting an error:
> 
> ```
>  D:\a\ppsspp\ppsspp\UWP\CommonUWP\CommonUWP.vcxproj(635,7): error MSB4232: Items that are outside Target elements must have one of the following operations: Include, Update, or Remove.
> ```

Weird, couldn't replicate that issue on my end. I guess I can look into it later today.

hrydgard:
Wait what, didn't mean to merge it :/

PR #15711, opened 2022-07-20
(UWP) More Build Tweaks

ghost:
- A dummy Package.appxmanifest file was added. This is needed for Visual Studio to allow the user to create a package, and is overwritten during build time by either PackageNormal.appxmanifest or PackageGold.appxmanifest. So there shouldn't be any changes to Cl builds.
- The temporary key used to sign a finished package has been updated with one that expires at a later date. Also makes it simpler to make an installable package via Visual Studio.
- As a result of the above, the old key file has been deleted. It wasn't used by anything, so no reason to keep it.
- The version number in the PackageNormal and PackageGold manifests has been updated to PPSSPP's current version number.

For further clarification for the first change, here is the option in question:

image


hrydgard:
Thanks, looks good.

Would be nice if we could also document how to generate new temp keys easily..

GABO1423:
> Thanks, looks good.
> 
> Would be nice if we could also document how to generate new temp keys easily..

I can tell you how I do so if it helps. When you choose "Create App Packages" in Visual Studio, you are then given a few options before the Package is generated. This includes you it will be signed, and this is where you can generate a new key. I don't know if there are other methods, but that one is what I use.

unknownbrackets:
It'd be most ideal to have a GitHub Action (which runs on a X month schedule or manually) that updates the key and commits it to the repo, if it needs to be periodically updated.  That'd require a command line method of generating it.

-[Unknown]

GABO1423:
Yeah, the key lasts a year before it expires. So it could be updated yearly. But I sadly don't know how to generate keys via a command line.

unknownbrackets:
Looks like it's just a self-signed code signing cert:

```
       X509v3 extensions:
            X509v3 Basic Constraints: critical
                CA:FALSE
            X509v3 Extended Key Usage: critical
                Code Signing, 1.3.6.1.4.1.311.84.3.1
```

This doesn't seem to be exactly right, but what if you try this:

```
openssl req -x509 -nodes -newkey rsa:2048 -keyout key.pem -out cert.pem -days 7300 -subj "/CN=Henrik" -extensions v3_req -addext "extendedKeyUsage = codeSigning, 1.3.6.1.4.1.311.84.3.1"
openssl pkcs12 -export -out cert.pfx -in cert.pem -inkey key.pem -name "03d871fb-e260-4c26-bc87-c32adc53c8e6" -CSP "Microsoft Strong Cryptographic Provider" -passout pass:
rm cert.pem key.pem
```

If you use the cert.pfx that generates, does everything work?  Here's an example in case you don't have openssl:
[PPSSPP_TestKey.zip](https://github.com/hrydgard/ppsspp/files/9174143/PPSSPP_TestKey.zip)

-[Unknown]

GABO1423:
> 

Test key did not work sadly.

image

PR #15717, opened 2022-07-24
Allows "merging" render targets that overlap on the Y axis. Fixes Juiced 2

### LINKED ISSUE #7295: Juiced 2: Hot Import Nights, screen artifacts and missing half of race tracks
a

The game has problems with light sources blurring the image, staying forever on screen, also the game is missing half of the asphalt on some tracks. Gameplay-wise is perfectly fine, just graphical issues.
As suggested by unknown, here is the GE Debugger for the game right when the bug happens (after the loading screen and right at start of the race) with all the tabs screenshotted, i've stretched the windows as high as i could:
http://expirebox.com/download/c04b80309ba636314960c73b81fd00c7.html

EDIT: Also the menu is broken in Vulkan: 
screenshot_2017-11-21-23-31-09-460_org ppsspp ppsspp

hrydgard:
Fixes #7295.

Juiced 2 bloom works fine with this. It does some tricky overlap between two render targets.

To be safe, gating this behind the related `AllowLargeFBTextureOffsets` compat flag, which is also required for the effect to work.

Also improves logging of multiple matches when binding framebuffers as textures (which was the main clue).

Additionally, fixes the offset check for X offsets to take bytes-per-pixel into account, which I guess is a very small risk.

### DIFF
--- a/GPU/Common/FramebufferManagerCommon.cpp
+++ b/GPU/Common/FramebufferManagerCommon.cpp
@@ -306,7 +306,7 @@ VirtualFramebuffer *FramebufferManagerCommon::DoSetRenderFrameBuffer(const Frame
 			}
 			break;
 		} else if (v->fb_stride == params.fb_stride && v->format == params.fmt) {
-			u32 v_fb_first_line_end_ptr = v->fb_address + v->fb_stride * bpp;
+			u32 v_fb_first_line_end_ptr = v->fb_address + v->fb_stride * 4;  // This should be * bpp, but leaving like this until after 1.13 to be safe. The God of War games use this for shadows.
 			u32 v_fb_end_ptr = v->fb_address + v->fb_stride * v->height * bpp;
 
 			if (params.fb_address > v->fb_address && params.fb_address < v_fb_first_line_end_ptr) {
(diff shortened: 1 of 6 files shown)


Gamemulatorer:
Nice 👍

unknownbrackets:
Does Breath of Fire 3 still work (or maybe that was the other kind of overlap only?)  I think there was some Silent Hill game triggering this too with 1024 wide buffers in a framedump?

-[Unknown]

hrydgard:
Breath of Fire 3 still works fine, and doesn't seem to trigger this kind of x offset. Don't remember which game did... I guess could find out with some git digging.

unknownbrackets:
Looks like it was originally #6324.  I remember some game having interleaved renders, though, pretty sure it was a Silent Hill.

-[Unknown]

hrydgard:
I reverted it, but the shadows in GoW do work fine even pre-revert. Let's fix it after 1.13.

unknownbrackets:
Well, the 4 iirc was just to skip the checks inside if it definitely didn't overlap, the actual x offset and height were checked afterward.  I don't think it can break it either way.

-[Unknown]

Gamemulatorer:
While the graphics glitch is fixed the performance is regress :(
Screenshot_2022-07-26-07-10-13-446_org ppsspp ppsspp

hrydgard:
Hmm, indeed :(

Letting the two overlapping render targets merge into one fixed the correctness issue, but as a result, instead we detect "self-texturing" (the render target being drawn into is also the texturing being used) which currently forces us into making a copy of the whole render target, about 30 times... Not great. 

There are ways to fix this, but it'll have to wait for after the release. In the meantime, the old cheats to turn off the effect probably work.

PR #15727, opened 2022-07-25
Implement the PSP's equal-size mips "3D texturing"

hrydgard:
See #6357. Basically, the PSP has a curious ability to have mipmaps with the same size and use mipmap interpolation to interpolate between them, essentially a very limited version of multitexturing or 3D texturing. This implements that using 3D textures.

This functionality is used heavily in Misshitsu no Sacrifice, and for the water effect in the Macross series.

~~Vulkan-only currently, though all the other backends except ES 2.0 without `GL_OES_texture_3d` can support it with some work.~~ This is done, now works on all backends.

Potentially this could also replace the hack we use for Tactics Ogre translation's text, see #5350 - though we'd still need it as a fallback just for GLES 2.0 without 3D texture support.

~~Strangely I don't think it looks quite as smooth as on the real PSP. I'm wondering if the second level should somehow have its own CLUT, though I don't see how that would work given that the first level uses an 8-bit CLUT that covers the whole of CLUT memory.~~ YES there's space for a 2nd CLUT!

Remaining work before merge (but mostly ready for review):

- [x] Fix the flicker issue (probably give it its own uniform flag, just gotta be careful not to regress games that change the mip bias a lot like GT)
- [x] Implement compatibility checks for OpenGL, disable if 3D textures are not available

### DIFF
--- a/GPU/Common/FragmentShaderGenerator.cpp
+++ b/GPU/Common/FragmentShaderGenerator.cpp
@@ -276,8 +276,13 @@ bool GenerateFragmentShader(const FShaderID &id, char *buffer, const ShaderLangu
 			WRITE(p, "precision highp int;\n");
 		}
 
-		if (doTexture)
-			WRITE(p, "uniform sampler2D tex;\n");
+		if (doTexture) {
+			if (texture3D) {
+				WRITE(p, "uniform sampler3D tex;\n");
+			} else {
+				WRITE(p, "uniform sampler2D tex;\n");
+			}
+		}
 
 		if (readFramebufferTex) {
 			if (!compat.texelFetch) {
@@ -335,6 +340,11 @@ bool GenerateFragmentShader(const FShaderID &id, char *buffer, const ShaderLangu
 		if (doTexture && texFunc == GE_TEXFUNC_BLEND) {
 			*uniformMask |= DIRTY_TEXENV;
 			WRITE(p, "uniform vec3 u_texenv;\n");
+		}
+
+		if (texture3D) {
+			*uniformMask |= DIRTY_TEXCLAMP;
+			WRITE(p, "uniform float u_mipBias;\n");
 		}
 
 		WRITE(p, "%s %s lowp vec4 v_color0;\n", shading, compat.varying_fs);
--- a/GPU/Common/FragmentShaderGenerator.cpp
+++ b/GPU/Common/FragmentShaderGenerator.cpp
@@ -83,7 +83,7 @@ bool GenerateFragmentShader(const FShaderID &id, char *buffer, const ShaderLangu
 	bool flatBug = bugs.Has(Draw::Bugs::BROKEN_FLAT_IN_SHADER) && g_Config.bVendorBugChecksEnabled;
 
 	bool doFlatShading = id.Bit(FS_BIT_FLATSHADE) && !flatBug;
-	bool shaderDepal = id.Bit(FS_BIT_SHADER_DEPAL);
+	bool shaderDepal = id.Bit(FS_BIT_SHADER_DEPAL) && !texture3D;  // combination with texture3D not supported. Enforced elsewhere too.
 	bool bgraTexture = id.Bit(FS_BIT_BGRA_TEXTURE);
 	bool colorWriteMask = id.Bit(FS_BIT_COLOR_WRITEMASK) && compat.bitwiseOps;
 
@@ -559,6 +559,8 @@ bool GenerateFragmentShader(const FShaderID &id, char *buffer, const ShaderLangu
 						WRITE(p, "  vec4 t = tex2D(tex, %s.xy)%s;\n", texcoord, bgraTexture ? ".bgra" : "");
 					}
 				} else {
+					// Note that here we're relying on the filter to be linear. We would have to otherwise to do two samples and manually filter in Z.
+					// Let's add that if we run into a case...
 					if (texture3D) {
 						if (doTextureProjection) {
 							WRITE(p, "  vec4 t = %sProj(tex, vec4(%s.xy, u_mipBias, %s.z));\n", compat.texture, texcoord, texcoord);
--- a/GPU/Common/FragmentShaderGenerator.cpp
+++ b/GPU/Common/FragmentShaderGenerator.cpp
@@ -348,7 +348,7 @@ bool GenerateFragmentShader(const FShaderID &id, char *buffer, const ShaderLangu
 		}
 
 		if (texture3D) {
-			*uniformMask |= DIRTY_TEXCLAMP;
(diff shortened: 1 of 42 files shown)


GPU/Common/FragmentShaderGenerator.cpp
```diff
@@ -558,10 +559,19 @@ bool GenerateFragmentShader(const FShaderID &id, char *buffer, const ShaderLangu
 						WRITE(p, "  vec4 t = tex2D(tex, %s.xy)%s;\n", texcoord, bgraTexture ? ".bgra" : "");
 					}
 				} else {
-					if (doTextureProjection) {
-						WRITE(p, "  vec4 t = %sProj(tex, %s);\n", compat.texture, texcoord);
+					if (texture3D) {
+						WRITE(p, "  float bias = pow(u_mipBias, 1.0);\n");
```
hrydgard, line 563: note to self, remove, was experimenting to see if a curve affected the perception of smoothness.

Additionally, this won't combine with shader depal currently, should probably lock it out in the shaderid level.

hrydgard:
image

There's some seriously suspicious stuff there in the CLUT uploads. It DOES set the mode to separate CLUTs, which we thought to be impossible with 8-bit color indices. Then it does two CLUT loads, one supposedly zero-sized, but there's an extra bit there (000040) which we currently ignore...  Maybe it selects an upper CLUT register for mip1.

hrydgard:
Alright, this now looks completely right, though still Vulkan-only. Work on that after the release.

LunaMoo:
Sorry had to stay longer at work, soo far it seems to work great in Misshitsu, in Macross it works great initially, but after flying around fast it starts glitchin
including dumps of both games, but in macross it seems the dump is textureless at least where it uses mipmaps
[DUMP.zip](https://github.com/hrydgard/ppsspp/files/9185253/DUMP.zip)

and a video of macross water glitching(it happens after flying on the map)
https://user-images.githubusercontent.com/5485237/180888772-66725c29-da1b-4b29-b24c-03fd933532c7.mp4

unknownbrackets:
Hm, I think we should create more tests for the second CLUT, clearly.  I guess I must've forgotten to test that since I don't see any tests?  So many GE features to test, so many assumptions to validate...

Not sure why the textures wouldn't dump correctly?

-[Unknown]

hrydgard:
Definitely needs testing yeah, but heh, not hard to miss testing that, I haven't seen any speculation anywhere before that the CLUT memory might be larger than 1024 bytes, so this is a brand new discovery.

hrydgard:
@LunaMoo Thanks for testing! That glitching does look weird yeah, and the conditions for it happening are odd... but I have a theory. We might have to set clamp instead of wrap mode on the Z dimension of the texture, for this case.

LunaMoo:
I had a thought dump might be textureless because of some settings I used in my textures.ini, but even without textures folder it doesn't dump mipmap textures.
[ULJS00321_0008.zip](https://github.com/hrydgard/ppsspp/files/9188052/ULJS00321_0008.zip)


 As for the glitching I can now reproduce by simply unthrottling in one spot, so maybe it does happen over time. If there will be any changes I could test for that I will, but in few hours I'm going to work^_^;.

LunaMoo:
Unfortunately still glitching by just using unthrottle in one spot for a while.

hrydgard:
Hm, maybe a bug with state tracking of the mip bias constant... I'll clean that code up later and see if I can get it fixed. Thanks for testing so far!

LunaMoo:
With the latest updates to this branch the glitching still happens on vulkan, also OGL behaves weirdly not drawing some mipmap textures at all depending on the player location on the map:
ULJS00321_00008
[ULJS00321_0012.zip](https://github.com/hrydgard/ppsspp/files/9226749/ULJS00321_0012.zip)

hrydgard:
Yeah, I haven't actually done any new fixes, just rebased it upon the latest code. I've been cleaning house and refactoring a lot.

More work is coming on this one soon.

hrydgard:
I've now implemented it for all the backends. Remaining work:

- [ ] Fix the flicker issue (probably give it its own uniform flag)
- [ ] Implement compatibility checks for OpenGL, disable if 3D textures are not available

hrydgard:
@LunaMoo can you test Macross again? If it looks good now, let's get this in.

LunaMoo:
Works great in both vulkan and OGL, the glitching is fixed, in both d3d modes it's missing bugged in same way:
ULJS00321_00009 
(which part of the terrain are missing change while walking)

hrydgard:
That is super weird! Could you get a savestate of a situation where that happens? If lucky might be able to get a frame out of one even if the GE dump doesn't work...

LunaMoo:
Sure, made one right after starting the stage:
[PPSSPP_STATE.zip](https://github.com/hrydgard/ppsspp/files/9228639/PPSSPP_STATE.zip)

LunaMoo:
A video might give a better hint how d3d fails:
https://user-images.githubusercontent.com/5485237/182026453-1f3fdfc7-2f74-4c30-a1d7-e6f96cd92e60.mp4

hrydgard:
Thanks, will look into it some more soon. Likely some ordering issue in binding the 3D texture (d3d is kinda picky sometimes), or something...

hrydgard:
Alright, that should take care of it, I think.

hrydgard:
Also I just realized that to make this work with max quality filtering (auto gen missing mipmaps), for the animation to work we really need to mipmap-filter the levels separately (and not have them mush together in the distance as 3D mipmaps would), which means array textures plus manual filtering is probably the way to go instead of 3D textures. This would only be available on Vulkan and D3D11. Anyway, that's for another day...

LunaMoo:
Works great now.:) Thanks!

hrydgard:
Thanks for your testing!

unknownbrackets:
The dump from https://github.com/hrydgard/ppsspp/pull/15727#issuecomment-1194744504 looks strangely dark.  I guess this must be from the second CLUT being ignored by GE frame dump recording.

Anyway, software does render this dump slightly incorrectly.  Here's the PSP output:
#15727_ULJS00251_misshitsu_3dtex1

The software output (it's visually very close):
#15727_ULJS00251_misshitsu_3dtex1_softgpu

The same dump looks like this on Vulkan:
#15727_ULJS00251_misshitsu_3dtex1_vulkan

This already suggests to me that the second CLUT theory isn't true on the real PSP (confirming my prior attempts to reproduce it.)  If I revert the secondary CLUT change I get (Vulkan):
#15727_ULJS00251_misshitsu_3dtex1_vulkan2

That said, it's been reported that it looks better and I'm not against enhancements.  I'll just make this a bit safer in case some game accidentally triggers this behavior unintentionally...

-[Unknown]

PR #15734, opened 2022-07-27
Better pt-pt translation

daniemun:
Made this because pt-pt translation is incomplete and outdated.
I would say it's ~98% translated.
Also, had to merge the original branch so my fork stays updated.


hrydgard:
Thank you, this is great! It'll go in the 1.13.1 release.

PR #15749, opened 2022-07-29
Color format cleanups

hrydgard:
I'm on a deep [yak shaving](https://americanexpress.io/yak-shaving/#:~:text=But%20what%20is%20yak%20shaving,going%20down%20the%20rabbit%20hole.%E2%80%9D) quest here. The origin was the 3D texture thing (#15727 ), then when trying to expand that to more backends than Vulkan, I found I needed to refactor texture loading (#15743), which spawned a number of cleanup tasks when I ran into things that were silly and/or outdated.

This one removes the ReplaceTextureFormat enum (which had 7 unused members) and their related code, keeping only the case we are currently actively using. Then I found out that some parameters were always false these days now that we switch R/B in the shaders for the D3D backends, so cleaned up more stuff. Good to get in separately.

### DIFF
--- a/Core/TextureReplacer.cpp
+++ b/Core/TextureReplacer.cpp
@@ -635,50 +635,13 @@ void TextureReplacer::NotifyTextureDecoded(const ReplacedTextureDecodeInfo &repl
 
 	SimpleBuf<u32> saveBuf;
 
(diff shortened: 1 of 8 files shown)


Anuskuss:
bryan-cranston-replace-the-bulb

PR #15754, opened 2022-07-30
Unify stencil upload using thin3d

hrydgard:
We currently have four different code paths, one per backend, to upload a stencil buffer to the GPU (which is a complex operation, have to do it a bit at a time). This is required for Star Ocean.

This is a prime use case for thin3d, our internal wrapper for simple drawing, now that #15753 added the required stencil functionality. This also required some updates and fixes to the ShaderWriter so we can mostly share the shader code, too, included in this PR. This way it also lays the ground work for more easily writing new stuff using thin3d, ~~like the upcoming depth upload needed for #9576~~ (didn't happen, another approach was better).


hrydgard:
Not working on Mac, will have to fix that before merging.

PR #15770, opened 2022-07-31
CI: fix tag regex to match also shorter tags like v1.13

rofl0r:
the previous tag regex matched only tagnames with 2 dots in them,
that's why the build.yml didn't run for the v1.13 tag.

addressing https://github.com/hrydgard/ppsspp/issues/15737

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,7 +7,7 @@ on:
     # For testing.
     - actions
     tags:
-      - "v*.*.*"
+      - "v*.*"
     paths-ignore:
     - '*.{txt,md}'
     - 'Tools/**'
--- a/.github/workflows/tarball.yml
+++ b/.github/workflows/tarball.yml
@@ -7,7 +7,7 @@ on:
       - created
   push:
     tags:
-      - "v*.*.*"
+      - "v*.*"
 
 jobs:
   build:


hrydgard:
I'm terrible at regex, will the new one match two dots even though it only contains one? * matches anything right, but the old expression basically expressed that two dots were required, but it would also have matched three dots, for example?

rofl0r:
yes, i just tested and the actions are triggered now on both vX.X and vX.X.X style tagnames.
btw regex here is actually a misnomer, as the tag declaration is really using a fnmatch(3p) style matching syntax (e.g. like filename matching in shell).

ANR2ME:
When i tested it on regex101.com, both shows the same result for Match
image
image

unknownbrackets:
Hm, I thought somewhere we said we'd use the `.0` because of one of these probably requiring it.

This isn't a regex afaik, it's an fnmatch.  If it were regex, `v*.*.*` would mean "zero or more 'v' characters, then 0 or more characters, then zero or more characters."  In other words, as a regex, `v*.*.*` would match literally anything.  So would `v*.*`.

The safest thing is probably `v*`, since [according to the actions docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet) `*` matches anything anyway.  I'm hoping we don't go with `v2` or something, but presumably `v*.*` won't match that.  Note that `v*.* would match `value.data` just as well.

-[Unknown]

rofl0r:
yes, i stated earlier that it's fnmatch style matching.

as of the time of this writing, you can actually see that the action didn't run, by looking at the last item of https://github.com/hrydgard/ppsspp/actions?page=4  - the only action that was started for the v1.13 tag/release push was tarball.yml.

PR #15777, opened 2022-08-01
Deferred depth buffer copies

hrydgard:
Due to the fact that the PSP overlaps stencil with color alpha, and PC interleaves stencil with depth instead, to get the semantics right we often have to copy the depth buffer from a previously bound framebuffer.

This avoids doing that if we never use depth after binding a color buffer, before binding the next one. This results in eliminated depth copies in several and perhaps many games, including God of War, saving some GPU performance.

This was made possible after #15700 and was @unknownbrackets 's idea (see comments in the issue) so cred goes to him.

Don't really have any numbers for the improvement, but eliminating unnecessary operations is always good. There is some extra checking per draw though.. wonder if there's a way to reduce that.

NOTE: This is draft because it doesn't avoid splitting render passes on OpenGL yet, in case the depth draw is not the first in a pass. That probably doesn't happen very often though.

EDIT: Regressions found during testing, check off once fixed:

- [ ] #7810 breaks again
- [ ] Some new issue with Kuroyuo

EDIT: There's problem with the pass injection. It's too aggressive, it should not happen when we are on the first draw on a new framebuffer.

### DIFF
--- a/Common/GPU/Vulkan/VulkanRenderManager.cpp
+++ b/Common/GPU/Vulkan/VulkanRenderManager.cpp
@@ -1054,7 +1054,17 @@ void VulkanRenderManager::CopyFramebuffer(VKRFramebuffer *src, VkRect2D srcRect,
 	_dbg_assert_msg_(dstPos.x + srcRect.extent.width <= (uint32_t)dst->width, "dstPos + extent x > width");
 	_dbg_assert_msg_(dstPos.y + srcRect.extent.height <= (uint32_t)dst->height, "dstPos + extent y > height");
 
-	for (int i = (int)steps_.size() - 1; i >= 0; i--) {
+	bool injectBeforeCurrent = false;
+	int lastStepToCheck = (int)steps_.size() - 1;
+	if (aspectMask == VK_IMAGE_ASPECT_DEPTH_BIT && curRenderStep_) {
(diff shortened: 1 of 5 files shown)


hrydgard:
The color-to-depth mode for Kuroyuo complicates this a lot. Before I get this in, I'm going to remove it in favor of a copying-based scheme instead (powered by the new sequence numbers).

hrydgard:
OK, I've decided that I'm gonna have to get this in together with the color-to-depth replacement. 

Here's a note to self about my progress figuring out Kuroyou.

There's an issue in the beginning of the scene that somehow wasn't a problem before, but is now:

* Drawing the background to 0000000/0040000 (alternating)
* Drawing the depth buffer using RGB565 to 88000 (in four strips)
* Texturing from 288000 in 565 format, drawing to 116400 (in four strips). This is black.
* Texturing from 116400 in 565 format, drawing to 88000 (same addr is bound as depth buffer, should be ignored).
* Finally, drawing the scene with depth testing to 0000000/0040000, with depth buffer at 88000.

The 565 bounce through 116400 of the depth image is presumably to swizzle the color data into the standard depth buffer ordering. We misinterpret this as depth texturing, causing our nice depth-image to turn black.

Then other shenanigans happen later in the rendering.

hrydgard:
Closing in favor of #15858

PR #15789, opened 2022-08-05
Unify the framebuffer depal process

hrydgard:
"Depal" is when we take a rendered framebuffer and treat is as a paletted texture (CLUT). PSP games use this to do all kinds of tricks, so we need to reinterpret the colors as what the PSP expects to see, turn in into an index into the palette and do the lookup on the PC GPU. There are two modes, one that converts whole textures on-demand, another that does the lookup in the fragment shader per-pixel, which can be faster in some cases.

We previously had separate implementations per backend. This merges them all into one that uses thin3d, greatly reducing the amount of code and makes changes quicker and easier.

This was also the final user of the Vulkan2D utility library, which is deleted here, too.

This works on all backends now, and while not working in all games on D3D9, neither did the old implementation, so that's an assignment for another day.

Ooh, close to -2000 lines. That's not bad.


hrydgard:
What remains here is that the D3D backend's "DrawEngine"s are doing resource binding and state management a bit backwards compared to the others, which previously worked with depal + shader blending simultaneously due to careful non-overlapping texture bindings, but that gets harder with the refactor.

So before getting this in, I'm going to align the DrawEngine's closer in a different PR. Then we can finally get this in, which will make it possible to work on some of the really tricky problems like Silent Hill shadows, and the Ratchet & Clank z buffer problems etc, without having to duplicate a lot of work.

hrydgard:
Alright, this should be ready to go now. I'm gonna do a bit of extra testing tomorrow and then get it in.

PR #15793, opened 2022-08-05
Fix logging warning issue.

anr2me:
Fixes some of the logging warnings mentioned at https://github.com/hrydgard/ppsspp/issues/15792

But it still have some warnings i couldn't figured out what was the main issue, like why `%s` was marked for warning while `Memory::GetCharPointer(communicationIdPtr)` returns a `const char*` which should be valid for `%s`.

Anyone know what was the main cause?

### DIFF
--- a/Common/StringUtils.cpp
+++ b/Common/StringUtils.cpp
@@ -55,6 +55,10 @@ void truncate_cpy(char *dest, size_t destSize, const char *src) {
 	}
 }
 
+const char* safe_string(const char* s) {
+	return s ? s : "(null)";
+}
+
 long parseHexLong(std::string s) {
 	long value = 0;
 
--- a/Common/StringUtils.h
+++ b/Common/StringUtils.h
@@ -84,6 +84,8 @@ inline void truncate_cpy(char(&out)[Count], const char *src) {
 	truncate_cpy(out, Count, src);
 }
 
+const char* safe_string(const char* s);
+
 long parseHexLong(std::string s);
 long parseLong(std::string s);
 std::string StringFromFormat(const char* format, ...);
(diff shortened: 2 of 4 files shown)


fp64:
Well, GCC specifically complains that `warning: ‘%s’ directive argument is null`, which `GetCharPointer` *can* return.
Since the full error message mentions `WrapI_UUUI`, I suspect that [L551](https://github.com/hrydgard/ppsspp/blob/master/Core/HLE/sceNp2.cpp#L551) may be the problem.

You can probably use a wrapper along the lines of:
```c
const char *safe_string(const char *s)
{
    return s?s:"(null)";
}
```
to avoid it (GCC's implementation of `printf` already does something similar).

It's not impossible for it to be a false positive, too.

hrydgard:
I thought all printfs automatically replaced nulls with the string (null) or similar anyway... Maybe not standard?

fp64:
Yes, that's not standard. The code `printf("%s",(const char*)0);` invokes undefined behaviour.
I think I have seen implementations that just crash on that.

ANR2ME:
Yeah, looks like feeding `NULL` to %s could cause segmentation fault on Linux, while VS replaces it with `"(null)"` https://stackoverflow.com/questions/11589342/what-is-the-behavior-of-printing-null-with-printfs-s-specifier

ANR2ME:
Btw, since `Memory::GetCharPointer` can be used on various places, where should i put the `const char *safe_string(const char *s)` function? so it can be easily accessed too from those places, should i put it in StringUtils.cpp ?

ANR2ME:
Okay, this removes all the warnings.

fp64:
There are couple more places where the result of `GetCharPointer` is passed to formatted printing. Maybe it makes sense to stick `safe_string` on those too? Though probably in a separate commit.

ANR2ME:
> There are couple more places where the result of `GetCharPointer` is passed to formatted printing. Maybe it makes sense to stick `safe_string` on those too? Though probably in a separate commit.

but i didn't saw any other warning when i did `./b.sh`, should i deleted the `build` folder first in order for all of them to be shown?

PR #15835, opened 2022-08-13
[Apple] Better get property

Halo-Michael:
1. Now, we no longer need to judge if jit is usable or not by according to the ios version. so we no longer need to defind g_jitAvailable, g_iosVersionMajor and g_iosVersionMinor. Also, we can display the version number in more detail like 15.3.1, no longer just showing 15.3.
2. return the currect SYSPROP_LANGREGION, not just return "en_US".


Halo-Michael:
Also, I would like to add some details on getting SYSPROP_LANGREGION on macOS:

Halo-Michael:
On macOS, you can only get LC_MESSAGES environment on terminal but not application, so PPSSPP on macOS will always return empty(en_US).
E8663BF3-676D-4865-BDAA-7B942F124E35
So we need find another way to get this on macOS:
Way first: do the same on iOS, but it's objc code, we can't use it in cpp file;
Way second: Use CoreFoundation:
58825EA9-5905-4922-94DE-CCFE13F05D1C
[main.txt](https://github.com/hrydgard/ppsspp/files/9331175/main.txt)
But it looks like the compiler doesn't like me inserting it directly into SDLMain.cpp:
018BAE85-14C7-4E94-9072-27A1BB781F21
Clang will always return errors:
A1EED2C8-076A-4AEB-90BE-9CA34161D331
Since I don't know much about CMake, I don't know how I can fix these errors.

ios/main.mm
```diff
@@ -92,16 +92,20 @@ kern_return_t catch_exception_raise(mach_port_t exception_port,
 static float g_safeInsetTop = 0.0;
 static float g_safeInsetBottom = 0.0;
 
+// We no longer need to judge if jit is usable or not by according to the ios version.
```
hrydgard, line 95: Why, because there no longer are any relevant versions that allow JIT?

Halo-Michael: We do not acturally have such code now

Halo-Michael: Now we only use get_debugged()

hrydgard: Ah right. I haven't really been keeping up with the iOS changes too much.

Issue #13773, opened 2020-12-16
Coded Arms Contagion music stops playing in few seconds


Panderner:
There's a bug where Coded Arms contagion music stops playing immediately in few seconds:

Here's a Log:
```
FMOD Streame E[ME]: HLE/sceAtrac.cpp:571 avcodec_decode_audio4: Error decoding audio -1094995529 / bebbb1b7
```

After Stopped playing music. The music will start playing again.

Panderner:
Here's a video:
[Screenrecorder-2020-12-16-14-14-47-421.mp4.zip](https://github.com/hrydgard/ppsspp/files/5700683/Screenrecorder-2020-12-16-14-14-47-421.mp4.zip)

sum2012:
v1.10.3-1330-g73da378ef log:
https://gist.github.com/sum2012/2261ff5a3b769a17d9864cd98d6bcff9
This kind of thing I only know hack,don't know correct fix

sum2012:
Real PSP log: 
https://gist.github.com/sum2012/c2d7bbc87594f88f00d8a13b39126ba4

PPSSPP log: 
34:17:523 FMOD Streame D[ME]: hle\sceatrac.cpp:1318 00000000=sceAtracDecodeData(2, 08472b84, 09fafa18[00000400], 09fafa1c[00000000], 09fafa10[175])
34:17:523 FMOD Streame D[ME]: hle\sceatrac.cpp:1587 0=sceAtracGetRemainFrame(2, 09fafa14[000000af])
34:17:523 FMOD Streame D[ME]: hle\sceatrac.cpp:1664 0=sceAtracGetStreamDataInfo(2, 09fafa00[084863f0], 09fafa0c[00010770], 09fafa04[000317a0])
34:17:523 FMOD Streame D[ME]: hle\sceatrac.cpp:1138 0=sceAtracAddStreamData(2, 00010770)
34:17:523 FMOD Streame I[ME]: hw\mediaengine.cpp:87 FF: Sound Unit id != 0x28.
34:17:523 FMOD Streame I[ME]: hw\mediaengine.cpp:87 FF: Frame decoding error!
34:17:523 FMOD Streame E[ME]: hle\sceatrac.cpp:571 avcodec_decode_audio4: Error decoding audio -1094995529 / bebbb1b7
[...]

Issue #11586, opened 2018-11-25
Jackass - The Game (USA) music and reset problems


benderscruffy:
choose any game then when you start it the music just keeps on repeating one note and then when you choose to exit out of the game and back to the main menu it freezes up on the loading screen
[ppsspplog.zip](https://github.com/hrydgard/ppsspp/files/2612770/ppsspplog.zip)

ResurrectionRedmi9:
Related https://github.com/hrydgard/ppsspp/issues/13676

Catarax:
In-game music still looping in 1.16.1

FATCatAndroid12:
Try the new sceAtrac maybe this can fixed
https://github.com/hrydgard/ppsspp/pull/20098

TAbdiukov:
This issue is possibly fixed, see: #12083

hrydgard:
This one is probably fixed though, yeah. Let's leave it closed.

FATCatAndroid12:
v1.19 milestone

TAbdiukov:
I manually confirmed today again, that the issue is fixed! For all of: US, EU, and unofficial translations of the game. Thanks again!

hrydgard:
Thanks for testing @TAbdiukov !

PR #15838, opened 2022-08-14
Add atrac looping hack for 4 games

sum2012:
helps https://github.com/hrydgard/ppsspp/issues/13773
(leave this to check) https://github.com/hrydgard/ppsspp/issues/7601
helps https://github.com/hrydgard/ppsspp/issues/11586
helps https://github.com/hrydgard/ppsspp/issues/10139
helps https://github.com/hrydgard/ppsspp/issues/12083
We have tried 7 years still do not get fixed.
So that I think that we can add in compat.ini

### DIFF
--- a/Core/Compatibility.cpp
+++ b/Core/Compatibility.cpp
@@ -96,6 +96,7 @@ void Compatibility::CheckSettings(IniFile &iniFile, const std::string &gameID) {
 	CheckSetting(iniFile, gameID, "MaliDepthStencilBugWorkaround", &flags_.MaliDepthStencilBugWorkaround);
 	CheckSetting(iniFile, gameID, "ZZT3SelectHack", &flags_.ZZT3SelectHack);
 	CheckSetting(iniFile, gameID, "AllowLargeFBTextureOffsets", &flags_.AllowLargeFBTextureOffsets);
+	CheckSetting(iniFile, gameID, "AtracLoopHack", &flags_.AtracLoopHack);
 }
 
 void Compatibility::CheckSetting(IniFile &iniFile, const std::string &gameID, const char *option, bool *flag) {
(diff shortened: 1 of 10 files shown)


LunaMoo:
I'd say as for hacks like that compat.ini are better than toggle since people will just activate it and forget breaking stuff in other games.

Gamemulatorer:
Yeah compat.ini is better and I already said that also 😅 https://github.com/hrydgard/ppsspp/pull/13793#issuecomment-1009578212

hrydgard:
I changed "fix" to "helps" so that the issues don't automatically close. I still think this should be considered a temporary measure, and hopefully someone will come up with a real fix at some point...

AryzenRS:
> (leave this to check) #7601

Just tried it, it loops in the wrong part of the song. About 8 bars in for the title menu. If left for long enough it will start looping a different part of the song.

sum2012:
The atrac need to re-write by someone to fix.
#15838
> > (leave this to check) #7601
> 
> Just tried it, it loops in the wrong part of the song. About 8 bars in for the title menu. If left for long enough it will start looping a different part of the song.

Catarax:
Stable 1.14 (Windows)
Jackass (Europe), the music still loop every 2 seconds, and then you get no sound on the next game. Emulator crash on exit.
The main menu music also stop all the music after 2 seconds of listen.

sum2012:
please download newest version in
https://buildbot.orphis.net/ppsspp/index.php

在 2022年12月16日週五 上午12:22，Catarax ***@***.***> 寫道：

sum2012:
Let me check

FATCatAndroid12:
This hack atrac loop can be remove now because https://github.com/hrydgard/ppsspp/pull/20098 is fixes the 4 games that needed this compat?

hrydgard:
@FATCatAndroid12 nah, the old implementation will still be used for old savestates, so the hacks stay (but will not be enabled anymore for new game bootups)

PR #15840, opened 2022-08-14
Hack to Fix Earth Saver Plus font

sum2012:
Work around #11055


Panderner:
Wrong font using real PSP font
NPJH00052_00001

Correct font here:
https://github.com/hrydgard/ppsspp/issues/11055#issuecomment-678722585

sum2012:
@hrydgard how about this ?

sum2012:
@hrydgard This pr don't include v1.13.2 release ?

hrydgard:
Not really sure what do with this but thanks to being a compat flag, while ugly, it's at least very safe. So I'm leaning towards getting it in.

PR #15847, opened 2022-08-16
Remove storage fileloader and storage folderbrowser, also re-add bin and elf [UWP]

tunip3:
This PR removes the storage fileloader and storage folderbrowser from the UWP port as they are no longer needed.


tunip3:
Updated to add bin and elf back to the file picker as they now work.
(tested with cube.elf on xbox)

hrydgard:
Thanks!

hrydgard (approved):

PR #15888, opened 2022-08-22
Copy color from overlapping framebuffers on bind, under certain conditions

### LINKED ISSUE #15728: Better ways to deal with overlapping render targets

hrydgard:
Leads to much faster performance in Juiced 2, instead of expanding the framebuffer to encompass both two overlapping framebuffers and relying on copies for self-blending. Similar to how we're handling depth copies.

This will later be expanded to handle more things in a more elegant way, like the framebuffer overlap in God of War for the shadows and color reinterpretation.

For now, keeps checking the same compatibility flag as the previous solution, but a later goal is to get rid of it. We currently do seems to get spurious copies from this in other games, one example being Wipeout where it seems to further break the already broken lens flare. However I think that one will need to be fixed properly on its own...

Fixes #15728

### DIFF
--- a/GPU/Common/FramebufferManagerCommon.cpp
+++ b/GPU/Common/FramebufferManagerCommon.cpp
@@ -190,7 +190,7 @@ void FramebufferManagerCommon::EstimateDrawingSize(u32 fb_address, GEBufferForma
 		// Unless the game is using overlapping buffers, the next buffer should be far enough away.
 		// This catches some cases where we can know this.
 		// Hmm.  The problem is that we could only catch it for the first of two buffers...
-		const u32 bpp = fb_format == GE_FORMAT_8888 ? 4 : 2;
+		const u32 bpp = BufferFormatBytesPerPixel(fb_format);
 		int avail_height = (nearest_address - fb_address) / (fb_stride * bpp);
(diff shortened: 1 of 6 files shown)


hrydgard:
This is pretty safe, I'm just gonna get it in.

Gamemulatorer:
GOW Chain Of Olympus shadow is expanded.
Screenshot_2022-08-27-12-35-11-065_org ppsspp ppsspp

hrydgard:
@Gamemulatorer Thanks for reporting, I'll get it fixed again soon.

PR #15948, opened 2022-09-03
Remove setting: Slow framebuf effects

hrydgard:
We want to have as few settings as possible that break graphics, and they better provide a major speedup if they do.

This one doesn't meet the cut, especially now that the really slow parts have been moved behind the "Enable block transfer" setting.

If some games get too slow on some device with this, I'm willing to look into more specific workarounds. But blanket settings like this that generally do not provide much of a speed boost, well, they just cause too much trouble.


Panderner:
The bad news is some games that needs to turn on Disable slower effects fix issues like Indiana Jones and Call of Duty for older Mali GPUs (https://github.com/hrydgard/ppsspp/issues/11598 and https://github.com/hrydgard/ppsspp/issues/11573)

hrydgard:
True, but I'd rather go investigate and fix the root cause of those problems instead, removing workarounds like this will kinda force me to :) 

If games break temporarily on old hardware, that will have to be OK as long as I fix it for the next major release.

PR #16026, opened 2022-09-14
OpenXR - Ensure the head pose is locked during frame rendering

lvonasek:
There is an issue that during rendering one frame multiple head positions are used. This PR ensures there is just one head pose in a single rendered frame.

### DIFF
--- a/Common/GPU/OpenGL/GLRenderManager.cpp
+++ b/Common/GPU/OpenGL/GLRenderManager.cpp
@@ -236,6 +236,9 @@ bool GLRenderManager::ThreadFrame() {
 			INFO_LOG(G3D, "Running first frame (%d)", threadFrame_);
 			firstFrame = false;
 		}
+
+		// Creation of OpenXR frame. This updates user's head pose and VR timestamps.
+		// For fluent rendering, the time between PreVRRender and PostVRRender must be really short.
 		if (IsVRBuild() && !vrlock) {
 			if (PreVRRender()) {
 				vrlock = true;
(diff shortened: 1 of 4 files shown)


Common/GPU/OpenGL/GLRenderManager.cpp
```diff
@@ -200,6 +200,7 @@ bool GLRenderManager::ThreadFrame() {
 	std::unique_lock<std::mutex> lock(mutex_);
 	if (!run_)
 		return false;
+	bool vrlock = false;
```
hrydgard, line 203: I don't really understand why this flag fixes a problem. 

If the issue is that PreVRRender gets called multiple times in a frame, can you just lift it out of the do {} loop below?

lvonasek: `PreVRRender()` updates timestamps and the head pose. If I move it before the `do` loop then OpenXR calculates the waiting loop (inside of `do` loop) into rendering time. It ends with very shuttery rendering.

hrydgard: Okay. I'll merge this if you add a clear comment explaining it, because it's really not very obvious when just looking at the code.

lvonasek: done

Common/GPU/OpenGL/GLRenderManager.cpp
```diff
@@ -236,18 +236,24 @@ bool GLRenderManager::ThreadFrame() {
 			INFO_LOG(G3D, "Running first frame (%d)", threadFrame_);
 			firstFrame = false;
 		}
+
+		// Creation of OpenXR frame. This updates user's head pose and VR timestamps.
+		// For fluent rendering, the time between PreVRRender and PostVRRender must be really short.
```
hrydgard, line 241: This still doesn't explain the purpose of the vrlock variable. That's what I need you to explain clearly.

lvonasek: I think the variable name and also the method names were choosen bit incorrectly. I changed to make it clear.

hrydgard: Yeah, that's better.

hrydgard (approved):

PR #16089, opened 2022-09-23
[AdhocMatching] Fix assertion issue when playing Cars over public adhoc server.

anr2me:
This fixes the assertion where current thread getting `nullptr` when playing Cars (ULUS10073) over public adhoc server.
image

PS: It's strange that i can only reproduce this when using public adhoc server, localhost and VPN didn't have this issue according to @MojoJojoDojo , so it might be affected by the latency between 1P -> AdhocServer -> 2P, since localhost and VPN use built-in AdhocServer that have minimum latency on one of them (eg. similar to 1P -> 2P), compared to a distant public AdhocServer.


hrydgard:
Sorry, forgot about it somehow!

PR #16116, opened 2022-09-26
Fix color test.

hrydgard:
Fixes the new color test bug reported in #13324, though doesn't fix that issue (didn't confirm it still is one).


unknownbrackets (approved):
Oops, the case I was looking at was against zero (I disabled the special against zero opt) so I forgot it was prescaled to 255.

-[Unknown]

hrydgard:
Ah, makes sense.

PR #16125, opened 2022-09-28
Remove alpha ignore in xbrz texture shaders.

stuken:
Gpu scaling ignores texture alpha and this is resulting in some upscaled but unfiltered textures. This tends to be mostly hud items but some ui elements can have a mix of filtered and unfiltered elements and it can be quite jarring. Just a small change to roll back an earlier commit that unifies the behaviour accross the two implementations.

Master:
UCES00422_00002
PR:
UCES00422_00001

Master CPU:
UCES00422_00000

Master:
ULES00111_00002
PR:
ULES00111_00001
Master CPU:
ULES00111_00000


hrydgard:
Yeah, looks good, let's have it this way.

PR #16154, opened 2022-10-03
Change int2hexstr method

sum2012:
After this merge,I will open another pull request for crc module's prx


sum2012:
I directly commit with the code

PR #16168, opened 2022-10-06
Shader uniforms (VK/D3D11): Fix issue where we could overwrite values in the fourth component padding.

hrydgard:
Several potential bugs here, things seems to have worked by some luck.

### DIFF
--- a/Common/Data/Convert/SmallDataConvert.h
+++ b/Common/Data/Convert/SmallDataConvert.h
@@ -82,6 +82,20 @@ inline void Uint8x3ToFloat4(float f[4], uint32_t u) {
 #endif
 }
 
+inline void Uint8x3ToFloat3(float f[4], uint32_t u) {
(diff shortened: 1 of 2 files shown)


Common/Data/Convert/SmallDataConvert.h
```diff
@@ -82,6 +82,20 @@ inline void Uint8x3ToFloat4(float f[4], uint32_t u) {
 #endif
 }
 
+inline void Uint8x3ToFloat3(float f[4], uint32_t u) {
+#if defined(_M_SSE) || PPSSPP_ARCH(ARM_NEON)
+	float temp[3];
```
iota97, line 87: Is this safe? Shouldn't it be `float temp[4]`?

hrydgard: good catch, thanks!

PR #16186, opened 2022-10-09
OpenXR - Initial work to support multiple render APIs

lvonasek:
This modifies the VR code to support multiple renderer APIs. This enables to integrate Vulkan API later.


hrydgard:
Cool to see you working on this! 

I think for the Vulkan stereo rendering, we really should try to make it work with the VK_KHR_multiview extension. 

I might take a crack at doing the groundwork in the next week, using it to implement red-green stereo, or something, then this PR can be adjusted to make use of that.

lvonasek:
> Cool to see you working on this!
> 
> I think for the Vulkan stereo rendering, we really should try to make it work with the VK_KHR_multiview extension.
> 
> I might take a crack at doing the groundwork in the next week, using it to implement red-green stereo, or something, then this PR can be adjusted to make use of that.

I am curious, if I make it working at all, I have no experience with Vulkan at all. If you want to help then you can translate the GL code in VRFramebuffer.cpp into Vulkan. There are todos on places where is it needed.

Multiview is something what I would really love to have working but I spent a lot of time on GL to make it working. The same multiview code works everywhere else but on PPSSPP it renders just black screen :(

hrydgard (approved):

PR #16197, opened 2022-10-10
More uniform optimization, fixes

hrydgard:
Removes another 16 bytes from the uniform structs, getting us down to 448 bytes.

More importantly, cleans up the situation where we used fragmentShaderInt32 support as a substitute for checking for bitwiseOps, though in practice they're likely to be the same (and should probably both be checked for some of these use cases...).

Also fixes running the shader compilation test in debug mode.


unknownbrackets (approved):
It seems like in GLES we checked that highp was actually >= 30 bits.  Do we need to worry about that (low precision uints) for any of these uniforms?

-[Unknown]

hrydgard:
Possibly, though in practice, I don't think it's an issue on any implementation that supports bitwise integer ops.

Issue #13845, opened 2020-12-31
[Wayland] Blank window with Vulkan backend


michelesr:
### What happens?

On Arch Linux, running `PPSSPPSDL` (v1.10.3) with `SLD_VIDEODRIVER=wayland` and Vulkan backend set in the config, the following message is printed and execution halts.

```
Vulkan subsystem 6 not supported
```

Full output:

```
I: VulkanLoader.cpp:268: VulkanMayBeAvailable: Device allowed ('SDL:Linux')
I: VulkanLoader.cpp:275: VulkanMayBeAvailable: Library loaded ('libvulkan.so')
I: VulkanLoader.cpp:328: VulkanMayBeAvailable: Enumerating instance extensions
I: VulkanLoader.cpp:339: VulkanMayBeAvailable: Instance extension count: 18
I: VulkanLoader.cpp:347: VulkanMaybeAvailable: Instance extension found: VK_KHR_device_group_creation (00000001)
I: VulkanLoader.cpp:347: VulkanMaybeAvailable: Instance extension found: VK_KHR_display (00000017)
I: VulkanLoader.cpp:347: VulkanMaybeAvailable: Instance extension found: VK_KHR_external_fence_capabilities (00000001)
I: VulkanLoader.cpp:347: VulkanMaybeAvailable: Instance extension found: VK_KHR_external_memory_capabilities (00000001)
I: VulkanLoader.cpp:347: VulkanMaybeAvailable: Instance extension found: VK_KHR_external_semaphore_capabilities (00000001)
I: VulkanLoader.cpp:347: VulkanMaybeAvailable: Instance extension found: VK_KHR_get_display_properties2 (00000001)
I: VulkanLoader.cpp:347: VulkanMaybeAvailable: Instance extension found: VK_KHR_get_physical_device_properties2 (00000001)
I: VulkanLoader.cpp:347: VulkanMaybeAvailable: Instance extension found: VK_KHR_get_surface_capabilities2 (00000001)
I: VulkanLoader.cpp:347: VulkanMaybeAvailable: Instance extension found: VK_KHR_surface (00000019)
I: VulkanLoader.cpp:347: VulkanMaybeAvailable: Instance extension found: VK_KHR_surface_protected_capabilities (00000001)
I: VulkanLoader.cpp:347: VulkanMaybeAvailable: Instance extension found: VK_KHR_wayland_surface (00000006)
[...]

PR #16219, opened 2022-10-15
Vulkan: fix wayland swapchain size 1x1

catsout:
From Vulkan spec,  
surfaceCapabilities.currentExtent can be 0xFFFFFFFF(-1)  

Define a callback to get drawsize from window creator  

relate:  
https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkSurfaceCapabilitiesKHR.html  
https://github.com/KhronosGroup/Vulkan-Docs/issues/590#issuecomment-347932321  

issue:  
https://github.com/hrydgard/ppsspp/issues/16163  
https://github.com/hrydgard/ppsspp/issues/13845


hrydgard (approved):
That's a good catch and a decent implementation. Let's get it in.

PR #16247, opened 2022-10-17
OpenXR - Code refactor

lvonasek:
In the future I would like to create from the VR code a submodule. `Common/VR/VR*` should be independent on other classes. `PPSSPPVR` is the connection between PPSSPP code and the future submodule.

Advantage of having such a submodule is that it could be easily used in other projects and it could help with the device support.

This PR is mainly getting rid of `VRTweaks` class which doesn't fit to the concept.
* Get rid of pseudo-math class `VRTweaks`
* Reduce uniform calls
* Remove two `#include "Core/.."` from `Common/`


hrydgard (approved):
I'll have to adjust https://github.com/hrydgard/ppsspp/pull/16246 a bit, but that's fine :)

I'm not sure about breaking it out in a submodule, but maybe it could work...

lvonasek:
Ok, sorry, I didn't see you start the big cleanup already.

PR #16255, opened 2022-10-18
SDL: fix glew on wayland by ignore glx

catsout:
glXGetProcAddress works on wayland if linking with x11.   
As long as we don't use glx related funcs, no_glx_error can be ignored.  
SDL should handle the no_glx_error brefore glewInit.  
Glew is only used to init gl funcs.

related:
https://github.com/nigels-com/glew/issues/172

issue:  
https://github.com/hrydgard/ppsspp/issues/16252  
https://github.com/hrydgard/ppsspp/issues/16237


gucio321 (approved):
I've tested and it started working with OPENGL backend on wayland

hrydgard:
Nice. That's really not a great API interface, heh.

PR #16262, opened 2022-10-20
OpenXR - Add an option to adjust camera distance

lvonasek:
A new option to adjust VR camera distance. VR settings reordered.


iota97:
Any reason for this not just being a float?

lvonasek:
> Any reason for this not just being a float?

UX. Floats are really hard to enter in VR.

hrydgard:
The solution to that should probably be to improve the UX around setting float values.

However, this is fine for now, can be improved later.

hrydgard (approved):

PR #16264, opened 2022-10-21
Update zh_CN.ini & Update zh_TW.ini 

leoxxx:
https://github.com/hrydgard/ppsspp/pull/15694
Translate "artifacts" to a straightaway meaning.
Change 2 words to TC.


hrydgard:
Thinking about changing that word in English too heh... It's a bit confusing.

leoxxx:
Then, how about change it to "map error"?

PR #16270, opened 2022-10-22
OpenXR - Refactor compatvr.ini file

lvonasek:
The previous compatvr.ini list was exported from PCVR version of PPSSPP. This updated one was manually reviewed to:
* Allow loading the file from the filesystem
* Add all known regions for supported games
* Group the game build IDs together

This makes much easier to maintain the list and it enables collaborations on extending it.


hrydgard:
This is nice. Ideally I guess we should find some format where we don't need to repeat the float values for every ID, but yeah, can think about that some other time...

I'm gonna squash-merge if you don't mind, that's a lot of commits.

hrydgard (approved):

PR #16310, opened 2022-10-31
OpenXR - Map controller gestures

lvonasek:
Use VR controller movements to generate an action which can be mapped in the controller mapping.


UI/GameSettingsScreen.cpp
```diff
@@ -1169,7 +1169,12 @@ void GameSettingsScreen::CreateViews() {
...
 		vrSettings->Add(new PopupSliderChoiceFloat(&g_Config.fCameraSide, -10.0f, 10.0f, vr->T("Camera side adjust", "Camera side adjust"), 1.0f, screenManager(), ""));
 		vrSettings->Add(new PopupSliderChoiceFloat(&g_Config.fCanvasDistance, 1.0f, 10.0f, vr->T("Distance to 2D menus and scenes", "Distance to 2D menus and scenes"), 1.0f, screenManager(), ""));
-		vrSettings->Add(new PopupSliderChoiceFloat(&g_Config.fFieldOfViewPercentage, 100.0f, 200.0f, vr->T("Field of view scale", "Headset's field of view scale"), 10.0f, screenManager(), vr->T("% of native FoV")));
+		vrSettings->Add(new PopupSliderChoiceFloat(&g_Config.fFieldOfViewPercentage, 100.0f, 200.0f, vr->T("Field of view scale", "Field of view scale"), 10.0f, screenManager(), vr->T("% of native FoV")));
+
+		vrSettings->Add(new ItemHeader(vr->T("VR controllers")));
+		vrSettings->Add(new CheckBox(&g_Config.bEnableMotions, vr->T("Map controller movements to keys")));
+		PopupSliderChoiceFloat *vrMotions = vrSettings->Add(new PopupSliderChoiceFloat(&g_Config.fMotionLength, 0.3f, 1.0f, vr->T("Motion needed to generate action", "Motion needed to generate action"), 0.1f, screenManager(), vr->T("m")));
```
iota97, line 1176: Not a big deal but when you wrap strings in vr->T(...) for translation if key and value are the same you may just pass 1 argument :)

lvonasek: Good point, thanks.

hrydgard:
This is pretty cool.

There's a minor UX issue where it's using the old message "Map a key for [control]", where in fact you're mapping a movement, not a key. On the other hand, we already have this problem with things like joystick movements, but maybe could be made more informative. Not gonna require changing it as part of this PR, just something to think about.

Common/Input/KeyCodes.h
```diff
@@ -261,7 +261,14 @@ typedef enum _keycode_t {
...
 	NKCODE_EXT_MOUSEWHEEL_UP = 1008,
-	NKCODE_EXT_MOUSEWHEEL_DOWN = 1009
+	NKCODE_EXT_MOUSEWHEEL_DOWN = 1009,
+
+	// Virtual reality controller motion
+	NKCODE_EXT_MOTION_UP = 1101,
+	NKCODE_EXT_MOTION_DOWN = 1102,
+	NKCODE_EXT_MOTION_LEFT = 1103,
+	NKCODE_EXT_MOTION_RIGHT = 1104,
+	NKCODE_EXT_MOTION_FORWARD = 1105,
```
hrydgard, line 271: Should we separate these for right vs left controller?

Common/VR/PPSSPPVR.cpp
```diff
@@ -201,6 +202,58 @@ void UpdateVRInput(bool(*NativeKey)(const KeyInput &key), bool(*NativeTouch)(con
...
 
+	//motion control
+	if (g_Config.bEnableMotions) {
+		for (int j = 0; j < 2; j++) {
+			bool activate;
+			float limit = g_Config.fMotionLength; //length of needed movement in meters
+			XrVector3f axis = {0, 1, 0};
+			float center = ToRadians((float)VR_GetConfig(VR_CONFIG_MENU_YAW));
+			XrQuaternionf orientation = XrQuaternionf_CreateFromVectorAngle(axis, center);
+			XrVector3f position = XrQuaternionf_Rotate(orientation, IN_VRGetPose(j).position);
```
hrydgard, line 213: When we're checking the position here against the limit, what is it relative too? Some kind of default hand position?

Common/Input/KeyCodes.h
lvonasek: That would require some code rewritting because the left and right controller are two different devices.

I find it cool that the gesture could be done by more devices.

Common/VR/PPSSPPVR.cpp
lvonasek: It is relative to the head position.

Common/Input/KeyCodes.h
hrydgard: Yeah you're right, it's better to keep one definition here, and to instead send different device code from left or right. Didn't think about that.

Common/VR/PPSSPPVR.cpp
```diff
@@ -201,6 +202,58 @@ void UpdateVRInput(bool(*NativeKey)(const KeyInput &key), bool(*NativeTouch)(con
 		}
 	}
 
+	//motion control
+	if (g_Config.bEnableMotions) {
```
hrydgard, line 206: Come to think of it, do we really need an enable option? Wouldn't this just not do anything if no motion codes are mapped? Sure there's a little bit of computation here that can be saved by really not a lot.

Common/VR/PPSSPPVR.cpp
hrydgard: Ah I see, that's why the other limit check is * 1.5, your controll base position is normally below the head... Seems to me like an offset first, and then use the same limit, would make more sense than multiplying one of the limits, but this stuff is easily tweakable (doensn't change semantics) so it's fine as-is.

Common/VR/PPSSPPVR.cpp
lvonasek: If people use a 3rd party controller then the movement could do some madness. This is there to avoid it.

hrydgard: Okay, I suppose. Just trying to minimize the amount of options, as always...

hrydgard (approved):

Issue #16317, opened 2022-11-01
Crash when clicking "Settings" in home screen and in game 


yaringal:
### Game or games this happens in

All

### What area of the game / PPSSPP

Settings menus doesn't load and instead freezes app

### What should happen

Should not crash

### Platform

Android

### Mobile phone model or graphics card

Pico 4

### PPSSPP version affected

v1.10 onwards

### Last working version

v1.9.4

### Graphics backend (3D API)

OpenGL / GLES

### Checklist

- [X] Test in the [latest git build](https://buildbot.orphis.net/ppsspp/) in case it's already fixed.
- [X] [Search for other reports](https://github.com/hrydgard/ppsspp/search?q=my+issue&type=issues) of the same issue.
- [X] Try resetting settings or older versions and include if the issue is related.
- [X] Try without any cheats and without loading any save states.
- [ ] Include logs or screenshots of issue.

lvonasek:
Hi, the versions you tested are very old and unlikely to work on Pico 4. I would recommend to try the latest version made for your headset:
https://drive.google.com/drive/folders/1GgyJqk6WTYGTFfQF6kojJMIHItdzrjaM?usp=sharing

lvonasek:
I guess something went wrong with the settings file. Could you try to delete ppsspp.ini from the filesystem and then open the latest VR version?

yaringal:
Yep I tried that (deleted the entire psp folder between testing different versions)

lvonasek:
Could you try this build?
https://drive.google.com/file/d/19iiE5W7AYsXCjI_VaCZl1xL6x4jRexf9/view?usp=sharing

yaringal:
I just tried the version you shared - it still crashes the same way when clicking Settings

lvonasek:
Ok thanks. I will try to fix it one more time. I might be able to share an APK tonight.

lvonasek:
I am sorry, I am not able to easily resolve it. I will take a look into this again, once I get Pico 4 myself (in about 2 weeks).

yaringal:
I can confirm this works now - many thanks!

PR #16318, opened 2022-11-01
OpenXR - File access fix

lvonasek:
New headsets (Pico 4, Quest Pro) use Android 11 + but the storage behavior is not according to Google specifications.

This PR forces Android 10 SDK for VR builds to avoid the storage issues. Both Meta and Pico do not require compiling apps under specific SDKs.

Resolving issue https://github.com/hrydgard/ppsspp/issues/16317

### DIFF
--- a/android/AndroidManifest.xml
+++ b/android/AndroidManifest.xml
@@ -47,9 +47,7 @@
         android:label="@string/app_name" 
         android:logo="@drawable/ic_banner"
         android:isGame="true"
-        android:banner="@drawable/tv_banner"
-        android:requestLegacyExternalStorage="true"
-        android:preserveLegacyExternalStorage="true">
+        android:banner="@drawable/tv_banner">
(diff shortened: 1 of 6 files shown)


hrydgard (commented):
Seems reasonable, as long as the request/preserve flags really survive the include process on Android 11, as we really don't want to change anything on regular Android.

lvonasek:
Thanks. Waiting for confirmation from a Pico 4 user and then I set it to "Ready for review".

hrydgard:
It would be interesting to get logs from the crash, too - ideally, we should at least handle such problem more gracefully if we come across weird platforms.

lvonasek:
I will get Pico 4 in about two weeks. I can then pick the stacktrace.

hrydgard:
Ah, very good.

lvonasek:
Ok, this didn't resolve the issue.

I will revisit the problem once I have Pico 4 in my hands.

PR #16329, opened 2022-11-03
OpenXR - Add an option to disable 72Hz update

lvonasek:
Add an option to disable 72Hz update.

72Hz update is needed to resolve laggy VR experience on Oculus Quest. In non-VR and on Pico headsets this problem doesn't exist. And I found that in SImpsons cutscenes 72Hz makes audio/video out of sync.


hrydgard:
Quest 2 also supports 120hz, we might be able to make good use of that by modifying the "duplicate frames to 60hz" setting..

lvonasek:
Yes but only as an option. It would double the intensity of lagging on framedrop.

PR #16333, opened 2022-11-05
OpenXR - Defines cleanup

lvonasek:
This PR cleans up Common/VR from ugly ifdefs. This helps with:
* Controller support
* Handling platform specific code
* PCVR porting
* The code will be always compiled


hrydgard (approved):
Looks alright to me! Maybe the flags should be named VR_PLATFORM_FLAG_* instead of just VR_PLATFORM, but in the context they're used it's pretty obvious anyway.

lvonasek:
I would prefer to keep it short to prevent too long lines in the code.

PR #16336, opened 2022-11-05
Support Android file intents with up to 3 dots in path

lvonasek:
The file intent doesn't support paths with more than one dot in path. That means e.g. this path is not resolved (because of a dot in version number): `file:///storage/emulated/0/PSP/GTA - Chinatown Wars (v1.01).iso`

This PR adds support up to three dots in the path.

### DIFF
--- a/android/AndroidManifest.xml
+++ b/android/AndroidManifest.xml
@@ -70,37 +70,28 @@
 				
 				<category android:name="android.intent.category.DEFAULT" />
 				<category android:name="android.intent.category.BROWSABLE" />
-				
-				<data
-					android:host="*"
-					android:mimeType="*/*"
-					android:pathPattern=".*\\.iso"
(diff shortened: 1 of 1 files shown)


hrydgard:
Wow, that's an odd one!

Sure we don't just have bad syntax in our filters? Having to duplicate the filters like this seems not obvious to be the best way...

lvonasek:
There is not a better way to solve it as it isn't a proper regex:
https://developer.android.com/guide/topics/manifest/data-element#path

lvonasek:
Note that this way of opening files is pretty deprecated (the launcher app has to be compiled with API 23 to make it working).

The new way gives you Inputstream instead of path. So probably a no-go.

However in VR in my launcher it looks and work pretty well:
IMG_20221105_215528_636

PR #16343, opened 2022-11-06
VR: Separate float parameters, to avoid ugly/dangerous scale hacks

hrydgard:
Some of these dropped important precision on the floor, too, I think - but also a bit afraid of some of them overflowing an int with the heavy scaling.

~~I think I made a mistake somewhere though because 6dof doesn't seem to be working correctly..~~


Common/VR/VRRenderer.cpp
```diff
@@ -338,37 +342,37 @@ bool VR_InitFrame( engine_t* engine ) {
 			}
 
 			vrMatrix[matrix] = ovrMatrix4f_CreateFromQuaternion(&invView.orientation);
-			float scale = (float)VR_GetConfig(VR_CONFIG_6DOF_SCALE) * 0.000001f;
+			float scale = VR_GetConfigFloat(VR_CONFIG_6DOF_SCALE) * 0.000001f;
```
hrydgard, line 345: ok, think I found my mistake

lvonasek (approved):
LGTM

PR #16345, opened 2022-11-06
Remove "hardware" skinning

hrydgard:
After this, we only skin directly during decode. Much simpler, and is fast enough. "Hardware" (vertex shader skinning) didn't make good use of the hardware due to OpenGL limitations, and it's not really worth making a faster Vulkan implementation since skinning is never really a bottleneck these days.

Part of option cleanup - and as a benefit, lets us get rid of quite a large amount of code.

Will only reconsider if someone can show a popular device with a game that runs noticeably faster with hardware skinning than software skinning, which I think is unlikely.


Gamemulatorer:
This will affect bleach the heat soul 7 and monster hunter games in my mali 450 gpu https://github.com/hrydgard/ppsspp/issues/10775

hrydgard:
Oh right, we tried this once before, I forgot...

hrydgard:
Just weird to me that it makes such a difference in those games. Wonder if anything has changed....

Gamemulatorer:
> Just weird to me that it makes such a difference in those games. Wonder if anything has changed....

I will test again using the recently build on my mali 450 gpu phone.

Gamemulatorer:
Only 5+ FPS is gain using hardware skinning in recently build 😅

HW
Screenshot_2022-11-07-00-34-20

SW
Screenshot_2022-11-07-00-32-59

hrydgard:
Well, still noticeable for sure. Although it'll only affect some games, and only one some old phones... Undecided, heh.

libretro/libretro.cpp
```diff
@@ -823,15 +823,6 @@ static void check_variables(CoreParameter &coreParam)
          g_Config.bHardwareTransform = true;
    }
 
-   var.key = "ppsspp_software_skinning";
```
unknownbrackets, line 826: Hm, this doesn't actually remove the option in the other file I think?

-[Unknown]

Gamemulatorer:
Bleach The Heat Soul 7 on my vivo y51a adreno 306 is also benifit in HW Skinning.

SW
Screenshot_20221107_010406

HW
Screenshot_20221107_010435

GPU/GPUCommon.cpp
```diff
@@ -66,8 +66,8 @@ const CommonCommandTableEntry commonCommandTable[] = {
 	{ GE_CMD_BEZIER, FLAG_EXECUTE, 0, &GPUCommon::Execute_Bezier },
 	{ GE_CMD_SPLINE, FLAG_EXECUTE, 0, &GPUCommon::Execute_Spline },
 
-	// Changing the vertex type requires us to flush.
-	{ GE_CMD_VERTEXTYPE, FLAG_FLUSHBEFOREONCHANGE | FLAG_EXECUTEONCHANGE, 0, &GPUCommon::Execute_VertexType },
+	// Changing the vertex type requires us to flush, unless we just change the weight count - need to handle in a func.
+	{ GE_CMD_VERTEXTYPE, FLAG_EXECUTEONCHANGE, 0, &GPUCommon::Execute_VertexTypeSkinning },
```
unknownbrackets, line 70: nit: We should probably just rename it.

-[Unknown]

GPU/Common/VertexDecoderCommon.h
```diff
@@ -300,7 +300,7 @@ class VertexReader {
 	bool hasNormal() const { return decFmt_.nrmfmt != 0; }
 	bool hasUV() const { return decFmt_.uvfmt != 0; }
 	bool isThrough() const { return (vtype_ & GE_VTYPE_THROUGH) != 0; }
-	bool skinningEnabled() const { return vertTypeIsSkinningEnabled(vtype_); }
+	bool skinningEnabled() const { return false; }
```
unknownbrackets, line 303: Well, these can be removed (`numBoneWeights()` too), which I did in the half-step pull.  Software uses them.

-[Unknown]

unknownbrackets (approved):

unknownbrackets:
Hm, 25% is a lot.  I feel like that's probably a case where it's using the same bones for a lot of (maybe offscreen?) drawing, which is the best case for hardware skinning I can imagine.

I wonder if there's a world in which we could decode using threads (for large draw calls, if that game even uses large ones) to close the gap?

I like the binding thing, though.

-[Unknown]

GPU/Common/SoftwareTransformCommon.cpp
```diff
@@ -168,6 +168,13 @@ void SoftwareTransform::SetProjMatrix(float mtx[14], bool invertedX, bool invert
 	projMatrix_.translateAndScale(trans, scale);
 }
 
+static void ReadWeightedNormal(Vec3f &source, VertexReader &reader, u32 vertType) {
```
unknownbrackets, line 171: I guess we can still have this func, but don't really need it now.  If we do keep, we should probably use it for the initial read too.

-[Unknown]

unknownbrackets (approved):

iota97:
Just to note MH games probably show the most difference in the village area that have a lot of NPC with bones. That area is just a management hub so not really terrible if it lag a bit on old device.

AFAIK on my old phone HW was 30% slower than SW tho' (but many many build ago).

Issue #16351, opened 2022-11-06
Vulkan validation layers just got pickier about input attachments


hrydgard:
Reminder to self for tomorrow.

We fail the new check https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/7acdcd6304b34aac41ef0b0e4bffaf8b7849aca4 (which is in the latest SDK validation layers), since we end up binding a different view of the same image as the input attachment and the descriptor, which is no longer ok. Shouldn't be too hard to fix. (VKRImage rtView vs texAllLayersView, I believe).

hrydgard:
I've fixed what I can, and then forcibly disabled these checks for now, and Hans-Kristian agrees with me that the validation (and spec rule) isn't right - so closing, nothing to do other than waiting for updated validation layers.

PR #16354, opened 2022-11-07
Vulkan: Use the very same view as input attachment and color attachment, not just the same image

hrydgard:
Fixes most of the new validation errors seen in #16351, though one remains.

I believe that to be a bug in the validation layers, will investigate later.


hrydgard:
Hm, still something weird going on..

hrydgard:
Oh, I see. The validator wants every draw to pass the input attachment in a descriptor, regardless of whether it's actually used in that draw, or not. That's a bit annoying and I think overzealous - I'm gonna post a question in the validator repo.

PR #16365, opened 2022-11-08
Fix GLSL compilation issues on really old Adreno drivers.

hrydgard:
They don't seem to accept unsigned integers as switch cases.

Fixes some stuff in #16295 , not sure if all of it.

### DIFF
--- a/GPU/Common/VertexShaderGenerator.cpp
+++ b/GPU/Common/VertexShaderGenerator.cpp
@@ -1054,11 +1054,11 @@ bool GenerateVertexShader(const VShaderID &id, char *buffer, const ShaderLanguag
 				p.C("        ldot = pow(max(ldot, 0.0), u_matspecular.a);\n");
 				p.C("      }\n");
 				p.C("    }\n");
-				p.C("    switch (type) {\n");  // Attenuation
-				p.C("    case 1u:\n");  // GE_LIGHTTYPE_POINT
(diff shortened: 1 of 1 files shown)


unknownbrackets (approved):

Gamemulatorer:
Yes this fixes #16295
Screenshot_20221109_125911

Gamemulatorer:
But not this https://github.com/hrydgard/ppsspp/issues/16295#issuecomment-1307804512
might be a separate issue.

unknownbrackets:
Do you get any shader compilation error for that one?

-[Unknown]

Gamemulatorer:
> Do you get any shader compilation error for that one?
> 
> -[Unknown]

Nope

Gamemulatorer:
This is how to reproduce the issue
https://drive.google.com/file/d/1-7VhP0kEpwoYNR4VJNSkYtDwuoL5GK6K/view?usp=drivesdk

Both happen in opengl/vulkan mali 450 gpu, adreno 306 and 505

hrydgard:
Oh, I think that's a result of removing some fallback code we had, in #16347. Should be an easy fix (force software skinning when hardware transform is disabled).

unknownbrackets:
Oops, my mistake - #16369 should handle the cases of fallback.

-[Unknown]

Issue #16244, opened 2022-10-17
VR user experience problems


hrydgard:
Alright, I've finally tried Lubos' VR code on a Quest 2, and what a trip! It's rough but still working surprisingly well, obviously with a lot of glitches and stuff but it sure adds a new dimension to some of these games.

There's a bunch of UX things I'd really like to fix over time, I'm going to start listing them here.

* We need to have Pause mapped by default to a button, so you can get to the menu. Maybe click-right-stick on Quest, not sure what other similar apps would use.
* I sometimes had problems getting a cursor working to point at things, though of course arrow navigation through PPSSPP's menus still worked.
* Control mappings need to be displayed somehow, or a quick way to access them from the pause menu
* "Known-good" games could be highlighted in a special way in the game selector, maybe
* PPSSPP's UI is workable in VR but not ideal. The square dimensions look odd, but don't matter that much, but it might make sense to restrict to a smaller rectangle.

PR #16394, opened 2022-11-17
OpenXR - UX improvements

lvonasek:
Resolves a few points from https://github.com/hrydgard/ppsspp/issues/16244.

* Allow higher canvas distance
* Camera adjust using external controller fixed
* Controller support feels more native now
* Do not show cursor in game mode
* Force non-VR rendering when dialog shown
* Menu with PSP aspect ratio
* Prevent misclicks when swtiching left/right hand cursor


Common/VR/PPSSPPVR.h
```diff
@@ -17,13 +17,25 @@ enum VRCompatFlag {
 	VR_COMPAT_MAX
 };
 
+enum VRAppMode {
+	VR_CONTROLL_MAPPING_MODE,
```
hrydgard, line 21: nitpicking: CONTROL not CONTROLL

hrydgard (changes requested):

PR #16410, opened 2022-11-21
Move AsyncImageFileView to Common

hrydgard:
Required some shuffling around of stuff, like also moving ManagedTexture. Just a minor cleanup.

Also had to move GameIconView somewhere (from ManagedTexture.h), it ended up in SaveDataScreen. Should possibly also have its own file, but meh.

### DIFF
--- a/UI/PauseScreen.h
+++ b/UI/PauseScreen.h
@@ -24,7 +24,11 @@
 #include "Common/UI/UIScreen.h"
 #include "Common/UI/ViewGroup.h"
 #include "UI/MiscScreens.h"
-#include "UI/TextureUtil.h"
+
+enum class PauseScreenMode {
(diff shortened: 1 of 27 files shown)


UI/PauseScreen.h
```diff
@@ -24,7 +24,11 @@
 #include "Common/UI/UIScreen.h"
 #include "Common/UI/ViewGroup.h"
 #include "UI/MiscScreens.h"
-#include "UI/TextureUtil.h"
+
+enum class PauseScreenMode {
+	MAIN,
+	DISPLAY_SETTINGS,
+};
```
hrydgard, line 31: self-review: oops, this is intended for a later PR, but meh, doesn't hurt.

PR #16412, opened 2022-11-21
Display edit mode - initial prototype

hrydgard:
Currently contains some commits from other PRs, will rebase when #16409 is in.

This adds a "Display settings" button to the in-game pause screen, which switches it to a mode where we can see the game screen fully, and (currently) just change the aspect ratio between Stretch and the other modes. More controls will be added.

This will eventually be able to fully replace the somewhat confusing "Display layout editor".

Transparency notes:

Screen stack examples:

* Popup - transparent, visible
* Graphics settings - transparent, visible
* Pause screen - transparent, not visible
* Emu screen - opaque, visible
* Main screen - opaque, not visible

Example 2:

* Popup - transparent, visible
* Pause screen - transparent, visible
* Emu screen - opaque, visible
* Main screen - opaque, not visible

I think the rules are, for the background visibility I want:

* Popups don't affect transparency of below. Only a single non-popup transparent screen is drawn, on top of the first opaque screen.


hrydgard:
I have a better plan now than these icky modes. Closing.

PR #16414, opened 2022-11-22
Pause menu background: Remove complicated transparency tricks

hrydgard:
Instead draw game as part of background.

This makes the game visible through many menu screens if in-game, not just the pause screen.


iota97:
This doesn't yet works for all screens, things like controls mapping (and many other) are not subclass of `UIDialogScreenWithGameBackground` but of `UIDialogScreenWithBackground`.

hrydgard:
Yeah, I know. I'll expand support.

PR #16418, opened 2022-11-23
(Libretro) Add VS 2022 Build Support

ghost:
This attempts to allow the libretro core to be built with Visual Studio 2022, while also adding more specific instructions on what build commands to use depending on your version of Visual Studio. The core built just fine, and it also worked perfectly when tested under RetroArch. Although I attached a test build in case I missed something.

However, since VS 2022 now installs in the `C:/Program Files` directory instead of the x86 folder older versions used, I had some difficulties with getting the path to be recognized. I'm not sure if the current solution is optimal, hence the draft status. Any feedback is certainly welcomed here.

Test build: [ppsspp_libretro.zip](https://github.com/hrydgard/ppsspp/files/10072546/ppsspp_libretro.zip)

### DIFF
--- a/libretro/Makefile
+++ b/libretro/Makefile
@@ -533,6 +533,115 @@ else ifneq (,$(findstring windows_msvc2019,$(platform)))
 	GL_LIB := opengl32.lib
 	LDFLAGS += ws2_32.lib user32.lib shell32.lib avcodec.lib avutil.lib swresample.lib swscale.lib avformat.lib advapi32.lib winmm.lib gdi32.lib d3d9.lib d3dx9.lib Iphlpapi.lib
 
+# Windows MSVC 2022 all architectures
+else ifneq (,$(findstring windows_msvc2022,$(platform)))
+
+	PlatformSuffix = $(subst windows_msvc2022_,,$(platform))
+	ifneq (,$(findstring desktop,$(PlatformSuffix)))
+		WinPartition = desktop
+		MSVC2019CompileFlags = -DWINAPI_FAMILY=WINAPI_FAMILY_DESKTOP_APP -D_UNICODE -DUNICODE -DWINVER=0x0600 -D_WIN32_WINNT=0x0600
+		LDFLAGS += -MANIFEST -NXCOMPAT -DYNAMICBASE -DEBUG -OPT:REF -INCREMENTAL:NO -SUBSYSTEM:WINDOWS -MANIFESTUAC:"level='asInvoker' uiAccess='false'" -OPT:ICF -ERRORREPORT:PROMPT -NOLOGO -TLBID:1
+		LIBS += kernel32.lib user32.lib gdi32.lib winspool.lib comdlg32.lib advapi32.lib
+	else ifneq (,$(findstring uwp,$(PlatformSuffix)))
+		WinPartition = uwp
+		MSVC2019CompileFlags = -DWINAPI_FAMILY=WINAPI_FAMILY_APP -DWINDLL -D_UNICODE -DUNICODE -DWRL_NO_DEFAULT_LIB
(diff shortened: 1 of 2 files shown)


hrydgard:
I don't mind it, though I guess it would be nice if the various vs configs in the file could share more of their setup, as they are pretty similar.

ANR2ME:
I think instead of assuming VS2022 to always be installed on it's default directory, you should retrieve the `InstallDir` from registry https://stackoverflow.com/questions/30504/programmatically-retrieve-visual-studio-install-directory

GABO1423:
Line 574 was change from `ProgramFilesw := C:/Program Files` to `ProgramFilesw := $(shell cmd /c "echo %PROGRAMFILES%")`, this avoids the hardcoded path and the core still built just fine. Does this seem like a satisfactory solution?

hrydgard:
Whether it's satisfactory or not is really for the libretro people to decide, the libretro directory is basically their domain :)

As long as CI keeps building both here and there I'm happy.

Good to see you found a cleaner way though.

GABO1423:
Managed to get this to work after hours of attempts (the main takeaway from this is that I now hate bash with a passion :D), however, the solution requires the addition of a new file to the libretro directory: `vswhere.exe`. I could not for the life of me get the compiler to use the `Program Files (x86)` directory, so this was the next thing.

I'll make a new PR with these changes, since this approach is decisively different from what the PR originally wanted to accomplish.

unknownbrackets:
Can't you use `%ProgramFiles(x86)%\Microsoft Visual Studio\Installer\vswhere.exe`?  Or in bash, `$(env | awk -F= '/^ProgramFiles\(x86\)=/ { print $2; }')`.

-[Unknown]

GABO1423:
> Can't you use `%ProgramFiles(x86)%\Microsoft Visual Studio\Installer\vswhere.exe`? Or in bash, `$(env | awk -F= '/^ProgramFiles\(x86\)=/ { print $2; }')`.
> 
> -[Unknown]

The former led to a syntax error nightmare no matter how much I tried. So no, that path didn't work. I ended up doing something else that I found much simpler. Should be getting close to finishing with testing that alternative.

ANR2ME:
Someone use this kind of cmake script:
```
  find_program(_vswhere_tool 
    NAMES vswhere 
    PATHS "$ENV{ProgramFiles\(x86\)}/Microsoft Visual Studio/Installer")
  if (NOT ${vswhere})
    message(FATAL_ERROR "Could not locate vswhere.exe - unable to source vc redistributable")
  endif()

  execute_process(
    COMMAND "${_vswhere_tool}" -latest -products * -requires Microsoft.VisualStudio.Component.VC.Tools.x86.x64 -property installationPath
    OUTPUT_VARIABLE _vs_install_loc_out
    RESULT_VARIABLE _vs_where_exitcode
    OUTPUT_STRIP_TRAILING_WHITESPACE
  )

  file(TO_CMAKE_PATH "${_vs_install_loc_out}" _vs_install_loc)
```

GABO1423:
> Someone was using this kind of cmake script:
> 
> ```
>   find_program(_vswhere_tool 
>     NAMES vswhere 
>     PATHS "$ENV{ProgramFiles\(x86\)}/Microsoft Visual Studio/Installer")
>   if (NOT ${vswhere})
>     message(FATAL_ERROR "Could not locate vswhere.exe - unable to source vc redistributable")
>   endif()
> 
>   execute_process(
>     COMMAND "${_vswhere_tool}" -latest -products * -requires Microsoft.VisualStudio.Component.VC.Tools.x86.x64 -property installationPath
>     OUTPUT_VARIABLE _vs_install_loc_out
>     RESULT_VARIABLE _vs_where_exitcode
>     OUTPUT_STRIP_TRAILING_WHITESPACE
>   )
> 
>   file(TO_CMAKE_PATH "${_vs_install_loc_out}" _vs_install_loc)
> ```

Cmake is not used by this build process, so it can't be used. Kind of a moot point anyway considering the new PR.

PR #16425, opened 2022-11-24
(Libretro) Huge MSVC Build Overhaul

ghost:
This is an evolution of what PR #16418 tried to accomplish, that rabbit hole led to this conclusion: a significant redo of the build process. Here's a summary of the changes:

- [VSWhere](https://github.com/microsoft/vswhere) is now used to detect the current location of the latest version of Visual Studio the user has installed. So for instance, if both 2019 and 2022 are installed, 2022 will be used to build the core. But if only 2019 is installed, that will be used instead.

- Since we now can get access to the Visual Studio folder, we no longer need dedicated lines to detect which version of Visual Studio the user has. So these were removed.

- While we're at it, since those lines are gone, and the rest of the code is basically the same across the 2017 and 2019 MSVC sections of the Makefile, let's combine these into one! This also allows Visual Studio 2022 to be used seamlessly, and possibly allows future versions to be used automatically.

- Since we now have a single MSVC section, we now have a universal build command that can be used with any version. So the README has been updated to reflect this.

With Visual Studio 2022, the core built perfectly fine this way, but Visual Studio 2017, and **___especially___** 2019 testing is welcomed here.
Like the previous PR, I also attached a test build in case of a regression.

Besides testing to see if older versions build correctly, any suggestions on how to further improve this PR are also welcomed!

Test build: [ppsspp_libretro.zip](https://github.com/hrydgard/ppsspp/files/10080868/ppsspp_libretro.zip)

### DIFF
--- a/libretro/Makefile
+++ b/libretro/Makefile
@@ -311,28 +311,28 @@ else ifeq ($(platform), emscripten)
 					-Drglgen_resolve_symbols=mupen_rglgen_resolve_symbols
 	PLATFORM_EXT = unix
 
-# Windows MSVC 2017 all architectures
-else ifneq (,$(findstring windows_msvc2017,$(platform)))
+# Windows MSVC all architectures
+else ifneq (,$(findstring windows_msvc,$(platform)))
 
-	PlatformSuffix = $(subst windows_msvc2017_,,$(platform))
+	PlatformSuffix = $(subst windows_msvc_,,$(platform))
 	ifneq (,$(findstring desktop,$(PlatformSuffix)))
 		WinPartition = desktop
-		MSVC2017CompileFlags = -DWINAPI_FAMILY=WINAPI_FAMILY_DESKTOP_APP -D_UNICODE -DUNICODE -DWINVER=0x0600 -D_WIN32_WINNT=0x0600
+		MSVCCompileFlags = -DWINAPI_FAMILY=WINAPI_FAMILY_DESKTOP_APP -D_UNICODE -DUNICODE -DWINVER=0x0600 -D_WIN32_WINNT=0x0600
 		LDFLAGS += -MANIFEST -NXCOMPAT -DYNAMICBASE -DEBUG -OPT:REF -INCREMENTAL:NO -SUBSYSTEM:WINDOWS -MANIFESTUAC:"level='asInvoker' uiAccess='false'" -OPT:ICF -ERRORREPORT:PROMPT -NOLOGO -TLBID:1
 		LIBS += kernel32.lib user32.lib gdi32.lib winspool.lib comdlg32.lib advapi32.lib
 	else ifneq (,$(findstring uwp,$(PlatformSuffix)))
 		WinPartition = uwp
-		MSVC2017CompileFlags = -DWINAPI_FAMILY=WINAPI_FAMILY_APP -DWINDLL -D_UNICODE -DUNICODE -DWRL_NO_DEFAULT_LIB
+		MSVCCompileFlags = -DWINAPI_FAMILY=WINAPI_FAMILY_APP -DWINDLL -D_UNICODE -DUNICODE -DWRL_NO_DEFAULT_LIB
 		LDFLAGS += -APPCONTAINER -NXCOMPAT -DYNAMICBASE -MANIFEST:NO -OPT:REF -SUBSYSTEM:CONSOLE -MANIFESTUAC:NO -OPT:ICF -ERRORREPORT:PROMPT -NOLOGO -TLBID:1 -DEBUG:FULL -WINMD:NO
 		LIBS += WindowsApp.lib
 	endif
 
 	ifeq ($(DEBUG), 1)
-		MSVC2017CompileFlags += -DEBUG
+		MSVCCompileFlags += -DEBUG
 	endif
 
-	CFLAGS += $(MSVC2017CompileFlags) -nologo
-	CXXFLAGS += $(MSVC2017CompileFlags) -nologo -EHsc -DARMIPS_USE_STD_FILESYSTEM
+	CFLAGS += $(MSVCCompileFlags) -nologo
+	CXXFLAGS += $(MSVCCompileFlags) -nologo -EHsc -DARMIPS_USE_STD_FILESYSTEM
 
 	TargetArchMoniker = $(subst $(WinPartition)_,,$(PlatformSuffix))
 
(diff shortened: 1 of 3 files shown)


hrydgard:
Looks nice as-is!

Take it out of draft and I'll merge it.

GABO1423:
Done, although I would like some confirmation that 2019 works first. I would also like to know if this affects the libretro buildbot somehow first.

hrydgard:
@LibretroAdmin, how about this?

unknownbrackets:
I realize you were frustrated by the way bash handles environment variable names, and got it working and don't want to reopen pandora's box.  Personally, I'd really prefer not committing a built exe when it's entirely unnecessary.  I don't see why `$(env | awk -F= '/^ProgramFiles\(x86\)=/ { print $2; }')` wouldn't work here.  Bash can't access an env var with parenthesis in its name, but using awk you can still get it.

That said, libretro has always continued to surprise me with its limitations and shortcomings, maybe expecting a basic level of sanity from its build scripts is unreasonable too.

-[Unknown]

GABO1423:
> I realize you were frustrated by the way bash handles environment variable names, and got it working and don't want to reopen pandora's box. Personally, I'd really prefer not committing a built exe when it's entirely unnecessary. I don't see why `$(env | awk -F= '/^ProgramFiles\(x86\)=/ { print $2; }')` wouldn't work here. Bash can't access an env var with parenthesis in its name, but using awk you can still get it.
> 
> That said, libretro has always continued to surprise me with its limitations and shortcomings, maybe expecting a basic level of sanity from its build scripts is unreasonable too.
> 
> -[Unknown]

I indeed lost a bit of patience yesterday, but I think it's important to keep an open mind. I'll try what you suggested and see if it works. The tricky part is that vswhere can only work when using cmd, so you basically have to input a command that makes bash happy, and also works with cmd. I had trouble balancing those two, I managed to get bash to accept the command, but cmd failed to execute it. While using a command cmd perfectly accepts resulted in syntax errors. I'll see if this changes anything though, thanks for the suggestion.

unknownbrackets:
For me, in a bash shell, `"$(env | awk -F= '/^ProgramFiles\(x86\)=/ { print $2; }')/Microsoft Visual Studio/Installer/vswhere.exe" -latest -requires Microsoft.VisualStudio.Component.VC.Tools.x86.x64 -property installationPath` provides the installation path I think you want from VSInstallPathw?

If it has to be a cygpath, you could use something like `cygpath "$("$(cygpath "$(env | awk -F= '/^ProgramFiles\(x86\)=/ { print $2; }')/Microsoft Visual Studio/Installer/vswhere.exe")" -latest -requires Microsoft.VisualStudio.Component.VC.Tools.x86.x64 -property installationPath)"`, I suppose.  If Makefile escaping gets in the way, you could put it in a small helper .sh file and run that, which would still be better than committing vswhere.exe itself (becomes another thing we have to update from time to time.)

-[Unknown]

GABO1423:
Using `"$(env | awk -F= '/^ProgramFiles\(x86\)=/ { print $2; }')/Microsoft Visual Studio/Installer/vswhere.exe" -latest -requires Microsoft.VisualStudio.Component.VC.Tools.x86.x64 -property installationPath` as the VSCommand resulted in a ```unexpected EOF while looking for matching `''``` error.

```cygpath "$("$(cygpath "$(env | awk -F= '/^ProgramFiles\(x86\)=/ { print $2; }')/Microsoft Visual Studio/Installer/vswhere.exe")" -latest -requires Microsoft.VisualStudio.Component.VC.Tools.x86.x64 -property installationPath)"``` did not fare much better sadly.

**Comment by unknownbrackets** [comment]
If it's Makefile escaping, did you try just putting it in a `vswhere.sh`?  Something like:

```sh
#!/bin/sh
"$(cygpath "$(env | awk -F= '/^ProgramFiles\(x86\)=/ { print $2; }')/Microsoft Visual Studio/Installer/vswhere.exe")" "$@"
```

Then you could just use that in place of the committed `vswhere.exe`.

-[Unknown]

GABO1423:
Funny that you should mention that, because that is exactly what I was trying right now.

It thankfully worked, I'll push the changes now.

GABO1423:
Changes have been pushed now, thanks for the assistance @unknownbrackets, I think everything is good now from the PPSSPP side of things, hope the libretro side of thinga isn't problematic.

hrydgard (approved):

hrydgard:
Let's merge and see if they complain :)

bslenul:
Hey there, hopefully you won't mind me asking here. No one seemed to know on the Libretro discord server.

So I got a new PC recently and had to reinstall Visual Studio, MSYS2, etc. on Windows, it used to build absolutely fine by just doing
```
cd libretro
make -j$(nproc) platform=windows_msvc2019_desktop_x64
```
but since I reinstalled, it just stays stuck doing nothing when I enter that command:

image

and if I open the task manager I can see 2 new "bash.exe" processes but that's pretty much it, nothing happens. I even tried to wait a bit, thinking maybe it was building in the background or something for whatever reason but no, CPU usage doesn't move and nothing is built.

After messing a bit I noticed that it seems to hang on this command from the Makefile: https://github.com/hrydgard/ppsspp/blob/edaa5246d144015f0c81c4a0b5029e53e23bdf10/libretro/Makefile#L353

and simply replacing it with `VSInstallPathw := $(shell cmd //c "$(VSCommand)")` fixed the issue and it built successfully:

image

Any idea why it worked on my previous setup without doubling that `/`? Something I forgot to do/reinstall?

hrydgard:
Honestly no idea whatsoever. But we should possibly just merge that change? Does it matter if you run it from a cmd shell or from mingw?

GABO1423:
> Hey there, hopefully you won't mind me asking here. No one seemed to know on the Libretro discord server.
> 
> So I got a new PC recently and had to reinstall Visual Studio, MSYS2, etc. on Windows, it used to build absolutely fine by just doing
> 
> ```
> cd libretro
> make -j$(nproc) platform=windows_msvc2019_desktop_x64
> ```
> 
> but since I reinstalled, it just stays stuck doing nothing when I enter that command:
> 
> image
> 
> and if I open the task manager I can see 2 new "bash.exe" processes but that's pretty much it, nothing happens. I even tried to wait a bit, thinking maybe it was building in the background or something for whatever reason but no, CPU usage doesn't move and nothing is built.
> 
> After messing a bit I noticed that it seems to hang on this command from the Makefile:
> 
> https://github.com/hrydgard/ppsspp/blob/edaa5246d144015f0c81c4a0b5029e53e23bdf10/libretro/Makefile#L353
> 
> and simply replacing it with `VSInstallPathw := $(shell cmd //c "$(VSCommand)")` fixed the issue and it built successfully:
> 
> image
> 
> Any idea why it worked on my previous setup without doubling that `/`? Something I forgot to do/reinstall?

I couldn't replicate this issue, using that command without the change in the Makefile still built fine on my end

GABO1423:
> Honestly no idea whatsoever. But we should possibly just merge that change? Does it matter if you run it from a cmd shell or from mingw?

I can test if it still builds fine with that command just to make sure, if it does (which I don't suppose there's a reason it shouldn't work) I can make a quick PR. Also it does have to be run from MSYS2 to use the MinGW tools, with cmd you would need some workarounds to get the tools to work (like using GnuWin32, which is as far as I know way out of date compared to MinGW).

bslenul:
> Honestly no idea whatsoever. But we should possibly just merge that change? Does it matter if you run it from a cmd shell or from mingw?

I tried running the command from MinGW, from a regular command prompt, from a VS dev command prompt and from PowerShell and same result, it just hangs. And it works fine as soon as I add that double `/`.

No idea if merging this would break builds for people where it works just fine without touching the Makefile tho :x 

> I couldn't replicate this issue, using that command without the change in the Makefile still built fine on my end

Yeah, to be clear I'm not saying this PR broke anything, far from it! It used to work like a charm for me too when you made these changes, it just started to happen when I reinstalled VS, MSYS2, etc. Just trying to understand *why* I can't get it to work anymore. Could be an update from MSYS2, could be something on my hand, I have no idea :(

GABO1423:
@bslenul Thanks for the report, the core should now build fine with that command.

bslenul:
Awesome, thank you! ❤️

PR #16428, opened 2022-11-24
Vulkan validation cleanups, remove "GetCurrentStepId" based invalidation

hrydgard:
Needed more detailed invalidation now that we have frame-global descriptors, plus I never liked the GetCurrentStepId solution. So cleaned it up with a callback instead.


hrydgard:
Hm, this breaks the water in Outrun in D3D11 only. Don't merge yet.

PR #16431, opened 2022-11-24
Snap from screen center

iota97:
This changes the snap to make it easier to have 2 buttons at the same distance from 2 opposite screen borders (e.g. L and R buttons were triggering a bit my OCD).

Changes the grid rendering to have a thicker line at the center too, it make it easier to place a button to the exact center (e.g. SELECT in the bottom row of 3).

### DIFF
--- a/UI/TouchControlLayoutScreen.cpp
+++ b/UI/TouchControlLayoutScreen.cpp
@@ -330,15 +330,11 @@ public:
 			float yOffset = bounds_.y;
 
 			dc.Draw()->Rect((x1+x2)/2 + xOffset - pixel_in_dps_x, y1 + yOffset, 3.0*pixel_in_dps_x, y2-y1, col);
-			for (int x = (x1+x2)/2 - g_Config.iTouchSnapGridSize; x > 0; x -= g_Config.iTouchSnapGridSize)
-				dc.Draw()->vLine(x + xOffset, y1 + yOffset, y2 + yOffset, col);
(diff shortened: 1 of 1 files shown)


UI/TouchControlLayoutScreen.cpp
```diff
@@ -327,10 +328,19 @@ class SnapGrid : public UI::View {
 			dc.BeginNoTex();
 			float xOffset = bounds_.x;
 			float yOffset = bounds_.y;
-			for (int x = x1; x < x2; x += g_Config.iTouchSnapGridSize)
+
+			dc.Draw()->Rect((x1+x2)/2 + xOffset - pixel_in_dps_x, y1 + yOffset, 3.0*pixel_in_dps_x, y2-y1, col);
+			for (int x = (x1+x2)/2 - g_Config.iTouchSnapGridSize; x > 0; x -= g_Config.iTouchSnapGridSize)
+				dc.Draw()->vLine(x + xOffset, y1 + yOffset, y2 + yOffset, col);
+			for (int x = (x1+x2)/2 + g_Config.iTouchSnapGridSize; x < x2; x += g_Config.iTouchSnapGridSize)
 				dc.Draw()->vLine(x + xOffset, y1 + yOffset, y2 + yOffset, col);
```
hrydgard, line 336: I think with a tiny bit of math you can get this down to one loop, and I also think it will be clearer that way :)

Same below.

PR #16461, opened 2022-11-29
Fix Burnout Dominator lens flare on OpenGL ES

hrydgard:
Not the prettiest solution, but works: Use `R8_UNORM` as a fallback if `R16_UNORM` is not available.

Fixes #16081 .


Gamemulatorer:
This is also fixes https://github.com/hrydgard/ppsspp/issues/11100#issuecomment-1255361960

PR #16467, opened 2022-11-30
Redo the control UI screen layout, slightly update the look

hrydgard:
If #16466 didn't take caer of #15773, maybe this will.

Basically a rewrite of the screen layout of the control layout screen.

image

Hopefully this will avoid whatever bug we used to have. Plus I think it looks slightly beter.

### DIFF
--- a/Common/UI/ViewGroup.cpp
+++ b/Common/UI/ViewGroup.cpp
@@ -1071,7 +1071,8 @@ float ScrollView::ClampedScrollPos(float pos) {
 	}
 
 	float childSize = orientation_ == ORIENT_VERTICAL ? views_[0]->GetBounds().h : views_[0]->GetBounds().w;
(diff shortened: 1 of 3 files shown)


bizz91:
Why not make the control ui similar to aethersx2 just so it becomes easier to adjust?

I mean the menu can be moved to a center button similar to ps2 android emu.

hrydgard:
Because I've never used AetherSX2? Though I probably should check it out, heh.

bizz91:
Some form of a drop down menu place above center and the whole fullscreen for accurately adjusting the controls.

bizz91:
> Because I've never used AetherSX2? Though I probably should check it out, heh.

Yes you should, its a great emu of the best selling console of all times.

bizz91:
Any progress?

hrydgard:
Don't nag please. Any further redesign will be after 1.14, I'm concentrating on fixing bugs now.

bizz91:
Sorry, its alright. I am looking forward to 1.14. 😀

PR #16501, opened 2022-12-04
Input (inc. mouse and keyboard) api for plugins

ThirteenAG:
Ideas, fixes, suggestions are welcome.

Added 

```
EMULATOR_DEVCTL__GET_AXIS
EMULATOR_DEVCTL__GET_VKEY
```

Usage example:
```
    float MouseX;
    sceIoDevctl("kemulator:", EMULATOR_DEVCTL__GET_AXIS, (void*)JOYSTICK_AXIS_MOUSE_REL_X, 0, &MouseX, sizeof(MouseX));
    return Mouse;
```

```
#define NKCODE_C 31
short CPad::GetLookBehindForPed()
{
    char C = 0;
    sceIoDevctl("kemulator:", EMULATOR_DEVCTL__GET_VKEY, (void*)NKCODE_C, 0, &C, sizeof(C));
    return C > 0 ? 255 : 0;
}
```

Useful for achieving some out of the ordinary things, like aiming with mouse (just emulated joystick for this test)

https://user-images.githubusercontent.com/4904157/205516948-2adeec97-1c79-4107-bc09-777c7454df56.mp4

Or driving slowly

https://user-images.githubusercontent.com/4904157/205517183-41ec3f28-e81b-48e0-9148-871291dc1cec.mp4

Entering cheats like in PC version, etc...


Core/HLE/sceIo.cpp
```diff
@@ -2067,11 +2069,15 @@ static u32 sceIoDevctl(const char *name, int cmd, u32 argAddr, int argLen, u32 o
 				Memory::Write_Float(scale, outPtr);
 			}
 			return 0;
-		case EMULATOR_DEVCTL__GET_LTRIGGER:
-			//To-do
+		case EMULATOR_DEVCTL__GET_AXIS:
+			if (Memory::IsValidAddress(outPtr)) {
+				Memory::Write_Float(HLEPlugins::PluginDataAxis[argAddr], outPtr);
```
hrydgard, line 2074: argAddr (which is being used as an index) needs range checking here, otherwise the plugin can easily crash the emulator here. Same below.

hrydgard (changes requested):
Sorry for the late review!

Will need to look into making a cleaner way to transmit the mouse deltas into where the plugin can access them, but I'm ok with as it is for this PR.

Just one minor comment.

hrydgard (approved):

hrydgard:
Hm, a conflict seems to have arisen. Other than that, looks good now.

PR #16506, opened 2022-12-05
Fix the Qt/SDL window icon path

xuzhen:
The assets directory will be installed into `${CMAKE_INSTALL_DATADIR}/ppsspp` https://github.com/hrydgard/ppsspp/blob/4fd336f5ef458708ce89631f4582af3427e0de01/CMakeLists.txt#L2520-L2522

But the Qt/SDL window will try to load window icon from the PPSSPP binary's directory https://github.com/hrydgard/ppsspp/blob/4fd336f5ef458708ce89631f4582af3427e0de01/Qt/mainwindow.cpp#L29 https://github.com/hrydgard/ppsspp/blob/4fd336f5ef458708ce89631f4582af3427e0de01/SDL/SDLMain.cpp#L773

My patch is to use the ASSETS_DIR defined in CMakeLists.txt to get the correct path https://github.com/hrydgard/ppsspp/blob/4fd336f5ef458708ce89631f4582af3427e0de01/CMakeLists.txt#L117

### DIFF
--- a/Qt/mainwindow.cpp
+++ b/Qt/mainwindow.cpp
@@ -5,6 +5,7 @@
 #include <QApplication>
 #include <QDesktopServices>
 #include <QDesktopWidget>
+#include <QFile>
 #include <QFileDialog>
 #include <QMessageBox>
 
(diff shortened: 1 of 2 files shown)


xuzhen:
And the `external_dir` variable may also set to a incorrect path in MacOS. https://github.com/hrydgard/ppsspp/blob/4fd336f5ef458708ce89631f4582af3427e0de01/SDL/SDLMain.cpp#L712-L722
But I'm not using Mac and didn't know the purpose of the `external_dir` variable, so I didn't touch this part.

iota97:
Well this will break the icon if we just build without installing... e.g: `./b.sh && ./build/PPSSPPSDL`

xuzhen:
OK. Patch updated

PR #16516, opened 2022-12-06
Rework the post-shader list

### LINKED ISSUE #16463: Post-processing UI needs improvement
Well known, just to have something to tag as a reminder for 1.14.

The current look and way of editing the post processing list is pretty awkward, plus very bulky with the parameters as well.

I'm gonna try to make it more like a regular selectable list and add "Add" and "Remove" buttons, I think - and it'll only show the parameters of the selected shader, instead of the whole chain.

hrydgard:
Will fix #16463 

image

Now you can show/hide settings for each shader with a little icon, they can be moved up and down in the list, and there's a delete button.

Another thing I need to fix is vertical screen layout...


hrydgard:
Added a screenshot above to show off the new context menu you can pop up for each entry in the list.

The "Off" entry will be renamed to "Add" or something.

iota97:
Looks a lot better now, the only thing I don't really like of this screen is that it lack the darkening background (for good reasons) and that makes it kinda hard to read on "white" game scenes. Should probably add a shadow to the headers and double the darkness of the buttons, not a big deal anyway :)

There is a bug (?) with duplicate shaders as it disable the settings only if the other have the settings drop down opened. It's probably simpler to just disable the drop down button in this case (I don't really want someone to expect it will edit only the open one).

On a side note we are adding 4 arrows rather than one and rotate it during rendering like the other image we have, shouldn't really matter, just pointing it out :)

PS: I would prefer the delete button directly on the item as is way more common to delete rather than move I think

PPS: we could add few of the new images to the custom touch button in `GamepadEmu.h`, I can do that after this is merged anyway

hrydgard:
Yeah I'm gonna do something about the backgrounds, I'll have a darkening background but only underneath the columns of buttons, so you can see the full effect of your shaders in the center at least.

Right hehe, we could rotate the arrows, just thought this was the easiest solution to not have to add rotation support everywhere..

Should we maybe simply disallow duplicate shaders? It's not that useful, especially since they can't have different settings...

I'll move the delete button onto the items directly.

iota97:
Disabling duplicate shaders might be fine, but I'm not sure if is a good idea to test if it is fine next to a stable release :)

Also there is no easy way to preview upscaling shader like `SSAA (Gauss)` right? We could add a little warning about not being able to preview it so people won't think it's just useless/broken :/

hrydgard:
Yeah true, that's a problem - those can't be previewed without rendering a new frame, moving game time forward, which might be unexpected.

I'm thinking of deleting a bunch of the less useful shaders too. Grayscale can be replaced with color correction, for example.

hrydgard:
Anyway, this is a good checkpoint, I'll get it in as-is and continue work in new branches.

PR #16520, opened 2022-12-07
LittleBigPlanet: Compat flag to work around smoke rendering problem

hrydgard:
See #16030

Hopefully just for the release. There might be a tweak of our heuristics that can fix it too, but much riskier at this point.

I might give that a try and hide tweaks behind the flag instead, too...


hrydgard:
Yeah I know there are, not gonna get rid of the functionality. I will fix this more properly after the release.

PR #16522, opened 2022-12-07
"Motion blur" - LCD persistence shader

iota97:
Well, should help testing if we broke it at least, 2008 feels...


hrydgard:
It's not really motion blur (that requires motion vectors), more like image persistence or slow LCD. So maybe just call it "LCD simulation"? :)

hrydgard:
Just tried it on Vulkan, seems to work perfectly fine?

jcchikikomori:
Best test scenario for this is Tekken 6. I attempt to fake the motion blue experience with the PS3 & the Arcade machines, including this texture pack. It's almost 1:1 with the PS3 but HD

https://github.com/kharij/Tekken-6-HD-Texture-Pack

iota97:
I really thought I had tested Vulkan too, my bad.

Anyway Vulkan works fine, OpenGL was broken in: bd674c47b66b62bfb00b8a2b1f57ee59092be2b8.

The previous frame is in `sampler2`, and get pushed to `queries`:
https://github.com/hrydgard/ppsspp/blob/b06ea9784580302ba3456712a8d2aec770687b1c/Common/GPU/OpenGL/thin3d_gl.cpp#L1248

After that commit the check is on `samplers_` rather than `queries` (reverting just this change on the commit fix it):
https://github.com/hrydgard/ppsspp/blob/b06ea9784580302ba3456712a8d2aec770687b1c/Common/GPU/OpenGL/thin3d_gl.cpp#L1257

I have no idea why this was changed, so I'll leave the fix to you :)

hrydgard:
I'm replacing this PR with #16531, which includes the commits from this one.

PR #16523, opened 2022-12-07
Post shader cleanup

hrydgard:
Delete Grayscale and InverseColors shaders and translation strings.

Grayscale can trivially be replaced with the Color Correction shader.

Adds background color to the left and right scroll views of the display layout screen in order to make things more readable.

image


assets/lang/bg_BG.ini
```diff
@@ -824,8 +824,8 @@ Undo last save = Undo last save
 [PostShaders]
 (duplicated setting, previous slider will be used) = (duplicated setting, previous slider will be used)
 4xHqGLSL = 4×HQ GLSL
-5xBR = 5xBR
-5xBR-lv2 = 5xBR-lv2
+5xBR = 5xBR pixel art upscaler
```
iota97, line 827: Only in this file? Intentional?

NABN00B:
Is InverseColors also replacable?

iota97:
It's not replaceable with existing ones. It was removed because was quite useless I think

Technically it might be possible to get something similar expanding the gamma range (for example using gamma = -0.3 and lightness = -1) but would be really confusing that way.

NABN00B:
> It was removed because was quite useless I think

I think the usefulness of shaders in video games in general is subjective anyway.

> Technically it might be possible to get something similar expanding the gamma range (for example using gamma = -0.3 and lightness = -1) but would be really confusing that way.

The UI doesn't allow you to set these values in 13.2, not sure if that has been changed since.

PR #16551, opened 2022-12-11
Fix Hz casing

iota97:
Just a bit of OCD :)


hrydgard:
Oh oops, i did that just to check that the translation worked :p

PR #16558, opened 2022-12-12
Update th_TH.ini

Saramagrean:
and Game crashed string should remove? as Unknown said.

GC-TRAN


hrydgard:
I'll figure out what to do with the Game Crashed string later. For now this is good.

PR #16592, opened 2022-12-16
Update zh_CN.ini

leoxxx:


leoxxx:
What' the meaning of "Green level" & "Color preservation" ?

hrydgard:
They are only used to tweak the output of the experimental red/blue-glasses stereo VR support, that you can find in developer settings.

Probably no need to translate at this time, I might get rid of them.

leoxxx:
I see.

PR #16604, opened 2022-12-16
New translation

Arashvscode:
More words translated


Arashvscode:
It is surprising that it has been confirmed in all the output versions except one. What is the reason? It is just an ini file

hrydgard:
That's just a random error that happens sometimes, we know about it but not the cause.

Arashvscode:
> 

Hello, I hope you are well. Thank you for supporting me. After several simple researches, I found out that ini is sensitive to spacing in some systems and leads to errors in making the best spacing is 1 to 2 spaces, I think.  I did not respect this distance in a translation and it led to an error!

PR #16605, opened 2022-12-16
Update Russian translation

TotalCaesar659:
P.S. I found that when being ingame, PPSSPP uses transparent game background for almost all menu points:

But in "Savedata manager", "System information" and "Remote disc streaming" it becomes usual not-ingame one:

Is it a bug or expected behavior?


hrydgard:
It's a bug, thanks for the reminder!

TotalCaesar659:
I found more untranslated string on Android: "UI Resolution" and "dpi" nearby (should it be "DPI"?) Please add them to translations.
Sorry can't make screenshot, phone doesn't let me do it (tried it about an hour). It's in system info under "Native resolution".

hrydgard:
Thanks @TotalCaesar659 , I'll take care of it soon.

TotalCaesar659:
Thank you very much, [updated](https://github.com/hrydgard/ppsspp/pull/16613). I hope background fix will arrive in 1.14.1.

TotalCaesar659:
Merry Christmas @hrydgard! Should background fix be waited in 1.14.2 or in 1.15.0?
P.S. I found one more untranslatable string for Android: "Scope Storage Enabled" in "Tools" - "System Information" - "Storage". It can be changed to just "Scope Storage" and states "On"/"Off", so there will be no need to add them to translations.

hrydgard:
I'll fix the backgrounds for 1.4.2.

Thanks for pointing out the scoped storage string - I'll follow your suggestion.

PR #16634, opened 2022-12-20
Android OpenGL: delete dead code

hrydgard:
We had an C++-EGL code path on Android a long time ago that doesn't seem to have any particular advantage, only disadvantages which is why it's been disabled for a long time. Delete it.


unknownbrackets (approved):

hrydgard:
Hm, I seem to have broken something here. Will fix tomorrow.

Issue #16646, opened 2022-12-23
Using the "IP Address" option in "Pro Ad Hoc Server IP Address" doesn't work


Gabriel2Silva:
### Game or games this happens in

ULES-01213 - Monster Hunter Freedom Unite

### What area of the game / PPSSPP

Tests done in Monster Hunter Freedom Unite.

Basically, networking works fine on the Libretro core. When using a private server such as socom.cc, I can play with people on both PPSSPP standalone and Libretro flawlessly. However, if I decide to host a server myself (all ports correctly forwarded, DMZ enabled, even) or to manually input an IP Address using the "IP Address" option in "Pro Ad Hoc Server IP Address", it doesn't work.

In fact, it seems to me that it doesn't even try to connect, as the log always prints the same two errors:

`[libretro ERROR] [SCENET] Socket error (10060) when connecting to AdhocServer [135.148.124.223/135.148.xx.223:27312]
[libretro ERROR] [SCENET] Socket error (10060) when connecting to AdhocServer [135.148.124.223/135.148.xx.223:27312]
[libretro ERROR] [SCENET] Socket error (10038) when sending`

or

`[libretro ERROR] [SCENET] DNS Error Resolving 192.168.000.029`.

What also bothers me is that if I select the "socom.cc" option in "Pro Ad Hoc Server IP Address", it connects just fine, but if I manually input the DNS IP Address for this same server, it gives me this error.

The "Localhost (multiple instances)" option also works just fine. It seems to me that this is a problem specific to the "IP Address" option in the "Pro Ad Hoc Server IP Address" dropdown.

I attached two log files.

[self hosted.log](https://github.com/hrydgard/ppsspp/files/10291924/self.hosted.log)
[socom original ip.log](https://github.com/hrydgard/ppsspp/files/10291925/socom.original.ip.log)

### What should happen

It should connect just fine when using the "IP Address" option in "Pro Ad Hoc Server IP Address".

### Platform

libretro / Retroarch

### Mobile phone model or graphics card

NVIDIA GeForce GTX 1660 SUPER

### PPSSPP version affected

v1.12.0 - PPSSPP (ff4971f)

### Graphics backend (3D API)

Vulkan

### Checklist

- [X] Test in the [latest git build](https://buildbot.orphis.net/ppsspp/) in case it's already fixed.
- [X] [Search for other reports](https://github.com/hrydgard/ppsspp/search?q=my+issue&type=issues) of the same issue.
[...]

PR #16651, opened 2022-12-24
Libretro: skip leading zeros in IP address.

amverweij:
In the LibRetro settings GUI, we enter the IP address digit by digit (in order to avoid a drop down menu with 256 values). When we combine these digits into a string, we should skip leading zeros for every byte. Failing to do so seems to change the interpretation of the remainder of the byte to an octal number. This only yields the desired result for bytes in the range [000, 007]. Anything else would either be invalid (in case it contains an 8 or a 9) or misinterpreted (since the byte would be composed with a base of 8 instead of 10).

This PR intends to fix that.


amverweij:
This should fix #16646

unknownbrackets (approved):

PR #16670, opened 2022-12-29
DrawContext destructor: Need to call DestroyPresets earlier.

hrydgard:
Very minor memory/resource leak, but why not fix it.


unknownbrackets (approved):

hrydgard:
Somehow, this fixed the problem I was having with task switching on Galaxy S3, so I'm gonna move on from that one, heh.

PR #16721, opened 2023-01-04
Inject default vertex components to avoid needing separate vshaders for various formats.

hrydgard:
Part of #16567 and compounds with the other changes.

This one removes 3 vertex shaders from Outrun, for example, resulting in a slightly larger decrease in the overall pipeline count.

A majority of the vertices in most games do have UVs so this doesn't add a lot of extra vertex decoding work.


hrydgard:
I'll be experimenting in this branch with really minimizing the different shaders, to see if all this is worthwhile. I expect it will be on first-run on mobiles that are slow at compiling shaders, but I basically want to see the extremes of the tradeoff here. Then we might go with it or find a compromise, or something.

hrydgard:
Rebased this on master, removing the 2x commit since that's already been merged separately.

This still reduces the amount of pipelines to compile by quite a bit, although not super drastically - though reducing the amount of expensive vertex shaders is the biggest result. Will need some benchmarking to see if there's any noticeable performance loss...

hrydgard:
This has problems with D3D11 in software transform that need fixing (vec2 vs vec3 texcoords).

hrydgard:
I might dig this up again at a later time. Right now, with parallel compiles and so on, I'm happy with the current improvements I've achieved for 1.15.

PR #16735, opened 2023-01-05
OpenXR - Fixes for 3rd birthday game

lvonasek:
From user feedback:
* allow smaller HUD scale
* overlay effect is wrong


hrydgard:
The build error can be fixed by just rebasing on master, you branched at a bad commit, sorry.

PR #16769, opened 2023-01-10
Uniform dirty-bits: Make room for four more uniforms. 

hrydgard:
Then we're out and need to start merging related ones (which might be a good idea anyway).


hrydgard:
Like #16778 , this somehow causes trouble for D3D11, or I'm having some very weird local issues.

unknownbrackets (approved):

PR #16775, opened 2023-01-10
[Libretro] Prevent the removal of "huf_decompress_amd64.S" when doing "make clean" on Linux

bslenul:
Currently when doing `make clean` on Linux it deletes `ext/zstd/lib/decompress/huf_decompress_amd64.S`, I don't know if there's a better/cleaner fix for this but it works fine on my Linux Mint VM, also confirmed to work from another person running on a real Linux machine.

### DIFF
--- a/libretro/Makefile
+++ b/libretro/Makefile
@@ -500,7 +500,7 @@ else
 endif
 
 clean:
-	rm -f $(OBJECTS) $(TARGET)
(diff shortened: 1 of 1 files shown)


hrydgard (approved):

hrydgard:
I'm ok with this for the libretro build :)

PR #16778, opened 2023-01-10
Remove the rather redundant DoTexture flag from vshaders.

hrydgard:
Part of https://github.com/hrydgard/ppsspp/issues/16567 and compounds with the other changes.

Slightly reduces the number of unique vertex shaders but doesn't do much for the pipeline count, as the fragment shader has a tex flag (which will stay). Still worth doing for the simplification and reduction of vertex shader generation cost.


hrydgard:
Weirdly, this appears to break D3D11. Will need to look at that.

Issue #16792, opened 2023-01-11
Resolution 


Gelo021:
Yow again, just wanna ask if changing Display Resolution to Any Resolution or Any Resolution to native, kinda misplace ?
![Screenshot_2023-01-12-02-25-46-92_2f85358b2198d26f8aca533d68bee793](https://user-images.githubusercontent.com/67460268/211889221-872cb24b-2509-4053-9d1b-c69a7fe3387a.jpg)
After Changing...
![Screenshot_2023-01-12-02-26-07-02_2f85358b2198d26f8aca533d68bee793](https://user-images.githubusercontent.com/67460268/211889287-a3290156-7c94-48cb-8639-47924e634c7d.jpg)

Pressing Reset wont change, restarting the game make it normal again kindda wierd 🤔

hrydgard:
I would recommend not changing the display resolution at all. But yeah, there seems to still be some bugs...

Gelo021:
> I would recommend not changing the display resolution at all. But yeah, there seems to still be some bugs...

Okay hrydgard, im back at reporting bugs hihi

PR #16794, opened 2023-01-11
Remove the Android display resolution selector

hrydgard:
I think its benefit is absolutely tiny if any, and causes a lot more problems than it's worth.

Note that this isn't the rendering resolution selector, that will of course still be there, but it's the "hardware scaler" resolution.

See #16792

### DIFF
--- a/Core/Config.cpp
+++ b/Core/Config.cpp
@@ -669,34 +669,6 @@ static int DefaultFastForwardMode() {
 #endif
 }
 
-static int DefaultAndroidHwScale() {
-#ifdef __ANDROID__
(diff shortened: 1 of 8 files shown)


Gamemulatorer:
Question: Why when I set rendering resolution to X1 and then use skip buffer rendering the resolution still in native device resolution 🤔🤔

hrydgard:
When using skip buffer rendering, we should probably hide the rendering resolution selector since it does nothing, it just renders to the backbuffer which is at your device resolution.

Gamemulatorer:
But in HW Scaler the skip rendering resolution can be change

hrydgard:
That's true. But that's also the only real use of this feature, and it's not worth the testing and maintenance work.

PR #16800, opened 2023-01-12
Android: update measurements on UI thread

hrydgard:
This takes the main change from #16583, and simplifies it a bit. Seems to work fine.

@Arashvscode


Arashvscode:
> This takes the main change from #16583, and simplifies it a bit. Seems to work fine.
> 
> @Arashvscode

I'm sure it does the background operation well, currently my system is broken, I will help more in the future..... thank you

PR #16846, opened 2023-01-25
Add new ARM64 Crypto Extensions ( Need to test )

zminhquanz:
-About Vero4k line , because they have the check Aarch64 on below if report failed , not detect it , so i remove it because it's unnecessary 
-I need someone to test on real device , using Device Info HW app to compare it
-Already test on my Galaxy S10 Qualcomm version and it's correct

### DIFF
--- a/Common/ArmCPUDetect.cpp
+++ b/Common/ArmCPUDetect.cpp
@@ -304,7 +304,8 @@ void CPUInfo::Detect()
 	bFP = false;
 	bASIMD = false;
 #else // PPSSPP_PLATFORM(LINUX)
-	truncate_cpy(cpu_string, GetCPUString().c_str());
+    // Samsung Exynos , Qualcomm Snapdragon , Mediatek , NVIDIA Tegra
+    truncate_cpy(cpu_string, GetCPUString().c_str());
 	truncate_cpy(brand_string, GetCPUBrandString().c_str());
 
 	bSwp = CheckCPUFeature("swp");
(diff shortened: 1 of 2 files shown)


Gamemulatorer:
Can you also do this https://github.com/hrydgard/ppsspp/issues/10620#issuecomment-1216125424?

zminhquanz:
> Can you also do this [#10620 (comment)](https://github.com/hrydgard/ppsspp/issues/10620#issuecomment-1216125424)?

Later , because i need to another solution

unknownbrackets:
Indentation is wrong.

I suppose some of these might be useful, like ASIMDRDM, ASIMDFHM, ASIMDDP, FRINT, SVE2, RPRES, I8MM, and CRC32.  But most of these I just can't see us even using.  Is there a reason to detect all of them?

-[Unknown]

zminhquanz:
> Indentation is wrong.
> 
> I suppose some of these might be useful, like ASIMDRDM, ASIMDFHM, ASIMDDP, FRINT, SVE2, RPRES, I8MM, and CRC32. But most of these I just can't see us even using. Is there a reason to detect all of them?
> 
> -[Unknown]

ASIMDRDM is Advanced SIMD Rounding Double Multiply , from ARMv8.1 like Cortex A72 will have it
ASIMDDP is Advanced SIMD Dot Product , from ARMv8.2 will have it
CRC32 some of ARMv8.0 soc have it ,i have Nexus 9 using NVIDIA Tegra K1
but SVE or SVE2 most of mobile SOC don't have it , so it won't show
ASIMDFHM , FRINT , I8MM on my Xiaomi 13 using Snapdragon 8 gen 2 have it

unknownbrackets:
Well, I just mean that - without actually adding something to the emitter or etc., these won't be used anyway.  If they aren't useful enough for us to ever do that, such as SSBS, it doesn't seem very worth detecting or showing them.  Even the dot product one is questionably useful, because it only operates on 8-bit values iirc.  Maybe could use it in softgpu somewhere...

SVE/SVE2 could be more useful, wish more devices supported them...

-[Unknown]

hrydgard:
With #16871, is this still relevant?

unknownbrackets:
I added the few I could really see us using in #16862, so possibly not.

-[Unknown]

Gamemulatorer:
Question might be unrelated to this pr but on my MT6592 android phone it detect many cpu extension.
received_867607317875221
But on my vivo y11 snapdragon 439 detect only two cpu extension.
Screenshot_20230201_180440_2f85358b2198d26f8aca533d68bee793

hrydgard:
Your first phone there is ARM32, which has a lot of extensions for fairly basic functionality, which is included in ARM64 (which is your second phone) by default. So it's not like the first one has any capability that the second one doesn't.

Gamemulatorer:
Ok thanks for good explanation Sir @Henrik

PR #16847, opened 2023-01-25
Basic setup to be able to call the NVDA text-to-speech API on Win64 only.

hrydgard:
Work towards fixing #16218, text-to-speech integration in PPSSPP's UI.

NVDA is an open and free screen reader library for Windows. It's quite easy to support so it can be the first TTS library we work with.

https://github.com/nvaccess/nvda/

This is only the very initial step of getting something up and running on Windows x86-64.

TODO list:

- [ ] Actually hook it up our UI toolkit to say the name of the focused view
  - [ ] Say stuff on keyboard navigation
  - [ ] Say stuff on mouse hover (somewhat difficult, we don't track hover right now, due to touch-centric design)
- [ ] Investigate if we can avoid shipping a DLL by linking the support library statically, or do some dynamic DLL loading
- [ ] Support more platforms
  - [ ] Windows ARM64
  - ~~Windows 32-bit~~ (probably not gonna bother)


Waltlover:
Hey this would be really helpful! I use ppsspp and it would be cool to be able to use NVDA. Looking forward to how it plays out!

PR #16851, opened 2023-01-27
Ability to show battery % on Darwin platforms, transform FPS/Speed counter into a BitCheckbox

NSAntoine:


Core/Config.h
```diff
@@ -31,6 +31,12 @@ extern const char *PPSSPP_GIT_VERSION;
 
 extern bool jitForcedOff;
 
+// should this be here?!
+// for Config.iShowFPSCounter
+#define SHOW_FPS_COUNTER 1 << 1
+#define SHOW_SPEED_COUNTER 1 << 2
+#define SHOW_BATTERY_PERCENT 1 << 3
```
hrydgard, line 38: When using #define with expressions, you need to use parentheses. Like:

`#define SHOW_FPS_COUNTER (1 << 1)`

Consider

`int x = flags & SHOW_FPS_COUNTER;`

the result is otherwise

`int x = flags & 1 << 1;`  which actually evaluates as `int x = (flags & 1) << 1;` due to the low precedency of shift operators.

Core/Config.h
```diff
@@ -270,7 +276,7 @@ struct Config {
 
 	// UI
 	bool bShowDebuggerOnLoad;
-	int iShowFPSCounter;
+	int iShowFPSCounter; // Should we rename this now?
```
hrydgard, line 279: int iShowStatusFlags maybe? what do you think?

hrydgard (changes requested):
Isn't there now some translation string that can be deleted, like "Show FPS Counter"?

SerenaKit:
> Isn't there now some translation string that can be deleted, like "Show FPS Counter"?

What do you think of using the existing `Show FPS Counter` translated string in place of the `Show FPS` key I put in? Would save localizers time so that they wouldn't have to re-translate the key (though, considering that the `Show FPS Counter` menu used to present options for *speed* as well, there would be the possible issue of some translation of `Show FPS Counter` mentioning speed, which is unwanted for the purposes of this pull request)

Core/Config.h
SerenaKit: Good to me

Core/Config.h
SerenaKit: Ah, I forgot about that for bit masks, will do

hrydgard:
Reusing that translation should be fine. Could just rename the key (search/replace, or there's a tool in /tools/langtool that will do it).

Core/Config.h
SerenaKit: Addressed in https://github.com/hrydgard/ppsspp/pull/16851/commits/ca154879f05e829c6c333c8f27c1ea0f1f65d4d2

Core/Config.h
SerenaKit: Addressed in https://github.com/hrydgard/ppsspp/pull/16851/commits/ca154879f05e829c6c333c8f27c1ea0f1f65d4d2

iota97:
Probably nitpicking but wouldn't been better to use uint32_t rather than int for bitflags? We don't really know how big a int is exactly and so on...

It's just a bit weird to me, but will do anyway so feel free to ignore this :)

hrydgard:
@iota97 Well int is gonna be at least 32 bit on any platform that PPSSPP will build for, and this bit field is 3 bits so far, so it'a not much of a concern at this time.

hrydgard (approved):

Common/UI/View.h
```diff
@@ -870,7 +870,11 @@ class BitCheckBox : public CheckBox {
 	BitCheckBox(uint32_t *bitfield, uint32_t bit, const std::string &text, const std::string &smallText = "", LayoutParams *layoutParams = nullptr)
 		: CheckBox(nullptr, text, smallText, layoutParams), bitfield_(bitfield), bit_(bit) {
 	}
-
+    
+    BitCheckBox(int *bitfield, int bit, const std::string &text, const std::string &smallText = "", LayoutParams *layoutParams = nullptr) : BitCheckBox((uint32_t *)bitfield, (uint32_t)bit, text, smallText, layoutParams) {
+        
+    }
```
hrydgard, line 876: what happened with indentation here?

hrydgard:
I'm considering doing some kind of grouping for multiple related bitfield checkboxes like this, like maybe automatically create a dialog with a list of checkboxes... but that's for later.

Common/UI/View.h
SerenaKit: Certified Xcode Moment, fixed now

hrydgard:
I don't mean the blank line, I meant the constructor itself, it's at a different indentation level than the one above it.

But whatever, I'll fix it later. Apart from that looks good to me.

Gamemulatorer:
Can you make also the FPS/SPEED color follow system theme color?
FPs/Speed color is blue
Screenshot_20230129_003918_2f85358b2198d26f8aca533d68bee793
My system theme color is Magenta?
Screenshot_20230129_003928_2f85358b2198d26f8aca533d68bee793

Core/Config.h
```diff
@@ -31,6 +31,12 @@ extern const char *PPSSPP_GIT_VERSION;
 
 extern bool jitForcedOff;
 
+// should this be here?!
```
unknownbrackets, line 34: Would be better to put them in ConfigValues.h.  I'd recommend an `enum class` although you'd have to do casts, it'd namespace the values better.

-[Unknown]

Core/Config.h
```diff
@@ -270,7 +276,7 @@ struct Config {
 
 	// UI
 	bool bShowDebuggerOnLoad;
-	int iShowFPSCounter;
+	int iShowStatusFlags; // Should we rename this now?
```
unknownbrackets, line 279: It's renamed now, right?

-[Unknown]

Common/UI/View.h
```diff
@@ -870,7 +870,10 @@ class BitCheckBox : public CheckBox {
 	BitCheckBox(uint32_t *bitfield, uint32_t bit, const std::string &text, const std::string &smallText = "", LayoutParams *layoutParams = nullptr)
 		: CheckBox(nullptr, text, smallText, layoutParams), bitfield_(bitfield), bit_(bit) {
 	}
-
+    
+    BitCheckBox(int *bitfield, int bit, const std::string &text, const std::string &smallText = "", LayoutParams *layoutParams = nullptr) : BitCheckBox((uint32_t *)bitfield, (uint32_t)bit, text, smallText, layoutParams) {
```
unknownbrackets, line 874: nit: Indentation is off.

-[Unknown]

hrydgard: Fixed

Core/Config.h
hrydgard: I cleaned this up in 42873654117579a057b8741b203c17a71d09a9ef .

PR #16857, opened 2023-01-28
OpenXR - Control game camera using head rotation

lvonasek:
As many users attempted to play games in first person view (setting the camera in front of the game character head). This PR enables a true first person experience. It allows mapping head rotation to keys.


lvonasek:
Quick test: https://youtu.be/oasMr9ln7bw

hrydgard:
I haven't tested this but the concept seems cool. I'm good with merging it.

PR #16859, opened 2023-01-29
Fix for issue of disappear text on Shinobido Talese of The Ninja

nassau-tk:
Fix for #16845.


sum2012:
Confirm fixed

hrydgard:
Cool! What did you change in the font?

nassau-tk:
I Changed few characters location in the caracter drawing area.
It means that character moved more inside of the area.
Of course, I fixed characters only difference of last fixed.

PR #16871, opened 2023-01-31
#16869 Fix libretro Mac x64 build, add Mac arm64 build

warmenhoven:
libretro does use cmake for Linux and Android so this moves Mac builds to be in line with that.

The Mac Metal build of RetroArch now has four graphics drivers:
- The gl graphics driver mostly works, though crashes for some content
- The glcore graphics driver displays a black screen, though the proper audio can be heard underneath
- The metal graphics driver displays a black screen, though the proper audio can be heard underneath
- The vulkan graphics driver crashes, I think due to a RetroArch bug.

### DIFF
--- a/libretro/CMakeLists.txt
+++ b/libretro/CMakeLists.txt
@@ -22,7 +22,7 @@ if(ANDROID)
 endif()
 
 if(NOT MSVC)
-   if (IOS)
+   if (APPLE OR IOS)
    	target_link_libraries(ppsspp_libretro "-Wl")
    else()
    	target_link_libraries(ppsspp_libretro "-Wl,-Bsymbolic")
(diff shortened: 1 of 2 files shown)


warmenhoven:
I should also say, though I haven't done a lot of thorough testing, I don't see any of the above issues on the standalone app.

libretro/CMakeLists.txt
```diff
@@ -22,7 +22,7 @@ if(ANDROID)
 endif()
 
 if(NOT MSVC)
-   if (IOS)
```
hrydgard, line 25: APPLE OR IOS seems a bit redundant? Well, doesn't really matter.

PR #16874, opened 2023-01-31
Browse button on Darwin platforms, rename some classes/functions

NSAntoine:


UI/DarwinFileSystemServices.h
```diff
@@ -1,12 +1,12 @@
...
+//  DarwinFileSystemServices.h
 //  PPSSPP
 //
 //  Created by Serena on 20/01/2023.
 //
 
-#ifndef DarwinMemoryStickManager_h
-#define DarwinMemoryStickManager_h
+#ifndef DarwinFileSystemServices_h
+#define DarwinFileSystemServices_h
```
hrydgard, line 9: Since you're changing this anyway, let's switch to #pragma once, not that it matters much.

UI/DarwinFileSystemServices.mm
```diff
@@ -44,44 +42,59 @@ - (void)documentPicker:(UIDocumentPickerViewController *)controller didPickDocum
...
 #if PPSSPP_PLATFORM(MAC)
         NSOpenPanel *panel = [[NSOpenPanel alloc] init];
         panel.allowsMultipleSelection = NO;
-        panel.canChooseFiles = NO;
-        panel.canChooseDirectories = YES;
-        panel.allowedFileTypes = @[(__bridge NSString *)kUTTypeFolder];
+        panel.canChooseFiles = allowFiles;
+        panel.canChooseDirectories = allowDirectories;
+//		if (!allowFiles && allowDirectories)
+//			panel.allowedFileTypes = @[(__bridge NSString *)kUTTypeFolder];
```
hrydgard, line 55: What's with this commented out code?

UI/MainScreen.cpp
```diff
@@ -534,7 +538,23 @@ UI::EventReturn GameBrowser::LastClick(UI::EventParams &e) {
...
+		
+		if (info.isDirectory)
+			SetPath(thePathChosen); // user selected a dir, go there
+		else
+			LaunchFile(screenManager_, thePathChosen); // it's a file, try open it
+	};
+	
+	DarwinFileSystemServices services;
+	services.presentDirectoryPanel(callback, /* allowFiles = */ true, /* allowDirectorites = */ true);
+#else
```
hrydgard, line 555: I would prefer if we didn't have to call this from here directly, and instead could do this from the System_SendMessage handler, the "browse_folder" case. Like for Windows in https://github.com/hrydgard/ppsspp/blob/master/Windows/main.cpp#L397 .

hrydgard (changes requested):

UI/MainScreen.cpp
SerenaKit: > I would prefer if we didn't have to call this from here directly, and instead could do this from the System_SendMessage handler, the "browse_folder" case. Like for Windows in https://github.com/hrydgard/ppsspp/blob/master/Windows/main.cpp#L397 .

What about getting the needed `screenManager`? And how would we call `SetPath` / `LaunchFile`?

UI/DarwinFileSystemServices.mm
SerenaKit: Oops, forgot to remove this

UI/MainScreen.cpp
hrydgard: You'd send "browse_fileSelect" or "browse_folderSelect" messages from the callback, then those would get handled by the existing handlers that MainScreen already has.

hrydgard (approved):

PR #16878, opened 2023-01-31
Restore settings to default - let you choose what to reset

### LINKED ISSUE #16381: Add options when reseting PPSSPP settings to default
### What should happen

For example selecting parts that will be reset.

- [x] Settings
- [ ] Controller mappings
- [ ] Recent history

### Who would this benefit

People who don't want to loose their recent history, especially when testing for bugs, player often asked to reset their settings.

### Platform (if relevant)

_No response_

None

_No response_

### Checklist

hrydgard:
image

Simple. Fixes #16381, though could be made more granular.

Also does some refactoring so commits are to be reviewed separately.

(though need to tweak the layout a little...)


Felipefpl:
This is one of the best settings i saw. Never saw this in any emulator/program before. I hope this becomes a trend. WAY BETTER than deleting the .ini file and start to configurate everything again.

iota97:
I have no way to test this rn, but does it works? I mean if you delete Game settings it nuke the ini with the recent iso too afaict, how do they survive the load call? Seems weird, but I'm probably just missing something :)

hrydgard:
Hm. You're right, this is not actually working correctly, I deluded myself to think it would because of the separate recent clearing in the RestoreDefaults function.

I'll fix it.

PR #16883, opened 2023-02-01
Update Simplified Chinese

Narugakuruga:
Use 🛆☐◯╳ to replace words. Not sure if this can display on phones 

I will test on my 7 year old Redmi 4 after merged


hrydgard:
"Use 🛆☐◯╳ to replace words" << what does this mean?

Narugakuruga:
> "Use 🛆☐◯╳ to replace words" << what does this mean?

Triangle Square Circle Cross. 
Apparently 🛆 is not available on old phones. I will change to △

hrydgard:
Heh. On Mac that looks like 

image

Narugakuruga:
Okay. so switch back to words now. Maybe it's some system font coding issue....

hrydgard:
I think those characters are just not very commonly supported ones.. Oh well.

unknownbrackets:
Yeah, those symbols are not widely supported.  🔺 is a triangle that's pretty widely supported but its color may be confusing.

You can search for them online, i.e. here: https://emojipedia.org/triangle-with-rounded-corners/ (which shows how limited the support is for that one.)

-[Unknown]

PR #16887, opened 2023-02-01
Restore defaults in a better way

hrydgard:
First, does some refactoring (ConfigSetting struct can be const since they only contain metadata), then implements restoring defaults without deleting the ini, by iterating through the settings.

And now restoring default without touching recents should actually work @iota97 :)

Do you see any problems with this?

### DIFF
--- a/Core/Config.cpp
+++ b/Core/Config.cpp
@@ -121,11 +121,11 @@ struct ConfigSetting {
 	typedef uint64_t (*Uint64DefaultCallback)();
 	typedef float (*FloatDefaultCallback)();
 	typedef const char *(*StringDefaultCallback)();
-	typedef ConfigTouchPos(*TouchPosDefaultCallback)();
+	typedef ConfigTouchPos (*TouchPosDefaultCallback)();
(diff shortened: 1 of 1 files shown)


Core/Config.cpp
```diff
@@ -1835,10 +1805,6 @@ void Config::RestoreDefaults(RestoreSettingsBits whatToRestore) {
 			ClearRecentIsos();
 			currentDirectory = defaultCurrentDirectory;
 		}
-
-		if (whatToRestore & (RestoreSettingsBits::SETTINGS | RestoreSettingsBits::CONTROLS)) {
-			Load();
```
iota97, line 1840: Still can't test, but this should probably be done if we nuke the "controller ini" right? Truth to be told we have a reset for this in control mapping screen, it's probably better to use that path here too.

PS: might need to call save after all of this just to be sure?

hrydgard: Seems to work just fine without this, something else is taking care of it somehow. I did test this time :)

Issue #16881, opened 2023-02-01
Suggestions for touch controls


BenjiBenji69:
### What should happen

Hello, sorry for the long winding post i just want to add some more minor suggestions, couple days ago i suggested to add analog stick controls into combo keys for touch controls, that is minor, most games have the option to switch it up but most of the time it completely messes up the control layouts, thing is those are just small part to make touch controls in PPSSPP to be truly mobile-like

First though is the issue i had with L and R, its completely useless for me to use on touch controls, i can move it to below the screens closer to action buttons but without toggled mode, its simply wonky and jank, most of the time i just set the buttons to combo keys toggled mode and disabled them in every games which makes L and R even more useless, and i can't seem to customize the default controls to make it toggled

![Screenshot_20230201_094050_PPSSPP](https://user-images.githubusercontent.com/99275631/215935136-d849c075-461d-40c2-bb6a-f879640c1061.jpg)

Second is tilting gestures, there should be an option to set it to tilting only horizontally (for dpad and analog) or all directions because i noticed how no matter i tilted it, due to it registering up and down directions (especially in GTA) it moves up and down also and there's no option to set it only tilting left and right directions

![Screenshot_20230201_094102_PPSSPP](https://user-images.githubusercontent.com/99275631/215935858-160d6d46-c358-4971-964d-b18692c28cb9.jpg)

Third is gestures, most of the time i register the dpad to the gestures options, its handy for certain games, while i set the touch controls for directions in combo keys like here in Burnout with the basic layouts

![ULUS10025_00017](https://user-images.githubusercontent.com/99275631/215936627-07403e12-066a-4055-ba4a-c9a2db4948af.jpg)
8cb9.jpg)

But there should also be an option to make it as customizable as custom keys (make it toggled, rapid fire etc) i know it could be buggy but worth a try, especially the "double tap option", an option to make it toggled, handy for certain button mashing games or games with awkward timing for QTE (very uncomfortable to button smash on a touch screens)

And couple more things for the suggestion,
Im sure some people already suggested this, but analog controls on a touch screen should have more space to it and more dynamic, im sure you know what im talking about, kinda like most mobile games analog, dynamic analog where one side is fully for controls and the other for anything else, you touch anywhere on the left side of the screen and analog controls can register in that area

[...]

PR #16889, opened 2023-02-01
Tilt improvements: Add visualizer, better defaults

hrydgard:
Helps #16881 

Turns out it was nearly impossible to understand how tilt controls worked without a visualizer, so I added one.

And it was revealed that my understanding was wrong too (I didn't write the code originally), especially around calibration.

I'll followup with a change that will make it work better held at an angle. This will remove the Base Position setting and replace with just calibration.

Anyway, here's the first improvement to the UI, which already makes it much easier to configure:

image


iota97:
There might be some minor issues with the translations, for example in Italian one:

```
Calibrate = Calibra il D-Pad
To Calibrate = Mantenere il dispositivo su una superficie piatta e premere "Calibra".
```

"To Calibrate" will ask to press "Calibra", but the button will actually be called "Calibra il D-Pad". Same with few other languages. Probably it's quite confusing it's referencing D-Pad here at all :/

assets/lang/ja_JP.ini
```diff
@@ -35,7 +35,7 @@ Button Binding = 対応ボタンの割り当て（複数同時設定可能）
 Button Opacity = ボタンの透明度
 Button style = ボタンのスタイル
 Calibrate Analog Stick = アナログスティックをキャリブレートする
-Calibrate D-Pad = 十字キーをキャリブレートする
+Calibrate = 十字キーをキャリブレートする
```
Narugakuruga, line 38: キャリブレート

assets/lang/lt-LT.ini
```diff
@@ -35,7 +35,7 @@ Button Binding = Button Binding
 Button Opacity = Mygtukų opacity
 Button style = Mygtukų stilius
 Calibrate Analog Stick = Calibrate Analog Stick
-Calibrate D-Pad = Kalibruoti krypčių mygtukus
+Calibrate = Kalibruoti krypčių mygtukus
```
Narugakuruga, line 38: Kalibruoti

assets/lang/tr_TR.ini
```diff
@@ -35,7 +35,7 @@ Button Binding = Düğme Bağlama
 Button Opacity = Tuş görünürlüğü
 Button style = Tuş stili
 Calibrate Analog Stick = Analog Çubuğu Kalibre Et
-Calibrate D-Pad = D-Pad'i kalibre et
+Calibrate = D-Pad'i kalibre et
```
Narugakuruga, line 38: kalibre

assets/lang/ca_ES.ini
```diff
@@ -35,7 +35,7 @@ Button Binding = Assignat de botons
 Button Opacity = Transparència dels botons
 Button style = Estil dels botons
 Calibrate Analog Stick = Calibrate Analog Stick
-Calibrate D-Pad = Calibrar la creueta
+Calibrate = Calibrar la creueta
```
Narugakuruga, line 38: Calibrar

PR #16890, opened 2023-02-01
Fix restore, was not working on my machine (TM)

iota97:
I tested and on my machine this was not resetting key mapping. Not really sure why it worked for @hrydgard and not for me tbh

In any case it's better to use this function I think :)


hrydgard (approved):

hrydgard:
Hm, I guess I just verified that I saw the file get deleted and added back :P

PR #16896, opened 2023-02-02
Tilt improvements 2: Fix/overhaul calibration, add more previews

hrydgard:
Followup to #16889

Helps #16881 

This further overhauls the tilt feature, making it quite a bit more usable (though for me, I feel it's still more of a cute demo than all that practical...)

* Math fixes, calibration fixes
* No more choice between horizontal and vertical, calibration to current angle supported instead
* Previews added for all mapping types (DPAD, action buttons, L/R)

TODO before merge:

- [x] Test portrait mode etc


Narugakuruga:
I think the sensitivity setting no longer work for D-pad? 
Some game use D-pad for camera. so it's good to have accurate D-pad control.

Another Idea: Use Tilt for Right Analog too. For PS3 HD games, right analog is used for camera too

PR #16901, opened 2023-02-03
OpenXR - Fix rendering on PicoOS 5.3

lvonasek:
This PR fixes the project for Pico 4 and Pico Neo 3 on PicoOS 5.3.

Quest 2 works fine. It might be slightly faster as it renders depthbuffer into renderbuffer instead of framebuffer texture.


hrydgard (approved):

hrydgard:
I guess the depth swapchain is if you want to enable clipping towards real world things if you have a laser depth buffer and a passthrough camera? For PPSSPP, indeed doesn't make sense to render to it.

lvonasek:
No idea, it worked as FBO depth buffer before the software update.

Issue #16899, opened 2023-02-03
Does ppsspp support gbm backend +vulkan?


zack-huangzihan:
### Game or games this happens in

games

### What area of the game / PPSSPP

In the code: https://github.com/hrydgard/ppsspp/blob/ceb4186d0ffc26091bb3fc25a1e71bce15b8902d/SDL/SDLVulkanGraphicsContext.cpp

We can see that there is some support for the back end: WINDOWSYSTEM_WAYLAND  WINDOWSYSTEM_XLIB  WINDOWSYSTEM_XCB WINDOWSYSTEM_METAL_EXT and so on, Can gbm+vulkan render?

Here's how I changed it：

```
diff --git a/SDL/SDLVulkanGraphicsContext.cpp b/SDL/SDLVulkanGraphicsContext.cpp
index 433f35b03..e08c30578 100644
--- a/SDL/SDLVulkanGraphicsContext.cpp
+++ b/SDL/SDLVulkanGraphicsContext.cpp
@@ -105,8 +105,9 @@ bool SDLVulkanGraphicsContext::Init(SDL_Window *&window, int x, int y, int mode,      
 #endif
 #endif
        default:
-               fprintf(stderr, "Vulkan subsystem %d not supported\n", sys_info.subsystem);
-               exit(1);
+               vulkan_->InitSurface(WINDOWSYSTEM_DISPLAY, nullptr, nullptr);
+               //fprintf(stderr, "Vulkan subsystem %d not supported\n", sys_info.subsystem);
+               //exit(1);
                break;
        }
```
But it can't work. Do you have any advice for me?

### What should happen

On a lot of embedded linux platforms, probably most of the way to use kms, hopefully can support

### Platform

Linux / BSD

### Mobile phone model or graphics card

arm mali

### PPSSPP version affected

latest

### Graphics backend (3D API)

Vulkan

### Checklist

- [X] Test in the [latest git build](https://buildbot.orphis.net/ppsspp/) in case it's already fixed.
- [X] [Search for other reports](https://github.com/hrydgard/ppsspp/search?q=my+issue&type=issues) of the same issue.
- [ ] Try resetting settings or older versions and include if the issue is related.
- [ ] Try without any cheats and without loading any save states.
- [ ] Include logs or screenshots of issue.

hrydgard:
GBM = https://www.reddit.com/r/linuxquestions/comments/ss4vhj/whats_the_gbm_api/ ?

[...]

PR #16921, opened 2023-02-06
vulkan: Add the function realization of arm device to kms vulkan

zack-huangzihan:
The implementation here is referenced：https://github.com/vanfanel/ppsspp
For the call to vulkan interface in kms on the arm linux system, I changed the changes inside to the form of dlopen，Let me explain how this code works:

First of all, in linux system, vulkan implementation is not open in the source code provided by gpu manufacturers, they use a hook called "vulkan loader", used to parse the ***.so of different manufacturers，It's usually "libvulkan.so"，So I need to read it with dlopen needed to sign, and then in "SDL/SDLVulkanGraphicsContext CPP" add a WINDOWSYSTEM_DISPLAY implementation.

These implementations are in Common/GPU/Vulkan/VulkanContext.cpp

See #16899 

Signed-off-by: ZiHan Huang <zack.huang@rock-chips.com>


hrydgard:
Thanks for implementing this!

Would like to change one thing though - we actually already do the `dlopen()` thing over in `VulkanLoader.cpp`, where we also load a lot of other functions. Maybe you could move the loading of the function pointers to there?

Common/GPU/Vulkan/VulkanContext.cpp
```diff
@@ -886,6 +889,197 @@ VkResult VulkanContext::ReinitSurface() {
...
+		VkDisplayKHR myDisplay = display_props[display_index].display;
+
+		// Get the list of display modes of the display
+		uint32_t mode_count = 0;
+		localGetDisplayModePropertiesKHR(phys_device, myDisplay, &mode_count, NULL);
+		if (mode_count == 0) {
+			_assert_msg_(false, "DISPLAY Vulkan couldn't find any video modes on the display");
+			return VK_ERROR_INITIALIZATION_FAILED;
+		}
+		mode_props = (VkDisplayModePropertiesKHR*)malloc(sizeof(VkDisplayModePropertiesKHR) * mode_count);
```
unknownbrackets, line 980: Rather than malloc/free, I'd recommend new/delete here, or even just a vector.  Both ways the code would look clearer:

```c++
mode_props = new VkDisplayModePropertiesKHR[mode_count];
...
delete [] mode_props;
```

Mainly, it's easier to be sure you didn't make a mistake with the type inside the sizeof since the cast could be different by accident.  That didn't happen, but could happen in the future if there was a `VkDisplayModePropertiesKHR2` or something.

-[Unknown]

SDL/SDLMain.cpp
```diff
@@ -768,6 +768,7 @@ int main(int argc, char *argv[]) {
 #if !PPSSPP_PLATFORM(SWITCH)
 	} else if (g_Config.iGPUBackend == (int)GPUBackend::VULKAN) {
 		SDLVulkanGraphicsContext *ctx = new SDLVulkanGraphicsContext();
+		mode |= SDL_WINDOW_VULKAN;
 		if (!ctx->Init(window, x, y, mode, &error_message)) {
 			printf("Vulkan init error '%s' - falling back to GL\n", error_message.c_str());
 			g_Config.iGPUBackend = (int)GPUBackend::OPENGL;
```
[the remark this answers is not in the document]
zack-huangzihan: > 

SDL_WINDOW_VULKAN  It must be in ". Init(window, x, y, mode, &error_message) "before, otherwise it will crash in the sdl library, I'm not sure why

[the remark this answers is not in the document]
unknownbrackets: Oh, my mistake - it's passed in as a parameter.  Can we use `if (!ctx->Init(window, x, y, mode | SDL_WINDOW_VULKAN, &error_message)) {` then?  I just want to ensure the GL init that's run when Vulkan fails works properly.

-[Unknown]

zack-huangzihan:
@hrydgard @unknownbrackets Ok, I modified it according to your suggestion

unknownbrackets (approved):

Common/GPU/Vulkan/VulkanContext.cpp
```diff
@@ -886,6 +886,185 @@ VkResult VulkanContext::ReinitSurface() {
...
+
+		display.displayMode = display_mode;
+		display.imageExtent = image_size;
+		display.transform = VK_SURFACE_TRANSFORM_IDENTITY_BIT_KHR;
+		display.alphaMode = alpha_mode;
+		display.globalAlpha = 1.0f;
+		display.planeIndex = plane;
+		display.planeStackIndex = plane_props[plane].currentStackIndex;
+		display.pNext = nullptr;
+#endif
```
unknownbrackets, line 1067: nit: I think you never delete display_props or plane_props.  Small leak.

-[Unknown]

zack-huangzihan: thanks, Have been repaired

hrydgard (approved):
Very nice, thanks!

PR #16944, opened 2023-02-09
Update fi_FI.ini

styxnix:
Translation work on [fi_FI.ini](https://github.com/hrydgard/ppsspp/blob/master/assets/lang/fi_FI.ini)


hrydgard:
Kiitos!

PR #16952, opened 2023-02-12
OpenXR - Enable stereo in more games

lvonasek:
This PR removes limitation for stereo "`only if UnitsPerMeter is specified`".

Let me explain why I think this is a good decision. Getting `assets/compatvr.ini` as a statistical coverage, I figured out:
* about 50% of games have the `UnitsPerMeter = 1.0` - this is our default value
* about 40% of games have the `UnitsPerMeter > 1.0` - these games will have a low stereo effect (*)
* about 10% of games have the `UnitsPerMeter < 1.0` - these games will have a wrong stereo (*)
 
(*) when the `UnitsPerMeter` value isn't specified

To sum up:
* before this PR we supported stereo on about 10% of all PSP games
* after this PR we will support stereo on about 50% of all PSP games (meaing the correct stereo separation is in place)
* after this PR the stereo will be too low on 40% of all PSP games (this is not worse than than without this PR)
* after this PR the stereo will be broken on 10% of all PSP games (I will be fixing these when reported)


hrydgard:
Yeah, this is probably a good plan. Bigger chance you'll get reports of failure cases too.

PR #16970, opened 2023-02-15
support libretro-VFS

hsoftxl:


hrydgard:
Okay, looks straightforward enough.

PR #16982, opened 2023-02-18
Fix a UI string (Rewind Snapshot)

Narugakuruga:
Rewind Snapshot Frequency →Rewind Snapshot Interval
based on #16981


hrydgard:
I think something went wrong with line endings here, because the diffs are showing the whole files.

Narugakuruga:
I don't know how to use git 🤔
Maybe it's because I changed the codec or something

PR #16994, opened 2023-02-24
Debugger: copy PSP memory base to clipboard

Nemoumbra:
Requesting the current PSP memory base address was only possible through memory.base remote API command.


hrydgard:
It's a bit of an obscure feature and I might later move it somewhere else, but it seems to have some uses so let's get it in.

Translation will be a little problematic, heh.

hrydgard (approved):

PR #16997, opened 2023-02-24
Retarget UWP projects for latest VS version

hrydgard:
19061 is the oldest SDK installed by default now.

Also disable "SDL" (security development lifecycle) checks in SPIRV-Cross because it doesn't build with it...


hrydgard:
Just the ubuntu test failing, getting this in.

PR #17004, opened 2023-02-25
Move more stuff to GPUCommonHW

hrydgard:
Followup to #17002.


hrydgard:
As this is just moving code around without changing functionality, I'm just gonna get it in.

PR #17012, opened 2023-02-26
(UWP) Small Build Overhaul

ghost:
Since I was bored, I decided to do some tweaks to the UWP build process again, here's what I have to offer this time:

# The Changes

- **Some small code cleanup**: Some files have some conditionals depending on what build configuration, and CPU architecture are used. Here's an example:

https://github.com/hrydgard/ppsspp/blob/e685dc910ff325393102745eb610538632899581/UWP/UWP.vcxproj#L546-L559

However, these conditionals are redundant since they basically tell the compiler to do the same action no matter the configuration or architecture. Since the result will be the same no matter what, let's just cut the bloat and tell the compiler to just do the action without the useless checks:

```
    <ClCompile Include="..\git-version.cpp">
      <PrecompiledHeader>NotUsing</PrecompiledHeader>
    </ClCompile>
```

The results are the same for both code snippets, but the second is much cleaner without the conditionals.

- **Missing Assets**: I noticed some of the files in the [assets](https://github.com/hrydgard/ppsspp/tree/master/assets) folder were missing from what a final packaged build includes, most notably the optional menu sound effects. So I decided to add these back, the sounds effects now work as they should too!

https://user-images.githubusercontent.com/35014183/221444310-50d35c7c-5330-46ef-98d6-8befd2809823.mp4

Alongside the sound files, the `he_IL_invert.ini` and `hr_HR.ini` language files were added, all of the missing shader files were added (although shaders are currently non-functional, see the screenshot below), the whole `debugger` folder was added, and the `font_atlas` zim and meta files were added.

UWP should now have all of the assets available (minus the `mime` folder).

**Shader error screenshot**:

221444657-8b0db1e7-eaf3-4b11-a226-0a5800cab59d

- **Incorrect Asset Paths**: The `ui_atlas.meta`, `ppge_atlas.meta`, and `defaultthemes.ini` files were not being copied correctly to the `Content` folder, instead being left at the root of the package. The `defaultthemes.ini` file being in the wrong location also meant that you could not select the Dark Theme as well. So these files paths were corrected, which is how I got the Dark theme to work in the video above.

- **Robocopy command**: This is the command executed during build time to copy the contents of the [assets](https://github.com/hrydgard/ppsspp/tree/master/assets) folder to the `Content` folder. It currently looks like this:

https://github.com/hrydgard/ppsspp/blob/e685dc910ff325393102745eb610538632899581/UWP/UWP.vcxproj#L490

However, since we don't use the `mime` folder in the final package, we can just tell the command to ignore this folder and just copy what we need:

`robocopy "..\Assets\." "Content\." /xd "mime" /E /R:0 /W:0`

# Conclusion

And that sums up what this PR has to offer, any comments and thoughts about any of the changes are welcomed.

### DIFF
--- a/UWP/UWP.vcxproj.filters
+++ b/UWP/UWP.vcxproj.filters
@@ -53,6 +53,21 @@
     <Filter Include="Content\themes">
       <UniqueIdentifier>{da2113a6-e616-4fc9-8be3-ff75b963664b}</UniqueIdentifier>
     </Filter>
+    <Filter Include="Content\debugger">
+      <UniqueIdentifier>{55147387-68fc-487e-abed-7a763f850ef1}</UniqueIdentifier>
+    </Filter>
+    <Filter Include="Content\debugger\static">
+      <UniqueIdentifier>{e5d2759b-3511-4ab9-a9ca-070dd9715f3b}</UniqueIdentifier>
+    </Filter>
+    <Filter Include="Content\debugger\static\css">
+      <UniqueIdentifier>{0fb972a9-20a9-4eba-be70-c408f9306939}</UniqueIdentifier>
+    </Filter>
+    <Filter Include="Content\debugger\static\js">
+      <UniqueIdentifier>{47d54a48-ca53-4d57-8bfd-9c86fd97d84b}</UniqueIdentifier>
+    </Filter>
+    <Filter Include="Content\debugger\static\media">
+      <UniqueIdentifier>{83476056-57b8-409f-b729-754d6401e3d5}</UniqueIdentifier>
+    </Filter>
   </ItemGroup>
   <ItemGroup>
     <ClCompile Include="App.cpp" />
@@ -101,6 +116,12 @@
     </Image>
     <Image Include="Content\asciifont_atlas.zim">
       <Filter>Content</Filter>
+    </Image>
+    <Image Include="Content\font_atlas.zim">
+      <Filter>Content</Filter>
+    </Image>
+    <Image Include="Content\debugger\favicon.ico">
+      <Filter>Content\debugger</Filter>
     </Image>
   </ItemGroup>
   <ItemGroup>
@@ -171,7 +192,13 @@
     <None Include="Content\lang\he_IL.ini">
       <Filter>Content\lang</Filter>
     </None>
+    <None Include="Content\lang\he_IL_invert.ini">
+      <Filter>Content\lang</Filter>
+    </None>
     <None Include="Content\lang\hu_HU.ini">
+      <Filter>Content\lang</Filter>
+    </None>
+    <None Include="Content\lang\hr_HR.ini">
       <Filter>Content\lang</Filter>
     </None>
     <None Include="Content\lang\id_ID.ini">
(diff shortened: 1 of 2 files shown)


unknownbrackets:
Isn't the debugger API available even on UWP?  Why wouldn't someone want to use it?

-[Unknown]

hrydgard:
Yeah keep the debugger, and the shaders as well - I'll look into that separately later, no reason it should be impossible to get them to work.

GABO1423:
> Yeah keep the debugger files, and the shaders as well - I'll look into that separately later, no reason it should be impossible to get them to work.

Sure, I'll get those back. Although the debugger files were never used in a finalized package anyway. Think I should add them just for the Debug Configurations or for all of them?

And back to the shaders, currently not all files are copied, want me to fix that and make all the files be copied?

GABO1423:
> Isn't the debugger API available even on UWP? Why wouldn't someone want to use it? Do people not debug things if they have an Xbox?
> 
> -[Unknown]

It's a valid use case, but not a lot of Xbox users even bother with debugging. Which is why I think no one even bothered to comment on how a UWP build never came with the debugger files at all. But I can add them no problem.

hrydgard:
As for the shaders, add 'em all, though the stereo shaders will likely not be useful, they're tiny.

And I think the debugger can be served from the built-in webserver, `ServeDebuggerFile` etc. Though it's not enabled by default, I think.

GABO1423:
> As for the shaders, add 'em all, although the stereo shaders will likely not be useful, they're tiny.
> 
> And I think the debugger can be served from the built-in webserver, `ServeDebuggerFile` etc. Though it's not enabled by default, I think.

I'll take as a yes for adding the debugger files regardless of configuration. All right, I'll add those files back then.

GABO1423:
All of the shaders, and the debugger folder have been added now! I think this should be good to go now, unless CI complains I guess...

UWP/UWP.vcxproj.filters
```diff
@@ -327,48 +429,81 @@
...
+    <None Include="Content\debugger\asset-manifest.json">
+      <Filter>Content\debugger</Filter>
+    </None>
+    <None Include="Content\debugger\index.html">
+      <Filter>Content\debugger</Filter>
+    </None>
+    <None Include="Content\debugger\manifest.json">
+      <Filter>Content\debugger</Filter>
+    </None>
+    <None Include="Content\debugger\static\css\main.3eab8a01.css">
```
unknownbrackets, line 480: Ah, it's unfortunate if the only way is to include each filename in the vcxproj.  These filenames will by default change every time there's an update.

-[Unknown]

GABO1423: You can thank how vcxproj files wanted to work, there's no easy workaround for that sadly.

GABO1423: Actually, I think I have an idea of how to fix. I'll push it soon.

UWP/UWP.vcxproj.filters
```diff
@@ -327,48 +429,81 @@
...
+    </None>
+    <None Include="Content\debugger\index.html">
+      <Filter>Content\debugger</Filter>
+    </None>
+    <None Include="Content\debugger\manifest.json">
+      <Filter>Content\debugger</Filter>
+    </None>
+    <None Include="Content\debugger\static\css\*.css">
+      <Filter>Content\debugger\static\css</Filter>
+    </None>
```
GABO1423, line 482: By replacing the file name with just the asterisk, it will copy any file that has that file extension regardless of the name. This should solve that issue quite nicely now.

hrydgard:
Nice, thanks!

PR #17023, opened 2023-02-28
(UWP) Fix Skipped Asset Files

ghost:
A follow up of my previous PR #17012, turns out these lines were also needed for some debugger files to be in the final package.
The complete debugger folder should now be available on the final build.


unknownbrackets (approved):

unknownbrackets:
Thanks.

-[Unknown]

PR #17031, opened 2023-03-01
Bring something back really

Halo-Michael:
It's basiclly like https://github.com/hrydgard/ppsspp/pull/15108 but won't break Xcode build.


Halo-Michael:
What did this do:
1. Bundle Settings.bundle currect.
2. Compile Launch Screen.storyboard with gnu-make.
3. Bundle Frameworks with Xcode currect.

Halo-Michael:
Xcode build not really work. See the error log:
280FA994-22E3-4273-B9DC-F2D59388863C
git-version.cpp does generated before xcodebuild running into that.
I have to running xcodebuild twice or create a git-version.cpp file before running xcodebuild.
gnu-make seems works fine.

Halo-Michael:
CF8A5564-34C9-490B-832D-14463850946A

hrydgard:
Closed? Gonna try a different way?

Halo-Michael:
> Closed? Gonna try a different way?

No, just because of that: https://github.com/hrydgard/ppsspp/pull/17031/commits/0c5cfd19054658e1ac2fa1b34bd5c1ee169bff7c

Halo-Michael:
I think we may need create git-version.cpp when running cmake not in building.

Halo-Michael:
Also git-version.cpp should generated in ${CMAKE_SOURCE_DIR} not ${CMAKE_CURRENT_SOURCE_DIR} due to macbundle.sh need that.

unknownbrackets:
> I think we may need create git-version.cpp when running cmake not in building.

Well, it sucks if cmake has to re-run literally every time to regenerate the project, though.  Otherwise it won't detect updated git versions.

I thought I just saw someone build this fine, although there were other problems (the iOS toolchain file didn't work and had to be essentially replaced with https://github.com/leetal/ios-cmake/blob/master/ios.toolchain.cmake with some tweaks - but it was the latest Xcode.)

-[Unknown]

Halo-Michael:
> replaced

But I didn't replaced iOS toolchain, it still works fine, and I'm using Xcode 14.2 (newest is 14.3 or 14.3.1, not very different)

unknownbrackets:
I'm not sure.  This was on an M2 host, and some of the build settings in Xcode were weirdly set to macOS or something until using the other toolchain file.  It was hitting weird link errors too.

-[Unknown]

Halo-Michael:
After cmake create project, I use command `xcodebuild build -project PPSSPP.xcodeproj -sdk iphoneos -configuration Release CODE_SIGNING_ALLOWED=NO` to build.

PR #17040, opened 2023-03-02
Update gradle

hrydgard:
Here we go again, with a new Android Studio version we get another update of the Android build system.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -22,7 +22,7 @@ dependencies {
 
 android {
 	flavorDimensions "variant"
-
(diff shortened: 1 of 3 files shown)


sum2012:
According to https://developer.android.com/studio/releases?hl=zh-tw#android_gradle_plugin_and_android_studio_compatibility
now need to use Android Studio Electric Eel | 2022.1.1 to build ?

hrydgard:
Not sure how hard a requirement it is. But shouldn't be much of an issue, just let it upgrade. It currently wants to auto-update to Electric Eel anyway.

PR #17046, opened 2023-03-03
(UWP) More Cleanups

ghost:
- Removes some more redundant conditionals, although a side effect is that the app won't longer be compiled with whole program optimization in the `Release` and `UWP Gold` platforms. Apparently PPSSPP does not benefit from this anyway. So no harm, no foul.
- Fixed some indentation for the `Command` element lines.

### DIFF
--- a/UWP/UWP.vcxproj
+++ b/UWP/UWP.vcxproj
@@ -64,68 +64,14 @@
     <WindowsTargetPlatformVersion>10.0</WindowsTargetPlatformVersion>
   </PropertyGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />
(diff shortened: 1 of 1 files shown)


hrydgard:
I wouldn't say it doesn't benefit at all from whole program optimization, but indeed, it's likely a very small benefit, like, a lot of time on the main emulation thread is spent in the JIT which is obviously not affected. Should be fine, anyhow.

unknownbrackets:
Well, it wasn't that recently, but I actually tried several years ago to compile everything under it and there was exactly zero FPS benefit.  I couldn't find anything really that was faster, and it made compiling slower.

Of course, I've written other code that has benefited from it but I don't think the patterns we use within PPSSPP benefit much.  Though, maybe it has changed.

-[Unknown]

PR #17053, opened 2023-03-04
(UWP) More Build Overhauls

ghost:
A lot of stuff was done here, so let's break down the changes one by one:

- **Kill all Redundant conditionals with fire**: This is the main change here, since the other UWP project files were marred by these redundant conditionals. The `CoreUWP` project file in particular was debloated significantly.
- **New Default Project file**: Before. the `CommonUWP` project file was the startup project for the UWP solution file, which does not make much sense considering you'd want to deploy the main UWP project file instead. So I changed the solution file to make the `UWP.vcxproj` file the startup project.
- **Package signing**: For some reason, this line: `<PackageCertificateKeyFile>PPSSPP_UWP_TemporaryKey.pfx</PackageCertificateKeyFile>` was removed from the `UWP.vcxproj` file, which meant that the pfx file went completely unused, since you'd have to create a new key every time you wanted to make an installable package. Adding this line back makes the pfx file actually do something again. And while I'm at it, let's update the file to one that expires next year instead of in a few months.
- **expandedResources**: I removed this capability since it seemed to cause a bug with the Xbox UI. Turns out the bug isn't caused by this capability, but is just a quirk present when using the application on Retail mode. Since the officially supported mode is Developer Mode, where this does not occur, I figured I might as well add it back.

I tested the build process carefully locally, but the CI builds will show if this actually works or not. So until I see what CI has to say about this one, I'm keeping this as a draft just in case more changes are needed.


GABO1423:
CI is happy now, so this is now ready for reviewing!

hrydgard:
Very nice, thanks for cleaning all this stuff up! I'm not sure exactly how the project files got to be that bad..

unknownbrackets:
When you change settings in Visual Studio, by default, it sets them per config even if you set the same thing for each config.  It generates all these conditionals.  So `<ConfigurationType>StaticLibrary</ConfigurationType>` for example is just a natural result of setting that in the UI instead of modifying the XML directly.

-[Unknown]

PR #17070, opened 2023-03-07
Updated brazilian portuguese

Felipefpl:
Updated all strings based on the several changes that were made.


hrydgard:
Great, thanks!

PR #17073, opened 2023-03-07
Use VFS for file access in the texture replacer

hrydgard:
In order to be able to add more texture formats, really have to get rid of this duplication, otherwise it's just too hard to write and debug.

There's a possible performance impact in the way we get width/height from pngs, but #17074 implements an even faster method anyway.

### DIFF
--- a/Core/TextureReplacer.cpp
+++ b/Core/TextureReplacer.cpp
@@ -136,7 +136,8 @@ bool TextureReplacer::LoadIni() {
 				iniLoaded = overrideIni.LoadFromVFS(*dir, overrideFilename);
 				if (!iniLoaded) {
 					ERROR_LOG(G3D, "Failed to load extra texture ini: %s", overrideFilename.c_str());
-					// TODO: Should we just move on instead?
+					// Since this error is most likely to occure for texture pack creators, let's just bail here
+					// so that the creator is more likely to look in the logs for what happened.
 					delete dir;
 					return false;
 				}
@@ -523,7 +524,6 @@ static ReplacedImageType Identify(VFSBackend *vfs, VFSOpenFile *openFile, std::s
 bool TextureReplacer::PopulateLevel(ReplacedTextureLevel &level, bool ignoreError) {
 	bool good = false;
 
-	// Everything in here is within a lock, so we don't need to relock.
 	if (!level.fileRef) {
 		if (!ignoreError)
 			ERROR_LOG(G3D, "Error opening replacement texture file '%s' in textures.zip", level.file.c_str());
@@ -553,6 +553,8 @@ bool TextureReplacer::PopulateLevel(ReplacedTextureLevel &level, bool ignoreErro
 			uint32_t flags;
 		} header;
 		good = vfs_->Read(file, &header, sizeof(header)) == sizeof(header);
+		level.w = header.w;
+		level.h = header.h;
 		good = (header.flags & ZIM_FORMAT_MASK) == ZIM_RGBA8888;
 	} else if (imageType == ReplacedImageType::PNG) {
 		PNGHeaderPeek headerPeek;
@@ -911,9 +913,9 @@ public:
 	}
 
 private:
+	VFSBackend *vfs_;
 	ReplacedTexture &tex_;
 	LimitedWaitable *waitable_;
-	VFSBackend *vfs_;
 };
 
 bool ReplacedTexture::IsReady(double budget) {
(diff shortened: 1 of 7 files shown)


Core/TextureReplacer.cpp
```diff
@@ -1181,31 +1053,13 @@ void ReplacedTexture::PrepareData(int level) {
...
-
-			if (!png_image_begin_read_from_memory(&png, &pngdata[0], pngdata.size())) {
-				ERROR_LOG(G3D, "Could not load texture replacement info: %s - %s (zip)", info.file.c_str(), png.message);
-				cleanup();
-				return;
-			}
-		} else {
-			_assert_(false);
+		pngdata.resize(fileSize);
+		vfs_->Read(openFile, &pngdata[0], fileSize);
```
unknownbrackets, line 1058: I still think it'd be more correct here to use `pngdata.resize(vfs_->Read(openFile, &pngdata[0], fileSize));` since the call to `png_image_begin_read_from_memory()` is more likely to error properly if the read is truncated somehow.

-[Unknown]

Core/TextureReplacer.cpp
```diff
@@ -553,123 +494,87 @@ enum class ReplacedImageType {
...
-			good = (flags & ZIM_FORMAT_MASK) == ZIM_RGBA8888;
-		}
+		struct ZimHeader {
+			uint32_t magic;
+			uint32_t w;
+			uint32_t h;
+			uint32_t flags;
+		} header;
+		good = vfs_->Read(file, &header, sizeof(header)) == sizeof(header);
+		good = (header.flags & ZIM_FORMAT_MASK) == ZIM_RGBA8888;
```
unknownbrackets, line 555: I think this is missing actually setting `level.w` and `level.h` to the loaded w/h values.  They're read but not used.  So this probably breaks zims.

-[Unknown]

unknownbrackets:
I'm also a bit worried that for Android users using non-zip packs (which they might still be, since those worked for a while before zips, although slow), this might increase IO and be slower (more GetFileInfo)... But it seems like zips ought to be okay...

-[Unknown]

hrydgard:
Yes, I'm considering removing that extra GetFileInfo again. I think we can make it work.

Core/TextureReplacer.cpp
hrydgard: Oops, fixing.

Core/TextureReplacer.cpp
hrydgard: Hm, fair enough. Though it's a buffered read, not a syscall read, so it should only be truncated if the actual file was truncated in the meantime, which isn't too likely. Still, I'll revert these.

Core/TextureReplacer.cpp
```diff
@@ -193,26 +132,26 @@ bool TextureReplacer::LoadIni() {
...
-				if (zip_.z) {
-					std::lock_guard<std::mutex> guard(zip_.lock);
-					iniLoaded = LoadIniZip(overrideIni, zip_.z, overrideFilename);
-				} else {
-					iniLoaded = overrideIni.LoadFromVFS(g_VFS, (basePath_ / overrideFilename).ToString());
-				}
+				iniLoaded = overrideIni.LoadFromVFS(*dir, overrideFilename);
 				if (!iniLoaded) {
 					ERROR_LOG(G3D, "Failed to load extra texture ini: %s", overrideFilename.c_str());
+					// TODO: Should we just move on instead?
```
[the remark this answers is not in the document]
hrydgard: Good point. Might even want to surface the error even more, actually, but I'll leave that for later.

Core/TextureReplacer.cpp
```diff
@@ -193,26 +132,26 @@ bool TextureReplacer::LoadIni() {
...
 					return false;
 				}
 			}
 		}
 	}
 
-	// The ini doesn't have to exist for it to be valid.
+	vfs_ = dir;
+
+	// The ini doesn't have to exist for the texture directory or zip to be valid.
```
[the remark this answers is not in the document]
unknownbrackets: Well, I thought of this when we started supporting zips.  I left not requiring textures.ini for directories because I didn't want to break BC.  That said, I think most texture packs do use textures.ini in some way so it probably wouldn't be that bad.  I think it's a reasonable BC break...

-[Unknown]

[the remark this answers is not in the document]
hrydgard: Oh, right, backwards compatibility..

No, I think I'm leaning towards printing a warning in this case instead (and disallowing ini-less texture zips).

hrydgard:
Alright, extra GetFileInfo gone, and added back the resize in png load.

PR #17086, opened 2023-03-09
OpenXR - Foveation implemented (Quest headsets only)

lvonasek:
https://developer.oculus.com/documentation/unity/unity-fixed-foveated-rendering/


hrydgard:
Should it be configurable, or did you choose a level where it's just not very noticeable?

hrydgard (approved):

lvonasek:
> Should it be configurable, or did you choose a level where it's just not very noticeable?

It is noticable only in screen recordings.

PR #17088, opened 2023-03-10
Replacer: Avoid tracking video textures

hrydgard:
Our texture replacer is not made for replacing video, so let's avoid bloating the cache and saved textures with video frames.

Also ignore a Vulkan validation message (yes, it's an ok one).


hrydgard:
Hm, just noticed we already have a mechanism for avoiding saving video textures, but it still tracks more info than we need.

Is there actually any case at all where we want to save them?

PR #17091, opened 2023-03-10
Replacement: Do all I/O on threaded tasks

hrydgard:
This should reduce I/O stutter during texture replacement considerably (especially on slow I/O like on Android 11+), no more I/O is done on the main thread by the replacer anymore except at startup.

~~We still open each file twice, I'll try to merge the Populate/Prepare steps later to avoid it.~~ Never mind, went ahead and merged them, now we only open each file once. We still first detect size, then pad the size if needed, rewind and read again.

I believe decimation of the "level cache" can still race against texture loading in some interesting ways. Need to look into that later.

### DIFF
--- a/GPU/Common/ReplacedTexture.cpp
+++ b/GPU/Common/ReplacedTexture.cpp
@@ -90,13 +90,13 @@ bool ReplacedTexture::IsReady(double budget) {
 
 	_assert_(!threadWaitable_);
 	threadWaitable_ = new LimitedWaitable();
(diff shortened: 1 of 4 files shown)


Gamemulatorer:
What happens to the recently build
Screenshot_20230311_192514
Screenshot_20230311_181432

hrydgard:
I have no idea, can't be related to this issue, i wouldn't think ...

Issue #17092, opened 2023-03-11
Texture replacement improvements for 1.15 (meta issue)


hrydgard:
# Plans for 1.15

- [x] Simplify state tracking (#17089)
- [x] Move all file reading to the thread, to avoid stutters even without zip. Avoid opening files twice (better performance) (#17091)
- [x] Bugfix basic mipmap support
- [x] Prepare for mipmapped texture file formats (reworked the cache a bit)
- [x] DDS support (desktop texture compression)
  - [x] Vulkan  (#17083)
  - [x] Other backends where possible
  - [x] Mipmaps
- [x] Universal texture compression - Basis/UASTC support through KTX [link](https://github.com/KhronosGroup/3D-Formats-Guidelines/blob/main/KTXDeveloperGuide.md). Will probably not bother supporting the .basis file format - "basisu.exe" can nicely generate .ktx files with both compression methods anyway, and it's likely the more future proof option.
- [x] Move texture size/format to the data cache (#17134, #17139)
- [x] Force mipmapping on, if present in replacement texture (#17144)
- [x] Verify thoroughly that clearing the cache doesn't corrupt textures or anything like that (see #16966)
- [x] Test and bugfix all of the above. There will be bugs :P
- [x] Document the new features

# Future plans

- [ ] Also support direct ETC/ASTC/BC through KTX (maybe)
- [ ] In Vulkan, offload image creation and the final upload to the thread, too! This is possible but will be a bit tricky with synchronization of the VRAM allocation.

Gamemulatorer:
Texture replacement in zip file format doesn't work using the recently build https://github.com/hrydgard/ppsspp/commit/8e5825b7af34e4dc9c11eeb714ab465badbe6f43

hrydgard:
Thanks, I'll check it out.

hrydgard:
Should be fixed now.

Gamemulatorer:
Replace texures disable/enable while in-game crashes the emulator.

```
F DEBUG   : Softversion: PD1930CF_EX_A_6.8.55
03-15 07:59:11.764 17235 17235 F DEBUG   : Time: 2023-03-15 07:59:11
03-15 07:59:11.764 17235 17235 F DEBUG   : *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
[...]

PR #17097, opened 2023-03-12
Texture Replacement: Support compressed textures in D3D9 as well

hrydgard:
D3D9 doesn't support BC7 but does support BC1-3, so I went ahead and added some format support checking, plus displaying supported formats in the system information screen.

Part of #17092 .


hrydgard:
This is pretty basic, I'll just get it in. I'll address the comments on the other PR in the next PR (that will add full DDS mipmap loading support)

PR #17098, opened 2023-03-12
OpenXR - Enable user to switch between topdown and fps camera

lvonasek:
This is a feature which the 2016 version of PPSSPP VR has and users were asking for that:
https://www.youtube.com/watch?v=gDrqVlCrlTs

### DIFF
--- a/Common/VR/PPSSPPVR.cpp
+++ b/Common/VR/PPSSPPVR.cpp
@@ -366,14 +366,14 @@ void UpdateVRInput(bool haptics, float dp_xscale, float dp_yscale) {
 			for (auto& axis : device.second) {
 				switch(axis.first) {
 					case JOYSTICK_AXIS_X:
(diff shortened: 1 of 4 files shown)


hrydgard:
That's an interesting trick, didn't know it had that. Just curious, how's game compatibility with this?

lvonasek:
It is basically just rotating the camera by 90 degrees. Users need to use the camera adjust feature to fit the camera view.

To sum up the game compatibility, it is any game where:
" has world unit scale defined or not extremely off
* game camera doesn't change the view too much
* culling is somehow acceptable
* geometry isn't 2D or missing

hrydgard:
Hehe, alright. Users are already presumably aware that running PSP games in VR has some limitations..

lvonasek:
Yeah, this feature is under VR settings group "Experts only".

PR #16856, opened 2023-01-28
Update Russian translation

TotalCaesar659:
P.S. ["Scope Storage Enabled"](https://github.com/hrydgard/ppsspp/pull/16605#issuecomment-1364669599) is not fixed yet.
P.P.S. Theme names are untranslatable too:

PR #17106, opened 2023-03-13
Add translation strings for dark theme and "Scoped Storage"

hrydgard:
Fixes the issues reported in #16856 .

I don't know that the string "Scoped Storage" is all that interesting to translate, it's a very technical Android-specific concept. But it was requested, so here we go.

@TotalCaesar659


hrydgard:
Oh, I misunderstood. Oh well.

PR #17131, opened 2023-03-16
[what][style][gpu] some coding style refine

HR1025:
[how]
1. use comment instead of unnecessary logic comparison
2. remove some useless code
3. move some implemention code from header to source
4. some more assert

### DIFF
--- a/Common/GPU/OpenGL/GLQueueRunner.cpp
+++ b/Common/GPU/OpenGL/GLQueueRunner.cpp
@@ -109,14 +109,14 @@ static std::string GetInfoLog(GLuint name, Getiv getiv, GetLog getLog) {
 	return infoLog;
 }
 
-int GLQueueRunner::GetStereoBufferIndex(const char *uniformName) {
+static int GetStereoBufferIndex(const char *uniformName) {
 	if (!uniformName) return -1;
 	else if (strcmp(uniformName, "u_view") == 0) return 0;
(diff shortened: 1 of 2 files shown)


HR1025:
I also see the memorandum for `GLRRenderData`:
```cpp
// TODO: Bloated since the biggest struct decides the size. Will need something more efficient (separate structs with shared
// type field, smashed right after each other?)
// Also, all GLenums are really only 16 bits.
```

I think meta programming container is a good practise. Something like `Poco::Any` , example follow as: `https://github.com/pocoproject/poco/blob/devel/Foundation/testsuite/src/AnyTest.cpp`.

It has some advantages:
1 -  safe type convert check (both copy and reference ability)
2 -  no memory align problem like union

> And its implementation is lightweight, requiring only a few hundred lines of code

hrydgard:
All looks good and correct, I'll merge.

As for metaprogramming, I'm not such a big fan of it. If I redo this stuff, I'll do something very custom to tightly pack the commands instead of the union, but it won't be through Poco::Any. too complex.

HR1025:
> All looks good and correct, I'll merge.
> 
> As for metaprogramming, I'm not such a big fan of it. If I redo this stuff, I'll do something very custom to tightly pack the commands instead of the union, but it won't be through Poco::Any. too complex.

Jsut an idea when I saw it. So I didn't change it.

hrydgard:
Yeah, I got that. Ideas are always welcome.

PR #17140, opened 2023-03-17
Change "Scoped Storage" values for more clarity

TotalCaesar659:
From "Yes"/"No" to "Enabled"/"Disabled".


hrydgard:
I suppose this makes it more readable, but it's not something most people will look at, is it really worth adding two new translation strings to all languages?

TotalCaesar659:
Not so much, it can be closed then, but changed again if there'll be another dialogue with these values.

hrydgard:
That I can agree on.

PR #17156, opened 2023-03-21
Remove graphics init/shutdown stuff from "Host"

hrydgard:
On most platforms it did nothing, and in Windows, it was trivial to just move the logic.

~Headless got slightly uglier (couple of ifdefs), though might clean that up later in a different way.~ Cleaned it up for now, the headless only gained a couple of ugly casts.

Followup to #17155

### DIFF
--- a/Core/Host.h
+++ b/Core/Host.h
@@ -27,9 +27,6 @@ class Host {
 public:
 	virtual ~Host() {}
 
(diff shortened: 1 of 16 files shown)


hrydgard:
Actually, don't like how this turned out. Gonna rewind a bit and simplify.

PR #17161, opened 2023-03-22
Rtc fix

### LINKED ISSUE #17119: Some pspautotests/tests/rtc/ failures
### Platform

Linux / BSD

cmake version 3.23.3, g++ (Debian 10.2.1-6) 10.2.1 20210110

### Operating system version

Linux

### Build commands used

```
./b.sh --headless
python3 test.py -g --graphics=software
```

### What happens

Was interested why it kept failing rtc tests:
```
Failed tests:
  rtc/arithmetic
  rtc/lookup
```

It looks like `time_t` is signed 32-bit integer on my system (32-bit Linux). And 2^32 seconds is just ~136.102208 years, hence failing bunch of tests, since `__RtcTicksToPspTime` and `__RtcPspTimeToTicks` just use `time_t` to adjust things (extra 400 years adjustment messes up?).

Logs of a test run:
[rtc.txt](https://github.com/hrydgard/ppsspp/files/10970548/rtc.txt)

Not sure if any games are affected.

### PPSSPP version affected

v1.12.3-5098-g3a6e48a9b

### Last working version

_No response_

### Checklist

fp64:
This *should* fix #17119 (it does on, my machine). Only briefly tested in actual emulator (outside autotests) - the date/time in saves seems sensible.

The code is also simplified, and less dependent on library functions. Some of the uses of `time_t` are still not purged. It should be possible to express `rtc_timegm` entirely in terms of `days_from_civil` (but change return type to `s64`), which is also more portable. Same for all uses of `gmtime`. All local/timezone stiff still needs standard library.

Not quite sure what the license for http://howardhinnant.github.io/date_algorithms.html is (the accompanying https://howardhinnant.github.io/date/date.html library is MIT). The text does say
> But the formulas are good. Consider these donated to the public domain.

So I hope a single-line comment attribution is enough.


hrydgard (approved):

fp64:
As an aside, I now get *different* pspautotests failures on my machine:
```
pspautotests/tests/cpu/vfpu/convert.prx:
O vus2i.p pattern 3: 007F8000, 7E678000, 7FC88000, 08800000
E vus2i.p pattern 3: 007F8000, 7E678000, 7F888000, 08800000
pspautotests/tests/gpu/bounding/viewport.prx:
O [x]   Y inside scissor negative: outside
E [x]   Y inside scissor negative: inside
O [x]   Y inside region negative: outside
E [x]   Y inside region negative: inside
```
Everything else passes.

fp64:
Well, I can tell that they were not happening in https://github.com/hrydgard/ppsspp/commit/dc4190ebf0fd4d7b4513a0493ce3c3bd01400750, but are happening in
https://github.com/hrydgard/ppsspp/commit/d87f03f90b80763763e13d5262529b8daba116a3.

The `pspautotests/tests/cpu/vfpu/convert.prx` seems like sNaN being quieted, which is finicky regardless - it depends on whether compiler decides to inline functions and/or spill variables.

Not sure `about pspautotests/tests/gpu/bounding/viewport.prx`.

hrydgard:
OK so here's the range:

https://github.com/hrydgard/ppsspp/compare/dc4190ebf...d87f03f

Mostly texture replacement and high level emulator framework stuff in there... Maybe an accidental change of FPU rounding mode or something? hm

fp64:
https://github.com/hrydgard/ppsspp/commit/d97790e28e2ac7e4e08d99cbd1148bf3e57872a5 explicitly affects `vi2us`/`vi2s`.

hrydgard:
Yeah, but only in the IR interpreter - which I don't think the tests run? or do they?

fp64:
Yes, fa1a4d26dc046e863f5923e8ea85048c0f8ab472 (i.e. d97790e28e2ac7e4e08d99cbd1148bf3e57872a5~1) also has these failing on my machine.

hrydgard:
Could you maybe do a quick gtt bisect then?

If you've never used it, do, it's awesome. basically:

* checkout the earliest known bad commit
* git bisect start ; git bisect bad
* checkout the latest known good commit
* git bisect good
* Then it'll automatically check out commits, and all you gotta do is tell it good or bad until the breaking one is found.

fp64:
Sorry, it seems I remembered wrong. Both dc4190ebf0fd4d7b4513a0493ce3c3bd01400750 and even earlier 4b73672b79807caa538a8a55f5a202250f6b78d4 *do* have these failures. #16984 does not, however.
If I simply add `-mfpmath=sse` (as I did in #16984) to 4b73672b79807caa538a8a55f5a202250f6b78d4, they are fixed.

That's not surprising for `pspautotests/tests/cpu/vfpu/convert.prx`, but a bit strange for `pspautotests/tests/gpu/bounding/viewport.prx`.

hrydgard:
Ah, interesting. Do add -mfpmath=sse, I don't see any reason to use traditional FPU as we require SSE anyway.

PR #17165, opened 2023-03-23
Add -mfpmath=sse to CMakeLists.txt

fp64:
Only affects x86+32bit+gcc combo.

Taken this stuff from #16984 into its own pull request.
Fixes (on my system) some pspautotests failures mentioned in #17161.
Left `-mstackrealign` there (but commented out) mostly to match #16984, but also as a remainder that it is a thing.


fp64:
Just a remainder, that clang neither understands, nor needs `-mfpmath=sse`.

unknownbrackets (approved):

PR #17168, opened 2023-03-23
Multiple key for single button

hrydgard:
Rebase of @iota97 's #14793.

I'm gonna go through it a bit more and add comments on how it works before merging.

### DIFF
--- a/Common/Input/InputState.h
+++ b/Common/Input/InputState.h
@@ -81,23 +81,25 @@ enum {
 // Represents a bindable key. Although it can be bound to two keys - you can bind key combinations as keys.
 class KeyDef {
 public:
-	KeyDef() : deviceId(0), keyCode(0), deviceId2(0), keyCode2(0), combo(false) {}
-	KeyDef(int devId, int k) : deviceId(devId), keyCode(k), deviceId2(0), keyCode2(0), combo(false) {}
-	KeyDef(int devId, int k, int devId2, int k2) : deviceId(devId), keyCode(k), deviceId2(devId2), keyCode2(k2), combo(true) {}
+	KeyDef() : deviceId(0), keyCode(0), deviceId2(0), keyCode2(0), dual(false) {}
+	KeyDef(int devId, int k) : deviceId(devId), keyCode(k), deviceId2(0), keyCode2(0), dual(false) {}
+	KeyDef(int devId, int k, int devId2, int k2) : deviceId(devId), keyCode(k), deviceId2(devId2), keyCode2(k2), dual(true) {}
 	int deviceId;
 	int keyCode;
+
+	bool dual;  // decides if deviceId2 and keyCode2 are valid.
 
 	int deviceId2;
 	int keyCode2;
 	
-	bool combo;
-
 	// If you want to use std::find and match ANY, you need to perform an explicit search for that.
 	bool operator < (const KeyDef &other) const {
 		if (deviceId < other.deviceId) return true;
 		if (deviceId > other.deviceId) return false;
 		if (keyCode < other.keyCode) return true;
 		if (keyCode > other.keyCode) return false;
+		if (dual < other.dual) return true;
+		if (dual > other.dual) return false;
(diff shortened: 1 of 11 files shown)


hrydgard:
Hm, it has a minor issue when mapping, and you just hold down one key for a bit:

image

hrydgard:
In the code there are a number of things called "combo"-something. I'm renaming these to dual, since we already have something called combo key mapping for touch screen controls.

hrydgard:
This is fine now except it doesn't work in the "Show PSP" preview in control mapping, since the code that loops through the keys to check for these dual keys isn't there.

Might refactor it a bit so we can re-use this detection.

Also "dual" doesn't quite feel right as a replacement for "combo", but it would be too confusing to have several "combo" concepts in the code base...

iota97:
FWIW we could rename the touch combo stuff to custom (as they arent even called combo anymore in the UI). Not sure if is worth the effort tbh :)

hrydgard:
That's actually a good point, these are more deserving of the combo name :)

Common/Input/InputState.h
```diff
@@ -78,24 +78,50 @@ enum {
...
+	KeyDef() : deviceId(0), keyCode(0), deviceId2(0), keyCode2(0), dual(false) {}
+	KeyDef(int devId, int k) : deviceId(devId), keyCode(k), deviceId2(0), keyCode2(0), dual(false) {}
+	KeyDef(int devId, int k, int devId2, int k2) : deviceId(devId), keyCode(k), deviceId2(devId2), keyCode2(k2), dual(true) {}
 	int deviceId;
 	int keyCode;
 
+	bool dual;  // decides if deviceId2 and keyCode2 are valid.
+
+	int deviceId2;
+	int keyCode2;
```
unknownbrackets, line 93: It kinda feels like these ought to be vectors instead, with the constructor pushing the single value for things like `hardcodedCancelKeys`.

Not saying we try to support it now, but it feels like if we need to redo any interfaces, it's better to design it so we could have three keys at once if we wanted.  I could see an argument for "ctrl + shift + space" being something.

-[Unknown]

UI/ControlMappingScreen.h
```diff
@@ -75,9 +76,12 @@ class KeyMappingNewKeyDialog : public PopupScreen {
 
 private:
 	int pspBtn_;
+	int devId_ = -2;
+	int keyCode_ = -2;
```
unknownbrackets, line 80: Did -1 have some special meaning?

-[Unknown]

unknownbrackets (commented):
Seems reasonable but thinking about the UI... it seems annoying to handle comboButton everywhere.

It seems like what would be most convenient is if some View or Screen getting a key event could just pass it to some class (not sure if global or specific) and ask for a list of `PSPKeyInput`s that represent PSP keys, if that's how it wants to think about things.

For example:

```c++
bool Key(const KeyInput &key) override {
	bool showInfo = false;

	for (auto pspKey : controlMapper_.ToPSPKeyInputs(key)) {
		if (pspKey.button == CTRL_TRIANGLE && (pspKey.flags & KEY_UP))
			showInfo = true;
	}
	if (hovering_ && key.deviceId == DEVICE_ID_MOUSE && key.keyCode == NKCODE_EXT_MOUSEBUTTON_2) {
		...
	}

	...
```

That would be clearer than having to consider `comboButton` all the time.

-[Unknown]

hrydgard:
I have come up with a different way to implement this, which I think will be a lot clearer. Just gotta get around to coding it up.

So no need to review this further right now.

hrydgard:
Hm. Now I'm considering trying to unify "Axis" and "Key" events all over the code base (where the latter simply has value = 1.0, and the former are split in to positive and negative events, like our mappings are currently...)

iota97:
About 3 button combination I guess nothing stops us from just register the key binds on the first key up. This would allow the user to press 3 button a time and just bind that one (any number tbh would work).

Probably the key name will be quite long with all the plus in between tho' :)

PR #17177, opened 2023-03-24
libretro: Bypass PPSSPP's resampler, since libretro already has one

hrydgard:
Now possible thanks to the recent refactorings.

Hopefully, this will fix the reported bad audio performance and behavior. Running two speed-compensating resamplers in a row can't be a good thing.

Not tested yet, I might set up retro stuff again, or I'll just go on the word of others.


hrydgard:
OK, apparently there are other problems too, but this cleans things up while not making anything worse, so I'm getting it on.

PR #17203, opened 2023-03-28
Added new option "Copy Float (32 bit)" to Windows Debugger UI

Nemoumbra:
image
Example behavior:
image


hrydgard (approved):
I'd probably have just used snprintf, but this is fine.

Nemoumbra:
> I'd probably have just used snprintf, but this is fine.

I really didn't like the fact that floats have multiple string representations and with snprintf you have to make a buffer of known size beforehand. I had no idea how to do that safely, so I just decided to go with something more high-level. std::to_string didn't work as it kept turning very small floats to 0.

PR #17211, opened 2023-03-30
Show a MIPS stack trace on crash screen

hrydgard:
~~Our stack tracer isn't very sophisticated (doesn't even look in ra for the first level, but that should be fixed), but why not at least try.~~  Actually, it works just fine, never mind.

Also write the stack trace to the log.

image

(bottom left)


unknownbrackets:
> doesn't even look in ra for the first level, but that should be fixed

Hm, what do you mean?  It scans up for the saved ra and uses `ra` if it doesn't find the entry, but thinks it's a leaf.  Obviously, the current value of ra outside a leaf is not super useful.

It's been there a while so of course it could have bugs, but I also haven't seen it be wrong (except within callbacks / interrupts / longjmps / etc. where it's not exactly "wrong") for a while, and step out uses it and pretty much always does what I want.

What case are you seeing where it's not working?

-[Unknown]

hrydgard:
Apologies, I take that back. During late night hacking I felt a frame was missing and got confused.

But it's working just fine.

I was also confused why we're not showing stack frames in the little drop down that I thought was intended for them next to PC in the debugger, but forgot that we have them in tab below instead, and that the drop down is just old Dolphin legacy that doesn't do anything. Shows how much I've used it recently... Argh.

hrydgard:
One thing that we could easily improve in the current stack display is that on lower frames, we could subtract 8 from the pc when doubleclicking. Like, right now, it shows you the future return address, rather than the actual jump two instructions above:

image

Can you think of any case where that would be wrong or unsafe? Seems to me that it'll always be correct to just subtract 8 from lower stack frame pcs.

unknownbrackets:
Well, that can be wrong but it'd be a pretty uncommon case.  Of course, it'll always be correct for a jal or bgezal, etc.

It's possible a game could set RA manually and then jump somewhere (like setjump or something.)  Probably these cases would confuse the stack walker anyway, though.  That said, I recall following a standard MIPS document when writing the stack walker so in theory the current behavior is as per standard...

That said, I think if we looked upward by 8 bytes and checked the op's `& OUT_RA`, the chances of it'd be wrong would very nearly approach zero.  And maybe it's a waste of time to even check that.

-[Unknown]

unknownbrackets:
Just to say, if we changed the results like that, we'd have to make sure to account for it in stepOut, etc.  It expects it to not have 8 subtracted, just the current PC in that frame.

-[Unknown]

hrydgard:
Maybe only subtract 8 in the UI layer then...

Issue #15561, opened 2022-05-29
[libretro] abysmal performance on android


andres-asm:
### Game or games this happens in

Tried several, RIDGE RACER in this case

### What area of the game / PPSSPP

Performance is a crapshoot on the libretro core now, but I don't think it's a performance issue per se, more like the core is implemented incorrectly or something like that because I get full speed sound but ~20fps video.

The core used to be fine ~6 months ago, I played ridge racer and peace walker to completion using this core.

Video of the issue (recording is not what is causing this at all)
https://photos.app.goo.gl/Fk5ydfiiLSBG7mnB7

### What should happen

Performance should be as before.

### Platform

libretro / Retroarch

### Mobile phone model or graphics card

Samsung Galaxy S22

### PPSSPP version affected

164c1b5

### Last working version

sadly the libretro versions are not stored

### Graphics backend (3D API)

Direct3D 9

### Checklist

- [X] Test in the [latest git build](https://buildbot.orphis.net/ppsspp/) in case it's already fixed.
- [X] [Search for other reports](https://github.com/hrydgard/ppsspp/search?q=my+issue&type=issues) of the same issue.
- [x] Try resetting settings or older versions and include if the issue is related.
- [x] Try without any cheats and without loading any save states.
- [x] Include logs or screenshots of issue.

ANR2ME:
Are you using Auto-frameskipping? if you are, try disabling it (ie. not using any frameskipping) and see if it makes a difference

hrydgard:
Could be related to https://github.com/hrydgard/ppsspp/pull/15541 ?

fr500:
no I don't think so, I have seen this for months, it was fine ~6 months ago but I didn't backup the core so it's hard to tell

Jetup13:
Didn't want to create another issue but I'm having the same issue with performance on the Android Libretro PPSSPP core. Doesn't matter what device I'm on (Retroid Pocket 2+, Retroid Pocket 3+, Samsung S10e SD855, Samsung Galaxy S22, ROG Phone 5) I seem to get around -25% to -30% performance hit on the core. 

Seen someone over on the Libretro discord mention it might of had a regression over time. Went back on what they mentioned that the last build that had better performance seems to be in RetroArch Plus (maybe -20% performance hit).  

Only reason why I want to use the core over the standalone is for RetroAchievements 😅

yudhiyou:
[...]

PR #17227, opened 2023-04-03
Libretro audio simplification

hrydgard:
There's just no reason to have a separate speed control mechanism when Libretro has one built-in. So just delete it.

(So in effect, libretro used to run with THREE audio speed control mechanisms bolted on top of each other, including PPSSPP's own..)

May help #15561.

### DIFF
--- a/libretro/Makefile.common
+++ b/libretro/Makefile.common
@@ -907,7 +907,7 @@ SOURCES_C += $(LIBRETROCOMMONDIR)/compat/compat_posix_string.c \
 endif
 
 GIT_VERSION_SRC = $(CORE_DIR)/git-version.cpp
-GIT_VERSION := $(shell git describe --always || echo v1.7.0-git)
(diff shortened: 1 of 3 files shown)


hrydgard:
@warmenhoven This might help.

yudhiyou:
I have just tested it and found an interesting info. Looked like this poor performance android ppsspp issue happened on spesific game. After set audio resampler to null, I tested on game Metal Slug XX and it still running normally. But when I tested it on game Street Fighter Alpha 3 Max it is running crazily.

yudhiyou:
I have tried the update. It is still same result on android phone. I set the audio resampler to null and the game is still running frame skipping everything like crazy.

hrydgard:
Right, libretro is not a focus for the major updates, it is not part of the main set of supported platforms (Windows, macOS, Android). I don't know what's wrong and I don't have time to fully supports libretro unfortunately, it's dependent on contributions.

PR #17228, opened 2023-04-03
Add back our older VFPU approximations, as fallbacks if the table files are missing

hrydgard:
PR #16984 added more accurate versions of these functions, but they require large lookup tables stored in assets/.

If these files are missing, PPSSPP would simply crash, which isn't good. It's very common for assets files to be missing when running as a libretro core, because those users just don't seem to be in the habit of copying these when they install a core.

We should probably try to warn the user somehow that these files are missing, though...

This also fixes deployment of the table files on UWP builds (some additional files snuck in too, without me explicitly adding them.. might as well let them, they're just gonna get auto-added again later and don't do any harm)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2067,6 +2067,8 @@ add_library(${CoreLibName} ${CoreLinkType}
 	Core/MIPS/MIPSTables.h
 	Core/MIPS/MIPSVFPUUtils.cpp
 	Core/MIPS/MIPSVFPUUtils.h
+	Core/MIPS/MIPSVFPUFallbacks.cpp
+	Core/MIPS/MIPSVFPUFallbacks.h
 	Core/MIPS/MIPSAsm.cpp
 	Core/MIPS/MIPSAsm.h
 	Core/MemFault.cpp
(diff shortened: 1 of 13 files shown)


GABO1423:
The CoreUWP.vcxproj file has the `ForcedInclude` double line bloat again :/

hrydgard:
It's MSVC screwing it up, gah.

GABO1423:
> Not sure it's worth removing since it keeps coming back, but did it anyway.

Guess we'll just give up on it next time it comes back.

GABO1423:
> > Not sure it's worth removing since it keeps coming back, but did it anyway.
> 
> Guess we'll just give up on it next time it comes back. Maybe it's more ideal long term to edit the files manually? That way the MSVC quirks aren't a factor.

unknownbrackets (approved):
It's a bit of a shame that it's on each call, maybe would be better to have a func pointer or something.  But probably not a big deal, considering the overall overheads in the func.

-[Unknown]

hrydgard:
Yeah, I think I'll re-do how this works. But this at least fixes the crash.

Issue #17245, opened 2023-04-06
Pressing a button on Ayn Odin Lite instantly exits


hrydgard:
But not on some other emulators.

Apparently it's the "M3" button, and some apps handle it and don't exit. Wonder what Android keycode it might be?

Reported through e-mail:

> I'm using it on AYN ODIN LITE and I'm extremely happy with it except for one small thing. 
> Every time I press the button marked as M3 (i.e. the back one on the left side - see photo), the emulator exits. 
> Nowhere have I found an option to turn it off

> I tried everything you advised :) Different remapping buttons, default control settings and nothing helped. 
> When I push it always throws me to the launcher, no matter where I am in the game or in the menu within the emulator. 
> It does this for PPSSPP, Doplhin and AetherSX2 (except I'm primarily playing PSP games now, which is why I asked you :)) 
Retroarch, Reddream for example can't be turned off by this.
> From what I've found, in Launcher (Daijisho, Odin) the key is dedicated to return to the emulator's home screen (and I also didn't find a remapping option) and it seems that some emulators ignore it and some don't. 
> Odin has its own physical key for returning to the splash screen, and that works in every emulator, but it's above the control buttons (ABYX) and can't be "accidentally" pressed. 

![image](https://user-images.githubusercontent.com/130929/230294824-c6b7be9b-d881-49dc-8828-3a26356a2b07.png)

hrydgard:
Hm, according to this article, it's not M3, but M2, and there's some kind of global M2 key mapping feature: https://clouddosage.com/ayn-odin-lite-review/

slynger:
Thank you for any advice.

slynger:
Just to clarify, it is not the M3 button, as I originally emailed Mr. Henrik, but the M1 button. 

M2 can be mapped to a specific numeric value on Odin.

See description here: https://clouddosage.com/ayn-odin-lite-review/

Only some emulators (PPSSPP, AetherSX2) respond to pressing M1 and returning to the Launcher. 
Others (RetroArch, Redream) stay running.

hrydgard:
I need to know which keycode the M1 button sends. Can't find it on google but in this thread they discuss using various button mapping software to get it: https://www.reddit.com/r/OdinHandheld/comments/wtd87x/question_about_the_rear_buttonsm1_m2_on_the_odin/

Maybe you can try some and see if one of them shows the keycode?

slynger:
Ok. I'll try to find out when I get back from work and put the info here.

[...]

PR #17247, opened 2023-04-06
Always tell Android we handled the BUTTON_MODE key

hrydgard:
Otherwise, as seen in #17245, some devies can decide to do something weird with it, like exiting the app.


kon0071235 (commented):
aaaa

PR #17254, opened 2023-04-08
Update Russian translation

TotalCaesar659:
P.S. Found more untranslatable strings:
1) "Controls" - "Auto-hide buttons after seconds"

2) "System" - "Rewind Snapshot Interval"

3) "System" - "Restore PPSSPP's settings to default" - "Controls"


hrydgard:
Thanks, I'll fix the cases of untranslatability soon.

hrydgard:
I merged the translatability fixes, feel free to rebase this and do the last translations on top.

TotalCaesar659:
Updated, can be merged now.

> Hm, any way to shorten the integer scale factor string? That's gonna look a bit cramped..

It looks not bad on screen.

I wouldn't want to shorten this by cutting words or changing them with losing some meaning.
P.S. After merging, could you check some?
1) Strings [108](https://github.com/hrydgard/ppsspp/blob/178fe2726362f3f04c9b61a294dbb2cc0c2c5a57/assets/lang/en_US.ini#L108) and [1148](https://github.com/hrydgard/ppsspp/blob/178fe2726362f3f04c9b61a294dbb2cc0c2c5a57/assets/lang/en_US.ini#L1148)? Looks like they differ only by space and may be combined to one.
2) "Seconds" word is repeated, first one isn't needed now.

hrydgard:
Good point about the repeated seconds string, that's kinda silly. Changing all those translations will result in some grammar mistakes I'm sure though..

PR #17263, opened 2023-04-11
Debugger: sceKernelPrintf improvement, QOL adjustments

Nemoumbra:
1. Now sceKernelPrintf can print %f floats. Note: it doesn't use the floating registers (if anyone has some evidence of a real PSP game that prepares f0 or something before the call, please do show).
Example with a1 equal to 0x3f8ccccd:
image

2. The "Copy Float (32 bit)" menu item has been moved to its own section because I suspect some people might have memorized the layout and thus experience slight inconveniences due to it now being the last element. Well, that's no longer a problem. We may also later add other floating types here.

image

3. [This commit](https://github.com/hrydgard/ppsspp/commit/8a74d95eb2bfbd8c17c09036257c6ae6f85283b3) does not fix the recently added menu item so I decided to do that.
I have also renamed it to have consistent terms across the code: 
image

4. Lastly, check the notes here: https://en.cppreference.com/w/cpp/io/basic_stringstream/str

### DIFF
--- a/Core/HLE/sceKernelMemory.cpp
+++ b/Core/HLE/sceKernelMemory.cpp
@@ -19,6 +19,7 @@
 #include <string>
 #include <vector>
 #include <map>
+#include <sstream>
 
 #include "Common/Thread/ParallelLoop.h"
 #include "Core/CoreTiming.h"
@@ -1100,6 +1101,8 @@ static int sceKernelPrintf(const char *formatString)
 	char tempStr[24];
 	char tempFormat[24] = {'%'};
 	std::string result, format = formatString;
+	std::stringstream stream;
+	float f_arg;
 
 	// Each printf is a separate line already in the log, so don't double space.
 	// This does mean we break up strings, unfortunately.
@@ -1170,6 +1173,18 @@ static int sceKernelPrintf(const char *formatString)
 			snprintf(tempStr, sizeof(tempStr), "%08x", PARAM(param++));
 			result += tempStr;
 			++i;
+			break;
+
+		case 'f':
+			static_assert(sizeof(float) == 4, "sizeof(float) != sizeof(u32)!");
+
+			// Maybe worth replacing with std::bit_cast when (if) we move to C++20
+			std::memcpy(&f_arg, &PARAM(param++), sizeof(u32));
+			stream << f_arg;
+			result += stream.str();
+			
+			++i;
+			stream.str(std::string()); // Reset the stream
 			break;
 
 		default:
--- a/Windows/Debugger/CtrlMemView.cpp
+++ b/Windows/Debugger/CtrlMemView.cpp
@@ -584,7 +584,8 @@ void CtrlMemView::onMouseUp(WPARAM wParam, LPARAM lParam, int button) {
 			auto memLock = Memory::Lock();
 			std::ostringstream stream;
 			stream << (Memory::IsValidAddress(curAddress_) ? Memory::Read_Float(curAddress_) : NAN);
-			W32Util::CopyTextToClipboard(wnd, stream.str().c_str());
+			auto temp_string = stream.str();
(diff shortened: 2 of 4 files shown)


hrydgard:
Yeah seems likely that printf (varargs) parameters go through regular registers. I do seem to recall that floating point arguments to regular functions can go through fp registers, but didn't check now.

Core/HLE/sceKernelMemory.cpp
```diff
@@ -1172,6 +1175,18 @@ static int sceKernelPrintf(const char *formatString)
...
+		case 'f':
+			static_assert(sizeof(float) == 4, "sizeof(float) != sizeof(u32)!");
+
+			// Maybe worth replacing with std::bit_cast when (if) we move to C++20
+			std::memcpy(&f_arg, &PARAM(param++), sizeof(u32));
+			stream << f_arg;
+			result += stream.str();
+			
+			++i;
+			stream.str(std::string()); // Reset the stream
```
hrydgard, line 1187: To clarify my previous comment about this, here's what I would have done here:

```cpp
char temp[64];
snprintf(temp, sizeof(temp), "%f", f_arg);
result += temp;
```

Though, to be fair, yours is better "C++", mine is just reflective of my somewhat excessive obsession with avoiding allocations. Doing that is important when running PPSSPP in debug mode for things that happen a lot, but since this shouldn't be very spammy, doing it the C++ way is fine and I'm not even sure in reality that there's more than one excessive allocation happening here (the stringstream has to allocate a buffer, while my way allocates it on the stack which is cheaper).

hrydgard (approved):

fp64:
The raw `"%f"` (which both `operator<<` and [`std::to_string`](https://en.cppreference.com/w/cpp/string/basic_string/to_string) use under the hood) in general gives you "wrong" float (produces different bitpattern when fed to e.g. [`strtof`](https://en.cppreference.com/w/c/string/byte/strtof)). The simplest version that guarantees (assuming round-to-nearest rounding mode) float->string->float round-trip is `"%.9g"`. Not sure what the desired behaviour for `sceKernelPrintf` is, but for right-click menu I would expect "value-preserving" serialization.

As for `stringstream`... why not use aforementioned [`std::to_string`](https://en.cppreference.com/w/cpp/string/basic_string/to_string)?

hrydgard:
That's a good point about copy - roundtrippable would be good there.

fp64:
> Here I explained my choice with using the stringstream and not std::to_string()

Oh, yes, I was wrong: `operator<<` uses (by default) `"%g"` (see https://en.cppreference.com/w/cpp/locale/num_put/put for details) and [std::to_string](https://en.cppreference.com/w/cpp/string/basic_string/to_string) uses `"%f"`.

Still, `"%g"` is `"%.6g"`, which is not enough for round-trip.

Nemoumbra:
Of course, we're not done with sceKernelPrintf. I only added the %f flag, but no precision params are available as of now. I guess, this is a PoC.
This is what I think we can do next:
1. Add support for things like "%.2f". I think this will downgrade the current precision we have for %f, but that's how the PSP works and... we want PPSSPP to emulate it, don't we?
2. If sceKernelPrintf supports this %g, it should also be added.
3. Finally, if the precision can be improved without breaking the correctness, improve the "Copy Float (32 bit)" option.

fp64:
> float->string->float round-trip is "%.9g"

For those curious: the formula is $1 + \lceil p \times \log_{10} 2 \rceil$ (and $\lfloor (p - 1) \times \log_{10} 2 \rfloor$ for *longest* safe string->float->string round-trip, which is 6 for `float`, which has $p = 24$). See e.g. [this](https://www.exploringbinary.com/number-of-digits-required-for-round-trip-conversions/).

PR #17287, opened 2023-04-15
Remove confusing setting fAnalogLimiterDeadzone

hrydgard:
As mentioned in comments for #17286 , there was some confusion about this setting.

Let's just get rid of it instead.


hrydgard:
Actually I'm reconsidering this. I confused myself even more it seems - forgot that analog limiter in this case means the thing that lets you emulate an analog stick from digital control by reducing the magnitude of the stick value by pressing a key.

Proper fix upcoming.

PR #17305, opened 2023-04-18
[what][style][gpu] some coding style refine

HR1025:


hrydgard (approved):
All of this looks correct, and you found some unused code too to delete. Nice!

PR #17329, opened 2023-04-24
Upgrade SDL for Mac to latest, prepare for notarization

hrydgard:
This removes the option to use system SDL and enforces use of the bundled one, which simplifies things.

Simplifies the macbundle.sh script accordingly.

I tried a number of different instructions I found around the web, but this is the only way I could get it working.

If there's a better way, I'm all ears.

This is needed for #17327 , and also, will allow us to support smooth mouse scrolling properly on Mac too.

This also prepares for signing & notarization (which is done by separate scripts, not included here), and a Mac gold build

PR #17331, opened 2023-04-25
SDL: Enable precise (smooth touchpad) scrolling.

hrydgard:
Silky smooth menu scrolling with mac touchpads now.

Made possible by #17329


hrydgard:
Turns out the scale factor is way off on my high-dpi Linux machine. Might do an ifdef mac for now.

PR #17342, opened 2023-04-27
fix open in folder on macOS, avoid spawning the open command to open URLs

NSAntoine:


SDL/CocoaBarItems.mm
```diff
@@ -51,6 +51,16 @@ void initializeOSXExtras() {
     [[BarItemsManager sharedInstance] setupAppBarItems];
 }
 
+void OSXShowInFinder(const char *path) {
+    NSURL *url = [NSURL fileURLWithPath:@(path)];
+    [NSWorkspace.sharedWorkspace activateFileViewerSelectingURLs:@[url]];
```
hrydgard, line 56: this is fine since it's following the tab-ness of the rest of the file (which we should change at some point, but meh)

hrydgard (approved):

Issue #16423, opened 2022-11-23
(UWP) Feature Request: Allow for changing the PSP Folder directory


GABO1423:
### What should happen

Currently, the PSP folder is hardcoded to the app's LocalState folder. However, there's been an issue with the Xbox's Developer Mode that makes it so any installed app cannot access any internal folders. It can only access two locations: the LocalState folder and a USB drive (if there is one plugged to begin with). So it's now difficult to use game DLC since there is now no way to easily access the folder, not to mention how you now can't export save files, or save states for safe keeping or for use with another platform.

The solution would be to allow users to change the default PSP folder directory to a USB folder, which would give users an accesible and simple location to store these critical files.

### Who would this benefit

Xbox users who can't access the LocalState folder.

### Platform (if relevant)

UWP / Xbox

### Games this would be useful in

Any that have any DLC.

### Other emulators or software with a similar feature

The RetroArch UWP port allows users to basically store everything but the main .cfg file in a USB Drive.

### Checklist

- [X] Check the [latest git build](https://buildbot.orphis.net/ppsspp/) in case it's already implemented.
- [X] [Search for other requests](https://github.com/hrydgard/ppsspp/search?q=my+issue&type=issues) of the same feature.

GABO1423:
This is possible thanks to #17350, so it can now be closed.

Issue #15595, opened 2022-06-13
Add URI Scheme to PPSSPP UWP Standalone for Retropass & other Frontends


danprice142:
### What should happen

A rom with per game settings should be able to be launched from a Frontend using URI Scheme for UWP and when the rom is closed it should close PPSSPP and return to the Frontend.

CLI can to be used with various frontends like LaunchBox, EmulationStation, etc...
In UWP, apps can't be started through command line, so right now there is no easy way to integrate it with other apps on for example Xbox.
However, there is a way for UWP app to start another UWP app through URI activation:
https://docs.microsoft.com/en-us/windows/uwp/launch-resume/handle-uri-activation

### Who would this benefit

This would open up PPSSPP on Xbox to integrate with other 3rd party Frontends.

### Platform (if relevant)

Other

### Games this would be useful in

All

### Other emulators or software with a similar feature

https://github.com/retropassdev/RetroPass
https://github.com/dolphin-emu/dolphin/commit/cf15db5b52f51a0f103e37840c3c2ec195a6c18c
https://github.com/PCSX2/pcsx2/commit/6067163a43fc40480003fe7e8a932ffc4122a7a0
https://github.com/libretro/RetroArch/issues/13581
(https://github.com/basharast/RetrixGold)

### Checklist

- [X] Check the [latest git build](https://buildbot.orphis.net/ppsspp/) in case it's already implemented.
- [X] [Search for other requests](https://github.com/hrydgard/ppsspp/search?q=my+issue&type=issues) of the same feature.

danprice142:
Started a bounty for this feature https://app.bountysource.com/issues/112701813-add-uri-scheme-to-ppsspp-uwp-standalone-for-retropass-other-frontends

PR #17350, opened 2023-04-28
UWP Improvements

basharast:
Hi,

Here a suggested changes to improve the UWP version

changes were carefully added to avoid breaking other platforms, it will be applied only for UWP

I don't know what is used exactly in libretro core, so all changes excluded from libretro build

## Main changes
- Custom memory stick location 
- Launch options (File, URI)
- UWP storage ~layer~ manager

these changes may address: #15595  & #16423

## Custom memory stick

I used the same UI in Android to make user able to change the location to another place

I did my best to enable this feature without breaking anything else.

MStick

## Launch options

- User can now launch PPSSPP (UWP) using file
- or using URI `ppsspp:` with support to switch back to the caller
`ppsspp:?cmd="fullpath"` or `ppsspp:?cmd="fullpath"&launchOnExit=targetURI`

## ~UWP storage layer~

Major part of this layer pulled out due to not required by latest windows builds, only builds before 10.0.17134

## Features enabled

I enabled some features to make it almost similar to the desktop version such as browse folder

now it will generate logs ~near to~ inside PSP folder ~inside LOGS folder~ (not sure if this was enabled before but in case it wasn't)

## Need to more work [Solved]

~Some places still can be better, like the way to detect launch request~

~I didn't find a better and effective place other than Render (loop) to keep track these requests~

~it need to be moved to proper place in future~.

I hope you will find it easy to check and helpful

already tested on XBOX, Desktop without any problem.

Thanks.


hrydgard:
This is really cool, thanks for doing this!

I'll review it in a bit more detail tomorrow, and will probably end up merging it after the 1.15 release process (we don't yet do official UWP releases anyway).

Common/File/DirListing.cpp
```diff
@@ -67,6 +71,11 @@ bool GetFileInfo(const Path &path, FileInfo * fileInfo) {
 	WIN32_FILE_ATTRIBUTE_DATA attrs;
 #if PPSSPP_PLATFORM(UWP)
 	if (!GetFileAttributesExFromAppW(path.ToWString().c_str(), GetFileExInfoStandard, &attrs)) {
+#if !defined(__LIBRETRO__)
+	if (GetFileInfoUWP(path.ToString(), fileInfo)) {
+			return true;
```
hrydgard, line 76: nitpick: double deep indentation

Common/File/FileUtil.cpp
```diff
@@ -157,7 +160,18 @@ FILE *OpenCFile(const Path &path, const char *mode) {
 	}
 
 #if defined(_WIN32) && defined(UNICODE)
+#if PPSSPP_PLATFORM(UWP) && !defined(__LIBRETRO__)
+	// We shouldn't use _wfopen specificly for custom memory stick location
```
hrydgard, line 164: "specifically". Also, maybe can write a quick line about the reason.

Common/File/FileUtil.cpp
```diff
@@ -258,6 +272,10 @@ static bool ResolvePathVista(const std::wstring &path, wchar_t *buf, DWORD bufSi
 	if (getFinalPathNameByHandleW) {
 #if PPSSPP_PLATFORM(UWP)
 		HANDLE hFile = CreateFile2FromAppW(path.c_str(), GENERIC_READ, FILE_SHARE_READ, OPEN_EXISTING, nullptr);
+#if !defined(__LIBRETRO__)
+		//Use UWP StorageManager to get handle
+		hFile = CreateFileUWP(path, GENERIC_READ, FILE_SHARE_READ, OPEN_EXISTING);
```
hrydgard, line 277: So in this case, we first call CreateFile2FromAppW and then CreateFileUWP? Doesn't seem quite right, the CreateFile2FromAppW call should probably be moved into an #else here.

UWP/PackageNormal.appxmanifest
```diff
@@ -23,6 +23,21 @@
...
+        <uap:Extension Category="windows.fileTypeAssociation">
+          <uap:FileTypeAssociation Name="psprom">
+            <uap:SupportedFileTypes>
+              <uap:FileType>.cso</uap:FileType>
+              <uap:FileType>.bin</uap:FileType>
+              <uap:FileType>.iso</uap:FileType>
+              <uap:FileType>.elf</uap:FileType>
+            </uap:SupportedFileTypes>
+          </uap:FileTypeAssociation>
+        </uap:Extension>
```
hrydgard, line 36: not sure what happened to indentation here, but maybe doesn't really matter.

UWP/UWPHelpers/StorageExtensions.cpp
```diff
@@ -0,0 +1,316 @@
...
+		size = WideCharToMultiByte(CP_UTF8, WC_ERR_INVALID_CHARS, &segment[0], (int)segment.size(), NULL, 0, NULL, NULL);
+		std::string converted = std::string(size, 0);
+		WideCharToMultiByte(CP_UTF8, WC_ERR_INVALID_CHARS, &segment[0], (int)segment.size(), &converted[0], (int)converted.size(), NULL, NULL);
+		ret.append(converted);
+	}
+
+	return ret;
+}
+
+std::wstring make_wstring(const std::string& string)
```
hrydgard, line 79: we have utilities for this kind of stuff already, in Common/Data/Encoding/Utf8.h . I guess it can still make sense to have it here if you want to make the StorageManager code reusable between apps.

UWP/UWPHelpers/StorageExtensions.cpp
```diff
@@ -0,0 +1,316 @@
...
+std::string replace2(const std::string str, const std::string& from, const std::string& to) {
+	std::string cpy = str;
+	size_t start_pos = cpy.find(from);
+	if (start_pos == std::string::npos)
+		return str;
+	cpy.replace(start_pos, from.length(), to);
+	return cpy;
+}
+
+std::vector<std::string> split(const std::string s, char seperator)
```
hrydgard, line 123: Similarly, we have this in Common/StringUtils.h.

UWP/UWPHelpers/StorageExtensions.h
```diff
@@ -0,0 +1,93 @@
...
+// Telegram: @basharastifan
+// GitHub: https://github.com/basharast/UWP2Win32
+
+// Functions:
+// replace(std::string& str, const std::string& from, const std::string& to)
+// replace2(const std::string str, const std::string& from, const std::string& to)
+// split(const std::string s, char seperator)
+// 
+// isChild(std::string parent, std::string child)
+// isParent(std::string parent, std::string child, std::string childName)
```
hrydgard, line 13: I don't really understand why you repeat the function declarations in comments - they'll just quickly drift out of date when changes are made.

UWP/UWPHelpers/StorageFileW.h
```diff
@@ -0,0 +1,281 @@
...
+		return information;
+	}
+
+	// Get main storage file
+	StorageFile^ GetStorageFile() {
+		return storageFile;
+	}
+
+    time_t  filetime_to_timet(LARGE_INTEGER ull) const {
+		return ull.QuadPart / 10000000ULL - 11644473600ULL;
```
hrydgard, line 246: indentation problem

hrydgard (changes requested):
First round of comments!

Also, I'm wondering if all those libretro checks is needed in callers - maybe it's possible to have the storagemanager use dummy implementations in libretro mode, or something.

basharast:
Thanks man, I really appreciate your time to check this PR, 
applying the changes now, will submit them later today.
as for (string functions/helpers) I'm in very early stage in C/C++ it wasn't easy to learn that and understand the project structure in short time,
so I forgot to look in the code for possible matches, I noticed that later after I finished, I will replace these as well.
and yeah libretro check better to be moved to the storage manager instead of the current case.

basharast:
So apparently I had few things mixed up due to the old builds of windows and lack of information in MS docs
We actually don't need to go through the slow UWP async functions
once the folder/file is added to the future access list then:

## API direct access
it will be allowed for direct access using the API that made for UWP (those ends with `FromApp`)
this is allowed even without FileSystem enabled.

## Get StorageFile with full path
also functions such as `StorageFolder::GetFolderFromPathAsync` allowed to get files from outside local data
as long the target in the future access list

## PR
I had to pull out many changes, let me know if I should re-submit another pull and close this in case this going to cause confusing.

GABO1423:
> ## Code formating
> for some reason in VS lines look aligned and good, but in github some of them appear not aligned.
> 
The good old tabs vs spaces indentation issue, in GitHub a tab is equal to around 8 spaces (IIRC) while in Visual Studio a tab is much less than that. It's kind of annoying to deal with but you get used to it once you know about this.

UI/NativeApp.cpp
```diff
@@ -756,7 +786,11 @@ void NativeInit(int argc, const char *argv[], const char *savegame_dir, const ch
 
 	DEBUG_LOG(SYSTEM, "ScreenManager!");
 	g_screenManager = new ScreenManager();
+#if PPSSPP_PLATFORM(UWP) && !defined(__LIBRETRO__)
+	if (g_Config.memStickDirectory.empty() || (g_Config.bFirstRun && isLocalState(g_Config.memStickDirectory.ToString()))) {
```
unknownbrackets, line 790: Since UWP controls this (`g_Config.memStickDirectory = Path(memstickFolderW);`), why not just set it to blank there if this case matches?

I guess the intention here is to always show this prompt on first run, right?

-[Unknown]

UWP/App.cpp
```diff
@@ -14,8 +14,11 @@
 
 #include <ppltasks.h>
 
-using namespace UWP;
+#include <UWPHelpers/LaunchItem.h>
```
unknownbrackets, line 17: Please use `""` around files that are local to the project, and `<>` around system files.  This is our file.

-[Unknown]

UWP/App.cpp
```diff
@@ -235,9 +252,10 @@ void App::OnSuspending(Platform::Object^ sender, SuspendingEventArgs^ args) {
 	auto app = this;
 
 	create_task([app, deferral]() {
+		g_Config.Save("App::OnSuspending");
 		app->m_deviceResources->Trim();
-		deferral->Complete();
-	});
+	    deferral->Complete();
```
unknownbrackets, line 257: Spaces and tabs mixed here.  I'm not sure what IDE you're using, but look for a plugin that enables EditorConfig.  It'll make it easier not to use the wrong indentation style across projects - even when they have different rules.

Also, please prefer:
```c++
create_task([app, deferral]() {
	asdf();
	bar();
});
```

Over:

```c++
create_task([app, deferral]() {
	asdf();
	bar();
	});
```

-[Unknown]

UWP/App.cpp
```diff
@@ -277,7 +295,8 @@ void App::OnVisibilityChanged(CoreWindow^ sender, VisibilityChangedEventArgs^ ar
 		// MainScreen::OnExit and even App::OnWindowClosed
 		// doesn't seem to be called when closing the window
 		// Try to save the config here
-		g_Config.Save("App::OnVisibilityChanged");
+		// OnSuspending should now do the job
+		//g_Config.Save("App::OnVisibilityChanged");
```
unknownbrackets, line 299: If this isn't needed, I recommend removing the entire if.  Git will remember.

-[Unknown]

UWP/PPSSPP_UWPMain.cpp
```diff
@@ -37,6 +37,12 @@
 #include "UWPUtil.h"
 #include "App.h"
 
+// UWP Storage helper includes
+#include "UWPHelpers/StorageManager.h"
+#include "UWPHelpers/StorageAsync.h"
+#include <UWPHelpers/LaunchItem.h>
```
unknownbrackets, line 43: nit: Should be `""` style include.

-[Unknown]

UWP/PPSSPP_UWPMain.cpp
```diff
@@ -367,12 +392,15 @@ std::vector<std::string> System_GetPropertyStringVec(SystemProperty prop) {
 		tempPath.resize(sz);
 		result.push_back(ConvertWStringToUTF8(tempPath));
 
-		if (getenv("TMPDIR") && strlen(getenv("TMPDIR")) != 0)
+		// As per below 'getenv' isn't supported
+		// https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/getenv-wgetenv?view=msvc-170
+		/*if (getenv("TMPDIR") && strlen(getenv("TMPDIR")) != 0)
```
unknownbrackets, line 397: FWIW, this is used to tell a user there's a problem when their savedata is being stored in a temp folder (this happens if you launch the non-UWP build from within a zip file) since your savedata will get lost between runs.

We can just remove these lines for UWP and rely only on `GetTempPath()`.  It should be fine.

-[Unknown]

UWP/PPSSPP_UWPMain.cpp
```diff
@@ -462,48 +501,95 @@ void System_Notify(SystemNotification notification) {
...
 			}
+			else {
+				g_requestManager.PostSystemFailure(requestId);
+			}
+			});
+		return true;
+	}
+	case SystemRequestType::BROWSE_FOR_FILE:
+	{
+		std::vector<std::string> supportedExtensions = { ".cso", ".bin", ".iso", ".elf", ".zip" };
```
unknownbrackets, line 541: Probably PBP as well if we have ELF.

-[Unknown]

UWP/PPSSPP_UWPMain.cpp
```diff
@@ -462,48 +501,95 @@ void System_Notify(SystemNotification notification) {
...
+		switch ((BrowseFileType)param3) {
+		case BrowseFileType::INI:
+			supportedExtensions = { ".ini" };
+			break;
+		case BrowseFileType::DB:
+			supportedExtensions = { ".db" };
+			break;
+		case BrowseFileType::ANY:
+			supportedExtensions = {};
+			break;
```
unknownbrackets, line 551: I recommend using a `default:` here.  For defensive programming, the best would be:

```
switch ((BrowseFileType)param3) {
...
case BrowseFileType::BOOTABLE:
	break;
default:
	ERROR_LOG(FILESYS, "Unexpected BrowseFileType: %d", param3);
	return false;
}
```

This is called "defensive programming."  It's a technique where you assume other parts of the code will change, and the person who changes them will not realize your code also needs to change.  By logging and returning false, it becomes more likely someone will figure out the cause of the error instead of making them spend hours debugging why something doesn't work, not realizing it's because the case is missing here.

It's almost a certainty that this will eventually change in PPSSPP and the person who changes it will very likely forget to update UWP.

-[Unknown]

UWP/PPSSPP_UWPMain.cpp
```diff
@@ -528,7 +614,13 @@ bool System_MakeRequest(SystemRequestType type, int requestId, const std::string
 }
 
 void System_ShowFileInFolder(const char *path) {
-	// Unsupported
+	auto ver = Windows::System::Profile::AnalyticsInfo::VersionInfo;
+	if (ver->DeviceFamily == "Windows.Xbox") {
+		// Unsupported
+	}
+	else {
+		OpenFolder(std::string(path));
```
unknownbrackets, line 622: Technically, the provided path is to a file, and we should be opening its parent folder.  On non-UWP this prompts Explorer to open its parent folder and to highlight the specified file.  Is `OpenFolder()` going to be happy with a filename?

-[Unknown]

UWP/UWP.vcxproj
```diff
@@ -89,12 +89,14 @@
     <AppxAutoIncrementPackageRevision>True</AppxAutoIncrementPackageRevision>
     <AppxSymbolPackageEnabled>False</AppxSymbolPackageEnabled>
     <AppxBundle>Never</AppxBundle>
-    <PackageCertificateThumbprint>C8DEB388B9BC89D1DC61324E4E9D9FE6A796B7AA</PackageCertificateThumbprint>
```
unknownbrackets, line 92: People keep changing this line, I haven't really involved myself in it.  Is this change intentional?

-[Unknown]

ext/libzip/zip_source_file_win32_utf16.c
```diff
@@ -95,7 +119,8 @@ utf16_create_file(const char *name, DWORD access, DWORD share_mode, PSECURITY_AT
 
     return CreateFile2((const wchar_t *)name, access, share_mode, creation_disposition, &extParams);
 #else
-    return CreateFile2FromAppW((const wchar_t *)name, access, share_mode, creation_disposition, NULL);
+    HANDLE h = CreateFile2FromAppW((const wchar_t *)name, access, share_mode, creation_disposition, NULL);
+	return h;
```
unknownbrackets, line 123: This change doesn't look very necessary?  We should limit patches to ext/ code.

-[Unknown]

ext/libzip/zip_source_file_win32_utf16.c
```diff
@@ -41,6 +41,29 @@ static HANDLE __stdcall utf16_create_file(const char *name, DWORD access, DWORD
 static void utf16_make_tempname(char *buf, size_t len, const char *name, zip_uint32_t i);
 static char *utf16_strdup(const char *string);
 
+#ifdef MS_UWP && !defined(__LIBRETRO__)
```
unknownbrackets, line 44: Why would we want to exclude LIBRETRO?

Also, you may want to propose these changes to https://github.com/nih-at/libzip.

-[Unknown]

UWP/UWPHelpers/StoragePickers.cpp
```diff
@@ -0,0 +1,74 @@
...
+	auto folderPicker = ref new Windows::Storage::Pickers::FolderPicker();
+	folderPicker->SuggestedStartLocation = Windows::Storage::Pickers::PickerLocationId::Desktop;
+	folderPicker->FileTypeFilter->Append("*");
+
+	return concurrency::create_task(folderPicker->PickSingleFolderAsync()).then([](StorageFolder^ folder) {
+		auto path = ref new Platform::String();
+	    if (folder != nullptr)
+	    {
+			AddItemToFutureList(folder);
+	    	path = folder->Path;
```
unknownbrackets, line 29: Still some mixing of spaces and tabs in this file.  I recommend searching for "    " and cleaning up that way.  Should be quick.

-[Unknown]

UWP/UWPHelpers/StorageManager.cpp
```diff
@@ -0,0 +1,331 @@
...
+		}
+	}
+	if (!state) {
+		state = IsRootForAccessibleItems(driveName);
+	}
+	return state;
+}
+
+bool isWriteMode(const char* mode) {
+	return (!strcmp(mode, "w") || !strcmp(mode, "wb") || !strcmp(mode, "wt") || !strcmp(mode, "at") || !strcmp(mode, "a"));
```
unknownbrackets, line 146: These are more or less "flags", so the order is not actually fixed.  I recommend instead using:

```c++
return strpbrk(mode, "wa+") != nullptr;
```

Which is roughly equivalent to the more obviously named:
```c++
return std::string(mode).first_first_of("wa") != std::string::npos;
```

UWP/UWPHelpers/StorageManager.cpp
```diff
@@ -0,0 +1,331 @@
...
+	auto access = GENERIC_READ;
+	auto share = FILE_SHARE_READ;
+	auto creation = OPEN_EXISTING;
+	bool isWrite = isWriteMode(mode);
+	bool isAppend = isAppendMode(mode);
+
+	if (isWrite) {
+		access = GENERIC_WRITE;
+		share = FILE_SHARE_WRITE;
+		creation = isAppend ? OPEN_ALWAYS : CREATE_ALWAYS;
```
unknownbrackets, line 167: Actually, `"a"` means that all writes should be forced to the end of the file.

If the file was opened for "r+" (or rb+, or r+b, or rt+, or r+t), it would be writable but would not be created if it doesn't exist.  "w+" would create still.

-[Unknown]

UWP/UWPHelpers/StorageManager.cpp
```diff
@@ -0,0 +1,331 @@
...
+				auto folderPath = Path(sItem);
+				auto attributes = FILE_ATTRIBUTE_DIRECTORY;
+				File::FileInfo info;
+				info.name = folderPath.GetFilename();
+				info.fullName = folderPath;
+				info.exists = true;
+				info.size = 1;
+				info.isDirectory = true;
+				info.isWritable = (attributes & FILE_ATTRIBUTE_READONLY) == 0;
+				info.atime = 1000;
```
unknownbrackets, line 254: 0 might be a better "I don't know".

-[Unknown]

UWP/UWPHelpers/StorageAccess.cpp
```diff
@@ -0,0 +1,117 @@
+// UWP STORAGE MANAGER
+// Copyright (c) 2023 Bashar Astifan.
+// Email: bashar@astifan.online
+// Telegram: @basharastifan
+// GitHub: https://github.com/basharast/UWP2Win32
+
+#include "StorageAsync.h"
+#include "StorageAccess.h"
+#include "UWPUtil.h"
+#include <Common/File/Path.h>
```
unknownbrackets, line 10: nit: Should use `""` style include for our files.

-[Unknown]

UI/MainScreen.cpp
```diff
@@ -1167,11 +1167,11 @@ void MainScreen::CreateViews() {
 		}
 	}
 
-#if !PPSSPP_PLATFORM(UWP)
-	// Having an exit button is against UWP guidelines.
+//#if !PPSSPP_PLATFORM(UWP)
+	// Having an exit button is against UWP guidelines?.
```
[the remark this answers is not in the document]
basharast: No, but:
1- Emulators are not allowed anyway in the store (I see the exist button is important to have)
2- Instead of `CoreApplication::Exit()` we are using [TryConsolidateAsync](https://learn.microsoft.com/en-us/uwp/api/windows.ui.viewmanagement.applicationview.tryconsolidateasync?view=winrt-22621)

> This method is a programmatic equivalent to a user initiating a close gesture for the app view.

I remember that there was a remarks regarding `CoreApplication::Exit()` but not able to find that anymore.

UI/NativeApp.cpp
```diff
@@ -491,7 +495,33 @@ void NativeInit(int argc, const char *argv[], const char *savegame_dir, const ch
 	if (!System_GetPropertyBool(SYSPROP_ANDROID_SCOPED_STORAGE)) {
 		CreateDirectoriesAndroid();
 	}
-
+#elif PPSSPP_PLATFORM(UWP) && !defined(__LIBRETRO__)
+	Path memstickDirFile = g_Config.internalDataDirectory / "memstick_dir.txt";
+	auto pathResolved = Path(ResolvePathUWP(memstickDirFile.ToString()));
```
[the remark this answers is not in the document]
basharast: True, I was checking this yesterday and noticed that if I assigned `LocalState` location to `g_Config.internalDataDirectory` would be way better for other places too, I'm getting to understand the project's structure better by time.

UI/NativeApp.cpp
basharast: Yes, I think if `g_Config.bFirstRun` is enough here, I don't remember why I made the other checks

UWP/PPSSPP_UWPMain.cpp
basharast: Oh, then this need to be solved within `OpenFolder`

UWP/UWP.vcxproj
basharast: No, I don't know actually why VS is changing this, since I never touched the certificate at all.

Core/PSPLoaders.cpp
```diff
@@ -418,6 +418,11 @@ bool Load_PSP_ELF_PBP(FileLoader *fileLoader, std::string *error_string) {
 
 	std::string homebrewName = PSP_CoreParameter().fileToStart.ToVisualString();
 	std::size_t lslash = homebrewName.find_last_of("/");
+#if PPSSPP_PLATFORM(UWP) && !defined(__LIBRETRO__)
```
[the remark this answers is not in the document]
basharast: I was testing this on RetroArch (PSP) version with custom memStick location, and the path was with `\` so I don't know if that happen due to the changes I made.

Core/System.cpp
```diff
@@ -767,6 +767,9 @@ void InitSysDirectories() {
 	File::CreateDir(GetSysDirectory(DIRECTORY_SAVEDATA));
 	File::CreateDir(GetSysDirectory(DIRECTORY_SAVESTATE));
 	File::CreateDir(GetSysDirectory(DIRECTORY_SYSTEM));
+#if PPSSPP_PLATFORM(UWP) && !defined(__LIBRETRO__)
```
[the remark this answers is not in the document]
basharast: Some Xbox users asked for this actually, also for `Cheats` folder for easy access, to avoid creating it manually.
if it's unnecessary we can drop this, I don't see it that important, but since I don't use Xbox don't know if it's really needed.

UI/NativeApp.cpp
```diff
@@ -491,7 +495,33 @@ void NativeInit(int argc, const char *argv[], const char *savegame_dir, const ch
...
+			Path memstickPath(memstickDir);
+			if (!memstickPath.empty() && File::Exists(memstickPath)) {
+				g_Config.memStickDirectory = memstickPath;
+				g_Config.SetSearchPath(GetSysDirectory(DIRECTORY_SYSTEM));
+				g_Config.Reload();
+				INFO_LOG(SYSTEM, "Memstick Directory from memstick_dir.txt: '%s'", g_Config.memStickDirectory.c_str());
+            }
+			else {
+				ERROR_LOG(SYSTEM, "Couldn't read directory '%s' specified by memstick_dir.txt.", memstickDir.c_str());
+				if (System_GetPropertyBool(SYSPROP_ANDROID_SCOPED_STORAGE)) {
```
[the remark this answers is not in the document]
basharast: I can show question dialog, I totally missed this part of the code, will do a test soon.

UWP/UWPHelpers/StoragePickers.cpp
basharast: Sorry for asking this, but I'm a bit confused from the formating, in VS it look like below
Capture

is there any tool can help to fix these before submitting to GitHub?

unknownbrackets: You can press Ctrl-R, then Ctrl-W to show whitespace indicators.  I usually search for a tab and it highlights where the indentation is funny.  But that said, most of the time I just don't have this problem, except maybe when pasting code from outside or something.

-[Unknown]

UI/MemStickScreen.cpp
```diff
@@ -124,7 +137,11 @@ MemStickScreen::MemStickScreen(bool initialSetup)
 		}
 	} else {
 		// Detect the current choice, so it's preselected in the UI.
+#if PPSSPP_PLATFORM(UWP) && !defined(__LIBRETRO__)
+		if (g_Config.memStickDirectory == Path(GetLocalFolder())) {
```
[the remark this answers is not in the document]
basharast: Hmm, I guess I will use `g_Config.internalDataDirectory` which is will be assigned as `LocalState`
What is the deference between `g_extFilesDir` and `internalDataDirectory` usually?
can I just depend on `internalDataDirectory` or this variable is for `Installation` folder?

UI/MemStickScreen.cpp
```diff
@@ -666,7 +711,11 @@ UI::EventReturn ConfirmMemstickMoveScreen::OnConfirm(UI::EventParams &params) {
 				// TODO: Handle failure listing files.
 				std::string error = "Failed to read old directory";
 				INFO_LOG(SYSTEM, "%s", error.c_str());
+#if PPSSPP_PLATFORM(UWP) && !defined(__LIBRETRO__)
+				progressReporter_.Set(error.c_str()); // it's throwing exception with 'iz->T'
```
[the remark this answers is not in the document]
hrydgard: Good catch, that's pretty bad. Fixing.

UWP/UWPHelpers/StorageManager.cpp
basharast: After the commit below it should be more accurate, I didn't know there are too many cases, hopefully all covered in the new function

UWP/UWPHelpers/StorageManager.cpp
```diff
@@ -0,0 +1,384 @@
...
+	Platform::String^ freeSpaceKey = ref new Platform::String(L"System.FreeSpace");
+	Platform::Collections::Vector<Platform::String^>^ propertiesToRetrieve = ref new Platform::Collections::Vector<Platform::String^>();
+	propertiesToRetrieve->Append(freeSpaceKey);
+	Windows::Foundation::Collections::IMap<Platform::String^, Platform::Object^>^ result;
+	ExecuteTask(result, ApplicationData::Current->LocalFolder->Properties->RetrievePropertiesAsync(propertiesToRetrieve));
+	int64_t remainingSize = 0;
+	if (result != nullptr && result->Size > 0) {
+		try {
+			auto it = result->First();
+			auto sizeString = it->Current->Value->ToString();
```
basharast, line 360: I'm not sure if [RetrievePropertiesAsync](https://learn.microsoft.com/en-us/uwp/api/windows.storage.fileproperties.storageitemcontentproperties.retrievepropertiesasync?view=winrt-22621) works perfectly, very little information about it,
based on docs [System.FreeSpace](https://learn.microsoft.com/en-us/windows/win32/properties/props-system-freespace) should return: 

> The amount of free space in a volume, in bytes

when I tried to cast this value (`Platform::Object`) to `int64_t` (`safe_cast<int64_t>(it->Current->Value)`)
it throwing WinRT exception `0x80004002 No such interface supported`,  
so I used `ToString()` instead, I guess this is not correct
it always report "2GB" maybe because of `ToString()`, 
it need someone with better experience to check the conversion

If we can get this to work, it will help for other drives too.

UI/MemStickScreen.cpp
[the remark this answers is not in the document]
unknownbrackets: Normally, `g_extFilesDir` is only used by Android currently.  Android has this weird concept of "external" and "internal" storage, neither of which refer to USB or microSD storage.  It's essentially app-private storage.  From a UWP perspective, you could use both to mean `GetLocalFolder()`.

But I think the change to internalDataDirectory is fine too.

-[Unknown]

UI/MainScreen.cpp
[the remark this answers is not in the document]
unknownbrackets: Well, seems reasonable to me then.

-[Unknown]

Core/System.cpp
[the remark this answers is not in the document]
unknownbrackets: Well, I'm fine with it, I just think there's no reason for it to be different for UWP.

I do think cheats should be easier to import...

-[Unknown]

UWP/PPSSPP_UWPMain.cpp
unknownbrackets: Hm, wait, you didn't add `BrowseFileType::BOOTABLE`, so I think this will fail for that type of dialog currently.

-[Unknown]

UWP/UWPHelpers/StorageManager.cpp
unknownbrackets: Oh, I meant for the atime/mtime/ctime.  We might actually want isWritable to be true - this is only used by some games when they run `stat()`, but for that one `true` is probably safest if we don't know.

-[Unknown]

UWP/UWPHelpers/StorageManager.cpp
unknownbrackets: Hm, I'm not sure.  A quick search shows people using something more akin to `result["System.FreeSpace"]` in C#/etc.  What if you set a breakpoint or log the FullName of GetType() on the object?

But I guess if we can count on it returning `#(.##)?[PTGMK]B` we can parse it worst case... although the resolution of only "2GB" isn't ideal.

-[Unknown]

Common/File/DiskFree.cpp
```diff
@@ -19,13 +19,25 @@
 #include <cinttypes>
 
 #include "Common/Log.h"
+#include "Core/Config.h"
 #include "Common/File/Path.h"
 #include "Common/File/AndroidStorage.h"
 #include "Common/Data/Encoding/Utf8.h"
 
+#if PPSSPP_PLATFORM(UWP)
+#include <UWP/UWPHelpers/StorageManager.h>
```
unknownbrackets, line 28: nit: Should use `""` style include.

-[Unknown]

Common/File/DiskFree.cpp
```diff
@@ -19,13 +19,25 @@
...
 bool free_disk_space(const Path &path, int64_t &space) {
 #ifdef _WIN32
 	ULARGE_INTEGER free;
+#if PPSSPP_PLATFORM(UWP)
+	if (path == g_Config.internalDataDirectory) {
+		space = GetLocalFreeSpace();
+	}
+	else
+#endif
+	// Is 'GetDiskFreeSpaceExW' returning wrong values in UWP?
```
unknownbrackets, line 40: No idea.  I mean, it's usable on UWP, but I don't know if it's allowed for all paths.  Would almost seem like a security issue if it "worked" for paths you don't have access to, as it might allow you to probe file existence...

-[Unknown]

UWP/PPSSPP_UWPMain.cpp
basharast: Oops, I shouldn't forget to add it back sorry about that

UWP/UWPHelpers/StorageManager.cpp
basharast: solved, I fetched the result in wrong way, it should be like
```cpp
auto value = result->Lookup(L"System.FreeSpace");
remainingSize = (uint64_t)value;
```

hrydgard:
I'll take another look soon, sorry for the long, long delay!

basharast:
> I'll take another look soon, sorry for the long, long delay!

No problem at all, take your time, let me know if there will be any further changes
Many thanks :)

hrydgard:
Alright, this is looking really good now, there are some nice fixes here.

Just wondering why indentation is now inconsistent betwee different XML files - some get their tabbing changed in this PR, making diffs hard to read (sure, there are workaround but still) and some don't.

With that fixed, I'm happy to merge.

basharast:
> Just wondering why indentation is now inconsistent betwee different XML files - some get their tabbing changed in this PR, making diffs hard to read (sure, there are workaround but still) and some don't.

Those are 'manifest' files, lines were aligned with spaces instead of tabs, so now will look different
I used VS auto formatting to and it fix and aligned the lines again with tabs instead

<details>
<summary>
Old manifest
</summary>

</details>

<details>
<summary>
New manifest
</summary>

</details>

Hope they are better now after this commit

hrydgard:
When looking at the overall PR diff, I still see them with a formatting change. But eh, it's not that important.

PR #17429, opened 2023-05-07
Revert zh_TW.ini

Kaitul:
Hi @Neo1102, all these sentences and terms are written in the Microsoft TC Style, so it is best not to replace some of them with Apple style or zh_CN style words.


hrydgard:
I'll trust that this is right. If anyone objects, please just file a counter pull request with an explanation.

Neo1102:
Hi Kaitul
Thank you for your feedback.
Personally, I also do not like to see zh-CN translations being directly applied to translations in zh-HK and zh-TW. If I make such a mistake, please let me know.
What are Apple style sentences and terms? Could you tell me more?

Hi hrydgard
Thank you. Please let me know if there are any issues with my translations.

Neo1102:
https://learn.microsoft.com/zh-tw/dotnet/api/system.windows.input.keybinding?view=windowsdesktop-7.0

I have consulted Microsoft's documentation and Microsoft does indeed refer to "Binding" as "繫結" in Chinese. However, in my memory, "綁定" is more commonly used in the gaming industry. I am currently unable to determine which one is more appropriate. If you prefer Microsoft's translation, you can change it back to "繫結".

Kaitul:
Hi @Neo1102, the usage of the terms is as follows:

>Binding
zh_TW: 繫結(MS) 綁定(Apple)
zh_CN: 绑定(MS) 绑定(Apple)
Layout
zh_TW: 版面配置(MS) 佈局(Apple)
zh_CN: 布局(MS) 布局(Apple)

Memory Stick™ does not have an accurate official translation, so in order to distinguish it from memory card, it is translated as "記憶棒".

Neo1102:
The Memory Stick Duo used by the PSP has a shape more similar to a card. In Taiwanese culture, some translations may refer to the shape of the object, and the "Stick" in "Memory Stick" may lead people to mistakenly think that the storage medium is rod-shaped. However, when it is inserted, it is more similar to an SD card, which is also translated as "記憶卡(memory card)" or directly referred to as "SD卡(sd card)" in Taiwan. Therefore, I believe that translating "Memory Stick" as "記憶卡" is a better choice.

Kaitul:
"記憶棒" is easier to understand the original text, it also can be seen as a special translation from Sony. A native English speaker sees a Memory Stick, which is also not rod-shaped.

Neo1102:
> Hi @Neo1102, the usage of the terms is as follows:
> 
> > Binding
> > zh_TW: 繫結(MS) 綁定(Apple)
> > zh_CN: 绑定(MS) 绑定(Apple)
> > Layout
> > zh_TW: 版面配置(MS) 佈局(Apple)
> > zh_CN: 布局(MS) 布局(Apple)
> 
> Memory Stick™ does not have an accurate official translation, so in order to distinguish it from memory card, it is translated as "記憶棒".

Thank you for your correction. However, the term "配置" is widely used in Simplified Chinese in China, and is used to refer to various things such as specifications, configuration, layout, and setup. In Taiwan, there is a more specific way to describe each of these things. Therefore, I will try to avoid using the term "配置" as much as possible.

Neo1102:
> "記憶棒" is easier to understand the original text, it also can be seen as a special translation from Sony. A native English speaker sees a Memory Stick, which is also not rod-shaped.

https://www.mobile01.com/topicdetail.php?f=672&t=655567
This is our Taiwanese discussion forum, and in 2008 we also used "記憶卡" to describe the Memory Stick. Sony's official translation may not necessarily be the commonly used term among Taiwanese people. For example, the current Chinese translation of Nintendo Switch is handled by the Hong Kong team, but there are cultural differences between Hong Kong and Taiwan, so some non-Taiwanese terms may appear in the Chinese interface of the Switch.

I can confirm that in Taiwan, "記憶卡" is a more commonly used translation than "記憶棒" for card-shaped storage media.

Kaitul:
> 

"配置" is also used in Taiwan, likes "平面配置圖" or "鍵盤配置", but the "配置" you said is offen translated from "config", which is translated in China but "設定(v)/組態(n)" in Taiwan.

Neo1102:
> > 
> 
> "配置" is also used in Taiwan, likes "平面配置圖" or "鍵盤配置", but the "配置" you said is offen translated from "config", which is translated in China but "設定(v)/組態(n)" in Taiwan.

You are correct that the term "配置" is also used in Taiwan. However, due to its overuse and misuse in China, I tend to avoid using it when I am unsure of the most appropriate translation. Instead, I would prioritize using other terms or technical jargon, such as "佈局".

Issue #17431, opened 2023-05-08
Nayuta Endless Trails - slowdowns due to unnecessary GPU readbacks


hrydgard:
Also known as Nayuta no Kiseki.

This is another game we can likely just add to [BlockTransferAllowCreateFB] or [IntraVRAMBlockTransferAllowCreateFB] in compat.ini - though like [Narikiri Dungeon X], there might be things that break, too - hard to say without testing through.

Reported by Mikio through e-mail.

Screen recording, showing the problem quite clearly:

https://user-images.githubusercontent.com/130929/236752641-b581ba1b-619a-4756-8054-ab1b84b86492.mp4

LunaMoo:
Dunno if it's just me, but the blur effect after disabling redbacks seems to be sticking weirdly and not looking as it should. Also the game probably wouldn't run as bad in the first place if the user wasn't playing while having 17% battery.

Crisanto47:
The lag is because of pixeldepthrounding compat removing this game id to that the game run smoothly tried and tested on my redmi 4x

hrydgard:
Ah, yeah that option is quite expensive. But doesn't it result in graphical errors to remove it? We should have added it for a reason...

Also, since you're experimenting, how about [BlockTransferAllowCreateFB] ?

Crisanto47:
> Also, since you're experimenting, how about [BlockTransferAllowCreateFB] ?

No performance improvement.

PR #17432, opened 2023-05-08
Add Nayuta Endless Trails to BlockTransferAllowCreateFB compat option

hrydgard:
Should get rid of the slow GPU readbacks, fixing #17431 (see that issue for video and details)

However, I have not tested this myself, so would be good if someone could give it a try.

Otherwise, the start of the 1.16 process is a good time to merge this to see if we get any complaints about anything else in the game getting broken, like savegame thumbnails or something..

It would also be ideal if we could try IntraVRAMBlockTransferAllowCreateFB instead (which has a tighter check).

### DIFF
--- a/assets/compat.ini
+++ b/assets/compat.ini
@@ -671,6 +671,12 @@ ULJM05160 = true
 # Narikiri Dungeon X. See issue #16714.
 ULJS00293 = true
 
+# Nayuta Endless Trails. See issue #17431.
+ULJM06113 = true
+NPJH50625 = true
+ULJM08069 = true
+NPJH50625 = true
+
 [IntraVRAMBlockTransferAllowCreateFB]
 # Final Fantasy - Type 0
 ULJM05900 = true


hrydgard:
Well, it's actually very clear from the supplied video, given what happens when GPU readbacks are turned off, that this is about a block transfer like this. But yes, testing through would obviously be the best way to verify that this is OK, and logs could strengthen the case.

LunaMoo:
Maybe compat options like that which aren't used for compatibility fixes should be reverted to speedhacks activated by UI with maybe mere suggestions which games can benefit from it instead of activating them for everyone for performance reasons based on mobile users reports since as we all know they often sacrifice everything for the sake of fps.

hrydgard:
Well, but allowing these copies to be read back to PSP RAM instead of creating a buffer at the copy destination and copying to that means reducing their resolution to 1x, which makes the experience both slower and worse-looking. By copying to a GPU buffer instead of reading back, we follow the actual intent of the game. Also, if the game had rendered to the copy destination before, we would have considered it a GPU buffer and skipped the readback anyway.

hrydgard:
I am not opposed to adding an option here, for the reasons you bring up. Naming it in a self-explanatory way will be tricky though... I have no good ideas there. Force GPU readback on block copy?

Of course, it'll still be useful as something we can tell users to toggle to avoid issues.

Maybe it should actually be three-state:

  1. Follow compat.ini
  2. Force creating framebuffers (avoiding the readback) even if not in compat.ini
  3. Force GPU readback on every block transfer

Such a three-stater would actually be even more useful as a user-toggle to diagnose stuff, I suppose.

Anyway, for the record in this specific case, the user has now played for a few hours with the compat setting enabled and everything seems tip-top. Of course, that's not solid proof, just an indication.

PR #17439, opened 2023-05-09
Revert "Merge pull request #16628 from hrydgard/remove-fog-fshader-flag"

hrydgard:
This reverts commit 10dee90c8311cbf0cd728e34011464aa3787f6d6, reversing changes made to 34c11c8acfa59cb19f5665f6304779f793c43b46.

Changes the fog back to be controlled by a fragment shader ID bit instead of by uniforms.

While this did reduce the shader count, I should have carefully benchmarked on older mobile GPUs - while not at all on newer GPUs, on ones like Mali-400, the hit is quite noticeable.

Same with the lmode flag, so I'm gonna have to revert that too. While risky just before a point release, these changes are fairly easy to see if they break..

Then for 1.16 I'll make a "fragment ubershader" flag, so we can choose between max fragment shader performance or minimizing the shader count for less stutter. Or maybe even compile specialized shaders in the background when using the ubershader..


hrydgard:
I'll do a little extra testing before merge btw.

PR #17440, opened 2023-05-09
Revert lmode variant reduction

hrydgard:
Followup to #17439 , see that for explanation.

With that and this, we recover most of the performance loss on Mali-400. We still keep some of the variant reductions.

I'll experiment a little more and see if we can recover all of it... would be nice.


hrydgard:
No, I think I'll stop here, we're almost there. For 1.16, I'll do like the vertex shaders and add a "fragment shader ubershader flag" so we can adjust between lowest shader count/longest shaders vs highest shader count / shortest shaders, based on the device we're running on. When we have the spare fragment shader power, optimizing for lowest shader count will provide the smoothest gameplay, but such is not the case on these very old devices.

PR #17442, opened 2023-05-09
Shrink the GLRRenderCommand struct from 152 to 88 bytes

hrydgard:
Turns out the VR work bloated it a bit, which can't be good. Though 88 bytes is also bloated..

Think it's fine to allocate these view matrices on the heap to get them out of the way, there won't be that crazy many per frame usually.


unknownbrackets (approved):

hrydgard:
I realized that this doesn't work with the current play-back-twice stereo implementation since it'll delete the matrix while rendering the first eye. Just needs some extra logic though.

PR #17477, opened 2023-05-17
OpenGL: Avoid rebind between glDrawArrays calls when possible

hrydgard:
Relatively profitable optimization for DrawArrays-heavy games like GTA on slow devices.

Can't do the same with glDrawElements (indexed draws) due to the lack of base vertex index. Well, some devices do support it as an extension (`glDrawElementsBaseVertex`). But there might be a better way by simply counting up the indices written to the index buffer.
 
Will do the same later in Vulkan but expecting a smaller benefit since rebinding the vertex buffer is less expensive there. I actually tried this on Vulkan before but had weird driver issues when applying a base vertex index.

Gonna get some numbers.

EDIT: Some very unscientific testing on an old Adreno (Nexus 5X), drawing in GTA gets about 20-30% faster (!) (though the framerate improves with ~15%). Though, the device has severe thermal throttling issues.

Now also tried it on Vulkan on Adreno and the difference is not really measurable so far, so maybe not worth it.

### DIFF
--- a/Common/GPU/OpenGL/GLFrameData.h
+++ b/Common/GPU/OpenGL/GLFrameData.h
@@ -39,6 +39,7 @@ struct GLQueueProfileContext {
 	bool enabled;
 	double cpuStartTime;
 	double cpuEndTime;
+	int drawArraysRebindsAvoided;
 };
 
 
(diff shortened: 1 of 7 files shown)


hrydgard:
It might also be possible to be smarter in DrawEngineCommon/DrawEngineGLES, and do some preparatory logic there to simplify (don't move the buffer binding, instead move first vertex index). Could avoid the divisions and help integrating it with similar tricks for indexed draws (which some games use extensively).

hrydgard:
Yeah, I have a redesign of this in mind that optimizes both array and element draws. So closing this, but will cherry-pick the first two commits in a different PR.

PR #17497, opened 2023-05-24
DrawEngineGLES: reduce buffer rebinds

hrydgard:
For some reason, not quite seeing the performance uplift I hoped to see given previous experiments like #17477. Will experiment more with this.

Works by "reserving" memory and then suballocating from it, computing vertex offsets as we go, instead of "allocating" every time. Works around the OpenGL deficit (lack of basevertex parameter) by nudging indices instead.


hrydgard:
Giving up on this optimization, closing the PR.

PR #17502, opened 2023-05-25
previous slot shortcut

nadenislamarre:


hrydgard (approved):

UI/GamepadEmu.h
```diff
@@ -307,6 +307,7 @@ namespace CustomKeyData {
 		{ ImageID::invalid(), VIRTKEY_REWIND },
 		{ ImageID::invalid(), VIRTKEY_SAVE_STATE },
 		{ ImageID::invalid(), VIRTKEY_LOAD_STATE },
+		{ ImageID::invalid(), VIRTKEY_PREVIOUS_SLOT },
```
iota97, line 310: Been a long time, but I think the saved config is based on the order of these arrays, new entry should probably be placed at the end to avoid breaking existing configs.

Should probably decouple the order and the configs eventually when I get the time...

PR #17506, opened 2023-05-25
Emulating HW vrnd

fp64:
The vrnd stuff from https://github.com/hrydgard/ppsspp/issues/16946 (see https://github.com/hrydgard/ppsspp/issues/16946#issuecomment-1467261209 onwards).

Still needs more testing, just putting it there to get overall comments.
Only tested in couple of games (including Ridge Racer, which is known to use vrnd). Seems to work fine, with no visual differences.
Bumps save state version for `MIPSState`, so save states are probably incompatible.

### DIFF
--- a/Core/MIPS/MIPS.cpp
+++ b/Core/MIPS/MIPS.cpp
@@ -269,7 +269,7 @@ void MIPSState::UpdateCore(CPUCore desired) {
 }
 
 void MIPSState::DoState(PointerWrap &p) {
-	auto s = p.Section("MIPSState", 1, 3);
+	auto s = p.Section("MIPSState", 1, 4);
 	if (!s)
 		return;
 
--- a/Core/MIPS/MIPS.cpp
+++ b/Core/MIPS/MIPS.cpp
@@ -306,6 +306,12 @@ void MIPSState::DoState(PointerWrap &p) {
 		Do(p, fcr0_unused);
 	}
 	Do(p, fcr31);
+	if (s <= 3) {
+		uint32_t dummy;
+		Do(p, dummy); // rng.m_w
+		Do(p, dummy); // rng.m_z
+	}
+
 	Do(p, inDelaySlot);
 	Do(p, llBit);
 	Do(p, debugCount);
--- a/Core/MIPS/MIPS.cpp
+++ b/Core/MIPS/MIPS.cpp
@@ -209,8 +209,6 @@ void MIPSState::Init() {
 	llBit = 0;
 	nextPC = 0;
 	downcount = 0;
-	// Initialize the VFPU random number generator with .. something?
-	rng.Init(0x1337);
 
 	std::lock_guard<std::recursive_mutex> guard(MIPSComp::jitLock);
 	if (PSP_CoreParameter().cpuCore == CPUCore::JIT) {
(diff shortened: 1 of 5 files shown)


Core/MIPS/MIPS.cpp
```diff
@@ -308,8 +306,6 @@ void MIPSState::DoState(PointerWrap &p) {
 		Do(p, fcr0_unused);
 	}
 	Do(p, fcr31);
-	Do(p, rng.m_w);
-	Do(p, rng.m_z);
```
hrydgard, line 312: Here, you need to check if the savestate version is 3 or below, in which case you still do the old steps, for example to local dummy variables. There are many examples around the code for this.

In this case it would simply be, I think:

```
if (s <= 2) {
   uint32_t dummy;
   Do(p, dummy);
   Do(p, dummy);
}

With that, old savestates should load fine (though new savestates won't load in old versions, but that's fine.)

}
```

Core/MIPS/MIPSVFPUUtils.cpp
```diff
@@ -769,6 +769,85 @@ float vfpu_dot(const float a[4], const float b[4]) {
...
+// The code below attempts to exactly match behaviour of
+// PSP's vrnd instructions. See investigation starting around
+// https://github.com/hrydgard/ppsspp/issues/16946#issuecomment-1467261209
+// for details.
+
+// Redundant currently, since void MIPSState::Init() already
+// does this on its own, but left as-is to be self-contained.
+void vrnd_init_default(uint32_t *rcx)
+{
+	rcx[0] = 0x00000001;
```
hrydgard, line 782: for new code we use this bracing style:

```
void vrnd_init_default(uint32_t *rcx) {
}
```

Core/MIPS/MIPSVFPUUtils.cpp
```diff
@@ -769,6 +769,85 @@ float vfpu_dot(const float a[4], const float b[4]) {
...
+	uint32_t C=(rcx[2] & 0xFFFFu) | (rcx[6] << 16);
+	uint32_t D=(rcx[3] & 0xFFFFu) | (rcx[7] << 16);
+	uint32_t E=(((rcx[0] >> 16) & 0xF) <<  0)|
+	           (((rcx[1] >> 16) & 0xF) <<  4)|
+	           (((rcx[2] >> 16) & 0xF) <<  8)|
+	           (((rcx[3] >> 16) & 0xF) << 12)|
+	           (((rcx[4] >> 16) & 0xF) << 16)|
+	           (((rcx[5] >> 16) & 0xF) << 20)|
+	           (((rcx[6] >> 16) & 0xF) << 24)|
+	           (((rcx[7] >> 16) & 0xF) << 28);
```
hrydgard, line 816: space around operators, including around = and |

hrydgard (changes requested):
This is great! Really only have a couple of code style tweaks to suggest, and a savestate compatibility fix.

I suppose you could also file an issue to https://github.com/jpcsp/jpcsp/issues and tell them that you've cracked it :)

Core/MIPS/MIPS.cpp
fp64: You mean `if (s <= 3)`, not `if (s <= 2)`?

hrydgard: Oh sorry yes, you are increasing from 3 to 4, not from 2 to 3.

fp64:
Fixed, I think.

hrydgard (approved):

hrydgard:
Are you working in the future btw? :)

image

Think you've got your timezone set wrong.

PR #17508, opened 2023-05-25
Merge drawcalls if they have the same output format

hrydgard:
Broken out from https://github.com/hrydgard/ppsspp/pull/17479

Seems to work just fine and still reduces the draws in GTA by 10%, but I'm not seeing much in the way of speed improvement in practice unfortunately :(   Not even on Nexus 5, where the GLQueueRunner submit loop is the # 1 bottleneck and an improvement would be expected.


hrydgard:
This just doesn't seem worth it. Closing.

Issue #14664, opened 2021-07-31
Carsh on Retroarch When set RenderMode as non-buffered rendering


xiejinglan:
Some game need set rendermode as  non-buffered  ,When i set is on Retroarch, it's carshed

Panderner:
@xiejinglan any games affected?

xiejinglan:
For example  <Mortal Kombat : Unchained> is Running too slow, And my cpu was Qualcomm snapdragon 865

xiejinglan:
For example  Mortal Kombat : Unchained  is Running too slow
And my cpu was qualcomm snapdragon 865

Gamemulatorer:
> Some game need set rendermode as non-buffered ,When i set is on Retroarch, it's carshed

Which games need non-buffered? 🤔

bslenul:
To add a bit more info...

Vulkan (full log: https://pastebin.com/kVdXD5dE):
```
[libretro ERROR] [SYSTEM] (..\Common/GPU/Vulkan/VulkanRenderManager.h:VulkanRenderManager::BindPipeline:155) Critical: [curRenderStep_ && curRenderStep_->stepType == VKRStepType::RENDER] *** Assertion ***
```
GL/GLcore (rest of the log is pretty much the same as Vulkan):
```
[libretro ERROR] [SYSTEM] (..\Common/GPU/OpenGL/GLRenderManager.h:GLRenderManager::SetNoBlendAndMask:764) Critical: [curRenderStep_ && curRenderStep_->stepType == GLRStepType::RENDER] *** Assertion ***
```
Black screen with D3D11 but audio can be heard in the background.

Only tested on Windows 10, I could try on my Linux VM if needed (GL only tho, it doesn't support Vulkan).

Panderner:
> For example Mortal Kombat : Unchained is Running too slow
> And my cpu was qualcomm snapdragon 865

Mortal Kombat: Unchained running smoothly for me. I'm using POCO M3 using buffered rendering. 
![Screenshot_2021-07-31-23-25-57-312_org ppsspp ppsspp](https://user-images.githubusercontent.com/37503397/127744790-5ad1ab70-c52e-4354-a3ce-e81d880f9027.jpg)

Panderner:
And same goes to RetroArch version
[...]

PR #17515, opened 2023-05-26
[Libretro] Remove "Skip Buffer Effects" core option

bslenul:
See #14664


hrydgard:
I'm sure someone is going to show up who runs libretro on some very slow device... But I'd just recommend that they switch to standalone if they want to use this hacky feature.

PR #17530, opened 2023-05-29
Draw/thin3d: Show the class name in refcount debug asserts

hrydgard:
Might help debug some refcount errors by telling which type it came from.

Ideally we should have stack traces in asserts, but alas...


hrydgard:
Argh, wait, this doesn't work, can't do virtual calls from the destructor... Will change it up.

PR #17532, opened 2023-05-29
Experiment with handling VK_LOST_DEVICE

hrydgard:
The idea is to gracefully teardown and recreate the Vulkan device if this happens due to either emulator or driver bug.

Currently, we don't fully implement LostDevice/RestoreDevice for Vulkan (as called from Android task switching), as in reality we do keep around the same device. So we don't delete/recreate everything, just most things, and finding the last things might be tricky.

However I was able to get this working to some degree on Intel GPU on Windows. On NVIDIA, I can't seem to create a new device in the same process after I've crashed one.

So overall, this is a bit questionable. Might be better to try to root out the remaining causes of lost device, although it would be nice to have this as a safety net.

### DIFF
--- a/Common/GPU/D3D11/thin3d_d3d11.cpp
+++ b/Common/GPU/D3D11/thin3d_d3d11.cpp
@@ -137,7 +137,7 @@ public:
 	void DrawUP(const void *vdata, int vertexCount) override;
 	void Clear(int mask, uint32_t colorval, float depthVal, int stencilVal) override;
 
-	void BeginFrame() override;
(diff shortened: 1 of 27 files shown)


FATCatAndroid12:
This is similar to dolphin emulator https://github.com/dolphin-emu/dolphin/pull/12191

K0bin:
@FATCatAndroid12 Not really. My PR was meant to help debug the cause of device lost errors so they can be fixed later rather than handle them.

PR #17534, opened 2023-05-30
Add a cache for MakePixelsTexture

hrydgard:
This allows us to not recreate the textures every frame for the purposes of DrawPixels, instead reuse them from previous frames. We don't use textures from the last 3 frames to be sure (should actually check the buffered frame count, perhaps).

We're still quite aggressive at garbage collecting these for now, currently after 10 frames but can easily tweak.

Previously we couldn't update the content of thin3d textures, but I've now added a path to do that.

Basically, it turns out that creating/destroying textures in Vulkan was more expensive than I thought, and in God of War which does a lot of DrawPixels for whatever reason, this is a really good speed boost, at least on PC, I'm seeing 20%. Working smoothly on mobile, too.

~~D3D11 implementation still missing, so draft.~~


hrydgard:
Oops, discovered I forgot some barriers on Vulkan. Will fix later today - but doesn't seem to affect it in practice on the devices I tried on.

PR #17545, opened 2023-06-04
VFPU: Some micro-optimizations. Don't fall back to interpreter path for vexp/vlog/vrexp.

hrydgard:
Overall this reduces CPU time in ReadVector/WriteVector by about 30% in GTA. They're actually somewhat hot in GTA (3-4% of execution time) due to the unknown prefix problem, forcing us to fallback a lot of first-vfpu-op-in-block to interpreter.

Just noticed this when doing a little profiling.

I think we can reduce the amount of asserts we have in the various VFPU decode functions like GetVecSize etc by the way, often we can never reach the INVALID case in practice. Added a bit of this.

### DIFF
--- a/Core/MIPS/JitCommon/JitBlockCache.cpp
+++ b/Core/MIPS/JitCommon/JitBlockCache.cpp
@@ -83,7 +83,7 @@ JitBlockCache::~JitBlockCache() {
 	Shutdown();
 }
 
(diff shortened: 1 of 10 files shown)


hrydgard:
Huh, I really don't understand why the Mac build has a problem linking, with simple inlined functions...

unknownbrackets (approved):

PR #17548, opened 2023-06-04
OpenXR - CI build for VR apk

### LINKED ISSUE #17065: Setup Orphis' Buildbot for VR builds
@hrydgard , @Ivonasek
### What should happen

Add dev VR builds to Buildbot

### Who would this benefit

VR community

### Platform (if relevant)

Android

Medievil Resurrection,Motostorm Arctic, etc...

_No response_

### Checklist

lvonasek:
An attempt to fix https://github.com/hrydgard/ppsspp/issues/17065. 

It seems this doesn't run until it is merged.

### DIFF
--- a/.github/workflows/manual_generate_vr_apk.yml
+++ b/.github/workflows/manual_generate_vr_apk.yml
@@ -1,0 +1,61 @@
+name: Manual Generate APK
+on:
+  workflow_dispatch:
+    inputs:
+
+      buildVariant:
+        type: choice
+        description: 'Build Variant'     
+        required: true
+        default: 'VrOptimized'
+        options: 
+        - VrOptimized
+        - VrDebug
+
+jobs:
+
+  apk:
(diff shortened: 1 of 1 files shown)


unknownbrackets:
Well, this would add a "manual" step to generate an APK, but the buildbot doesn't use GitHub Actions at all currently, and I don't know that anyone would be manually clicking this button:

image

It might be better to use a push trigger if you're goal is to make builds available.  Something like:

```
on:
  push:
    branches:
      - master
    tags:
      - v*
```

-[Unknown]

lvonasek:
Thanks, I somehow cannot figure out where the Android CI build is called from.

lvonasek:
I close this for now, will reopen it if I figure out how the Android building works.

unknownbrackets:
The Android building is handled on a separate server unrelated to GitHub Actions, buildbot.orphis.net.  It's got its own configuration which is not managed within GitHub which is why Henrik @'d him.

-[Unknown]

PR #17549, opened 2023-06-05
Fix vrnd to the current understanding

fp64:
Followup to #17506.


fp64:
Ah, so checks also get a whole lot of `/usr/include/features.h:194:3: warning: "_BSD_SOURCE and _SVID_SOURCE are deprecated, use _DEFAULT_SOURCE" [-W#warnings]`, not just my local build (well, it's `-Wcpp` locally).
I've seen a [recommendation](https://stackoverflow.com/questions/29201515/what-does-d-default-source-do) to use both `_BSD_SOURCE` and `_DEFAULT_SOURCE`, which [CMakeLists.txt](https://github.com/hrydgard/ppsspp/blob/master/CMakeLists.txt) does in one place, but not the other. Not sure if appropriate.

hrydgard:
I can't claim to understand how those defines are really meant to be used..

PR #17592, opened 2023-06-17
Use _mm_movemask_ps for AnyMask

fp64:
Probably very minor speed improvement, but it's rather neat.


unknownbrackets (approved):
Well, tricky things can happen with overall performance and this is extremely hot code, so it'd be best to test whether it's a tiny improvement or even a tiny harm.

That said, this is the non-SSE4 path so I'm not that concerned - especially because I don't currently have a device conveniently available to test real-world performance on pre-SSE4.

-[Unknown]

PR #17605, opened 2023-06-20
On-screen messages refactoring

hrydgard:
Redo how OSD messages are posted, allowing for more semantic control (like if it's an error, warning etc) and moves the data holder to Common, so it can have a nice public API accessible from anywhere, while rendering still will live in UI. Rendering can now happen on top of PPSSPP's menus too, it's no longer confined to EmuScreen.

Doesn't actually change the rendering yet, apart from that detail. 

(This will eventually be used for RetroAchievement (see #17589) popups and similar, in addition to the usual messages).

### DIFF
--- a/Common/System/Request.cpp
+++ b/Common/System/Request.cpp
@@ -2,8 +2,76 @@
 #include "Common/System/System.h"
 #include "Common/Log.h"
 #include "Common/File/Path.h"
+#include "Common/TimeUtil.h"
 
 RequestManager g_requestManager;
+OnScreenDisplay g_OSD;
+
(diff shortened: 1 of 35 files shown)


LunaMoo:
Could this be used to render info directly from the game(ie. in the future by custom cwcheat or alike) and refresh it every frame at specified position with choosen font and size?

hrydgard:
I suppose, but I think we'd want to do that a bit differently since the game display might be moved relative to screen space.

Instead we'd want to introduce a way to draw stuff onto the framebuffer before the post-processing chain (though I guess sometimes you might want to draw at the end of it instead..)

PR #17616, opened 2023-06-24
align should be 3 when format is GL_RGB and type is GL_UNSIGNED_BYTE

HR1025:
04b6d12435d7653fe994ca2315cd114

>according to `es_spec_3.2.pdf` - Table 8.2: Valid combinations of format, type, and sized internal format.


HR1025:
Also I found that GL_ LUMINANCE may not be applicable on certain ES3.

https://github.com/hrydgard/ppsspp/commit/ce79649c02edd1

Maybe this would be more appropriate.

hrydgard (approved):

PR #17625, opened 2023-06-26
[Nintendo Switch] Lower distance

m4xw:
First commit just does a fix-up for my last PR
Second commit brings in the required MemoryUtil changes for the Nintendo Switch platform
Could probably argue to move the Switch platform into its own MemoryUtil because of its simplicity.

Tried to fix all whitespaces, _assert_msg_ was changed due to compiler warn which doesnt seem to be Switch specific?


hrydgard (approved):

hrydgard:
Looks fine to me. Take it out of draft and I'll merge.

PR #17630, opened 2023-06-27
Common & CodeBlock changes for Nintendo Switch

m4xw:
- The first commit adds a implementation of W^X dualbase for the Nintendo Switch platform in `CodeBlock`.

- The second commit fixes CommonType for the Nintendo Switch platform, avoiding name conflicts and declaring certain types.

- The third commit adds the `cstring` include to the `ArmCPUDetect`, fixing a compilation error on some platforms.

- The fourth commit adds a placeholder implementation of the `Crash()` method for the Nintendo Switch platform.

### DIFF
--- a/Common/CodeBlock.h
+++ b/Common/CodeBlock.h
@@ -10,6 +10,11 @@
 #include "Common/Log.h"
 #include "Common/MemoryUtil.h"
 
+#if PPSSPP_PLATFORM(SWITCH)
+#include <cstdio>
+#include <switch.h>
+#endif // PPSSPP_PLATFORM(SWITCH)
+
 // Everything that needs to generate code should inherit from this.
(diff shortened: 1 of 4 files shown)


hrydgard (approved):

hrydgard:
If possible, another thing we really could use would be a github CI setup for Switch, so we always test-build for it.

Common/CodeBlock.h
```diff
@@ -66,8 +71,19 @@ template<class T> class CodeBlock : public CodeBlockCommon, public T {
 	void AllocCodeSpace(int size) {
 		region_size = size;
 		// The protection will be set to RW if PlatformIsWXExclusive.
```
m4xw, line 73: Minor nit, forgot to move the commend to the #else

PR #17634, opened 2023-06-27
Streamline x86 SSE workaround

fp64:
Seems clearer than using #ifdef's at each site. Also rationale is clearly spelled out, one 'Go to definition' away from any instance.


hrydgard (approved):
I think this looks alright, at least. Don't have any better ideas.

What do you think @unknownbrackets ?

unknownbrackets (approved):

Issue #17631, opened 2023-06-27
RetroAchievements: Remaining features to implement


hrydgard:
The main work was done in #17589 . Here's a checklist for additional and future development.

- [x] Leaderboards UI: #17629 
  - [x] Show avatars in leaderboard
- [x] Switch to the new rc_client_t API layer once it's merged: https://github.com/RetroAchievements/rcheevos/pull/242
  - [x] Get the basics working
  - [x] Finish up the integration
  - [x] Challenge/hardcore mode (savestates and cheats disabled)
  - [x] Wait for the merge
  - [ ] ~~Always disable cheats when achievements are enabled?~~
- [x] Move settings to the front tab
- [x] Allow alternate speeds above 1.0x (slowdown is not permitted in challenge mode)
response_body" - see https://github.com/RetroAchievements/rcheevos/wiki/rc_client-integration#server-errors )
- [x] Refactor so we can play sounds on top of gameplay #17703 
  - [ ] Copy custom sound effects into the memstick dir
- [x] Sound effect for unlocks and leaderboards submits #17705 
- [x] Correctly call do_frame even during fast-forward #17710
- [x] Cache icon images properly (#17709 )
- [x] CSO file support (#17699)
- [x] Make it work even when launched from a frontend (#17708 )
- [x] HTTPS support, so we can use the standard RA URLs
- [ ] Correctly handle HTTP timeout ("When a request times out, the server callback should be called with a NULL
- [x] More UI tweaks
  - [x] Customizable sound effects #17725 
  - [x] Filtering or collapsible sections in achievement viewer #17728
  - [x] Improved text formatting in achievement list
  - [x] In leaderboards, show just the range you are in by default. Allow choosing to view top entries. #17828
  - [x] Allow configuring in which corner for notifications to show
- [ ] Disable more things in challenge mode
  - [ ] Interactive debugger
  - [ ] Remote debugger
- [x] "RAIntegration", for easier development of achievements on Windows?
  - [ ] Use the [RAInterface](https://github.com/RetroAchievements/RAInterface) library for loading the DLL
  - [x] Alternatively wait for rcheevos to integrate RAIntegration, this is planned
  - [x] Debug setting to allow saving, but not loading, state in challenge mode (only meaningful after RAIntegration is added): #18425
  - [ ] Allow changing mode (softcore etc) in-game

Additional feature requests:

[...]

PR #17638, opened 2023-06-27
Remove the RAIntegration stuff for now (wasn't actually enabled)

hrydgard:
This is just the unused, untested achievement editor integration code that came along with the ride when I adapted RetroAchievement code from DuckStation.

Removing this code will make the conversion to rc_client_t easier. Might eventually add this capability back properly in one way or another.

Part of #17631


ScottFromDerby:
This is still pretty exciting stuff though! Can't wait to see more :)

PR #17642, opened 2023-06-29
(UWP) Code Cleanup: Round 2

ghost:
# Main Changes

The sequel to #17053, this PR removes the remaining bloated code that I noticed across the UWP vcxproj files, here's a summary of the changes:

- **Combined `ItemDefinitionGroup` when applicable**: A lot of these groups were duplicated to oblivion, with the worst case scenario being that we could have a single group for every single configuration. So when applicable, I combined many of these groups into a single one. Of course, not all of these had shared code with all configurations, so when needed we still keep different groups for different CPU architectures and build configurations.

- **Simplify the UWP Gold setting**: A lot of the UWP Gold code in these files were identical to the Release configuration code, to the point where some files don't even have UWP Gold configurations, instead specifying in the sln file to use the Release configuartion when building a UWP Gold build. So I standardized this approach across every project with the exception of the main UWP project file. This file has specific code for the UWP Gold configuration (mainly regarding assets), so I could not really change much here.

As an added bonus, this also fixes the Release ARM64, UWP Gold x64, UWP Gold ARM, and UWP Gold ARM64 builds, which were broken since the sln file used the wrong build and CPU architecture configurations of the rcheevos library for these builds (oops).

# Observations

While testing if every configuration built locally, I got build errors when trying to build all of the Win32 configurations, as well as the Debug ARM64 configuration. The specific errors are the following:

Release/UWP Gold Win32:

image

Debug Win32:

image

Debug ARM64:

image

These issues don't seem to be related to this PR, since they still occur when using the current master branch. The ARM64 issue needs further investigating, although I don't think there's a lot of merit to Win32/x86 UWP applications nowadays, but that's just me.


GABO1423:
The failed Ubuntu builds don't seem to be related to the PR either:

Screenshot_20230629-080124_Chrome

GABO1423:
> Yeah, I think we can ignore those errors.
> 
> This looks good, let's get it in.

While on topic, do you know if there's any use for the Win32/x86 builds? I've been considering removing them since I can't see any particular use case for them.

PR #17647, opened 2023-06-30
Postshader - Fake reflections

lvonasek:
Demos:
GTA - https://twitter.com/LubosVon/status/1674769471959977986
NFS - https://www.youtube.com/watch?v=YqykcQXXiSE

### DIFF
--- a/UWP/UWP.vcxproj
+++ b/UWP/UWP.vcxproj
@@ -835,6 +835,9 @@
     <None Include="Content\shaders\defaultshaders.ini">
       <DeploymentContent>true</DeploymentContent>
     </None>
+    <None Include="Content\shaders\fakereclections.fsh">
+      <DeploymentContent>true</DeploymentContent>
+    </None>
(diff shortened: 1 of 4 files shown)


hrydgard:
Thanks to the quite static camera angle in GTA, it looks better than it has any right to, hah.

Though, it'll reflect regardless of weather, which is hardly ideal...

lvonasek:
It is a very simple trick but it works quite well. In NFS in VR it does it's job, on 2D it is not so noticable. In some games is the effect better.

I was tuning it quite lot to have it fast but still effective. The wet effect is the only way to make it work. Without it it would be immediately noticable that the reflections are wrong.

hrydgard:
I don't much like the idea of so game-specific post shaders, but we don't really have another way to deal with it either. So I'll go ahead. However, I'm a bit afraid of a huge proliferation of these, which will get confusing. We will need categories or similar.

PR #17651, opened 2023-06-30
SDL: support HiDPI on wayland

hch12907:
Not sure if this is correct, I am familiar with neither SDL nor the codebase of PPSSPP. Tested on X11 and Wayland though, both seem to work just like before.

### DIFF
--- a/Core/Core.cpp
+++ b/Core/Core.cpp
@@ -167,7 +167,7 @@ bool UpdateScreenScale(int width, int height) {
 	float g_logical_dpi = System_GetPropertyFloat(SYSPROP_DISPLAY_LOGICAL_DPI);
 	g_display.dpi_scale_x = g_logical_dpi / g_display.dpi;
 	g_display.dpi_scale_y = g_logical_dpi / g_display.dpi;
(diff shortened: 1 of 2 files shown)


hrydgard:
Cool, I'm gonna test on my Linux laptop where I had some DPI issues soon.

hrydgard:
It doesn't help the weird DPI I'm getting, but also it doesn't break anything. Let's get it in.

hrydgard (approved):

hrydgard:
Hm, this somehow made things gigantic on OSX, so gonna have to add some checks, or figure out what's going on..

hrydgard:
Actually, I'm going to revert this. It's clearly not tested enough. For example, on OSX, you get a different rendered DPI after the first window resize, plus it's wrong.

hch12907:
That's a bummer, because I don't have a MacOS machine to test the code on :/

hrydgard:
I may resurrect and debug this myself on Mac in the near future, I just didn't want to leave it in such a broken state. That's probably a good place to start looking yeah.

PR #17660, opened 2023-07-03
Backport rcheevos build setup to libretro/Makefile.common from the rc_client branch

hrydgard:
For some reason this was only needed on Windows?

Either way, I just backported the fix I had already made for the rc-client branch, #17646 .


m4xw (approved):
Works on our CI

PR #17661, opened 2023-07-03
Added new Windows UI Debugger window

Nemoumbra:
This is a PoC PR that is no longer a draft. It adds a new context menu item for summoning a window through which the users can execute `hle.func.scan` or `hle.func.removeRange` themselves. The interface is certainly not final, don't worry.
image
I used the breakpoint window as a template for the ppsspp.rc file entry: 
image

Let's see if it builds... ~~There are some JIT problems with this feature: I made Patapon 3 EU crash while running tests and this doesn't happen when I call hle.func.scan through the Remote API.~~ It was my skill issue with a break inside the case of a switch.


Nemoumbra:
Alright, this is the newest version.

image

And I've also added the ellipsis to the context menu item to indicate that clicking on it will open a window.

hrydgard:
I would probably name the window something like the Symbol Editor or similar, "scanremove" is a really odd name that doesn't really allow for the addition of more functionality that might be convenient to have nearby.

Nemoumbra:
I agree that the naming is far from optimal (and thus is not final). We have to decide on the name for the internal window class (EditSymbolsWindow sounds really great), the window's caption and the context menu item's name (can be both "Edit symbols"). Once you respond to this idea, I'll make a renaming commit.

hrydgard:
I'm fine with EditSymbolsWindow, etc.

hrydgard:
Unrelated, I've long been thinking of adding a little command line debug console for stuff like this, that would be easy to add more obscure commands to... Sometimes would make more sense than adding UI to the Windows debugger.

Problem is that our own UI doesn't work while a game is paused, which would be a pretty important use case, so would have to make it a Win32 window anyway, ugh. Well, that's for another time.

PR #17662, opened 2023-07-04
Simplify projective texcoord calculation

fp64:
As mentioned in https://github.com/hrydgard/ppsspp/issues/17613#issuecomment-1613583152 .


hrydgard (approved):

hrydgard:
@unknownbrackets how about this?

unknownbrackets (approved):
Thanks, sorry - have been a bit busy.

Quickly checked the TOCA2 framedump just to be sure - looks the same, so no breakage.  Unfortunately didn't seem to affect perf much (probably just not enough projection mapping happening in that frame.)  Also checked the gpu/texmtx/* tests quickly to make sure it didn't change them (they currently fail due to some precision issue from the interpolation.)

-[Unknown]

PR #17665, opened 2023-07-04
Postshader - New version of bloom

lvonasek:
The old bloom basically blurres everything (bright colors more, dark color less). This updated version reduces(~removes) the blur on dark colors so it is usable in VR (but also other platforms).

Demo video:
https://twitter.com/LubosVon/status/1676326319133540352?s=20

Complexity is +/- the same (before 19 texture reads, now it is 17).

### DIFF
--- a/UWP/UWP.vcxproj
+++ b/UWP/UWP.vcxproj
@@ -820,6 +820,9 @@
     <None Include="Content\shaders\bloom.fsh">
       <DeploymentContent>true</DeploymentContent>
     </None>
+    <None Include="Content\shaders\bloomnoblur.fsh">
+      <DeploymentContent>true</DeploymentContent>
+    </None>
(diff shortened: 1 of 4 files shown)


LunaMoo:
I wouldn't call removing blurr from darker colors an improvement. Blurr combined with resolution above display res is very nice AA effect.

 Though I'm just stating an opinion knowing how bad can be changing an existing effect for old users. Personally I'd just branch the code based on ui changeable parameter and call it "vr compatibility" or better call it by literally what it does instead of changing how it works by default for everyone based on personal preference and needs.

lvonasek:
That makes sense. The shader is way too different from original to change it using parameter. I will add it as a separate shader.

hrydgard:
Was gonna do categories first, but meh, let's get it in.

PR #17684, opened 2023-07-10
RetroAchievements: Revert to the old behavior for bad accesses (return 0).

hrydgard:
The new smarter achievement-disabling feature has a lot of overhead and doesn't quite seem to do what I expected. More work seems to be needed, so let's fall back to the old behavior that most sets have been developed against, anyway.

Should fix the Valkyrie Profile issue reported in #17631


hrydgard:
A proper fix in rc_client is coming soon, will revert this once it's there.

PR #17698, opened 2023-07-11
Fixed Polish translation

Patrxgt:
I've fixed some incorrect lines and translated most of the new ones.


hrydgard:
Thank you!

There's gonna be some more changes to the new lines, by the way.

PR #17752, opened 2023-07-21
OpenXR - Enable 6DoF in cinema mode

lvonasek:
I did expect it but it is pretty good experience:
https://twitter.com/LubosVon/status/1682415161620004865?t=HrZQp-XEQdVyb6VMFxwJeg&s=19


hrydgard:
Interesting, so head movement is smooth even though the display isn't, but you still get some VR effect? cool!

lvonasek:
Only headset position is applied, not orientation.

The jittering/headaches in VR are always caused by out of sync orientation.

PR #17757, opened 2023-07-22
Fix a segfault in ReadAllWithProgress

hch12907:
On Linux, when `progress->cancelled` is NULL, this results in a segfault.

### DIFF
--- a/Common/Net/NetBuffer.cpp
+++ b/Common/Net/NetBuffer.cpp
@@ -76,8 +76,8 @@ bool Buffer::ReadAllWithProgress(int fd, int knownSize, RequestProgress *progres
 	int total = 0;
 	while (true) {
 		bool ready = false;
-		while (!ready && (!progress || !*progress->cancelled)) {
-			if (*progress->cancelled)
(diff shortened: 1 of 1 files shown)


Common/Net/NetBuffer.cpp
```diff
@@ -76,7 +76,7 @@ bool Buffer::ReadAllWithProgress(int fd, int knownSize, RequestProgress *progres
 	int total = 0;
 	while (true) {
 		bool ready = false;
-		while (!ready && (!progress || !*progress->cancelled)) {
+		while (!ready && !progress) {
 			if (*progress->cancelled)
```
ANR2ME, line 80: When `progress` is null we shouldn't try to access it's member isn't?

hch12907: Hmm... Let me rethink this a little, I am still trying to understand RequestProgress.

hch12907: Seems like `progress->cancelled` is some sort of shared state? At least that's what I think it is... changed the code to reflect this.

hrydgard (approved):
We should actually make sure progress->cancelled isn't null. Though, for now, let's get this in.

PR #17758, opened 2023-07-22
SDL: support HiDPI on wayland

hch12907:
So this is really just #17651 with additional fixes to try and nail down where the bugs are on MacOS.

I still don't own a Mac so I am marking this as a draft until that platform is tested.

(I included a commit from #17757 because I can't test the build otherwise)

### DIFF
--- a/Core/Core.cpp
+++ b/Core/Core.cpp
@@ -167,7 +167,7 @@ bool UpdateScreenScale(int width, int height) {
 	float g_logical_dpi = System_GetPropertyFloat(SYSPROP_DISPLAY_LOGICAL_DPI);
 	g_display.dpi_scale_x = g_logical_dpi / g_display.dpi;
 	g_display.dpi_scale_y = g_logical_dpi / g_display.dpi;
-#elif PPSSPP_PLATFORM(WINDOWS) && !PPSSPP_PLATFORM(UWP)
+#elif (PPSSPP_PLATFORM(WINDOWS) && !PPSSPP_PLATFORM(UWP)) || PPSSPP_PLATFORM(LINUX)
 	g_display.dpi = System_GetPropertyFloat(SYSPROP_DISPLAY_DPI);
 	g_display.dpi_scale_x = 96.0f / g_display.dpi;
 	g_display.dpi_scale_y = 96.0f / g_display.dpi;
--- a/Core/Core.cpp
+++ b/Core/Core.cpp
@@ -162,20 +162,19 @@ static bool IsWindowSmall(int pixelWidth, int pixelHeight) {
 // TODO: Feels like this belongs elsewhere.
 bool UpdateScreenScale(int width, int height) {
 	bool smallWindow;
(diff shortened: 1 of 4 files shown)


hch12907:
Slight note: I forced the DPI to be 1.0 because it seems like the automatically calculated DPI would make everything look much bigger.

You can printf `g_DesktopDPI` in `UpdateScreenDPI()` if you want to know what the DPI is. Use `--dpi` to force the program to show in a specific DPI.

hrydgard:
This still breaks things badly unfortunately (mouse input not working or maybe clicking in the wrong place) and after a window resize, scale is busted.

hch12907:
Turns out MacOS needs a flag to be set in `Info.plist` for HiDPI.

Some sources also mention something about setting the `NSPrincipleClass` to `NSApplication`, but I'm not entirely sure about that...

hrydgard:
I'll check this again shortly. There's a merge conflict that needs resolving, btw.

hch12907:
You mean the UI was larger than the window itself? Strange...

What do the debug messages show? This is what mine looks like on Wayland (130% scale factor):
```
drawable_width = 1868, window_width = 1437, calculated DPI = 1.300626
drawable_width = 1877, window_width = 1444, calculated DPI = 1.300554
drawable_width = 1891, window_width = 1455, calculated DPI = 1.300344
drawable_width = 1920, window_width = 1477, calculated DPI = 1.300609
```

hrydgard:
Sorry, could you rebase again? I pushed some broken code...

hrydgard:
Sorry, I wasn't on the latest commit. Here are the current results on Mac:

* On startup:
image

drawable_width = 1692, window_width = 846, calculated DPI = 2.001182 (clicks are wrong)

* After resize:

image

drawable_width = 1712, window_width = 856, calculated DPI = 2.001168 (clicks match. Looks the same as above though)

master:

image

hch12907:
I think I see why I couldn't reproduce the issue on Wayland now: SDL applications on Wayland receive a resize event immediately on the first frame, hence the (wrong) values used in initialization were overwritten.

hrydgard:
Working perfectly now.

hrydgard (approved):

hrydgard:
Actually, I take that back - we still have a problem with MoltenVK on Mac (Vulkan), for whatever reason:

image


Wonder if we should just add a check to ignore.. I'll try to take care of this soon.

hch12907:
FWIW, you can do `if (backend == Vulkan) { g_ForcedDPI = 1.0 }` at the meantime...

(also, it seems like font atlas is still used for font rendering, possibly lack of sdl2_ttf package?)

hrydgard:
yeah, I'll do a quickfix like that.

Gonna have to look into the font rendering. It indeed seems we'll want to ship a SDL2_ttf framework as well, it's not included in the SDL2 distribution we're missing. Or we should add native Mac font rendering.

PR #17767, opened 2023-07-23
Translation: Consolidate a few strings

hrydgard:
Don't want to have separate string for every option that may require a restart. Had to use Google Translate to fix things up, did the best I could.

Also adds more missing translation strings and simplifies some code.


GABO1423:
I could touch up the Spanish lang files if you need some help.

hrydgard:
@GABO1423 Please do!

PR #17772, opened 2023-07-23
Fix issue where achievements wouldn't always activate when starting from frontends

hrydgard:
My previous workaround was flawed - it didn't work at all since we run http callbacks on the main thread, so no progress was made in the background. This problem was not very easy to reproduce on my machine so I interpreted my repro failures as success.

Now we really delay starting the game until the achievement system is ready, if it's in the process of logging in.

Part of #17631

### DIFF
--- a/Core/RetroAchievements.cpp
+++ b/Core/RetroAchievements.cpp
@@ -647,12 +647,14 @@ struct FileContext {
 
 static BlockDevice *g_blockDevice;
 
+bool IsReadyToStart() {
+	return !g_isLoggingIn;
+}
+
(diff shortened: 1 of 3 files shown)


unknownbrackets (approved):
Yeah, if we need to finish login before calling `Achievements::SetGame(filename, loadedFile);`, this makes sense.

Though I do wonder if we could move `Achievements::SetGame(filename, loadedFile);` to a step within InitUpdate, so that we could start loading the game during login.  Might be more efficient.

-[Unknown]

hrydgard:
Yeah, I may try that later. I think getting this in as a correctness fix first makes sense.

PR #17779, opened 2023-07-24
Cleanup emscripten libretro target

### LINKED ISSUE #13567: Planning to build for WebAssembly?
## What is WebAssembly (WASM)
WebAssembly (WASM) is a format in which native code such as C, C++ runs on a web browser.
Today it is included in the latest versions of all major browsers.

## Why we need

It would be great if PPSSPP works in a web browser, and can be used without the need for installation.

It's worth noting that WebAssembly is also available in iOS Safari.

Until now, PPSSPP has not become an iOS native app.
This is probably due to the restrictions of the App Store rules that emulators cannot be listed.

However, once PPSSPP can work with WebAssembly, PSP Emulator will work for the first time on iOS without the need for jailbreaking.

This is amazing.

## Similar WebAssembly project

WebAssembly implementations of emulators other than the PSP, such as Game Boy Advance or PlayStation 1, already exist.
https://github.com/torch2424/wasmboy
https://github.com/tjwei/pcsxjs

There are also WebAssembly implementations of existing complex and long programs written in C, C++ such as FFmpeg.
https://github.com/ffmpegwasm/ffmpeg.wasm

---

Please let me know if you are planning the WebAssembly version as a future plan.
If you have already decided not to plan the WebAssembly version, please tell us why.

ethanaobrien:
This pr cleans up the emscripten libretro build target. It's a little messy in my opinion, so let me know if anything needs to be cleaned up.

This pr fixes the ability to compile the emscripten `.bc` file (to be linked with RetroArch) by running `emmake make platform=emscripten` in the libretro folder.

When building with retroarch, it must be built with the pthread value at least 8, (I built mine with 12).

**Important note:** For it to compile, you must modify `ext/armips/ext/filesystem/include/ghc/filesystem.hpp` to tell it to use emscripten as a linux target (emscripten is not linux, but uses the same filesystem)
It should look like:
image

I also had to change `GetElfHwcapFromGetauxval` in `ext/cpu_features/src/hwcaps.c` to always return just a number instead of actually doing what it needs to... theres got to be a way to do this in emscripten... right?

Image of the RetroArch main menu, loaded in microsoft edge, with the ppsspp core loaded:
image

This pr does **not** close #13567! While this pr fixes up the ability to compile, jit is broken for emscripten. When running, I get the following error in the console:
image

and the screen just goes black...

I attempted to compile without jit (using `FakeJit`), but I got this error when linking to retroarch (and on runtime):
image

I don't have the knowledge to continue further, but if anyone wants to look into this. -
This branch of RetroArch is what I used to link: https://github.com/EmulatorJS/RetroArch
This project is built to be able to load the ppsspp assets (if used correctly): https://github.com/EmulatorJS/EmulatorJS/


hrydgard:
Well the JITs are obviously never going to work since they don't emit wasm code. I don't know if wasm has some provision for emitting small chunks of code, otherwise I guess some kind of whole program translation needs to be used to compile the whole game from mips to wasm - and full program translation wouldn't work very well on PSP since many games load code dynamically.

Does it run if you go into developer settings and switch the CPU core to interpreter?

Common/MemArenaPosix.cpp
```diff
@@ -107,12 +107,16 @@ void MemArena::ReleaseSpace() {
 
 void *MemArena::CreateView(s64 offset, size_t size, void *base)
 {
+#ifdef __EMSCRIPTEN__
+	void *retval = mmap(NULL, size, PROT_READ | PROT_WRITE, MAP_FILE | MAP_SHARED, fd, offset);
```
unknownbrackets, line 111: Well, I'm not very interested in emscripten for PPSSPP one way or another, so I won't get in your way.  But is this safe?

The intent of this function is to create address space that matches PSP address space.  That is, when PSP code accesses these addresses, it refers to exactly the same memory:

0x08800000
0x48800000
0x88800000

It's actually super important that if I write to 0x48800000 (which is commonly used to bypass CPU caches for interop with GPU or etc.) that a later read from 0x08800000 will read exactly the same value.

Normally, this is made to work by passing the same `fd, offset` parameters for both, but different addresses so we can setup this mapping.  If it allocates where ever it wants (which is what null means here), then these later mirrors won't work, and I'd expect some things to not work right in games.

-[Unknown]

Core/MIPS/MIPSTables.cpp
```diff
@@ -1057,6 +1057,10 @@ int MIPSInterpret_RunUntil(u64 globalTicks) {
 			// DEBUG_LOG(CPU, "Hit the max ticks, bailing 1 : %llu, %llu", globalTicks, CoreTiming::GetTicks());
 			return 1;
 		}
+#ifdef __EMSCRIPTEN__
+        //Never block the main loop... But this seems to cause the emulator to crash after a bit..?
```
unknownbrackets, line 1061: PPSSPP should normally do as much as it can until it gets a frame, which will bail out of this function.  Probably if you're seeing it never exit this function, it's some failure in processing graphics commands.  This could be related to the uncached mirror problem, since uncached memory accesses are used heavily by games to synchronize with the GPU (which doesn't see things in the CPU caches.)

-[Unknown]

Common/MemArenaPosix.cpp
ethanaobrien: Thats true, I thought about that and forgot to go back and look at this. On runtime without this I'd get an invalid argument error, did this temporarily to get it to at least run, and forgot to look at it. I'll look at this

Core/MIPS/MIPSTables.cpp
ethanaobrien: That makes sense. Thanks for your input. That should lead me somewhere

Common/MemArenaPosix.cpp
hrydgard: On platforms where we can't set up mirror mappings (or don't have enough address space) we apply a different workaround by simply &-ing every address used for a memory access with 0x3FFFFFFF. See MASKED_PSP_MEMORY

ethanaobrien: So basically, the way psp works is it creates a 4GB file, which is the emulated psp's memory, and maps that into memory, correct? And enabling `MASKED_PSP_MEMORY` keeps the memory in memory? Did I understand that correctly?

unknownbrackets: As I mentioned, we want multiple pointers to point to the same data.  The good news is: most operating systems already have a way to do this.  It's faster to just let the operating system handle it, instead of doing it manually.

But it's not like operating systems have a function to just map memory to two addresses.  That'd be too easy and convenient.  No, what they have is a tool that I believe was originally meant for inter-process communication: shared file memory mapping.  It just maps shared memory, but we can map the "file" to multiple places and it all points to the same RAM.  This is often supported, but only works if we can get 4 GB of address space reserved all in a row.

On 32-bit operating systems, that's not possible.  We still use the mapping for VRAM there (which has multiple mirrors inside 0x3FFFFFFF of VRAM), but we allocate less address space, and the mirrors I mentioned above aren't "needed" because we mask pointers instead.

So if emscripten doesn't support shared memory mapping (maybe because they almost explicitly don't want inter-process communication?), your best bet is to define MASKED_PSP_MEMORY (see where it's defined for 32-bit) on emscripten and hope the VRAM mirrors don't cause problems.  Most of the time they'd only be used for depth reads and writes from the CPU.

-[Unknown]

ethanaobrien:
After some research, I found the reason why `mmap` was failing was because we only have 4gb of storage (memory + storage) and thats not enough at the moment. I did some cleanup, removing some things I either cant debug / dont need to be debugged, I guess this pr is ready, and we just wait for the day when the wasm memory cap is raised

libretro/Makefile
```diff
@@ -304,7 +304,7 @@ else ifeq ($(platform), emscripten)
 	TARGET := $(TARGET_NAME)_libretro_emscripten.bc
 	GLES := 1
 	CPUFLAGS +=
-	PLATCFLAGS += -msimd128 -msse3 -mssse3 -msse4.1 -pthread -D__EMSCRIPTEN__ -DUSE_ASAN
+	PLATCFLAGS += -msimd128 -msse3 -mssse3 -msse4.1 -pthread -D__EMSCRIPTEN__
```
hrydgard, line 307: when doing emscripten, surely sse* flags aren't relevant?

ethanaobrien: They are, actually. I just tested, making sure they were. When removing them, it requires that you have at least `-msimd128` set, which basically implies all the other sse flags.

ethanaobrien:
I had a quick question,

So after more testing, I found that I think the issue is actually the emscripten filesystem doesn't seem to either support more than one mmap at a time, or at least more than one mmap on the same file, which somewhat makes sense.

Would it, theoretically, function the same if I used malloc to allocate the data into memory, since either way it's in memory in emscripten? If so, I'd probably make a separate mem handling file in emscripten. Would this have any side effects?

There are a few other issues I've found. You can't currently build ppsspp with gles3, which I think is because of webgl's weirdness, and this causes graphical issues in game. Building with gles2 works to an extent.

If you want to see the current status or the port, you can view it here - https://testing.emulatorjs.org/

If you wouldn't mind, I'd actually prefer to close this pull request, cleanup my commits, and re-open a new one once everything is ready, since a lot of these commits are really messy and changed things that didn't need to be changed

unknownbrackets:
You can always just rebase and force push.  GitHub won't get confused if you force push a brand new branch to the same old name.  That's usually the way of doing it rather than opening a new pull request.

It won't work for the VRAM mirrors to just malloc().

The important thing to understand here is the difference between "memory" and "address space".  Think of memory as a grid, like in the game Battleship.  There are things in the grid in certain places, and each place has some way you could reference it: a row and a column.
battleships

Address space gives you a number - a "pointer" as it's called - to access the memory.  For example, a pointer on a PC might be 0x3e0457e800.  Address space is normally allocated in "pages" which you can think of as the rows in the grid above.  The operating system isn't managing a table of address space for each individual square (that'd be a lot of work), it's just managing it per page.  Each address points to a row, and that address plus one is just the next square (next column) in the grid.

In our case, we need two numbers (pointers) - and they must be different numbers, in fact very specifically different numbers - to point to the same memory.  Row 6 on that battleship grid.  We need it to be that both 0x3e0457e800 AND 0x3e0477e800 both see exactly the same battleships.  If they don't - well, you'll never be able to sink my battleships, and I'll never be able to sink yours.  We have to be able to share the memory using different addresses.

malloc() doesn't do this.  In fact, it fails at both the things we need:
 - It can't put memory at a certain place, so it's basically useless to us from the beginning.  We need the address space to look the same way it looked on a PSP - games are written for THAT address space.  Otherwise, you should give up on emulation and start 1,400 individual game porting projects to recompile the games individually in enscriptem, redoing how they manage memory themselves.  Sounds like a lot of work.
 - It can't point two allocations at the same memory.

If it's impossible to have two addresses point at the same memory in enscriptem, that's bad, but some games will still work.  It'd be better to still use mmap(), but just not use the same "file", in that case.  As I mentioned - this will cause bugs.  You won't be able to sink my battleship, remember?  The actual bugs will probably be missing textures and things like that, and won't affect that many games.  What will affect more games is if memory has the wrong "layout" - the numbers in address space need to be correct.  Without that, many games that would otherwise work will crash.

PSP games are hardcoded to expect this memory layout:
 - base addr + 0x00010000 - scratch pad (16 KB)
 - base addr + 0x04000000 - VRAM (4 mirrors of 2 MB each, different swizzle)
 - base addr + 0x08000000 - RAM (32 MB)
 - all the above + 0x40000000 - CPU cache bypass (heavily used by games) - can workaround via MASKED_PSP_MEMORY
 - all the above + 0x80000000 - kernel-mode access (rarely used by games) - can workaround via MASKED_PSP_MEMORY

So the important thing is that you need to construct that address space.  Games directly access VRAM with that number because they know "the VRAM will always be here."  They don't use malloc() to get pointers to VRAM - they just "know" that address space is already this way.  So it has to be.

-[Unknown]

ethanaobrien:
That all makes perfect sense. Sorry for the confusion and thank you.

I decided to do some digging into emscripten's filesystem and found this



Looks like I'll need to try implementing this...

libretro/Makefile
ethanaobrien: EDIT: compiling with only `-msimd128` produces this error:

Screenshot 2023-07-31 175532

ethanaobrien:
I believe this is ready for review.

Basically, thr mmap workaround is to directly handle the memory access, it's not unsafe when built properly (`-s GOLBAL_BASE=300` needs to be added to retroarch's `LDFLAGS` when compiling)

hrydgard:
There is actually a way to get a working memory space with a single contiguous allocation, if you accept a bunch of waste.

The highest address that will normally be used, once you chop off the mirroring bits at the top, is 0x08000000 + 32MB == 0x0A000000. So if you just create a single big allocation of that size, and point "base" at it, and enable MASKED_PSP_MEMORY, you're mostly good to go, though you'll have a bunch of extra memory by accident in there that shouldn't be accessible, like between VRAM and RAM. Shouldn't really hurt anything. 0x0A000000 is just 168MB so these days, which means about 130MB of waste, not such a big deal.

ethanaobrien:
>Shouldn't really hurt anything. 0x0A000000 is just 168MB so these days, which means about 130MB of waste, not such a big deal.

And it's still a better solution than what I have now. Basically, are you saying to use `malloc` for the base and return as I have now?


Should I keep a separate emscription memory handling file?

hrydgard:
Don't need a file.

ethanaobrien:
@hrydgard Does this look correct?

hrydgard:
It looks good to me, but if it's just a cleanup and not actually close to working yet, I'm not sure I want to merge.

Additionally, I'm not sure how Emscripten support is good for the project anyway... What would be the use case?

ethanaobrien:
>It looks good to me, but if it's just a cleanup and not actually close to working yet, I'm not sure I want to merge.

Thats fair, I believe the only issue is that the emulator seems to lock up the page after about 2 frames. I'm not sure how to trace or debug this. I removed the loop [here](https://github.com/EmulatorJS/ppsspp/blob/8426b35a8066a447e312de881053ca141361a3d4/Core/MIPS/MIPSTables.cpp#L1045C46-L1045C46) and it seemed to help it run (and yield to the main loop), but for some reason I can't seem to recompile a working version, even though nothing changed. If you know what might be blocking, I'd appreciate the help. I tried moving this into a worker to free up the main thread, although this issue still occurred. Thats the only issue as far as I can tell

>Additionally, I'm not sure how Emscripten support is good for the project anyway... What would be the use case?

Mainly just because it's possible. If you do decide to not support emscripten (which I fully understand and wouldn't argue with) I'd suggest removing the current emscripten target in the libretro Makefile

hrydgard:
Hm, that's fair enough - since the target already exists, this PR doesn't change any decision. I'll merge.

As for hanging, I wouldn't expect moving that loop to a worker to work...

hrydgard (approved):

ethanaobrien:
>As for hanging, I wouldn't expect moving that loop to a worker to work...

I wouldn't either, I was just wondering if it was blocking some needed javascript from running. Not sure what could be blocking it, since compiling without the loop worked for a time.

ethanaobrien:
And thank you!

PR #17780, opened 2023-07-25
SDL: implement TextDrawer using SDL2_ttf

hch12907:
For a quick comparison on DPI set to 2.0 (top is PR, bottom is master):

image


Common/Render/Text/draw_text_sdl.cpp
```diff
@@ -0,0 +1,263 @@
...
+	}
+}
+void TextDrawerSDL::DrawStringBitmap(std::vector<uint8_t> &bitmapData, TextStringEntry &entry, Draw::DataFormat texFormat, const char *str, int align) {
+	if (!strlen(str)) {
+		bitmapData.clear();
+		return;
+	}
+
+	std::string processed_str(str);
+	processed_str = std::regex_replace(processed_str, std::regex("&&"), "&");
```
hrydgard, line 170: we're not using std::regex_replace anywhere else in the code, so pulling it in just for this seems overkill.

hrydgard (approved):
Very nice!

One comment, will merge once addressed.

hch12907:
With fallback fonts working, CJK texts and even Thai renders properly. RTL scripts are still flowing left to right though.

image

The remaining task is to implement it for Macs. Walking through `/Library/Fonts/` and getting the suitable fonts should be enough.

hch12907:
There's a weird memory leak happening in WrapString().

I can consistently reproduce the leak by going to Settings -> Controls -> Control mapping. Commenting out WrapString() "fixes" it.

hrydgard:
The latest changes look pretty good to me! Let's get it in, and see how it goes.

Common/Render/Text/draw_text_sdl.cpp
```diff
@@ -0,0 +1,476 @@
...
+		FcNameParse((const FcChar8 *) "Droid Sans Bold"),
+		FcNameParse((const FcChar8 *) "DejaVu Sans Condensed"),
+		FcNameParse((const FcChar8 *) "Noto Sans CJK Medium"),
+		FcNameParse((const FcChar8 *) "Noto Sans Hebrew Medium"),
+		FcNameParse((const FcChar8 *) "Noto Sans Lao Medium"),
+		FcNameParse((const FcChar8 *) "Noto Sans Thai Medium")
+	};
+
+	for (int i = 0; i < ARRAY_SIZE(names); i++) {
+    	FcFontSet *foundFonts = FcFontList(config, names[i], os);
```
unknownbrackets, line 61: nit: Indentation mistake, some spaces here.

-[Unknown]

Common/Render/Text/draw_text_sdl.cpp
```diff
@@ -0,0 +1,476 @@
...
+			}
+		}	
+	}
+#else
+	// We don't have a fallback font for this platform.
+	// Unsupported characters will be rendered as squares.
+#endif
+}
+
+uint32_t TextDrawerSDL::CheckMissingGlyph(std::string& text) {
```
unknownbrackets, line 136: nit: Since you're not changing `text`, prefer to use `const std::string &text` here.

-[Unknown]

Common/Render/Text/draw_text_sdl.cpp
```diff
@@ -0,0 +1,476 @@
...
+	auto iter = fontMap_.find(fontHash);
+	if (iter != fontMap_.end()) {
+		fontHash_ = fontHash;
+		return fontHash;
+	}
+
+	const char *useFont = fontName ? fontName : "Roboto-Condensed.ttf";
+	const int ptSize = (int)((size + 6) / dpiScale_);
+
+	TTF_Font *font = TTF_OpenFont(useFont, ptSize);
```
unknownbrackets, line 199: We include a version of Roboto-Condensed in the repo.  We actually intentionally use an "older" version of the font, but using a consistent version across platforms improves the chances buttons won't wrap weird for some people but not others.

Would you be against this preferring the one from g_VFS?

-[Unknown]

Common/Render/Text/draw_text_sdl.cpp
```diff
@@ -0,0 +1,476 @@
...
+			entry->width = width;
+			entry->height = height;
+			sizeCache_[key] = std::unique_ptr<TextMeasureEntry>(entry);
+		}
+
+		entry->lastUsedFrame = frameCount_;
+		if (total_w < entry->width) {
+			total_w = entry->width;
+		}
+		total_h += entry->height;
```
unknownbrackets, line 296: This should probably be TTF_LineSkip() or something.  But maybe we can just pass the string with newlines to TTF_SizeUTF8(), it'll give correct height?  I'm not sure.  Just noting that fonts don't always have the same height as line spacing.

Especially since you're using Wrapped to render (which is probably a good thing), this could cause multiline panels to be have awkward spacing below them (assuming the text rendering func respects line spacing.)

-[Unknown]

Common/Render/Text/draw_text_sdl.cpp
```diff
@@ -0,0 +1,476 @@
...
+
+	// If a string includes only newlines, SDL2_ttf will refuse to render it
+	// thinking it is empty. Add a space to avoid that. 
+	bool isAllNewline = true;
+	for (int i = 0; i < processedStr.size(); i++) {
+		if (processedStr[i] != '\n') {
+			isAllNewline = false;
+			break;
+		}
+	}
```
unknownbrackets, line 368:
```c++
bool isAllNewline = processedStr.find_first_not_of('\n') == processedStr.npos;
```

-[Unknown]

unknownbrackets (commented):
Seems great - thanks for this, should mean practically all platforms have nicer fonts.

-[Unknown]

Common/Render/Text/draw_text_sdl.cpp
hch12907: TTF_SizeUTF8() doesn't handle wrapped text unfortunately.

Common/Render/Text/draw_text_sdl.cpp
hch12907: It might sound silly, but we actually hit the `if (!font)` case all the time. Unless the current working directory provides a `Roboto-Consensed.ttf`, the one from g_VFS is used.

unknownbrackets: Oh, I assumed it used a path.  Got it, that sounds good.

-[Unknown]

PR #17830, opened 2023-08-01
Updated brazilian portuguese translation

Felipefpl:
Included the latest changes.


assets/lang/pt_BR.ini
```diff
@@ -23,54 +23,41 @@
 # Happy translating.
 
 [Achievements]
-%1: Attempt failed = %1: Attempt failed
-%1: Attempt started = %1: Attempt started
 %d achievements, %d points = %d conquistas, %d pontos
+%1: Leaderboard attempt started = %1: A tentativa da tabela de classificação começou
+%1: Leaderboard attempt failed = %1: A tentativa da tabela de classificação falhou
```
hrydgard, line 28: Hm, looks like you switched to the old keys here? I don't think you started from the latest?

hrydgard: If you go look in the latest:

https://github.com/hrydgard/ppsspp/blob/master/assets/lang/en_US.ini#L27

You see that I've removed "Leaderboard" from these two lines for example, because it took up too much space. In your change though, Leaderboard is still there on the left side of the = sign.

Felipefpl: Fixed. Thank you for showing my mistake.

hrydgard: I think what happened is that your branch was from an old master in your repo.

Felipefpl: nope. it's not this, i made the sync today.

PR #17895, opened 2023-08-13
IniFile: Store sections in unique_ptrs, instead of directly.

hrydgard:
This fixes an issue when you create two sections consecutively and retain pointers to them, and then modify them, such as happens in the postshader ini initialization.

Previously, one of the section pointers could get invalidated since the section vector got resized. Now, the pointed-to sections don't move around in memory, only the list of them does.


hrydgard:
Tested and valgrinded on linux.

PR #17902, opened 2023-08-13
Some debug overlays don't make sense when not in-game, disable them

hrydgard:
Minor feedback fixes too.

### DIFF
--- a/Common/GPU/Vulkan/thin3d_vulkan.cpp
+++ b/Common/GPU/Vulkan/thin3d_vulkan.cpp
@@ -1104,7 +1104,6 @@ VKContext::~VKContext() {
 }
 
 void VKContext::BeginFrame(DebugFlags debugFlags) {
-	// TODO: Bad dependency on g_Config here!
 	renderManager_.BeginFrame(debugFlags & DebugFlags::PROFILE_TIMESTAMPS, debugFlags & DebugFlags::PROFILE_SCOPES);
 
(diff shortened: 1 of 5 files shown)


unknownbrackets (approved):

IrfanH495:
sorry I want to ask does this fix the issue in my comment.? #17878

hrydgard:
@IrfanH495 Yeah it does, should have mentioned that more clearly.

Some of the other overlay types intentionally display over the menu though, so I can debug issues there.

IrfanH495:
Ok thanks, I tried the last build showing up in the menu and missing in the game https://youtu.be/Mt6qvh6u3aU

hrydgard:
Oh, hah, may have flipped it ... thought I tested it though :/ I'll fix it up.

EdHerdman:
I get "Sas not initialized" when trying to bring up the Debug Stats overlay, which seems a bit cryptic. The frametimes graph is empty, but "Frame timing" still works (which I like, I find it interesting to see how much time Submit/Queue-p takes without a game loaded).

Apologies if it's off-topic but I've also been wondering about the eight side-by-side counters in the "frame timing" overlay. Is that intentional?

hrydgard:
The debug stats should only be visible in-game, it's currently a bug so it's the other way around, resulting in the error.

The eight counters are the last 8 frames. 

All of this is early preparation for some important changes to how timing works, and it'll change and I'll probably present it in cleaner ways as we go.

Stay tuned for updates.

PR #17917, opened 2023-08-16
WIP: New frame timing calculations

hrydgard:
Introduces some new frame timing logic, entirely replacing the old one if conditions are right (frame skipping is off, for example).

This does needs more work, it's currently not good enough to merge yet as it causes audio glitches due to too aggressive present time management, but it's looking quite encouraging.

This can greatly reduce display lag (and thus perceived input lag) in vsync mode on Android, and also on PC. However it doesn't correctly take some things into account and is indeed far too aggressive at the moment.

Also, it doesn't yet take into account the extra info we get from VK_GOOGLE_display_timing and VK_KHR_present_wait - this will be added later, and should allow to shave even more milliseconds, at the expensive of reduced resilience to cpu spikes.

Part of work on #17685 .


hrydgard:
I'm going to take another approach to this, starting with a smaller incremental improvement. Closing.

PR #17932, opened 2023-08-18
Update es_ES.ini

motta-git:
Hey i noticed that the archivements barely have any transalation to spanish so i worked on them! Also i noticed some mistranslations here and there. here is my fix for all of that!


hrydgard:
Muchas gracias!

PR #17952, opened 2023-08-22
[UWP] Improvements 2 (Configs, Render, Input)

### LINKED ISSUE #17639: (Xbox UWP) On-screen Keyboard is not working

None

When trying to change certain settings that require the user to type something (like changing the nickname of the PSP, or logging into RetroAchievements), no keyboard appears on screen. A USB Keyboard was tested and while it could control the UI just fine, it also did not work for typing stuff in. This issue occurs on the UWP port when used on an Xbox console, but I can't say if it works on Windows computers or not.

### What should happen

We can type when needed, with a USB keyboard or a controller.

### Logs

N/A

### Platform

UWP / Xbox

Xbox Series X GPU

### PPSSPP version affected

v1.15.4 (Commit c66182a)

### Last working version

None.

### Graphics backend (3D API)

Direct3D 11

### Checklist

basharast:
Hi, this is the second (phase2 of [PR17350](https://github.com/hrydgard/ppsspp/pull/17350)), almost last PR for UWP
(PR 17350 was not mentioned in the release log so I'm linking it here for future tracking)

this PR will address the following issues:

## Initialize

- Configs and App initialize now in better state
- this change was important to handle user's custom configuration on startup
- it shouldn't cause any further problems

## Render device

- I made some clean up on `DeviceResources.cpp`
- Multiple adapters now possible
- it will handle user's custom adapter selection

<details>
<summary>
Screenshot
</summary>

</details>

## Post shaders

- Tls problem for UWP was solved long time ago
- it doesn't require any hacks at all

Read the it at [TlsAlloc Remarks](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-tlsalloc#remarks)

<details>
<summary>
Screenshot
</summary>

</details>

## Input keyboard

- You can now use keyboard to write into text edit
- On-Screen keyboard also works
- It will help users to enter login data for RA
- Keyboard check no longer always `true`, it has actual check now

<details>
<summary>
Screenshot
</summary>

</details>

Plus other minor fixes.

## HWK & OSK
So now we can support Hardware keyboard and on-screen keyboard
the changes made on the UI was to add UI-Notification whenever:
- TextEdit got focus, State: `text_gotfocus`
- TextEdit lost focus, State: `text_lostfocus`
- Popup closed, State: `popup_closed`

With those events the front end was able to manage Input pane and to forward chars to the `NativeKey` when required
on XBox it's a bit complicated because input could be DPad or Keyboard, but thankfully done.

## Related issues
Fixes #17639 

Hope you will find this PR helpful,

Many thanks.


GABO1423:
@basharast Please link issue #17639 to this PR so that when it is merged it is closed automatically. You can do this by just typing "Fixes #17639" into any part of the PR's description.

GABO1423:
Made a build to test this on Xbox, OSK does not seem to work still. But USB Keyboards now work.

The Render Device changes work perfectly, although you obviously don't want to change from the default on Xbox lol:
Unknown-2023_08_22-19_11_21

And shaders work perfectly:
Unknown-2023_08_22-19_12_46

basharast:
Ah nice thanks for the test, yeah as mention in the details at the end, I prepared the events to show/hide the OSK, but need to be linked with the UI text edit (focus state), so I hope this will be easy to be implemented

hrydgard:
Really good stuff!

I think at least on Xbox, we should hide Microsoft Basic Render Driver, though.

basharast:
The only reason I want to implement the focus got/lost event for text edit is to avoid `OnCharacterReceived` and `OnKeyDown`, `OnKeyUp` to send the same call at the once (`NativeKey`)
and to prevent `OnCharacterReceived` from sending any call when no text edit has focus.

I will try to do this, and see how it can be better asap
as for `Basic` render, yeah as it's useless I will skip that from XBox

basharast:
I just updated the code to support the OSK case with few improvements, I hope I covered most cases, @GABO1423 could you please test this if it works fine on Xbox.

GABO1423:
Sorry for the late response, I'll make a build and test this in a bit, until then please hold off from merging this PR.

hrydgard:
I won't merge until I get green light from both of you, don't worry :)

GABO1423:
Sad news, OSK is not working. In fact, trying to use it only causes the app to freeze, softlocking the app and requiring the user to close it. Another issue I found is Menu sounds do not work anymore, but I would have to test further to see if this PR caused the issue or if it was a regression from another commit.

basharast:
No problem, probably need a few changes.. it's almost done, I will try to manage to test this in someway directly, will check the menu sounds if there is something wrong soon.

GABO1423:
> No problem, probably need a few changes.. it's almost done, I will try to manage to test this in someway directly, will check the menu sounds if there is something wrong soon.

I did some quick bisecting, and it seems that the issue is indeed caused by this PR. So at the very least it was not a regression from another commit. Should make narrowing down the issue a bit easier.

basharast:
True, actually this line should be called before configs load
```cpp
// Mount a filesystem
g_Config.flash0Directory = exePath / "assets/flash0";
```
I didn't move it to the new initial function
this was fixed now, will try to solve the OSK and will push another update.

basharast:
I think it freeze because if this line:
```cpp
if (IsXBox()) {
   return InputPane::GetForCurrentView()->Visible;
}
```
this may need to be called from UI thread,
it can be fixed but first need to be tested, let's hope, it's the problem.

GABO1423:
> this may need to be called from UI thread, it can be fixed but **first need to be tested**, let's hope, it's the problem.

Making a build to do just that right now.

GABO1423:
After a lot of work over at Discord, we finally made the OSK work! So when the relevant commits are pushed, this PR is ready to go.

basharast:
As mentioned we reached a stable state, many thanks to @GABO1423 for his time it wasn't possible without testing on real XBox, PR descriptions updated.

hrydgard:
Very cool, nice work guys!

PR #17973, opened 2023-08-25
(UWP) Fixed Skipped Debugger Files

ghost:
These files were not being included before since the files were not in the `UWP/Content/debugger` folder, and since these are only copied during build time, you would need to first build the solution, then make a package for the files to be included. Or copy the files manually before creating a package. So let's just use some tricks to workaround this, it will now look for the files in the original location `assets/debugger`, while still including them in the final package. 

In theory we could do this for every single file in the assets folder, removing the need to use this command:

https://github.com/hrydgard/ppsspp/blob/24ec26d9354462d554b44446d8e666da84354950/UWP/UWP.vcxproj#L110

But for now let's just use it for these files.

### DIFF
--- a/UWP/UWP.vcxproj
+++ b/UWP/UWP.vcxproj
@@ -74,7 +74,7 @@
   </ImportGroup>
   <PropertyGroup Label="UserMacros" />
   <PropertyGroup>
-    <AppxAutoIncrementPackageRevision>True</AppxAutoIncrementPackageRevision>
+    <AppxAutoIncrementPackageRevision>False</AppxAutoIncrementPackageRevision>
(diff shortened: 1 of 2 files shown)


unknownbrackets (approved):
The change looks good.  Is the `AppxAutoIncrementPackageRevision` change intentional?

-[Unknown]

GABO1423:
> The change looks good. Is the `AppxAutoIncrementPackageRevision` change intentional?
> 
> -[Unknown]

Yes, that option before increased the number version found in the .appxmanifest files which is supposed to be the same as the current PPSSPP version. So if you don't disable it, what is supposed to be version 1.16.0 becomes 1.16.1. So I disabled it by default.

PR #17974, opened 2023-08-25
(UWP) Another Round of Code Cleanups

ghost:
To quote myself on PR #17973:

> In theory we could do this for every single file in the assets folder, removing the need to use this command:
> https://github.com/hrydgard/ppsspp/blob/24ec26d9354462d554b44446d8e666da84354950/UWP/UWP.vcxproj#L110

Well, this is that idea put in practice. As the Git diffs show, we shaved ***a lot*** of code from these files, while still producing a complete package. An added side-effect of this approach is if a new shader, sound effect, or language file is added, we don't need to manually add it to the vcxproj files. They will be detected and used automatically! 

I also changed how the UWP Asset files are handled, it seemed more logical to handle them in a similar matter to how the `assets` folder and its contents were handled previously. This removes the need for two PreBuild commands as well: `mkdir Assets` and `robocopy "..\Assets\." "Content\." /xd "mime" /E /R:0 /W:0`.

Another change is the simplification of how we handle the appxmanifest files. Before, we had two files (Normal and Gold) that would overwrite a default manifest during build time. However, the Normal file was redundant since it's identical to the regular manifest file, so we can just use it instead of overwriting it when using the Release and Debug configurations. When using the UWP Gold configuration, the old behavior is still intact. As a bonus, I also added the Gold manifest to the Visual Studio Solution Explorer to allow for easier editing:

image

Finally, this PR also resolves a build warning that occurred near the end of the compilation, should improve the build speeds a bit.

This is meant to be a companion PR to #17973, which handles some of the debugger files already. So until that PR and ideally also #17952 are merged, I'll keep this as a draft. In the meantime, besides the Cl check, you can test if these changes work by simply making a appx/msix package and making sure that the `Content` folder is filled correctly.


hrydgard:
Both the other PRs are merged now :) feel free to rebase or merge from master, and I'll merge.

GABO1423:
@hrydgard This is ready to go now!

hrydgard:
It's not actively used, it was used once to convert some images. If we ever change the logo, might want to run it again :)

PR #17982, opened 2023-08-26
[Adhoc] An attempt to fix winsock error 10022 on Monster Hunter Freedom

### LINKED ISSUE #17969: Monster Hunter Freedom - multiplayer broke for some people
I got a report on the RetroAchievements discord that multiplayer broke for them in this game recently - after the introduction of RetroAchievements.

"Me and (other user) use Radmin to play multi, when we tried to upgrade the build it simply stopped working, no clue why"

~~Asking the user to try some builds now.~~

Apparently it broke between -581 and -598, pointing to #17744. That's weird though because that's the HTTPS upgrade, and it shouldn't be very relevant to multiplayer...

They are playing on Windows.

anr2me:
An attempt to fix https://github.com/hrydgard/ppsspp/issues/17969 assuming the socket error 10022 came from Winsock 1.1 when HTTPS Enabled.

If it was a real error (invalid argument being passed to `connect`) this PR won't makes a difference.

### DIFF
--- a/Core/HLE/proAdhoc.h
+++ b/Core/HLE/proAdhoc.h
@@ -84,7 +84,7 @@
 #define EALREADY WSAEALREADY
 #define ETIMEDOUT WSAETIMEDOUT
 #define EOPNOTSUPP WSAEOPNOTSUPP
-inline bool connectInProgress(int errcode){ return (errcode == WSAEWOULDBLOCK || errcode == WSAEINPROGRESS || errcode == WSAEALREADY); }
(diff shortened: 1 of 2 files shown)


ANR2ME:
Looks like all actions failed on macos build O.o

hrydgard:
Yeah there's some bizarre problem on Mac, I'll try to figure it out soon but it doesn't seem like a real error to me. It does build fine on Mac locally.

hrydgard:
So hm, we're getting upgraded or downgraded to winsock 1.1? I'm not really understanding how that's happening anyway. But let's try this.

ANR2ME:
We're using 2.2, but in case WinHTTP overrides it with 1.1 (not really sure whether it's possible or not tho)

hrydgard:
Hm, I suppose it might be an issue with who or what calls WSAStartup first...

Issue #18032, opened 2023-08-31
Gesture mapping L/R


Baduull:
### What should happen

[Controls]
gesture mapping, screen L/R
touch control button style, rounded rectangle & transparent
touch control, custom icon(suggest size)
touch control, dpad shape style
new button style, like a material
vibrate on click button with custom strength
sound on click button with custom volume

[ Context other emulators ] for analog, it feels uncomfortable because the movement does not follow the boundaries of the circle (fill box)

[User Interfaces]
make all background rounded rectangles
add button support via adsense (if it's possible right, I want to support the development of this emulator but I don't have much money)

### Who would this benefit

..

### Platform (if relevant)

Android

### Games this would be useful in

..

### Other emulators or software with a similar feature

Lemuroid

### Checklist

- [X] Check the [latest git build](https://buildbot.orphis.net/ppsspp/) in case it's already implemented.
- [X] [Search for other requests](https://github.com/hrydgard/ppsspp/search?q=my+issue&type=issues) of the same feature.

hrydgard:
There are some good ideas here, but can you try to make it a little more readable?

I don't understand "Gesture mapping L/R" at all, for example. What does it mean? What should be mapped to what?

Baduull:
sorry, for the current version, when there is a gesture on a part of the screen, the gesture will be executed, so I want when the left cannot be touched by a gesture and the right can be touched by a gesture (for example games like 3D fps), note: when I swipe, sometimes it happens twice touch likes executed

hrydgard:
Fixed by #21197

PR #18036, opened 2023-09-01
Analog stick gesture

### LINKED ISSUE #16820: Dynamic left analog stick
### What should happen

I propose to make a dynamic left stick.

I imagine this as some area where the stick can work(Stick area). Pressing in this zone will be considered the center of the stick (zero point), and any deviation from the center will already work as a classic stick.(move area)

In input settings make ability to change size for stick area and move area. 

Stick icon and areas doesn't draw in touchscreen input ui. (Draw only in settings)

### Who would this benefit

In shooters, you will no longer need to try to aim at the stick or select its optimal settings for a long time

### Platform (if relevant)

Android

any shooters

_No response_

### Checklist

iota97:
Fixes #16820. The area will be all screen with this, not the best but will get better if we get implement different gesture for the right and the left half of the screen as requested in #18032.

Could eventually support different button rather than only the analog stick, but that would require to "digitalize" the input, for now let's KISS.

Not sure if you had other ideas but this will at least kinda sidestep the UI issues a bit :)

The strings "Analog Stick" and "Sensitivity" are already in the language file, so only 1 string translation will be needed.

### DIFF
--- a/Core/Config.cpp
+++ b/Core/Config.cpp
@@ -804,6 +804,9 @@ static const ConfigSetting controlSettings[] = {
 
 	ConfigSetting("SystemControls", &g_Config.bSystemControls, true, CfgFlag::DEFAULT),
 	ConfigSetting("RapidFileInterval", &g_Config.iRapidFireInterval, 5, CfgFlag::DEFAULT),
+
+	ConfigSetting("AnalogGesture", &g_Config.bAnalogGesture, false, CfgFlag::PER_GAME),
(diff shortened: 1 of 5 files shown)


hrydgard:
I tested it briefly with the mouse and it works fine. Great to have as an option, and serves as a good replacement for #16820 indeed - except it kinda misses any visual feedback, but we can have it like that for now.

It's kinda confusing though that you can have both this and the regular stick enabled, though there's probably no way to avoid that.

Long term, I think we need to totally overhaul UI to make gestures more obvious as an option and a more natural alternative, because this is still a little obscure.

hrydgard (approved):

PR #18037, opened 2023-09-01
More axis input refactoring

hrydgard:
A couple of more steps towards faster input processing.

Again, this likely will have a negligible effect on performance on modern devices, but still good to minimize unnecessary overhead.


hrydgard:
I replaced this with #18251 , closing.

PR #18045, opened 2023-09-01
Update Russian translation

TotalCaesar659:
P.S. Almost every string in "Internals" is untranslatable.

Every notification test is too (just one example):


hrydgard:
I think I'm just gonna hide the internals section in non- debug build, it's just a place where I dumped some test stuff.

hrydgard (approved):

TotalCaesar659:
@hrydgard I found one string that's untranslatable and is just a copy of another one (but shortened). Is it really necessary there?

hrydgard:
It's kinda redundant, yeah. I'll get rid of them.

PR #18049, opened 2023-09-02
Add basic support for loongarch64

KatyushaScarlet:
Can run properly on loongarch64 CPU with FakeJIT for now.
Can get correct thread count and features of Loongson 3A5000 (4c4t) and 3A6000 (4c8t).
Because cpu_features added support for loongarch64 recently, it needs to be upgraded to the latest version.
Tested on 3A5000 and 3A6000


hrydgard (approved):

hrydgard:
ios build says:

```
/Users/runner/work/ppsspp/ppsspp/ext/cpu_features/src/impl_aarch64_macos_or_iphone.c:23:2: error: "Darwin needs support for sysctlbyname"
#error "Darwin needs support for sysctlbyname"
 ^
```

I think this is separate from the loongson changes and comes just from the cpu_features upgrade.

I'm going to make a separate PR that we should merge first, that just upgrades cpu_features.

hrydgard:
Right, so exactly as suspected: https://github.com/hrydgard/ppsspp/pull/18057 

I'll see if I can figure it out.

hrydgard:
OK, I've merged an upgrade of cpu_features.

Can you rebase this on master?

hrydgard:
Seems the conflict is about LOONGARCH64 vs LOONGARCH. What do you think?

KatyushaScarlet:
> Seems the conflict is about LOONGARCH64 vs LOONGARCH. What do you think?

I think using LOONGARCH would be fine.

KatyushaScarlet:
By the way, should the tag be "LoongArch64" ?

PR #18057, opened 2023-09-03
Bump the cpu_features submodule to latest

hrydgard:
Upgrading it separately from #18049 to see if we get the same build error.


unknownbrackets (commented):
This file should be updated:
https://github.com/hrydgard/ppsspp/blob/42741430bc3204248d6fc737f6f817910c73cd7c/ext/cmake/cpu_features/CMakeLists.txt#L69

There's some cpuinfo_loongarch.h stuff and HAVE_SYSCTLBYNAME is in there (just needs to move outside PROCESSOR_IS_X86.)  I think I had to rip out the testing stuff or something from here:
https://github.com/google/cpu_features/blob/fd4ffc1632db7b4e763bd28ffa6fc9d761cf3587/CMakeLists.txt

-[Unknown]

hrydgard:
Thanks, I think this will do the trick.

PR #18065, opened 2023-09-04
Add ability to dismiss notifications

hrydgard:
Adds the ability to dismiss notifications with a touch. 

Also hides the Internals tab in system info in non-debug-builds. Maybe I should move it to developer settings instead, but the point is, it does not need to be translated.


hrydgard:
Actually not blocking clicks from reaching underlying layers feels really bad, I'm not gonna merge this until I've solved that.

hrydgard:
There, that did it. Made sure to do the checking in AsyncTouch, to avoid slowing anything down.

PR #18090, opened 2023-09-07
[NetResolver] Implement sceNetResolver library

anr2me:
Yet another rough and experimental implementation taken from my stashed codes.

It works on most of the games i tested, also used for the test builds at https://github.com/hrydgard/ppsspp/issues/14256
However, since it use dummy id (1), it can only support one resolver created at a time.
It also enforce synchronous resolving (i think), including for the Async syscalls, this will lower the chance for the game to create more than one resolver, as the game will delete it once the result are out.

Leaving this as a draft until i can work on it later, at least to get the id properly generated to support multiple resolvers (ie. for games/app with multi-threading).

PS: Should also move it to a separate file (ie. sceNetResolver.cpp) as the general sceNet.cpp will grow bigger with infrastructure being implemented later.

### DIFF
--- a/Core/HLE/FunctionWrappers.h
+++ b/Core/HLE/FunctionWrappers.h
@@ -443,6 +443,11 @@ template<int func(int, u32, u32, int, int)> void WrapI_IUUII() {
 	RETURN(retval);
 }
 
+template<int func(int, u32, u32, int, int, int)> void WrapI_IUUIII() {
+	int retval = func(PARAM(0), PARAM(1), PARAM(2), PARAM(3), PARAM(4), PARAM(5));
+	RETURN(retval);
+}
+
 template<int func(int, const char *, int, u32, u32)> void WrapI_ICIUU() {
 	int retval = func(PARAM(0), Memory::GetCharPointer(PARAM(1)), PARAM(2), PARAM(3), PARAM(4));
 	RETURN(retval);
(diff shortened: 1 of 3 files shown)


ANR2ME:
As there are no information regarding the *Async syscalls i can find on the internet, will need to find out which games uses them first to be investigated.

hrydgard (approved):
Looks alright to me, I'm happy to merge after the 1.16.x release process is completed.

ANR2ME:
I intentionally not to save it in case it need additional data to be stored in the Resolver struct for the async version (may be used for the PollAsync syscall), but hopefully there is no additional data. So for now just simple reset it and let whatever the game were doing to fail when loading the state (as most of the time the connection won't be resumeable anyway), until i can get more info related to async syscalls.

ANR2ME:
Oops i forgot that loading a savestate won't trigger the init/shutdown functions.

hrydgard:
Hm, dunno why we didn't get this in, maybe I just forgot about it. Ready to merge? Because it looks like it.

anr2me:
I also forgot about this :) it should works, but since i'm no longer working on my old PR and someone else is working on it (ie. https://github.com/hrydgard/ppsspp/pull/18578 ), merging this could cause a conflict with @JustAWhiteBloodCell PR later.

hrydgard:
Oh right, I forgot that was incorporated there. Well, I'm going to close this then.

PR #18102, opened 2023-09-08
New logging channel "PRINTF" for SceKernelPrintf

hrydgard:
This is simply Nemoumbra's #18098 but with KERNELPRINTF changed to PRINTF and NOTICE changed to INFO.


unknownbrackets (approved):

Nemoumbra:
I was about to suggest manually disabling the channel by default just to keep the logs green.

Nemoumbra (approved):

PR #18109, opened 2023-09-08
Fix readme

iota97:
Not sure when it broke tbh.


hrydgard:
Oops, didn't notice. Thanks!

iota97:
FWIW it seems that the hovercard is back on GitHub readme, sadly seems to work only when there is a link to an issue and not just with the #id as it used to (good 'ld times).

PR #18143, opened 2023-09-12
Fix UMD disc swap with Retroachievements enabled

### LINKED ISSUE #17976: Disc swapping hang with RetroAchievements enabled
Reported on the RA discord:

> Loading DJMax Portable 2 first with achievements enabled, then loading Portable 1 through either the window menu or through the emu pause menu causes it to hang when the new ISO is selected. When doing it through the window menu I can still access the emu pause menu afterwards, but it'll freeze when selecting Continue

(this is when enabling Disc Link in the game)

hrydgard:
Shouldn't hang anymore. I think I'm talking to retroachievements the correct way now, so any errors would be in rcheevos' multi-disc support.

Fixes #17976

### DIFF
--- a/Core/HLE/sceUmd.cpp
+++ b/Core/HLE/sceUmd.cpp
@@ -487,12 +487,13 @@ static u32 sceUmdGetErrorStat()
 
 void __UmdReplace(const Path &filepath) {
 	std::string error = "";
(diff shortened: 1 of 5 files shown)


hrydgard:
Ugh, somehow forgot to merge this before launching 1.16.1. Oh well.

PR #18144, opened 2023-09-12
Update README.md for 1.16.1

hrydgard:


Felipefpl:
Isnt too soon for that? I mean, wouldnt be better to wait for 3 weeks before a release? Unless there are serious bugs that require an update.

hrydgard:
There is a very serious bug.

Felipefpl:
I understand, go for it then. :) :+1:

PR #18151, opened 2023-09-15
GPU, VFS, UI: Fixed minor memleaks

GermanAizek:
@hrydgard, hi.
Minor deletion allocated memory for objects.


hrydgard (approved):
Some of these could be solved by moving the new below the if, but this is fine.

PR #18154, opened 2023-09-15
Core, UI, Windows: Removed excess converting C-string in params

GermanAizek:
Minor code cleaning from unnecessary calls `c_str()`


hrydgard:
Did you find these with some static analyzer or some warning we should enable?

GermanAizek:
> Did you find these with some static analyzer or some warning we should enable?

I building in Clion with Clang 16 compiler, maybe some warnings were included.

hrydgard:
OK. Anyway, these should be safe - only time where removing .c_str() leads to badness is if you do it from a format string argument, and these all look good.

hrydgard (approved):

PR #18162, opened 2023-09-17
HTTPS: Proof-of-concept for linux

hrydgard:
Switches to using libcurl for https, instead of plain old http.

Works, but very slow for me for some reason.

Remaining work here is to update build instructions, add a proper compile time check for libcurl existence, etc.


hrydgard:
Rebased. Not seeing the slowness I complained about before.. maybe should get it in.

hrydgard:
Rebased this on master. Not sure where the previous bad performance came from, but now it seems to be fine.

However, has build errors, I assume we need some cmake/ci magic.

Installing the development library fixed all the CI except alpine.

Anyway, I'm holding off until post 1.20.4.

PR #18174, opened 2023-09-19
Disable Wayland if USE_WAYLAND_WSI is off

bentley:
If Wayland is installed, there is no way to disable Wayland support in the build. `USE_WAYLAND_WSI` is listed as a user‐editable option, but if Wayland is installed, setting `USE_WAYLAND_WSI=OFF` makes no difference; the build starts with Wayland enabled (problematic if, e.g., the system SDL2 does not have Wayland support).

This patch works for me: `USE_WAYLAND_WSI=ON` builds with Wayland support, and `USE_WAYLAND_WSI=OFF` builds without Wayland support, even if Wayland is installed.


hrydgard:
Hm, yeah, this looks like it makes more sense. 

@gucio321 , what do you think?

hch12907:
Is `USE_WAYLAND_WSI` on by default currently?

gucio321:
> Hm, yeah, this looks like it makes more sense.
> 
> @gucio321 , what do you think?

Yes, it looks more sensful.
Unfortunately I can't check it because I can't build the project at all :cry:
<details><summary>Details</summary>
<p>

I'm running `./b.sh` and getting this:
```console
/home/me/git/ppsspp/Core/RetroAchievements.cpp: In function ‘void Achievements::event_handler_callback(const rc_client_event_t*, rc_client_t*)’:
/home/me/git/ppsspp/Core/RetroAchievements.cpp:311:14: error: ‘RC_CLIENT_EVENT_ACHIEVEMENT_PROGRESS_INDICATOR_UPDATE’ was not declared in this scope; did you mean ‘RC_CLIENT_EVENT_ACHIEVEMENT_PROGRESS_INDICATOR_SHOW’?
  311 |         case RC_CLIENT_EVENT_ACHIEVEMENT_PROGRESS_INDICATOR_UPDATE:
      |              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |              RC_CLIENT_EVENT_ACHIEVEMENT_PROGRESS_INDICATOR_SHOW
/home/me/git/ppsspp/Core/RetroAchievements.cpp:315:14: error: ‘RC_CLIENT_EVENT_ACHIEVEMENT_PROGRESS_INDICATOR_HIDE’ was not declared in this scope; did you mean ‘RC_CLIENT_EVENT_ACHIEVEMENT_PROGRESS_INDICATOR_SHOW’?
  315 |         case RC_CLIENT_EVENT_ACHIEVEMENT_PROGRESS_INDICATOR_HIDE:
      |              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |              RC_CLIENT_EVENT_ACHIEVEMENT_PROGRESS_INDICATOR_SHOW

```

Should I open an issue or Its something simple that could be fixed in one step?

</p>
</details>

gucio321:
also @bentley I'd be greatful if there was a flag in `b.sh` to enable/disable wayland.
And I'd make USE_WAYLAND_WSI `ON` by default (or it is already set, I'm not sure)

hrydgard:
@gucio321 `git submodule update --init --recursive`

bentley:
> Is `USE_WAYLAND_WSI` on by default currently?

I don’t know how CMake works, but in git master, if Wayland is present it will be used no matter what.

My first commit builds without Wayland if `USE_WAYLAND_WSI` is unset (as opposed to on or off). With the second commit I just pushed, `USE_WAYLAND_WSI` defaults to `ON` (but unlike before, disables Wayland when turned off).

hch12907:
The latest commit builds and runs fine here (Arch Linux) with Wayland support turned on by default. Nice!

hrydgard:
Alright. Let's get this in for 1.16.3, and hope nothing else breaks :)

PR #18192, opened 2023-09-21
Fixed tmp string is cleared after moving variable (UB)

GermanAizek:
I think a useful change.

### DIFF
--- a/Common/File/DirListing.cpp
+++ b/Common/File/DirListing.cpp
@@ -185,7 +185,6 @@ bool GetFilesInDir(const Path &directory, std::vector<FileInfo> *files, const ch
 		while (*filter) {
 			if (*filter == ':') {
 				filters.insert(std::move(tmp));
-				tmp.clear();
 			} else {
 				tmp.push_back(*filter);
 			}


hrydgard:
I think this is actually needed to get the string back into a good state, otherwise the check after the while (*filter) loop is gonna run into undefined behavior itself?

Actually, I'll just remove the std::move stuff here, since I'm not feeling too confident.

fp64:
I don't think it was an UB. Moving out from stirng leaves it "in valid, but unspecified state" (see (8) [here](https://en.cppreference.com/w/cpp/string/basic_string/basic_string)).
In fact, the classic [move, simply](https://herbsutter.com/2020/02/17/move-simply/) by Herb Sutter is pretty clear, that it is supposed to hold in general: moving out should leave an object in a valid state, always.
Now, the state in question is "unspecified", so it is techincally legal to leave the string's value as `"good luck debugging this!"` which would than be promptly `insert`ed into `filters`. So `clear()` was on point.

I agree that just nuking `move` here makes more sense. The code wasn't particularly speed-oriented, anyway.

PR #18209, opened 2023-09-23
(UWP) Cleanup libchdr Files

ghost:
- Removes some redundant conditionals and duplicate code from the `libchdr_UWP.vcxproj` file, as well adding ARM64 support in place of the dead Win32 configuration.
- Removes all references to the dead x86/Win32 Build configuration in the `PPSSPP_UWP.sln` file, as well as fixing ARM64 builds (these were broken due to the libchdr project not defining this configuration) and the UWP Gold builds.


hrydgard (approved):

hrydgard:
thanks!

PR #18219, opened 2023-09-24
Make GetIndexBounds friendlier to autovectorization. Works on x86 at least.

hrydgard:
Not all games use indexed draws at all, but should be a small boost for those that do and draw a lot of geometry.

Although I noticed I was getting pmaxuw instructions on my machine, which are SSE4.1. Should the compiler really generate those? Weird! (MSVC)


fp64:
> Should the compiler really generate those?

Is it MSVC? Concerning either way.

fp64:
Ok, I can reproduce MSVC getting `pmaxuw` without SSE4: https://godbolt.org/z/KshjqTGGx, but it seems to be doing run-time detection, so it may be fine?

unknownbrackets (approved):

hrydgard:
Oh, I didn't read the disassembly very carefully (I just breakpointed and "show disassembly" in MSVC itself, which produces rather hard-to-read output). If it's runtime detected it's good.

PR #18240, opened 2023-09-25
UI/Localization: Partial rework of the Italian translation

Valtekken:
While I was fixing a few lines in the translation, I noticed more work was needed in multiple sections. This is a first revision/cleanup pass, more work will probably be needed. Will take a look whenever I have more time.

(Sorry for all these Localization PRs)


hrydgard:
No need to be sorry, keep them coming!

PR #18268, opened 2023-09-29
SDL: Use an "EmuThread" in Vulkan mode

hrydgard:
This avoids the problem of sampling input once per rendered frame, instead allowing inputs to arrive at any time, as seen from the emulation. Should help with the jittery feeling of input events reported on Linux by JMC for example.

Takes care of the Vulkan part of #17876 . Doing the same for OpenGL is iffier due to threading issues.

- [x] Add some missing functionality from the old loop
- [x] Test on Linux


fp64:
So, does https://github.com/hrydgard/ppsspp/pull/18056 needs fixing now?

hrydgard:
Hm, good point!

hrydgard:
Although they do seem to work as-is, I feel there might be a race conditoin.

PR #18289, opened 2023-10-02
Add SSE2 version of vfpu_dot

fp64:
See #18249. Speedup for this function ranges 10%..100%, depending on system. Updated verification and speed measurements: https://godbolt.org/z/W1z3sj6hz

Please feel free to double-check that it matches the current version (previous code - and verification - were wrong around NaNs, this is *hopefully* correct).

### DIFF
--- a/Core/MIPS/MIPSVFPUUtils.cpp
+++ b/Core/MIPS/MIPSVFPUUtils.cpp
@@ -879,8 +879,9 @@ static float vfpu_dot_sse2(const float a[4], const float b[4])
 float vfpu_dot(const float a[4], const float b[4]) {
 #if defined(__SSE2__)
 	return vfpu_dot_sse2(a, b);
-#endif
+#else
 	return vfpu_dot_cpp(a, b);
+#endif
 }
 
 //==============================================================================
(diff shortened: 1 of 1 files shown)


Core/MIPS/MIPSVFPUUtils.cpp
```diff
@@ -711,6 +712,177 @@ float vfpu_dot(const float a[4], const float b[4]) {
...
+}
+
+#endif // defined(__SSE2__)
+
+float vfpu_dot(const float a[4], const float b[4]) {
+#if defined(__SSE2__)
+	return vfpu_dot_sse2(a, b);
+#endif
+	return vfpu_dot_cpp(a, b);
+}
```
hrydgard, line 884: please do an #else, if nothing else for clarity.

fp64: Done.

fp64:
On a completely unrelated note, "JIT from IR" currently crashes in both local build and win32 build from buildbot (`ppsspp-v1.16.5-84-g008055d24-windows-x86`) in, apparently, any game (yes, I'm too lazy to file an issue properly).

hrydgard:
And in another unrelated note, Jit using IR is no longer crashing on Mac M1, for whatever reason.

PR #18299, opened 2023-10-03
Naruto video flicker heuristic take 2

hrydgard:
Revive the commits reverted in #18142 , for further work.

Not to be merged until we've actually solved the problems though! This broke for example Hot Shot Golf.


hrydgard:
Replaced by #18454

PR #18301, opened 2023-10-03
Fix memory leak in basisu_transcoder.cpp

visitorckw:
This commit addresses a memory leak issue by releasing the pPVRTC_work_mem resource when global codebooks are not unpacked yet.


hrydgard:
Is this an actual memory leak that you have detecte to be a problem?

Reason I'm asking is that we don't have any code trying to support using PVRTC compression and I wouldn't expect that code to ever run.

If this is indeed an acual bug, it should be reported to [upstream](https://github.com/BinomialLLC/basis_universal).

visitorckw:
No, this is not an actual memory leak that I've detected. I stumbled upon this memory leak issue while using the cppcheck static analysis tool to check other parts of the code. If this code segment is not expected to run, then it shouldn't be a problem.

hrydgard:
OK. I suggest you submit the fix to the upstream project then, we'll get it next time we update the basis transcoder.

PR #18306, opened 2023-10-04
Update ja_JP.ini

nishinji:
image
In Japanese, the order of the contents of %d is reversed
ex.
A of B
BのA
Please solve this issue.


hrydgard:
Oh, silly mistake. I'll make a fix so you can use %1 and %2.

hrydgard:
I'm gonna get https://github.com/hrydgard/ppsspp/pull/18308 in first, then you can update this to match.

nishinji:
@hrydgard done.
If you don't like the fix commit, i can discard it.

hrydgard:
It's alright.

PR #18320, opened 2023-10-08
Vulkan: Simplify pipeline and descriptor set layout, pool creation

hrydgard:
Wrap pipeline/descriptor-layout in a new object VKRPipelineLayout, and simplify the creation of those, and also of matching descriptor set pools.

This is early preparatory work for moving Vulkan descriptor set management off to the render thread, to speed up the main thread. This PR on its own shouldn't have any visible effect.


master1274:
v1.16.5-153-g43cdb88ac Completely non-working Vulkan.

hrydgard:
@master1274 thanks, I'll investigate soon.

PR #18325, opened 2023-10-08
(UWP) Update Manifest Files

ghost:
Updates version string to 1.16.5.

### DIFF
--- a/UWP/Package.appxmanifest
+++ b/UWP/Package.appxmanifest
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Package xmlns="http://schemas.microsoft.com/appx/manifest/foundation/windows10" xmlns:mp="http://schemas.microsoft.com/appx/2014/phone/manifest" xmlns:uap="http://schemas.microsoft.com/appx/manifest/uap/windows10" xmlns:rescap="http://schemas.microsoft.com/appx/manifest/foundation/windows10/restrictedcapabilities" IgnorableNamespaces="uap mp rescap">
-	<Identity Name="0ad29e1a-1069-4cf5-8c97-620892505f0c" Publisher="CN=Henrik" Version="1.16.0.0" />
+	<Identity Name="0ad29e1a-1069-4cf5-8c97-620892505f0c" Publisher="CN=Henrik" Version="1.16.5.0" />
 	<mp:PhoneIdentity PhoneProductId="0ad29e1a-1069-4cf5-8c97-620892505f0c" PhonePublisherId="00000000-0000-0000-0000-000000000000" />
 	<Properties>
 		<DisplayName>PPSSPP - PSP emulator</DisplayName>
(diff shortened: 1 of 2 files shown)


hrydgard:
I'm not sure we should add all these file extensions.. I noticed that on my Windows PC, the UWP build had somehow registered itself for them, but double clicking a game to launch didn't actually work for some reason.

GABO1423:
> I'm not sure we should add all these file extensions.. I noticed that on my Windows PC, the UWP build had somehow registered itself for them, but double clicking a game to launch didn't actually work for some reason.

Hmmm, alright then. I'll just make this a version string update then.

hrydgard (approved):

hrydgard:
I'll add updating this version number to my checklist so I don't forget next time. Or maybe we can find a way to automate it from the git tag...

ANR2ME:
> I'm not sure we should add all these file extensions.. I noticed that on my Windows PC, the UWP build had somehow registered itself for them, but double clicking a game to launch didn't actually work for some reason.

May be we're missing the "Handle the activated event"? https://superuser.com/questions/1599952/associate-file-extension-to-uwp-windows-10-app-not-in-open-with-list
But it only have C# and VB example, unlike the URI activation event that have C++ example too.

GABO1423:
> Or maybe we can find a way to automate it from the git tag...

I don't think that really can be done as-is, but I do know that if we generate the UWP vcxproj and sln files with CMake, it's possible to define the string based on the git tag via a CMake command. Of course that means adding CMake support for the UWP builds in the first place, but other than that it just has to be done manually sadly.

ANR2ME:
> > Or maybe we can find a way to automate it from the git tag...
> 
> I don't think that really can be done as-is, but I do know that if we generate the UWP vcxproj and sln files with CMake, it's possible to define the string based on the git tag via a CMake command. Of course that means adding CMake support for the UWP builds in the first place, but other than that it just has to be done manually sadly.

We can use powershell script to update the version isn't? https://www.codeproject.com/Articles/1023041/Automatically-Incrementing-the-Package-appxmanifes

GABO1423:
> > > Or maybe we can find a way to automate it from the git tag...
> > 
> > 
> > I don't think that really can be done as-is, but I do know that if we generate the UWP vcxproj and sln files with CMake, it's possible to define the string based on the git tag via a CMake command. Of course that means adding CMake support for the UWP builds in the first place, but other than that it just has to be done manually sadly.
> 
> We can use powershell script to update the version isn't? https://www.codeproject.com/Articles/1023041/Automatically-Incrementing-the-Package-appxmanifes

There's an inherent issue with that, when do we run it? We don't want to run it when using the workflow/making a build locally since this should only be increased when a new version of PPSSPP is released. And if we make a script at all, it would have to be run manually whenever a PPSSPP update is released, and at that point you might as well just manually update the string in the files anyway.

ANR2ME:
Yes, what i meant was updating the version string instead of incrementing it, just using the link as reference for the poweshell script to access the property.

PR #18335, opened 2023-10-09
Crazy Taxi rendering fix

hrydgard:
Fixes the hardware-renderer part of #18331


ANR2ME:
it's fixed now, thanks

PR #18337, opened 2023-10-10
Fix for Vulkan on 32-bit devices, oops.

hrydgard:
Should fix the issue on Shield Tablet reported in https://github.com/hrydgard/ppsspp/pull/18332#issuecomment-1754533756

Please test again after this is merged @master1274 .


master1274:
Thank you. Non-worker problem Vulkan corrected in v1.16.5-197-g4eecf5ab1

PR #18339, opened 2023-10-10
Remove the vertex cache option

hrydgard:
After the latest round of optimizations, which is now coming to a close, I can no longer measure a meaningful frame rate improvement ( > 1%) by using the vertex cache. I had hoped to reach this point, and I'm happy I did.

Since ~all it now does is to complicate the code, let's finally get rid of it.

Although, @unknownbrackets , might want to check if it still has a benefit on RISC-V, which doesn't do very well at vertex decoding?


unknownbrackets (approved):
Seems reasonable to me.  I didn't try again on RISC-V but archanox reported it didn't help all that much there anymore either, which doesn't surprise me.

At least on the device I have, most problems are what you'd expect: anything that you'd say "oh, I should vectorize this with SIMD" or "oh this might be using a lot of memory bandwidth" -> those things will be slow.  Verts fit that bill for sure, but hashing them is probably not cheap either.

-[Unknown]

PR #18347, opened 2023-10-12
v1.16.6 hotfixes - NOT TO BE MERGED YET

hrydgard:
This PR is here to check CI conveniently for the 1.16.6 branch.

It will not be merged until after 1.16.6 has been tagged.


hrydgard:
One remaining issue: On Mac, we seem to get in some trouble when exiting games now, likely due to the new sync somehow. I might have to find a different solution..

hrydgard:
Hm, why is CI not running??

PR #18350, opened 2023-10-12
[AdhocMatching] Detects leaving players better

### LINKED ISSUE #18242: Crazy Taxi Fare Wars - Multiplayer joining issue

ULUS10273

Apparently if you don't accept the invitation to join within around 500ms the room creator will be shown a connection error message thinking that the player has left, happened on both Crazy Taxi and Crazy Taxi 2.
image

The Logs between the received `JOIN` event to the `sceNetAdhocMatchingCancelTargetWithOpt` call didn't shows anything obvious, it didn't even use any adhoc/networking syscalls between them ~~(not sure whether i've enabled verbose logging or not on this build - will need to do more testing with verbose channel enabled on SCENET).~~

Suspected to be timeout issue where it goes too fast or miscalculated (ie. in case the "time goes backward" [issue](https://github.com/hrydgard/ppsspp/issues/13533) kicks in - will need to test this too if the game didn't use any networking syscall that return a timestamp).

PS: I haven't figured out how the game calculate the timeout yet.
PPS: In-game, after multiplayer game successfully started, the count down timer feels ticking faster (about twice faster)

### What should happen

~~It shouldn't detects that the players has left when it's still there waiting to be accepted.~~ should at least shows the same error message on both players.

### Logs

_No response_

### Platform

Windows

Geforce MX150

### PPSSPP version affected

v1.15.4 (probably latest version too)

### Last working version

not sure, never tested this game before

### Graphics backend (3D API)

Vulkan

### Checklist

anr2me:
This should allow Crazy Taxi & Crazy Taxi 2 to detects players that left during joining process, instead of waiting for a reply that will never come forever. Fixes https://github.com/hrydgard/ppsspp/issues/18242

Additionally, process packets and event messages quicker to handle more than 2 players better.

PS: Leaving this as a draft since i need to test this on more games


anr2me:
I didn't have the time to test this on other games, but this should probably okay.
Hopefully, other people could test this on games that use AdhocMatching to give feedback if anything broken.

hrydgard:
OK, let's get it in.

PR #18372, opened 2023-10-20
Add empty shape to custom keys

### LINKED ISSUE #18356: add "none" to button style: shape

iota97:
Long story short this is easier and less bug prone than adding checks all over the code.

No highlight effect sadly as the highlight shape is also used in the "Classic" button style (we could decouple that, but I prefer to keep this as simple as possible).

Note that the size of the image is used to calculate the context dimension and basically where you can press to activate the button.

Fixes #18356.

PS: you may want to merge this when you change atlas again to avoid binary bloat, even if without the font is about 71kB.


hrydgard:
Hm, I dunno, I actually think I would prefer a special case to just not draw the surrounding shape, rather than drawing empty space...

iota97:
I'm afraid is not just that: it would need at least a special case to add the entry in the UI selection, a special case in the UI preview rendering, a special case in the rendering and a special case in GetContextDimension.

The list of shapes is an array rn, the special case need to either be -1 as a value or if we use the next one handle new entry in that array to be shifted in the config value.

Is it really worth all this effort and added complexity?

hrydgard:
Okay, fair enough. Maybe not worth it... Let's get it in.

Though, I think you could revert the changes to the asciifont atlas, right?

iota97:
Oh sorry, I thought it wouldn't be updated and I didn't even realized that it got added - reverted.

PR #18376, opened 2023-10-23
DNS resolver warning logs fix

Nemoumbra:
The function `gai_strerrorA` returns a string localized in accordance with the system's culture. In my case this is the extended ASCII encoding called "866 (OEM - Russian)". The code that prints logs to the console expects a UTF8 encoding and fails here.

Old VS New:
image

I guess it wasn't found before because UTF8 encodes the English chars like in ASCII.


hrydgard:
Nice, thanks.

PR #18377, opened 2023-10-24
Game info screen: Add checks against the Redump database

### LINKED ISSUE #12469: Add option to verify ISOs against redump.org?

hrydgard:
The ReDump project keeps a database of title names, IDs and hashes of nearly all PSP games (and games for a lot of other platforms).

This adds a way to check your ISO against it, to make sure you managed to dump your UMD to ISO correctly through your PSP. By default, it just checks the ID and size, and if those are OK it'll show a notice to click Calculate CRC to verify. If not, it immediately shows a warning.

Needs a bit more polish before it goes in, both UI and code-wise.

We may consider doing a size check and warning users for bad ISOs on boot too, not just in the game info screen.

Fixes #12469


fp64:
Super-minor, but other overloads of `startsWith` can probably just call `std::string_view` version now?

fp64:
Also, unused parameter `delim` in `SplitCSVLine`.

Felipefpl:
> Fixes #12469

It's GREAT to see the oldest issues being fixed.

hrydgard:
Let's get this in and see how it works for people.

PR #18391, opened 2023-10-31
Add winhttp for cmake Windows builds

kotcrab:
I'm really not sure if this is the correct way to do it but I needed this change to build on Windows using CLion+MSVC since `naett` requires `winhttp`.


hrydgard (approved):

hrydgard:
If that build flag is set though, naett shouldn't be building. I think this change is ok for now, though.

hrydgard (approved):

Issue #16904, opened 2023-02-03
Driver '76 instability


hrydgard:
BLOC 15 on discord has been playing Driver '76 (ULES-00740), and encountered a hang right after this save state:

[Driver '76 ULES00740_1.01_0.zip](https://github.com/hrydgard/ppsspp/files/10581526/Driver.76.ULES00740_1.01_0.zip)

He also reports various random-feeling crashes.

Panderner:
Still happens when simulate UMD delays is set?

blocxv:
not work if i use this setting game freeze when i launch an savestate my trouble seem apear only in mission 3 when we collect 3 truck

Gamemulatorer:
Related issue https://github.com/hrydgard/ppsspp/issues/12054

sum2012:
the freeze problem should be sceAtrac problem.
29:25:305 ATRAC3 play  I[ME]: HW\MediaEngine.cpp:87 FF: JS mono Sound Unit id != 3.
29:25:305 ATRAC3 play  I[ME]: HW\MediaEngine.cpp:87 FF: Frame decoding error!
29:25:305 ATRAC3 play  E[ME]: HLE\sceAtrac.cpp:604 avcodec_decode_audio4: Error decoding audio -1094995529 / bebbb1b7
full log: https://gist.github.com/sum2012/4d1d9d6cdd95d813c8b1d9f2923fe8da

sum2012:
crash full debug v1.16.5-83-gb85f7e28a log
[ppsspplog.zip](https://github.com/hrydgard/ppsspp/files/12781727/ppsspplog.zip)
using interpreter or UMD delay also not work

sum2012:
I need a game save rather than save status to test

sum2012:
I play the game to myself, after 1 stage, auto save ->
[ULES00740001.zip](https://github.com/hrydgard/ppsspp/files/13257937/ULES00740001.zip)

In mission 2 save status->
[ULES00740_1.01_2.zip](https://github.com/hrydgard/ppsspp/files/13257940/ULES00740_1.01_2.zip)

Keep drive in font,you will freeze a while then car graphic error ->
 
![2](https://github.com/hrydgard/ppsspp/assets/2177532/804bcec8-b944-417f-a9c4-65b06b313b25)
![3](https://github.com/hrydgard/ppsspp/assets/2177532/a054437d-ed1c-4b31-a5f1-a4ae7f88439e)
[...]

PR #18396, opened 2023-11-04
Add AtracDecodeDataRemainFrameHack

sum2012:
keep return less than 1 RemainFrame in sceAtracDecodeData to get valid data .when RemainFrame  = 0 ,then psp will call sceAtracAddStreamData and sceAtracGetStreamDataInfo.

Work around  #15233

Modify log:
https://gist.githubusercontent.com/sum2012/2209e25bdc9cfe79f5a57337fbf67f11/raw/484d6e234e829bc393aefc8614996eba1d09497f/gistfile1.txt


ANR2ME:
I wondered whether this can also fix all of those audio issue on other games too or not, most of them seems to use atract3+ too.

sum2012:
@ANR2ME Init test ,this hack only apply that don't always use  sceAtracGetRemainFrame() (line 1661 of this pull request) and  sceAtracGetNextSample() (line 1606 of this pull request)

 #16904 may help but need a game save rather than save staus

sum2012:
I just find a game (K-ON_Houkago_Live_JPN) that jpcsp emulator decode Atrac audio also have problem (infinite loading after save )

sum2012:
in ppsspp side, I think that in looping with (atrac->bufferState_ == ATRAC_STATUS_STREAMED_LOOP_FROM_END) need to do something in some where because  sceAtracAddStreamData give corrupt atrac data after loop 2 times

sum2012:
I find a reasonable fix K-ON_Houkago_Live_JPN but unlucky not match JPCSPTrace log.
Amway , I will test all games that use sceAtracGetNextSample

sum2012:
I find better hack which can apply more game (which replace AtracLoopHack )
But I need test more Atrac problem games

} else if (bytes_read < 0) {
ERROR_LOG_REPORT(ME, "avcodec_decode_audio4: Error decoding audio %d / %08x", bytes_read, bytes_read);
//failedDecode_ = true;
//return ATDECODE_FAILED;
packet_->size = 0;
return ATDECODE_BADFRAME;
}

hrydgard:
Closing in favor of #20098

Issue #18404, opened 2023-11-10
 Xinput triggers mapped to analog sticks could pollute internal analog sticks state


Kethen:
### Game or games this happens in

all games, if triggers were mapped under xinput

### What area of the game / PPSSPP

It seems that Xinput devices would not send trigger values unless it's over XINPUT_GAMEPAD_TRIGGER_THRESHOLD

https://github.com/hrydgard/ppsspp/blob/2036f1c36a4023c2e4525002a3817f0d93a4ab83/Windows/XinputDevice.cpp#L240

So it never sends anything under XINPUT_GAMEPAD_TRIGGER_THRESHOLD after a higher value was sent, leaving a stuck analog value > `30.f/255.0f ~ 0.1176` on internal state

The stuck value can then further affect analog normalizing on control mapper

### What should happen

Idle trigger should not maintain an analog value

Maintaining a value could also be the cause of #16643

Perhaps when XINPUT_GAMEPAD_TRIGGER_THRESHOLD is not met, it should gradually send values until 0 so that control mapper can also drop back to under deadzone? Or maybe send `clamp((x - 30.0f)/(255.0f - 30.0f), 0.0, 1.0)` as trigger input and skip the check

### Platform

Windows

### Mobile phone model or graphics card

Not relevant

### PPSSPP version affected

v1.16.6 release, v1.16.6-290-g2036f1c36

### Graphics backend (3D API)

Vulkan

### Checklist

- [X] Test in the [latest git build](https://buildbot.orphis.net/ppsspp/) in case it's already fixed.
- [X] [Search for other reports](https://github.com/hrydgard/ppsspp/search?q=my+issue&type=issues) of the same issue.

Kethen:
Since https://github.com/hrydgard/ppsspp/commit/633a6f612b44b36e12ba34c7ded3147897d7e3af, it might make sense to just skip the deadzone checking for triggers as well

hrydgard:
Fixed by #18407.

Issue #16643, opened 2022-12-22
Analog stick stuck - MFHU - multiinstance - x360


Dschogo:
### Game or games this happens in

ULES01213 - Monster Hunter Freedom Unite (EUR)

### What area of the game / PPSSPP

When multiple instances are open in-game (2-4 tested), sometimes the right analogstick (which is bound to the d-pad to look around) gets stuck. It only happens if multiple instances are active. Flicking the stick/d-pad in the same direction again stops it. - I assume the mapped "key up" event gets "lost" somewhere? PPSSPP gets confused with the multiple-nearly parallel controller inputs, since it only happens if other player are actively playing/fighting

(controllers working fine in other non PPSSPP games like Movin out, emulated switch via yuzu, gampad-tester.com also shows no "stuck" xinput)

The four instances of PPSSPP are seperate portable installs, for individual saves.
Controlmappings are completly cleared except for the specific controller in each instance. (x360_something in the first, x360_1_somthing in second, etc)
Analog-Stick calibration doesn't help.

Emulation settings are default - except Adhoc to local and PostFX (Natural Color).

Additionally, is there a start parameter or setting to disable auto-muting other instances? I get the Idea but in MHFU its possible to disable the background music and we like the FX from every player. RN i made an ahk script with quickly pushes the volume to 10 after an instance gets booted.

### What should happen

Stick/cam shouldnt "stick"/ cam should stop spinning.

### Logs

Since it's no crash - don't think there is value - i'll still attach one in the following days

### Platform

Windows

### Mobile phone model or graphics card

win10 (22h2) - GTX1060 (517.56) - ryzen 1700 - 4 x360 controller

### PPSSPP version affected

1.14.1 (2d1f8f0)

### Last working version

none - issue already persisted in 1.7 (tested the custom adhoc build)

### Graphics backend (3D API)

Vulkan

### Checklist

- [X] Test in the [latest git build](https://buildbot.orphis.net/ppsspp/) in case it's already fixed.
- [X] [Search for other reports](https://github.com/hrydgard/ppsspp/search?q=my+issue&type=issues) of the same issue.
- [X] Try resetting settings or older versions and include if the issue is related.
- [X] Try without any cheats and without loading any save states.
- [x] Include logs or screenshots of issue.

Dschogo:
Example video of three instances:

- Controller 1 and 2 right analog stick random movement (by me for simulating ingame camera movement)
[...]

PR #18407, opened 2023-11-11
skip xinput trigger threshold check

Kethen:
Address #18404 , remove the hard coded 30.0f/255.0f deadzone on triggers

It should be fine since analog deadzone was moved to control mapper around https://github.com/hrydgard/ppsspp/commit/633a6f612b44b36e12ba34c7ded3147897d7e3af

To further address #16643 with analog mapped to digital, I'm guessing KeyMap thresholds should be evaluated (if that still happens), but this should already make it harder for analog triggers mapped to digital buttons to get stuck

https://github.com/hrydgard/ppsspp/blob/2036f1c36a4023c2e4525002a3817f0d93a4ab83/Core/KeyMap.h#L81-L83


hrydgard:
Huh, yeah, indeed forgot to remove this old deadzone handling when centralizing.

Should also just get rid of the function NormalizedDeadzoneDiffers, but I'll sneak that into the merge.

PR #18411, opened 2023-11-11
allow `ALL = true` then `gameID = false` blacklisting

Kethen:
Rearrange plugin enumeration code to allow list `ALL = true` to load a certain plugin on all games, then `<gameID> = false` to not load the plugin on certain games


hrydgard (approved):
The logic with the extra case and "continue" isn't that obvious, but looks like it'll work.

PR #18430, opened 2023-11-17
Handle SCE_UTILITY_SAVEDATA_TYPE_SAVE then SCE_UTILITY_SAVEDATA_TYPE_WRITEDATA

Kethen:
Modnation racers does a normal save for global profile, then writes an insecure file for keeping track of selected language. During SCE_UTILITY_SAVEDATA_TYPE_WRITEDATA it also provides no data for param.sfo, so I'm assuming that utility.prx does not update param.sfo data in that mode since the PSP save provided on the issue tracker has information.

This is an attempt to fix #6078

### DIFF
--- a/Core/Dialog/SavedataParam.cpp
+++ b/Core/Dialog/SavedataParam.cpp
@@ -496,13 +496,18 @@ int SavedataParam::Save(SceUtilitySavedataParam* param, const std::string &saveD
 	std::string sfopath = dirPath + "/" + SFO_FILENAME;
 	std::shared_ptr<ParamSFOData> sfoFile = LoadCachedSFO(sfopath, true);
 
+	bool subWrite = param->mode == SCE_UTILITY_SAVEDATA_TYPE_WRITEDATASECURE || param->mode == SCE_UTILITY_SAVEDATA_TYPE_WRITEDATA;
+	bool wasCrypted = GetSaveCryptMode(param, saveDirName) != 0;
+
 	// Update values
-	sfoFile->SetValue("TITLE", param->sfoParam.title, 128);
-	sfoFile->SetValue("SAVEDATA_TITLE", param->sfoParam.savedataTitle, 128);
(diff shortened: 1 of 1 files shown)


hrydgard:
Quite interesting, and that explanation of the behavior sounds plausible indeed.

Though, ideally, this should of course be tested on hardware using pspautotests. 

I assume most other games wouldn't be affected by this, since they don't combine SAVE and WRITEDATA like that?

Kethen:
I don't have a psp with me currently so it'll be difficult for me to expand autotest with secure save then secure/insecure, then run on a real device for expectations

@sum2012 would you be able to do that?

sum2012:
@Kethen Well done ,it is really fix https://github.com/hrydgard/ppsspp/issues/6078
I don't know about autotest

hrydgard:
This does make a lot of sense, thinking about it some more. And while we really should write a test, I'll get it in for now so we can confirm that it doesn't break other games somehow. It doesn't seem likely that it would, but you never know..

hrydgard:
OK, so it turns out that it subtly broke at least Harvest Moon, well, the game still works but the PARAM.SFO metadata isn't updated correctly. So, we'll have to come up with a different check.

hrydgard:
When ModNation has no info, it has a `param->overwriteMode` of 0, while Harvest Moon uses a `param->overwriteMode` of 1. Might be related (but no proof). Also "multiStatus" is set to 2, vs 0 in ModNation.

I guess we could also check for empty strings and ignore the param sfo write..

Kethen:
> When ModNation has no info, it has a `param->overwriteMode` of 0, while Harvest Moon uses a `param->overwriteMode` of 1. Might be related (but no proof). Also "multiStatus" is set to 2, vs 0 in ModNation.
> 
> Though, probably neither is related, because Wipeout Pure uses 0 for both parameters, too, while providing correct data.
> 
> I guess we should just check for empty strings and ignore the param sfo writes in that case...

To fully narrow this down it might require setting up multiple sdk version autotest on real hardware

PR #18433, opened 2023-11-20
Fix building on OpenGL ES 2.0 (RPI VideoCore IV)

joolswills:
GL_COMPRESSED_RGBA_ASTC_4x4_KHR is not defined and looks to be an OpenGL ES 3.0 extension.

This fixes building on the Raspberry Pi 1-3 using VideoCore IV OpenGL ES 2.0 headers.

### DIFF
--- a/Common/GPU/OpenGL/DataFormatGL.cpp
+++ b/Common/GPU/OpenGL/DataFormatGL.cpp
@@ -146,12 +146,14 @@ bool Thin3DFormatToGLFormatAndType(DataFormat fmt, GLuint &internalFormat, GLuin
 		alignment = 16;
 		break;
 
+#ifdef GL_COMPRESSED_RGBA_ASTC_4x4_KHR
 	case DataFormat::ASTC_4x4_UNORM_BLOCK:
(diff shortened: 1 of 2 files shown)


joolswills:
Not sure if this is the right fix, so please feel free to comment/change/disregard. But GL_COMPRESSED_RGBA_ASTC_4x4_KHR doesn't exist in the RPI VideoCore IV GLES2 headers, and from what I have read online ASTC Texture Compression is a GLES3 extension.

hrydgard:
Unfortunately our old define USING_GLES2 dates back to before GLES3, so it gates that too.

I'll try to come up with something else, but I think simply

```
#ifdef GL_COMPRESSED_RGBA_ASTC_4x4_KHR 

...

#endif
```

would work, as that's a define.

joolswills:
Thanks. I changed the commit as you suggested as that's a better fix (and have used that in RetroPie - https://github.com/RetroPie/RetroPie-Setup/pull/3810).

hrydgard (approved):

PR #18438, opened 2023-11-21
OpenXR - Projection matrix on Quest 3 fixed

lvonasek:
This fixes many games on Meta Quest 3. For some reason the previous calculation of projection matrix didn't work on Quest 3. This is a simplified version which is not compatible with FOV feature. I think it is good to remove the FOV feature, it worked only if watching forwards anyway. It would be good to have it on 2D screens but for VR it is bad.

### DIFF
--- a/Common/VR/PPSSPPVR.cpp
+++ b/Common/VR/PPSSPPVR.cpp
@@ -26,13 +26,6 @@
 #include "Core/KeyMap.h"
 #include "Core/System.h"
 
-enum VRMatrix {
-	VR_PROJECTION_MATRIX,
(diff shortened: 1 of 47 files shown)


hrydgard (approved):
Seconding sum2012's suggestion, otherwise seems fine.

Thank you for taking care of this.

lvonasek:
I cleaned the texts as suggested. VR build upoaded here:
https://drive.google.com/drive/folders/1609uufJhLZlc0GrcXxis45312erzHZMb?usp=drive_link

Feel free to post it on your Discord.

Issue #18465, opened 2023-12-02
Texture replacement priority regression


hrydgard:
Reported on Discord by ]Silver ❄「Suzuna」:

It seems that in the past, in this case:

* Texture 0962ec60df1a2fba49d1f4e8 
* 0962ec60df1a2fba49d1f4e8.png present in root
* The line `0962ec60df1a2fba49d1f4e8 = Map/Villa/Yukumo16.png` present in ini

We previously prioritized the named PNG, but now the hash-named one in the root has priority. This isn't intentional and should probably be fixed.

hrydgard:
Closing, fixed by #18466.

PR #18466, opened 2023-12-02
Texture replacement: Prioritize ini file [hashes] section over just files in the "root" folder.

hrydgard:
This reverts back to what the old behavior seems to have been, as detailed in #18465 .

I think this does make more sense, too.


unknownbrackets (approved):
Yeah, that behavior was previously very intentional.  I mean, it feels obvious - you're specifying the path for the files in the ini, so the default behavior shouldn't even activate.

-[Unknown]

hrydgard:
Yeah, my bad for screwing this up, it seems.

PR #18506, opened 2023-12-10
[iOS] Update JIT enable method from project UTM

Halo-Michael:
See at https://github.com/utmapp/UTM/blob/714871a85d8004c80a53cc1d79a1f17d3b2c4ff2/Services/UTMJailbreak.m
This pull will allowed PPSSPP can self-enable JIT with entitlements:
https://github.com/Halo-Michael/ppsspp-builder/blob/master/ent.xml
It's possible with project [TrollStore](https://github.com/opa334/TrollStore), which allowed people sign applications with custom entitlements till iOS 16.6.1/17.0.

### DIFF
--- a/ios/main.mm
+++ b/ios/main.mm
@@ -3,12 +3,16 @@
 #import <UIKit/UIKit.h>
 #import <dlfcn.h>
 #import <mach/mach.h>
+#import <mach-o/loader.h>
+#import <mach-o/getsect.h>
 #import <pthread.h>
+#import <spawn.h>
 #import <signal.h>
 #import <string>
 #import <stdio.h>
 #import <stdlib.h>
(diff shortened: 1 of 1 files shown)


Halo-Michael:
It's how will it works look like:

https://github.com/hrydgard/ppsspp/assets/45062120/db3c0b67-3fee-4734-988b-48cfb67c48b2

hrydgard:
Cool!

Will this affect anyone using older methods or devices?

Halo-Michael:
If PPSSPP installed with no special entitlements(but must have get-task-allow), they can still using xcode attach to enable JIT.

hrydgard:
OK. Not 100% I understand all the implications, but let's try it and see if we get any complaints.

Issue #18509, opened 2023-12-10
Brothers in Arms: D-Day - slow gameplay


hrydgard:
https://github.com/hrydgard/ppsspp/assets/130929/a9926168-914d-42c1-ae09-1c4569d5e8b8

penderfender.objectivebutter on Discord reports that gameplay in Brothers in Arms: D-Day is running weirdly slowly compared to on the real hardware. It sure looks lke in the video, PSP on top, PPSSPP on the bottom.

Panderner:
Movement Speed for this game is tied to frame rate.

Movement speed is slow if the game is running at 60 FPS: 
https://streamable.com/hqrreo

The Moving Speed is okay if running at 30 FPS using underclocks:
https://streamable.com/ip4lov

hrydgard:
Hm, interesting. Maybe we should just lock framerate to 30hz similarly to how we lock God of War to 60hz. 

If you put the ID under [GoWFramerateHack30], is it better? Do menus break or slow down or something?

LunaMoo:
Those GOW hacks are via function replacements since it worked better than the old method and worked around game bug, unlikely to affect different games. We'd need an alt version of ForceMax60FPS for more universal approach.

hrydgard:
Oh right, forgot about that. I'll set something up.

LunaMoo:
There's a secondary issue here althrough probably caused by same/similar thing. The game's often is running under 100%  even if hardware can is more than capable running it in hundreads of FPS. Looked at it and it's using sceKernelGetSystemTimeLow twice every frame, probably uses this time to decide how it should run. Maybe it doesn't eat enough cycles?

LunaMoo:
I believe checking if and how much scePowerGetCpuClockFrequencyInt eats cycles on real hardware would be a way to fix this, otherwise a compat hack that makes it eat 200 cycles would be fine. (That's just the third value I tested, not sure how little would be enough.)

Panderner:
Another game affected by high frame rate issue is Blitz: Overtime which runs too fast if running at 60 FPS.

LunaMoo:
>Blitz: Overtime

Unless we want to collect all issues caused by "PPSSPP running too fast" in here, it probably should get it's own issue.
[A workaround exist for it.](https://github.com/LunaMoo/PPSSPP_workarounds/blob/48711e22f63baf61252b20cde2abc68ca1f1547e/cheat.db#L1443)

Panderner:
[...]

PR #18513, opened 2023-12-10
Work around Brothers in Arms speed issue via compat hack

LunaMoo:
Technically overclocking emulated PSP CPU should be enough to work around #18509 issue, this compat hack just makes it animated with correct speed without OC to keep the game more friendly for lower end devices, however even with it, it'll still work better with OC PSP CPU due to dropping FPS whenever too much happens in-game with default clocks. Not sure about this solution since I had a lot of confusion during testing, but opening PR to make it easier for others to test.


hrydgard:
Such large delays in that function is a little too gross and not accurate.. Maybe there's something else the game calls a lot that should actually be slower by an amount that adds up to that. 

Or maybe we'll just need to accept to forcibly downclock, since we know our CPU emulation really is too fast, due to not emulating the impact of cache misses. Or, since it seems the game's behavior is quite confusing, maybe it's really something else..

LunaMoo:
Just to add, the theory about game speed being tied to FPS is false, it's either full speed or half speed. FPS is variable depending on how heavy the scene is.

 I think the confusion it's really more about me being ill and getting a fewer while looking at this and forgetting that PPSSPP setting for PSP CPU clock doesn't apply when leaving settings and returning to game, but only after game changes it or if we restart the game(loading savestates also work), so I kind of figured out how to work around the issue, but it wasn't perfect as it couldn't keep 60 fps constantly, so I OC PSP CPU, but saw no difference, so I then tried a different commonly repeated function and since the OC PSP CPU activated only then and I thought it didn't affect the issue at all, any results later just keep confusing me.

 Actual behaviour isn't that messed up, some game function probably activates at wrong order or triggers incorrect amount of times due to PPSSPP being too fast and we end up with half of the speed, we can counter it without any code changes by the following:
 - underclocking PSP CPU, but underclocking too much will ruin the framerate when more stuff happens on the screen, it'll also be variable,
 - overclocking PSP CPU, it's perfect if we can achieve 60/60, but that's not ideal for weaker/mobile devices.

 The hack here effectively works like underclocking, but without forcing variable/reduced fps to people that have hardware which can easily run OC PSP CPU.

anr2me:
Hmm... GetSystemTime should be a fast syscall isn't? since it's usually used to measure timeouts or other time-based calculation, if it's delayed by 70ms the outputted value might no longer be accurate.

LunaMoo:
If I understand, eating cycles does not mean delaying result, it just makes it heavier on the PSP CPU, at the same time lighter to emulate.

 Either way this isn't correct change, just a hackish workaround to make it work out of the box, I'm not saying it's a good change, but speaking from PC user perspective much better than enforcing 30fps due to missunderstanding that it's speed would be tied to fps.

 Correct change would mean to slow down every function to a realistic value, this would also fix other problems, but which is something that might never happen.

anr2me:
Ah right, the value is in cycles and not micro seconds, sorry i forgot,  how much delay does 70k cycles be (based on the clock the game use by default)?

LunaMoo:
I guess it's 0,0315ms, but again, this doesn't delay result, we have different function "hleDelayResult" for that, this is just using "hleEatCycles" and it just reduces budget.

 The problem is many PPSSPP functions doesn't eat out anything from that budget basically being free or using up a tiny constant value based on tests where in reality how long it takes might depend on things, when the game doesn't limit itself and just packs as much as it can within that budget, functions being free or very cheap makes it do much more work and also increases emulation requirements, ocassionally this also causes crashes or messes up animation speed or fps, it's pretty random and large variety of problems games in PPSSPP have still comes from that.

 And yes, the game does render as much as it can, but it's not like GOW where it would go to 100+, with the change in here being between 58-60+ when little to nothing happens and drops to 30 or maybe less when a lot stuff happens on the screen. Without the change it also runs poorly and not being able to hold 100% fps even on hardware where it can keep much more with unthrottle and I'm not sure why is that as it seems counter-intuitive to affect emulation speed %, but this change works around that as well.

hrydgard (approved):

PR #18517, opened 2023-12-11
Run-behind-pause: Fix some edge cases with the transparent background setting

hrydgard:
If you turn off transparent background, the background will now become transparent anyway if you unpause when the pause screen is up, so the EmuScreen can run.

Also, the DisplayLayoutScreen works again with the option disabled.

Followup to #18515

### DIFF
--- a/Common/UI/Screen.cpp
+++ b/Common/UI/Screen.cpp
@@ -257,13 +257,6 @@ void ScreenManager::sendMessage(UIMessage message, const char *value) {
 	}
 }
 
-Screen *ScreenManager::topScreen() const {
-	if (!stack_.empty())
-		return stack_.back().screen;
-	else
-		return 0;
-}
-
 void ScreenManager::shutdown() {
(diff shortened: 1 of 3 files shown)


IrfanH495:
just information version ppsspp-v1.16.6-497-gbf054203e-android
opengl black screen
the screen goes black
Screenshot_20231212-042552_PPSSPP
I have to go into the menu to see it
Screenshot_20231212-042559_PPSSPP

but this is a great feature, I wanted something similar but without the menu just a stop game button so I could read the subtitles.

hrydgard:
I'll fix that shortly, thanks!

Gamernumberone:
to me [v1.16.6-499-g8f75649b4] black screen

hrydgard:
@Gamernumberone yup, expected. I just merged it so it's not in the builds yet (but will be soon). As you can see in the listing right now when I'm typing this, the name of that PR is not in the list of builds yet.

PR #18525, opened 2023-12-12
[GPU/GLES] Using reserve if possible

GermanAizek:
@hrydgard minor changes using reserve() for vector if we know count elements.


hrydgard (approved):
Looks good to me.

PR #18526, opened 2023-12-12
[GPU/Vulkan] Fixed assignment index to itself

GermanAizek:
@hrydgard I'm not sure if I chose assigned variable correctly, but maybe index counter was meant. The place found is very strange and clearly meant something else.


hrydgard:
Oh, no that one I just put there temporarily to have somewhere to breakpoint. It shouldn't have been merged.

I'll just delete those three lines.

PR #18530, opened 2023-12-12
[Common/Core/GPU/UI] Code refactor C++17 range-based loop and const reference objects

GermanAizek:
@hrydgard more correct using range-based loop C++17, for strict types loop elements.

### DIFF
--- a/Core/HLE/sceNet.cpp
+++ b/Core/HLE/sceNet.cpp
@@ -450,7 +450,7 @@ void __NetApctlCallbacks()
 	// How AP works probably like this: Game use sceNetApctl function -> sceNetApctl let the hardware know and do their's thing and have a new State -> Let the game know the resulting State through Event on their handler
 	if (!apctlEvents.empty())
 	{
-		auto args = apctlEvents.front();
+		auto& args = apctlEvents.front();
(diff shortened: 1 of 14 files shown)


hrydgard (changes requested):
This looks mostly fine, but storing the concatenation of strings to references looks weird to me, so please revert those.

Core/HLE/sceNet.cpp
```diff
@@ -450,7 +450,7 @@ void __NetApctlCallbacks()
 	// How AP works probably like this: Game use sceNetApctl function -> sceNetApctl let the hardware know and do their's thing and have a new State -> Let the game know the resulting State through Event on their handler
 	if (!apctlEvents.empty())
 	{
-		auto args = apctlEvents.front();
+		auto& args = apctlEvents.front();
```
anr2me, line 453: As i remembered this is an intentional copy since below it the element is being popped out (`apctlEvents.pop_front();`), i wasn't sure whether it's safe to reference a popped out element thus not using `auto&`.
Hopefully it's safe.

hrydgard:
Replacing with https://github.com/hrydgard/ppsspp/pull/18541 .

PR #18534, opened 2023-12-12
Use the universally supported Vulkan "565" 16-bit texture format

hrydgard:
Previously mistakenly used the BGR format instead of the RGB, which has wide but not universal support, so sometimes we'd fall back to converting to 32-bit.

Probably won't make much of a difference for anything, but may affect #17881 if my theory about it is correct.

Also minor cleanups.


IrfanH495:
Sorry, want to ask is this an issue?
black screen after selecting shaders
I've tried the last build, 

https://github.com/hrydgard/ppsspp/assets/78854859/76ff205e-0e96-4c53-8522-6a4c8e242b16

hrydgard:
Hm, I'll have a look at that. Unrelated to this particular commit though I'm sure.

hrydgard:
That's a weird one, fortunately it resolves itself as soon as you resume playing, so it's not terrible.

But I'll look into it later this week.

IrfanH495:
> Unrelated to this particular commit though I'm sure.

Yes, right
I haven't found it yet
tried to install the one downloaded yesterday and didn't find it
Screenshot_20231213-053747_ZArchiver

don't worry, I rarely use it, only in games where the image is dark, there's no need to rush to fix it, I'm very lazy if I have to create a new issue.

PR #18536, opened 2023-12-12
OpenXR - Gran Turismo camera shaking workaround

lvonasek:
Since https://github.com/hrydgard/ppsspp/pull/18438 there is VR camera shaking in Gran Turismo. This PR fixes that on all headsets.

### DIFF
--- a/Common/VR/PPSSPPVR.cpp
+++ b/Common/VR/PPSSPPVR.cpp
@@ -44,7 +44,7 @@ static int vr3DGeometryCount = 0;
 static long vrCompat[VR_COMPAT_MAX];
 static bool vrFlatForced = false;
 static bool vrFlatGame = false;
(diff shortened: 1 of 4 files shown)


lvonasek:
Interesting. I wouldn't expect the shaking is an intention. I leave it on you but I find pity that the best looking PSP game don't work out of the box (even that it worked in the previous version). The new compat flag is basically switching between an old projection matrix Hack and new projection matrix hack.

hrydgard:
How did it work in VR previously?

lvonasek:
It used the projection matrix from OpenXR and changed the sign of all 16 values to match the sign of the game projection matrix. This for some reason is very broken on Quest3.

lvonasek:
I am closing this. Let's keep the VR integration simple.

PR #18545, opened 2023-12-14
Objects in function parameters have been replaced with constant references

GermanAizek:
@hrydgard do not be lazy to review PR changes, sometimes I accidentally forget to specify `&` for parameters.
I expect feedback, I understand that most fixes will not increase generic speed calling functions, but it is useful for refactoring code.


hrydgard (approved):
All this looks alright, though I don't understand what you mean regarding laziness.

Where possible (no need for a zero terminator) I'd like to move to `std::string_view` instead of `const std::string &`, but doesn't have to be in this PR.

GermanAizek:
Wow, how quickly did you check and accept commits.

PR #18548, opened 2023-12-14
Adrenotools followup

hrydgard:
This fixes most of the issues I commented on in #18532 . Most importantly, moves it to developer tools until we've gained more confidence in it - but it's looking good.

Remaining issues for future commits:

- [ ] Add PPSSPP doc here: https://github.com/K11MCH1/AdrenoToolsDrivers/tree/main/docs
- [ ] Fix driver bug "detection" to not detect bugs from the proprietary driver when using these
- [ ] Add Adreno drivers to PPSSPP Store? 
- [ ] Move back to Graphics settings?

### DIFF
--- a/Common/File/FileUtil.cpp
+++ b/Common/File/FileUtil.cpp
@@ -1209,12 +1209,11 @@ bool WriteStringToFile(bool text_file, const std::string &str, const Path &filen
 	return true;
 }
 
-bool WriteDataToFile(bool text_file, const void* data, const unsigned int size, const Path &filename) {
+bool WriteDataToFile(bool text_file, const void* data, size_t size, const Path &filename) {
 	FILE *f = File::OpenCFile(filename, text_file ? "w" : "wb");
 	if (!f)
 		return false;
-	size_t len = size;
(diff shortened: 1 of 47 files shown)


Crytonics:
PPSSPP doc are added in https://github.com/K11MCH1/AdrenoToolsDrivers

Crytonics:
> @Crytonics https://github.com/K11MCH1/AdrenoToolsDrivers/blob/main/docs/ppsspp.md still has the headline "Citra Usage" :)

Fixed

hrydgard:
I think I do want it to not be too easy to find. The reason is that there's a substantial amount of crash reports coming in from people loading bad adreno drivers.

Crytonics:
> I think I do want it to not be too easy to find. The reason is that there's a substantial amount of crash reports coming in from people loading bad (or not matching their device) adreno drivers.

Yea that is problem on all emulators, people just need to read desc for drivers. When drivers get released there is says what gpu can run that specific driver.

hrydgard:
people will never read the description :)

PR #18550, opened 2023-12-14
[Common Data Net/Core Debugger HLE/GPU Debugger] Fixed reduction data type size to strict 32-bit integer

GermanAizek:
@hrydgard It may be left over from old code, some functions accept size_t, reducing memsize type can in some situations ruin result algorithm.


hrydgard (approved):
None of this will change anything, but I do agree it's a slight improvement in clarity.

PR #18563, opened 2023-12-17
(UWP) Add Redump Asset File

ghost:
The `redump.csv` database was not included in the UWP package, so all the features that required this file were not working before. It is trivial enough to add the file however, so here we are.


hrydgard:
Thanks!

PR #18565, opened 2023-12-17
Add option to slow down mouse wheel to closer resemble button presses

### LINKED ISSUE #18564: Speed up the mouse wheel
### What should happen

 I noticed that the buttons are faster compared to the mouse wheel

### Who would this benefit

PPSSPP

### Platform (if relevant)

Windows

PPSSPP

_No response_

### Checklist

LunaMoo:
Should fix #18564
Default is unchanged, add's an option to delay it.

### DIFF
--- a/Core/Config.cpp
+++ b/Core/Config.cpp
@@ -838,6 +838,7 @@ static const ConfigSetting controlSettings[] = {
 	ConfigSetting("ConfineMap", &g_Config.bMouseConfine, false, CfgFlag::PER_GAME),
 	ConfigSetting("MouseSensitivity", &g_Config.fMouseSensitivity, 0.1f, CfgFlag::PER_GAME),
 	ConfigSetting("MouseSmoothing", &g_Config.fMouseSmoothing, 0.9f, CfgFlag::PER_GAME),
+	ConfigSetting("MouseWheelUpDelay", &g_Config.iMouseWheelUpDelay, 1, CfgFlag::PER_GAME),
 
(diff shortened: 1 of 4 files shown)


hrydgard:
Hm, this will only work on Windows. 

Maybe we can rework this to a more cross-platform solutions by doing some simple state tracking in NativeKey ?

LunaMoo:
Hah, sorry just followed wheelUp and forgot MainWindow is for windows, have to look how wheel is handled outside of platform specific code. Anyway the person requesting it might at least test if that's what he wants, because there's a bit of confusion there about mouse wheel being slow when it's imo more about the opposite.

hrydgard:
Yeah, good point about testing first.

LunaMoo:
Hmmm... not sure how to make it the way you said, I tried delaying it in NativeKey, but then if I just stop controlling the game it'll freeze in the state it was in since it stops updating.^_^; Anyway I'm going to family birthday soon, so will keep this on hold.

hrydgard:
You're gonna need to put one part of it in NativeFrame as well.

I can take care of it.

By the way, how meaningful do you think it is to have a setting? Maybe it would be alright to just do 0.1ms, most games should catch that. And if a second event comes in during that time, we'll just send a KEY_UP directly followed by a new KEY_DOWN.

hrydgard:
I made https://github.com/hrydgard/ppsspp/pull/18568 to replace this, have a look @LunaMoo .

hrydgard:
Because I've merged #18568, I'm closing this.

If a setting for the timer delay is still required, we can just copy it over from there.

PR #18568, opened 2023-12-17
Mouse wheel: Centralize the sending of KEY_UP events on a timer.

hrydgard:
We had platform-specific code for this, but it can simply be implemented directly in NativeApp.cpp, so let's do that.

This sets the delay to 0.1s, which should be alright to replace #18565 .

If you scroll the wheel quickly, events are not limited to 0.1s spacing, instead a KEY_UP will be sent just before the new KEY_DOWN.

This is good when mapping the mouse wheel to PSP buttons, they'll work better.

### DIFF
--- a/SDL/SDLMain.cpp
+++ b/SDL/SDLMain.cpp
@@ -708,31 +708,6 @@ static void EmuThreadJoin() {
 }
 
 struct InputStateTracker {
-	void TranslateMouseWheel() {
-		// SDL2 doesn't consider the mousewheel a button anymore
-		// so let's send the KEY_UP if it was moved after some frames
-		if (mouseWheelMovedUpFrames > 0) {
-			mouseWheelMovedUpFrames--;
-			if (mouseWheelMovedUpFrames == 0) {
-				KeyInput key;
-				key.deviceId = DEVICE_ID_MOUSE;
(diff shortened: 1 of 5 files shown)


LunaMoo:
Nice centralization, I couldn't make it that well.

 Not sure about behaviour change, it definitely deals with #18564 since it's basically the same max delay the user tested there, but without an option and defaults changing soo much could be bad for someone out there. I definitely can feel and even see the delay now it's not as snappy as it was, when someone had it mapped to something sensitive, it will work completely differently now possibly ruining some use case.

hrydgard:
I don't really think it'll be much of a problem because for games that you want to react fast, they usually react to keydown, not key up - so if the key release takes 100ms if you don't press it again, it doesn't matter. But, we can probably tune it down to 60ms or so quite safely. The idea is to cover a couple of frames at 30hz, so even games that really mishandle key input will not miss the events.

LunaMoo:
Yeah in general it might not, but just as with the use case where this change came from, it sometimes matters. For that user, it was preferable to have a keyboard-like keystroke, but there'll always be someone out there where a tick of the mouse wheel doing only one frame of a movement would be the only reason to even use the mouse wheel in the first place.

 But I guess it's fair to completely change behaviour of potentially rarely used feature due to complain from one person and then just wait if there will be any complains wanting the old behaviour back.

Issue #18561, opened 2023-12-16
DBZ Tenkaichi Tag Team: Textures disappear on modded maps


TagTeamEstudios:
### Game or games this happens in

ULUS10537 - Dragon Ball Z Tenkaichi Tag Team 

### What area of the game

Acontece nos

https://github.com/hrydgard/ppsspp/assets/106892701/40e12069-c249-497c-a6b4-1879a35a45ab

Anexarei um vídeo.

> Mapas que são editados por fãs, mapas mods. Os mapas originais não ocorrem esse bug apenas nos mapas que são mods. Observe o vídeo e verá os bugs nas texturas dos mapas.

Maps that are edited by fans, mod maps. This bug does not occur on original maps, only on maps that are mods. Watch the video and you will see the bugs in the map textures.

### What happens

>Bug na renderização das texturas dos mapas "Mods", do jogo Dragon Ball Z Tenkaichi Tag Team. Ao usar um mapa modificado no jogo, na gameplay as texturas dos mapas tendem a ficarem desaparecendo e reaparecendo conforme o personagem se movimenta na arena de batalha

Bug in the rendering of the textures of the "Mods" maps, from the game Dragon Ball Z Tenkaichi Tag Team. When using a modified map in the game, in gameplay the map textures tend to disappear and reappear as the character moves in the battle arena

![Screenshot_2023-12-16-16-57-56-059_org ppsspp ppsspp](https://github.com/hrydgard/ppsspp/assets/106892701/38117262-cc41-41a3-a033-94cb7c4f9504)
![Screenshot_2023-12-16-16-57-46-237_org ppsspp ppsspp](https://github.com/hrydgard/ppsspp/assets/106892701/f48fa274-e4e9-40a3-adba-6720f94140a7)
![Screenshot_2023-12-16-16-57-38-313_org ppsspp ppsspp](https://github.com/hrydgard/ppsspp/assets/106892701/50bb4e76-4d2a-4f61-a1c6-b8e506b9e4de)
![Screenshot_2023-12-16-16-57-30-251_org ppsspp ppsspp](https://github.com/hrydgard/ppsspp/assets/106892701/f3d58c29-0294-49bf-88e6-5989912edc57)

### What should happen

>Nas versões atuais do PPSSPP sempre acontece esse tipo de bug, porém, a versão do emulador 1.6.3 (A testada por mim), não acontece esse problema de texturas nos mapas mods, elas não desaparecem como nas versões atuais!

[...]

Issue #17061, opened 2023-03-06
Kidou Keisatsu Patlabor - Kamubakku MiniPato another screen problem


sum2012:
### Game or games this happens in

ULJS00026

### What area of the game

Game issue history: #16992
In-game

### What happens

![1](https://user-images.githubusercontent.com/2177532/223112576-695d2e15-0b05-42f8-83db-d694c06179a4.png)

### What should happen

This is softgpu
![2](https://user-images.githubusercontent.com/2177532/223112651-83ec16d5-7322-4b13-b186-925b59f50074.png)

### GE frame capture

[ULJS00026-BAD2.zip](https://github.com/hrydgard/ppsspp/files/10898230/ULJS00026-BAD2.zip)

### Platform

Windows

### Mobile phone model or graphics card

Nvidia Geforce 710

### PPSSPP version affected

v1.14.4-736-gebd8a6391

### Last working version

https://github.com/hrydgard/ppsspp/issues/16992#issuecomment-1445125401

### Graphics backend (3D API)

Vulkan

### Checklist

- [X] Test in the [latest git build](https://buildbot.orphis.net/ppsspp/) in case it's already fixed.
- [X] [Search for other reports](https://github.com/hrydgard/ppsspp/search?q=my+issue&type=issues) of the same issue.
- [X] Try resetting settings or older versions and include if the issue is related.
- [X] Try changing graphics settings to determine if one causes the glitch (especially speed hacks or enhancements/replacements.)
- [X] Include logs or screenshots of issue.

hrydgard:
This still happens. Checked the dump with renderdoc and the vertex data is full of NaNs after the vertex shader.

sum2012:
v1.16.6-333-gecab50346 still happens

hrydgard:
Tiny adjustment to the range Z culling range makes the GE dump work, so I think this is a precision issue where we our culling limits are just a smidge too tight. This single change is enough (on Vulkan): `maxValues[2] = z.second + 0.0001;` in ShaderUniforms.cpp.

hrydgard:
This can now be worked around with the new setting. Pushing forward to the next version, since any change to the culling will need a lot of testing.

sum2012:
Do I need open another issue for Nvidia gpu ? ( No color for police)
![1](https://github.com/hrydgard/ppsspp/assets/2177532/763e73f7-578c-4d3f-a3d7-3d20b731c6ad)
[...]

PR #18572, opened 2023-12-18
Add "Disable culling" as a speedhack for experimentation

hrydgard:
Range culling (which we'll just call culling in the setting name) still is problematic enough that it can be useful to have this way to turn it off as an escape route, until we really get it right. Since disabling it can break some games like famously TOCA, it needs to be under speedhacks, technically it can also be a small speedup though generally the performance difference is not detectable.

See for example #17061 or #18561 . In the first one, we are off by a small epsilon, that needs further testing to fix.

We will still try to avoid adding back a compat setting, since we know most of the details now.


JustAWhiteBloodCell (approved):

IrfanH495:
this fixes tenchu ​​san too
I've experienced it for a long time.
Finally, without creating issues, the problem can be resolved.
Thanks!
disable culling off
Screenshot_20231219-064651_PPSSPP

disable culling on
Screenshot_20231219-064704_PPSSPP

hrydgard:
Oh, I don't remember or didn't know that issue. That's another benefit of adding options like this - reports come in about more instances of the problem.

unknownbrackets:
I think any remaining range culling issues are down to rounding, fiddly nan/inf issues, or similar.  Overall, I'm pretty confident range culling widely happens.

I think that game might've been in the software renderer issues issue, and iirc it just barely ends up rounding Z slightly wrong.

-[Unknown]

hrydgard:
Yeah, I don't doubt that it happens consistently - as you say, it's more about the fiddly stuff.

#17061 for example needs a really slight amount of padding.

Kethen:
Interesting, was working on a fov patch for gran turismo and was scratching my head over some popping culling on the sides of the screen pretty close to the camera when fov is roughly above 50 (also applies to unmodified cockpit view, which has a fov of 52.xxxx, while follow/bumper has 33, roof has 35 so there wasn't much visible screen side popping for those)

No idea if that was intended by the game, but with this option enabled, it seems to have removed all of those popping culling

Kethen:
sceIoDevctl extension would indeed be handy for plugin development, tho currently the fov patches are done in cwcheat to leverage the built-in ui

eg. https://github.com/Kethen/RemasteredControls_GTpsp/blob/8b1178c30c65462c720ed7d08543dced8f0e5e3d/multiplayer_tweaks/UCES01245.ini#L19

PR #18575, opened 2023-12-18
Continue general code refactor for all project

GermanAizek:
@hrydgard I know that it is inconvenient for most to use std::move, std::swap, std::copy because developer should know that after std::move, variable value does not exist. So decide whether to use move semantics everywhere or only in critical performance places.

### DIFF
--- a/Common/Data/Encoding/Compression.cpp
+++ b/Common/Data/Encoding/Compression.cpp
@@ -53,7 +53,7 @@ bool compress_string(const std::string& str, std::string *dest, int compressionl
 		return false;
 	}
 
-	*dest = outstring;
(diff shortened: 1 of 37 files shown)


JustAWhiteBloodCell:
Just my $0.02: I think this should be restricted to only the hottest of code paths. Developers should know what happens after moving but it makes it significantly easier to make a mistake and if code is only executed once then the actual gain is very small.

hrydgard:
Adding std::move in a lot of places is a lot more risky than your other refactorings here. I'm not comfortable with such a large scale change doing that. Like @JustAWhiteBloodCell  says, std::move makes the most sense for very specific hotspots.

PR #18582, opened 2023-12-19
Optimize fbo_bind_fb_target and fbo_get_fb_target using constexpr template value C++17

GermanAizek:
@hrydgard,
I'm not sure of course, but maybe you're interested in rewriting some parameters of functions to create guaranteed C++17 metafunctions that are more optimized due to constant conditions and variables.
I doubt that the functions I fixed have a big impact on performance, but this is a step towards using such a method for functions that are frequently called.

More info here: https://stackoverflow.com/a/57211067
My tests on godbolt here: https://godbolt.org/z/4qq9n38Tz


hrydgard:
These are not even close to any kind of hot path (generally, creation functions aren't) so I think this just adds clutter. I like this kind of thing for actual hot paths though.

PR #18587, opened 2023-12-20
Minor lighting optimization in vshader

hrydgard:
Just a driveby optimization while looking for something else. Removes a max operation that's redundant due to the surrounding if statement.


hrydgard:
Ugh, this max operation isn't redundant at all, since ldot changes in between. Oops. Fix coming up.

PR #18616, opened 2023-12-28
[Workflow] Manually Generate UWP MSIX/MSIXBundle on-demand

anr2me:
Since we don't have UWP artifacts, here we go.

This will allow people to build UWP Package (Sideloading only) in `.msix` file when selecting a single platform, or `.msixbundle` file when selecting `Bundle` (contains all supported platforms).

PS: Symbol files only included on debug builds.

I also made minor changes to the APK and IPA workflows.


hrydgard:
Cool.

PR #18622, opened 2023-12-28
Updated German language

vajasoft-fressel:
Updated German language


hrydgard:
#18621  adds one more string :)

PR #18643, opened 2023-12-30
Update rcheevos submodule

hrydgard:


hrydgard:
The buildfix is dependent on an upstream change, so leaving this as draft.

Issue #15757, opened 2022-07-30
Metal Slug Anthology - texture issue at title screen


MojoJojoDojo:
### Game or games this happens in

ULUS-10154 - Metal Slug Anthology

### What area of the game

You can see at the title screen in the background.

### What happens

 there will be green lines around a certain area here :
![image](https://user-images.githubusercontent.com/33392696/181914155-9a774eec-30a5-4411-8781-cd5c21a59016.png)

Note: This will only happen when you upscale without texture upscaling (which is demanding to do).
Also, it won't appear when using software renderer unlike in [issue](https://github.com/hrydgard/ppsspp/issues/15755)
 

### What should happen

The lines should not appear.

### GE frame capture

[ULUS10154.ppdmp.zip](https://github.com/hrydgard/ppsspp/files/9226445/ULUS10154.ppdmp.zip)

### Platform

Linux / BSD

### Mobile phone model or graphics card

Radeon 530/Intel UHD 620

### PPSSPP version affected

1.13.1

### Graphics backend (3D API)

Vulkan

### Checklist

- [X] Test in the [latest git build](https://buildbot.orphis.net/ppsspp/) in case it's already fixed.
- [X] [Search for other reports](https://github.com/hrydgard/ppsspp/search?q=my+issue&type=issues) of the same issue.
- [X] Try resetting settings or older versions and include if the issue is related.
- [X] Try changing graphics settings to determine if one causes the glitch (especially speed hacks or enhancements/replacements.)
- [X] Include logs or screenshots of issue.

hrydgard:
This is likely the same issue as #15755 and many similar ones that have been reported over the years. Should make a meta-issue instead of a lot of specific ones...

PR #18646, opened 2023-12-30
Through-mode smart texture filtering

### LINKED ISSUE #18460: Auto-nearest-filter for 2D graphics with pixel mapping detection, in some games?
A lot of 2D UI in games, though certainly not all, is drawn like this:

* through mode (untransformed) RECT primitives
* 1:1 pixel mapping
* linear filter 

1:1 pixel mapping can, in theory, often be detected during rectangle-to-triangles expansion, for example if the data looks as simple as this we can just compare the UV and XY deltas and check if they match:

image

And we could just automatically turn on nearest filter for these draws. In many games, without texture replacement or upscaling enabled, this would look better. On the real PSP, the games often leave linear filtering on for no reason, because it doesn't make a difference when you map texels to pixels 1:1. But when we want to render 3D graphics at high resolution in PPSSPP, we end up with muddy UIs.

For example, this UI in Everybody's Golf:

image

Would look like this:

image

While the interior looks alright filtered, the borders are really destroyed by the undesired filtering.

Unfortunately we scale UVs a bit too early in the software transform pipeline that we use for RECT parameters to detect it trivially right now, but by moving the UV scaling in this case to the rectangle expansion part, we can still do cheap checks.

Anyway, just documenting an idea for some future hacking session :) It won't work with all games anyway.

hrydgard:
Adds a new option, "Smart 2D texture filtering" to switch to nearest (pixel) filtering mode if we detect "pixel-mapped" graphics in RECT draws.

Helps #18460 (the issue with the idea), though we might want a compat flag to argue that it fixes it.

Things that get fixed when enabled:

* Fixes the problem in #15757
* Weird color bleeding in dialog boxes in Final Fantasy gets fixed
* Text gets really sharp in Ridge Racer menus, but some filtering glitches remain
  image
  image

Things I hoped would be fixed but are NOT:

* Outrun text rendering and other glitches (it doesn't use through mode, plus draws with tristrips)
* #15755 (it doesn't use through mode, plus draws with tristrips)

Dubious results:

* Soul Calibur character select turns unfiltered, which doesn't really look much better, though subjective

Untested:

* https://github.com/hrydgard/ppsspp/issues/15449

~~Draft because there are some open questions:~~

* ~~Should this be a setting, or another texture filtering mode along with Auto etc? "Smart?". Combine with Auto max quality?~~ A new setting
* ~~To fix dialog frames in FFI, which are mistakenly filtered causing color fringes, we need to accept 1-dimensional stretching, which isn't true pixel mapping but due to the nature of the situation, it acts like it in this case.~~ (this seems ok, let's keep it)
* Would like to remove the gstate_c global, maybe

Additional fixes since original PR: 

- [x] Small 2D through-mode triangle draws now get checked. Helps multiple games, including some things the previous method missed in Ridge Racer menus

Possible future expansion:

* Detect this for multiple through-mode triangles.
* Detect if projection matrix/viewport params together set up something equivalent to through mode. Outrun has a good example of this, same with Wipeout Pure menus. Then we can use similar logic on top of that.


hrydgard:
Breaks BoF III due to DetectOffsetTexture not working correctly with unscaled coordinates.

LunaMoo:
Due to "dubious results" and the fact it's never soo black and white with people's preferences, it probably shouldn't replace any of the existing filtering types, but it would be fine either as an option or as a new type of filtering combined with auto max.

 Naming might be hard, but it could use tooltip for explanation or just use "smart" as you proposed, and be explained somewhere in the [docs](https://www.ppsspp.org/docs/category/reference).

hrydgard:
Yeah, probably the best. Just undecided if it should be separate, since it makes sense to combine it both with the old Auto and the newer Auto Max Quality..

Narugakuruga:
We can add a single tick box ☑️.
If this is an extra feature, it can be chosen when auto and auto max quality is selected.
And the tick box is supposed to hide when linear and nearest is selected, because it's "forced texture filter"

Narugakuruga:
Also if this proves to be useful for some 3D games, then auto max quality will benefit from both mipmap generation and smart filtering.

hrydgard:
This no longer breaks BOFIII (after #18653 ). Went ahead and added a new setting to play with.

It's quite nice now in Ridge Racer menus, and many other games too, if you prefer a sharper pixel-perfect look over sometimes-ugly bilinear.

We are not yet able to detect this in situations where games don't use through-mode for their UI, unfortunately.

hrydgard:
I feel this is useful enough already that we can include it in 1.17, though plan to improve it further in 1.18.

PR #18647, opened 2023-12-30
Enforce run-behind-pause-screen if ad-hoc multiplayer active

hrydgard:
This forces the game to keep running behind the menu if adhoc is connected, avoiding desyncs-by-mistake.

@anr2me What do you think? Is this the right approach?

### DIFF
--- a/Core/HLE/sceNetAdhoc.cpp
+++ b/Core/HLE/sceNetAdhoc.cpp
@@ -127,7 +127,7 @@ static int sceNetAdhocPdpSend(int id, const char* mac, u32 port, void* data, int
 static int sceNetAdhocPdpRecv(int id, void* addr, void* port, void* buf, void* dataLength, u32 timeout, int flag);
 
 bool __NetAdhocConnected() {
-	return netAdhocInited && netAdhocctlInited && adhocctlState == ADHOCCTL_STATE_CONNECTED;
+	return netAdhocInited && netAdhocctlInited && (adhocctlState == ADHOCCTL_STATE_CONNECTED || adhocctlState == ADHOCCTL_STATE_GAMEMODE);
 }
 
 void __NetAdhocShutdown() {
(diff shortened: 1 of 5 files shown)


UI/PauseScreen.cpp
```diff
@@ -447,7 +447,7 @@ void GamePauseScreen::CreateViews() {
 		playButton_->SetImageID(g_Config.bRunBehindPauseMenu ? ImageID("I_PAUSE") : ImageID("I_PLAY"));
 		return UI::EVENT_DONE;
 	});
-	rightColumnHolder->Add(new Spacer(10.0f));
+	//rightColumnHolder->Add(new Spacer(10.0f));
```
unknownbrackets, line 450: Hm, I feel like it's better to keep the space?  But otherwise should just delete this line, right?

-[Unknown]

Core/HLE/sceNetAdhoc.cpp
```diff
@@ -126,6 +126,9 @@ static int sceNetAdhocPdpCreate(const char* mac, int port, int bufferSize, u32 f
 static int sceNetAdhocPdpSend(int id, const char* mac, u32 port, void* data, int len, int timeout, int flag);
 static int sceNetAdhocPdpRecv(int id, void* addr, void* port, void* buf, void* dataLength, u32 timeout, int flag);
 
+bool __NetAdhocConnected() {
+	return netAdhocInited && netAdhocctlInited && adhocctlState == ADHOCCTL_STATE_CONNECTED;
```
anr2me, line 130: it should be `(adhocctlState == ADHOCCTL_STATE_CONNECTED) || (adhocctlState == ADHOCCTL_STATE_GAMEMODE)` as gamemode is also in connected state, similar to https://github.com/hrydgard/ppsspp/blob/d2ebd36ab963bdf21f8fae50c6cc5eec21cc2426/Core/HLE/sceNetAdhoc.cpp#L2285-L2287

hrydgard: Fixed it up, please test :)

UI/PauseScreen.cpp
hrydgard: Reverted.

Core/HLE/sceNetAdhoc.cpp
anr2me: I only tested this on 2 games, Naruto Shippuden Ultimate Ninja Heroes 3 (normal adhoc), and Bomberman (gamemode), seems to be okay.

hrydgard: Thanks for testing! I think it should be generally fine, I guess we'll get bug reports if it isn't.

Issue #6379, opened 2014-06-20
Tokimeki Memorial 4 graphic issue.


daniel229:
Look at the arms.

buffered rendering
![01](https://cloud.githubusercontent.com/assets/3481559/3339086/b19a2cc6-f865-11e3-8b9a-259fb1c897c4.png)

read framebuffers to memory fixes the problem,but have other problem.
![02](https://cloud.githubusercontent.com/assets/3481559/3339087/b6067206-f865-11e3-86d6-28988ef2c8e4.png)

log

```
GLES\Framebuffer.cpp:973 Creating FBO for 0015e000 : 256 x 272 x 3
GLES\Framebuffer.cpp:1016 FBO reusing depthbuffer, 0015e000/00110000 and 00000000/00110000
GLES\ShaderManager.cpp:139 Linked shader: vs 27 fs 34
GLES\ShaderManager.cpp:139 Linked shader: vs 36 fs 37
GLES\ShaderManager.cpp:139 Linked shader: vs 39 fs 37
GLES\Framebuffer.cpp:973 Creating FBO for 001a2000 : 256 x 240 x 3
GLES\Framebuffer.cpp:1016 FBO reusing depthbuffer, 001a2000/00110000 and 00000000/00110000
GLES\ShaderManager.cpp:139 Linked shader: vs 22 fs 37
GLES\Framebuffer.cpp:973 Creating FBO for 001de000 : 128 x 136 x 3
GLES\Framebuffer.cpp:1016 FBO reusing depthbuffer, 001de000/00110000 and 00000000/00110000
GLES\ShaderManager.cpp:139 Linked shader: vs 27 fs 42
GLES\Framebuffer.cpp:973 Creating FBO for 001f6000 : 32 x 34 x 3
GLES\Framebuffer.cpp:1016 FBO reusing depthbuffer, 001f6000/00110000 and 00000000/00110000
GLES\Framebuffer.cpp:973 Creating FBO for 001f0000 : 64 x 68 x 3
GLES\Framebuffer.cpp:1016 FBO reusing depthbuffer, 001f0000/00110000 and 00000000/00110000
GLES\ShaderManager.cpp:139 Linked shader: vs 27 fs 44
GLES\ShaderManager.cpp:139 Linked shader: vs 22 fs 46
GLES\Framebuffer.cpp:973 Creating FBO for 00154000 : 64 x 32 x 3
[...]

PR #18653, opened 2023-12-30
Use the Tokimeki / old Juiced 2 method for Breath of Fire III

hrydgard:
See #18650 / #6379

Merges the two framebuffers into one, instead of complex logic to pick the right texture from the coordinates.

Allows us to delete a piece of complex code from the renderers, but for now at least, requires us to enable the compat flag.

I don't think there are any other known games that rely on this?


hrydgard:
@unknownbrackets summoning for opinion since I'm deleting your old workaround here :) It's always a bit iffy to go from general hack to compat.ini option, but I still would argue that it's cleaner this way.

hrydgard:
I'm going to get this in, it'll make some things I want to do like #18646 easier.

Hopefully we can make the framebuffer-merge heuristic safe enough in the future to enable without a flag - the only known breakage is that it slows down Juiced 2 drastically since it does self-texturing within the merged range, which isn't an issue in BOF.

PR #18669, opened 2024-01-06
GPS: updates

Florin9doi:
- Simulate fresh updates to satisfy MAPLUS 1/2 when real GPS data isn't available
- Remove NmeaListener workaround for Android 29 SDK
- Add constants for GPGGA tokens

### DIFF
--- a/android/src/org/ppsspp/ppsspp/LocationHelper.java
+++ b/android/src/org/ppsspp/ppsspp/LocationHelper.java
@@ -17,6 +17,9 @@ import java.util.Iterator;
 
 class LocationHelper implements LocationListener {
 	private static final String TAG = LocationHelper.class.getSimpleName();
+	private static final int GPGGA_ID_INDEX = 0;
+	private static final int GPGGA_HDOP_INDEX = 8;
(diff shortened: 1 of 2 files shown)


Florin9doi:
MAPLUS apps also need GPS location from within Japan, but this is likely expected.

android/src/org/ppsspp/ppsspp/LocationHelper.java
```diff
@@ -74,17 +73,7 @@ public void onNmeaReceived(long timestamp, String nmea) {
 							onNmea(nmea);
 						}
 					};
-
-					// Use reflection to work around a bug in the Android 29 SDK.
-					// https://stackoverflow.com/questions/57975969/accessing-nmea-on-android-api-level-24-when-compiled-for-target-api-level-29
-					try {
```
hrydgard, line 80: Hm, how come we can remove this?

hrydgard: Oh wait, never mind, apparently it was indeed just the SDK that was buggy, not Android. it's ok.

PR #18670, opened 2024-01-07
Fix compilation against newer ffmpeg versions

Tatsh:
- `AVCodec*` -> `const AVCodec*`
- `#include "libavcodec/avcodec.h"` where required
- `stream->need_parsing` is moved to an internal header and should no longer be set by outside code
- `stream->codec` is replaced with call to `avcodec_find_decoder`, etc

### DIFF
--- a/Core/HW/MediaEngine.cpp
+++ b/Core/HW/MediaEngine.cpp
@@ -42,6 +42,11 @@ extern "C" {
 #endif // USE_FFMPEG
 
 #ifdef USE_FFMPEG
+
+#if LIBAVFORMAT_VERSION_INT >= AV_VERSION_INT(59, 16, 100)
+#define AVCodec const AVCodec
+#endif
+
 static AVPixelFormat getSwsFormat(int pspFormat)
 {
 	switch (pspFormat)
@@ -394,7 +399,7 @@ bool MediaEngine::addVideoStream(int streamNum, int streamId) {
 		// no need to add an existing stream.
 		if ((u32)streamNum < m_pFormatCtx->nb_streams)
 			return true;
-		const AVCodec *h264_codec = avcodec_find_decoder(AV_CODEC_ID_H264);
+		AVCodec *h264_codec = avcodec_find_decoder(AV_CODEC_ID_H264);
 		if (!h264_codec)
 			return false;
 		AVStream *stream = avformat_new_stream(m_pFormatCtx, h264_codec);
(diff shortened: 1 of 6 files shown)


hrydgard:
It's a bad idea to use system ffmpeg for PPSSPP in general. We intentionally ship a single fixed version that contains everything PPSSPP ever need in order to avoid this kind of work, and also things have changed in ffmpeg over time that has reduced our compatibility with PSP games if we use modern versions.

But, if you're really insistent on using it, you should fix the guards like:

#if LIBAVFORMAT_VERSION_INT >= AV_VERSION_INT(57, 33, 100)

to match when the API changes were made. Otherwise you'll break building on some intermediate versions, most likely.

Tatsh:
Right now this compiles but it does not work correctly. For Wipeout Pulse if you don't press X before the first video ends, the emulator seems to hang.

image

hrydgard:
Which is another reason we don't really support using system ffmpeg. Hopelessly many combinations to test.

Tatsh:
I'll add the version checks soon so it doesn't affect the builds.

Core/HW/MediaEngine.cpp
```diff
@@ -499,7 +510,7 @@ bool MediaEngine::setVideoStream(int streamNum, bool force) {
 
 		AVStream *stream = m_pFormatCtx->streams[streamNum];
 #if LIBAVFORMAT_VERSION_INT >= AV_VERSION_INT(57, 33, 100)
-		AVCodec *pCodec = avcodec_find_decoder(stream->codecpar->codec_id);
+		const AVCodec *pCodec = avcodec_find_decoder(stream->codecpar->codec_id);
```
iota97, line 513: When `#define AVCodec const AVCodec` get expanded how does this even compile? Is double const legal in some compiler? I'm quite confused here tbh...

hrydgard: Yeah I was saying they should match, which means adjust the values to what they should be. And I don't feel like being the one to test that they're right...

Double/triple/etc const is only legal if the consts are about different parts of the type, like a `const char *const my_var;` and means both that the pointer and the data pointed to is const. Not sure if that applies in this case, didn't look very closely.

Tatsh: This should be fixed in the latest push.

Tatsh: As for the version, `56.58.58` is ffmpeg 4.4.4, `57.59.59` is 5.1.4 and `58.60.60` is 6.0 (and remains with 6.1.1). I don't know which one should be used though I would guess 57.59.59 would be better? It's the last version of ffmpeg 5.x. https://ffmpeg.org/releases/?C=N;O=D

Tatsh: I determined the correct version of libavformat is `59.16.100`. This is ffmpeg 5.0, when functions accepting `AVCodec*` added `const`.

Tatsh:
@hrydgard please review again.

Haimiano67 (approved):

hrydgard (approved):

PR #18676, opened 2024-01-10
Made up symbols names -> the real ones

Nemoumbra:
The game Aces of War `ULES00590` turned out to have debug symbols and also not to be relocated (`e_type == 0x2`).
This was a perfect opportunity to make a copy of the executable, zero out the string `.symtab` in `.shstrtab` to confuse the emulator and manually match the symbol names with what we used to have until now.

I actually wanted to add `dl_write_light_color`, `vector_add_t`, `vector_sub_t`, `vmidt_q` and `vtfm_t` too, but I didn't, since they are mentioned a few times in the files with different hashes. Maybe that's not a big deal, but I've decided to do it slowly and carefully.

image


hrydgard:
These function names aren't really any more secret than the symbols the homebrew sdk has, so this should be ok.

PR #18678, opened 2024-01-10
Vulkan: Fix trying to compare uninitialized parts of packed descriptors

hrydgard:
Found by Valgrind.

Other minor cleanup, too.

This shouldn't have caused anything worse than slightly worse re-use of descriptors, fortunately.


hrydgard:
Hm, just realized that this might go badly in 32-bit builds. Will fix

PR #18682, opened 2024-01-12
More string_view optimizations

hrydgard:
Just some internal code cleanup I've wanted to do for some time, let's get it done. Will enable some other minor optimizations.

### DIFF
--- a/Common/Data/Encoding/Utf8.cpp
+++ b/Common/Data/Encoding/Utf8.cpp
@@ -383,7 +383,7 @@ bool AnyEmojiInString(std::string_view str, size_t byteCount) {
 	return false;
 }
 
-int UTF8StringNonASCIICount(const char *utf8string) {
+int UTF8StringNonASCIICount(std::string_view utf8string) {
 	UTF8 utf(utf8string);
 	int count = 0;
 	while (!utf.end()) {
@@ -394,7 +394,7 @@ int UTF8StringNonASCIICount(const char *utf8string) {
 	return count;
 }
 
-bool UTF8StringHasNonASCII(const char *utf8string) {
+bool UTF8StringHasNonASCII(std::string_view utf8string) {
 	return UTF8StringNonASCIICount(utf8string) > 0;
 }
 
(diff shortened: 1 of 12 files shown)


Common/Data/Encoding/Utf8.cpp
```diff
@@ -628,12 +570,13 @@ static size_t ConvertUTF8ToWStringInternal(wchar_t *dest, size_t destSize, const
...
+std::wstring ConvertUTF8ToWString(std::string_view source) {
 	std::wstring dst;
-	// utf-8 won't be less bytes than there are characters.  But need +1 for terminator.
-	dst.resize(source.size() + 1, 0);
-	size_t realLen = ConvertUTF8ToWStringInternal(&dst[0], source.size() + 1, source);
+	// utf-8 won't be less bytes than there are characters.
+	dst.resize(source.size(), 0);
+	size_t realLen = ConvertUTF8ToWStringInternal(&dst[0], source.size(), source);
 	dst.resize(realLen);
+	dst[realLen] = 0;
```
unknownbrackets, line 579: Something doesn't look right here.  You're setting `[realLen]` to 0, but you've intentionally not added one for the terminator.  In practice, this may be fine - I think std::wstring will always have a null there, but it seems unsafe?

Suppose the string were just ASCII codepoints, let's say 5 letters: hello.  `dst.resize(5);` and `realLen` is returned as 5 (no terminator written because it hit destSize), but then you'll write to `dst[5]`.

I guess this write isn't needed maybe?  But not sure if any callers somehow depend on it, as it was there before.

-[Unknown]

Common/Data/Encoding/Utf8.cpp
```diff
@@ -206,35 +206,22 @@ int u8_charnum(const char *s, int offset)
...
-uint32_t u8_nextchar(const char *s, int *index) {
+uint32_t u8_nextchar(const char *s, int *index, size_t size) {
 	uint32_t ch = 0;
 	int sz = 0;
 	int i = *index;
 	do {
 		ch = (ch << 6) + (unsigned char)s[i++];
 		sz++;
-	} while (s[i] && ((s[i]) & 0xC0) == 0x80);
+	} while (i < size && ((s[i]) & 0xC0) == 0x80);
```
unknownbrackets, line 217: A string may have nulls in it which should terminate, even if they are earlier than the expected size?  I think we've even seen this with savedata.

-[Unknown]

hrydgard: Hm, yes, you're right, this is a concern, I will add it back.

If a `string_view` is created from a `const char *`, it will already be truncated at the first null. But if you create it from an std::string that contains nulls, or explicitly with pointer range, it indeed can also contain nulls.

Common/Data/Encoding/Utf8.cpp
hrydgard: Uh yeah, I screwed this one up it seems. After resizing a string, there is indeed no need to write the terminator and I believe that doing so is undefined though will work in practice (but may be caught by debugging tools). Will fix.

PR #18683, opened 2024-01-12
Handle file type detection of extracted ISO directories better

hrydgard:
Fixes checking the size of these on the game info screen. Reported by Nemoumbra.


UI/GameInfoCache.cpp
```diff
@@ -576,7 +578,6 @@ class GameInfoWorkItem : public Task {
 
 		case IdentifiedFileType::PSP_DISC_DIRECTORY:
 			{
-				info_->fileType = IdentifiedFileType::PSP_ISO;
```
hrydgard, line 579: I'm not sure what the purpose of this was, but it doesn't seem to do anything other than breaking the checks when trying to figure out the size..

PR #18684, opened 2024-01-12
sceSircs/Infrared support on Android

Florin9doi:
Test program : https://gist.github.com/Florin9doi/c4ccbe7fa2e1a62762b75c61b914f71f

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2160,6 +2160,8 @@ add_library(${CoreLibName} ${CoreLinkType}
 	Core/HLE/sceSfmt19937.h
 	Core/HLE/sceSha256.cpp
 	Core/HLE/sceSha256.h
+	Core/HLE/sceSircs.cpp
+	Core/HLE/sceSircs.h
 	Core/HLE/sceSsl.cpp
 	Core/HLE/sceSsl.h
 	Core/HLE/sceUmd.cpp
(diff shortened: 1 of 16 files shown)


hrydgard:
That's really cool! Although extremely niche of course, but I'm happy to merge this.

Not worth implementing the init/deinit functions? Might maximize compatibility to return 0 from init, at least.

What else other than your test program uses it? :)

Florin9doi:
>What else other than your test program uses it? :)

Only similar homebrew programs. Afaik, it was never used by any official program😀

hrydgard:
To fix the build, you'll have to add the new file to the UWP vcprojs and also libretro/common.makefile.

Let me know if I should take care of it instead.

hrydgard:
OK, I think we can just skip them then.

anr2me:
> > Not worth implementing the init/deinit functions?
> 
> I don't know if they receive any parameter, they are not documented

You can try decompiling the official prx file using [pspdecompiler](https://github.com/anr2me/pspdecompiler/releases) to see whether the init syscall took any argument or not.

Nemoumbra:
@Florin9doi
I'm personally shocked to know there's an IR port in PSPs (I don't think my PSP Go possesses one). So you're saying this library is official, but no games have used it? Then how do we know of its existence? Where does said `sircs.prx` come from? Which SDK was it introduced in? So many mysteries...

Florin9doi:
Only PSP1000 (fat) have infrared, this library was known since 2005 and is still present in latest firmware.

PR #18697, opened 2024-01-15
Address feedback on Utf8.cpp changes

hrydgard:
Address some feedback from #18682, thanks [Unknown].


fp64:
Um... `((s[i]) & 0xC0) == 0x80` implies `s[i]!=0` so this feels redundant.
You can argue that it is clearer...

hrydgard:
Hm, that's a good point heh.

PR #18706, opened 2024-01-15
Remove a bunch of dead code from CoreTiming ("threadsafe" events)

hrydgard:
We haven't used these "threadsafe" events since we removed our first attempt at GPU threading, so like for 10 years, and maybe some experimentation in the networking code according to some comments. 

Regarding savestates, it's unlikely that any savestates that had some of these events stored would load anyway. By just storing a single byte empty list marker in savestates, and rejecting save states that do have events in this list, we avoid even bumping the savestate version here.

@unknownbrackets , agree? Might also keep enough code to load the events and throw them into the regular queue.

### DIFF
--- a/Core/CoreTiming.cpp
+++ b/Core/CoreTiming.cpp
@@ -63,15 +63,7 @@ struct BaseEvent {
 typedef LinkedListItem<BaseEvent> Event;
 
 Event *first;
-Event *tsFirst;
-Event *tsLast;
-
-// event pools
 Event *eventPool = 0;
-Event *eventTsPool = 0;
-int allocatedTsEvents = 0;
-// Optimization to skip MoveEvents when possible.
-std::atomic<u32> hasTsEvents;
 
 // Downcount has been moved to currentMIPS, to save a couple of clocks in every ARM JIT block
 // as we can already reach that structure through a register.
@@ -81,8 +73,6 @@ alignas(16) s64 globalTimer;
 s64 idledCycles;
 s64 lastGlobalTimeTicks;
 s64 lastGlobalTimeUs;
-
-static std::mutex externalEventLock;
 
 std::vector<MHzChangeCallback> mhzChangeCallbacks;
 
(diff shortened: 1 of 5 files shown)


Core/MIPS/JitCommon/JitBlockCache.cpp
```diff
@@ -239,7 +239,6 @@ void JitBlockCache::RemoveBlockMap(int block_num) {
 		}
 	}
 }
-
```
unknownbrackets, line 242: Unintentional?

-[Unknown]

Core/CoreTiming.cpp
```diff
@@ -719,10 +522,10 @@ void DoState(PointerWrap &p) {
 
 	if (s >= 3) {
 		DoLinkedList<BaseEvent, GetNewEvent, FreeEvent, Event_DoState>(p, first, (Event **) NULL);
-		DoLinkedList<BaseEvent, GetNewTsEvent, FreeTsEvent, Event_DoState>(p, tsFirst, &tsLast);
+		DoIgnoreUnusedLinkedList(p);
 	} else {
 		DoLinkedList<BaseEvent, GetNewEvent, FreeEvent, Event_DoStateOld>(p, first, (Event **) NULL);
-		DoLinkedList<BaseEvent, GetNewTsEvent, FreeTsEvent, Event_DoStateOld>(p, tsFirst, &tsLast);
+		DoIgnoreUnusedLinkedList(p);
```
unknownbrackets, line 528: I think a comment that this is for BC would be good.  Someone will come along and "optimize" this code to remove the do-nothing call otherwise.

-[Unknown]

unknownbrackets (approved):
The GE interrupts that were previously using this would probably have no pending events when save states are created (normally only at the start of a new frame.)

Pretty sure these pre-existed the GE thread stuff and maybe even came from Dolphin, so back when I just save stated everything indiscriminately so we could have stable save states (better to get it right first, without doing anything obviously slow, then optimize second.)  But now I think they would've been moved by this time anyway.

-[Unknown]

Core/CoreTiming.cpp
hrydgard: Good point, adding.

PR #18714, opened 2024-01-17
OpenXR - Shadow of Destiny VR camera fixed

lvonasek:
The same problem like in Tony Hawk. For some reason the method for detecting camera rotation mirroring (which works in 99% of games) doesn't work in this game.


hrydgard:
Tony Hawk does a lot of funky stuff with its camera, including runtime-patching its internal frustum culling code to account for the camera direction... (!)

PR #18718, opened 2024-01-17
Android: Fix running plain PRXs on scoped storage. Minor cleanup.

hrydgard:
Not sure when this regressed, but with this it's working again.


hrydgard:
Seems it broke the windows unit tests somehow? I'll narrow the fix for now..

PR #18730, opened 2024-01-19
Settings banner for game specific settings

hrydgard:
This shows a banner if you're editing game-specific settings:

image

Unfortunately, we don't have a way to display exactly which settings are game-specific, but this might reduce confusion a bit?

The banner also has room for a Help button that leads to for example https://www.ppsspp.org/docs/settings/graphics . 

However, I don't 100% like this, it's a bit bulky - and I'm also not really sure what to display there if editing global settings.

Thoughts?

### DIFF
--- a/Tools/langtool/src/section.rs
+++ b/Tools/langtool/src/section.rs
@@ -33,7 +33,7 @@ impl Section {
         }
     }
 
-    pub fn get_line(&mut self, key: &str) -> Option<String> {
+    pub fn get_line(&self, key: &str) -> Option<String> {
         for line in self.lines.iter() {
             let prefix = if let Some(pos) = line.find(" =") {
                 &line[0..pos]
--- a/UI/DevScreens.cpp
+++ b/UI/DevScreens.cpp
@@ -1020,7 +1020,7 @@ bool ShaderViewScreen::key(const KeyInput &ki) {
 }
 
 
-const std::string framedumpsBaseUrl = "http://framedump.ppsspp.org/repro/";
+const std::string_view framedumpsBaseUrl("http://framedump.ppsspp.org/repro/");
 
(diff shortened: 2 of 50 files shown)


Domiiniik:
For the global settings, maybe don't give it the name at all? And display only if you edit game settings, that'd be also good for separating it visually

anr2me:
May be using a different font color if it use a game-specific settings?

jcchikikomori:
I understood that this is kind of bulky, as i already imagine what will look like on smaller-smallest screens.

Maybe we can put the icon as a indicator on each sections, e.g.: `⚙️Rendering mode`.
Or maybe we can put the Game ID after each section, e.g.: `Rendering mode - for UCES00304`.

> For the global settings, maybe don't give it the name at all? And display only if you edit game settings, that'd be also good for separating it visually

For this one, i also agree, we don't need to display (except Help maybe?) for Global Settings.

NABN00B:
In case you're still collecting opinions on this (and the conversation at #19387 ), here's what I think:
- for every setting whose current value is different from that of global settings: concatenate something simple to the start of the string, such as an asterisk `* ` character (the gears icon is way too heavy for this)
- displaying the game name/id here might be unnecessary -- you either get here from pause menu while game is running or from game info screen which displays the games' backgrounds (which is something even most homebrew games should have)
- the Help button would look good slightly above the Back button -- although mobiles don't have much vertical space left here, so maybe not?

~~Instead of displaying a separate Help button we could also try displaying the info icon from uiatlas appended at the end of the currently selected tab, and when the tab is pressed again have the user directed to the webpage.~~ This would look scuffed though with translations like German.

hrydgard:
Closing in favor of #20976

Issue #18694, opened 2024-01-15
CwCheats loading error after a while


hrydgard:
This is a strange one, reported by DJ on Discord.

> After I've been playing for a while on my handheld device (Android), trying to go into the menu to toggle a cheat on leads me to a blank menu and I have to restart the app. Is that just a consequence of something, or might it be a bug?

> Playstore PPSSPP Golden on a [Retroid Pocket 3+](https://www.goretroid.com/products/retroid-pocket-3-handheld-retro-gaming-system-1) (Android 11). I tend to notice it after a handful of hours, but I've also gotten into the habit of saving in-game and then restarting the app when I know I need to toggle a cheat. It's the cheat menu itself that is blank. Let me see if I have screencaps. 

![image](https://github.com/hrydgard/ppsspp/assets/130929/3b8f0123-e42f-4f94-9595-33d94dde9945)
![image](https://github.com/hrydgard/ppsspp/assets/130929/25cfffa7-f670-480c-81f4-72b0a070facb)
![image](https://github.com/hrydgard/ppsspp/assets/130929/6529f55f-2506-4eef-b3ea-fd1068539d5b)
![image](https://github.com/hrydgard/ppsspp/assets/130929/50823747-b66a-49b5-a82d-62dc2c9b945f)

>  In this particular instance, I'm using HD texture packs but this was happening even before I found those packs to use in the first place. The game itself is Fate/Extra (ULUS10576) and then its sequel did the same. 

> I have a personal cheats file with the ones I personally think to use, and then the cheat.db to load from if I need one from the community list.

> I always have to hit "import" to load the cheat.db

> This is a copy of it. When I'm not experiencing this bug, it's just there automatically since I put it in PPSSPP\PSP\Cheats.

[ULUS10576.ini.zip](https://github.com/hrydgard/ppsspp/files/13936311/ULUS10576.ini.zip)

biomagnet:
Seems to happen when I open the app panel (to swap to something else), or when I lock the device and reopen it. Can confirm the same thing happens on my Samsung Fold 5.

hrydgard:
It's quite likely that #18736 fixed this - can you test?

biomagnet:
I absolutely definitely forgot to say, but this seemed fixed as far as my Fold 5 goes. Haven't been able to install the dev build on my other device yet, but for now seems to work just fine.

PR #18733, opened 2024-01-20
CwCheatScreen::LoadCheatInfo wait for GameInfo to be ready

Kethen:
Fixes a type of https://github.com/hrydgard/ppsspp/issues/18694 caused by onResume()

```
01-20 14:44:59.127  3094  4393 I MediaProvider: Open with lower FS for /storage/emulated/0/PPSSPP/PSP/Cheats/.ini. Uid: 10085
```

It may or may not be exactly https://github.com/hrydgard/ppsspp/issues/18694 however since the original report noted that a full app restarted is needed, but on my end backing out of the `CwCheatScreen` and going back in was enough to list cheats again on android 13, and this PR only makes it so onResume() does not break it in the first place

### DIFF
--- a/UI/CwCheatScreen.cpp
+++ b/UI/CwCheatScreen.cpp
@@ -33,6 +33,8 @@
 #include "UI/GameInfoCache.h"
 #include "UI/CwCheatScreen.h"
 
+#include <thread>
+
 static const int FILE_CHECK_FRAME_INTERVAL = 53;
 
 static Path GetGlobalCheatFilePath() {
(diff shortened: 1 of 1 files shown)


hrydgard:
Thanks, this is a very good find!

Though, I am going to rework it so we don't block the UI thread, by polling in Screen::update() instead.

Kethen:
@hrydgard Got it, in that case feel free to close this PR and replace it with the new rework :)

hrydgard:
Alright, closing in favor of #18736.

PR #18753, opened 2024-01-23
Update to the brazilian portuguese translation

Felipefpl:
Additions of the latest strings were made.


assets/lang/pt_BR.ini
```diff
@@ -199,7 +199,7 @@ Cheats = Trapaças
 Edit Cheat File = Editar arquivo de trapaças
 Import Cheats = Importar do cheat.db
 Import from %s = Importar do %s
-Refresh interval = Taxa de atualização
+Refresh Rate = Taxa de atualização
```
hrydgard, line 202: don't change the key please (left side of = )

hrydgard: I think it was a conflict resolution..

PR #18759, opened 2024-01-24
Update Russian translation

TotalCaesar659:
P.S. This [string](https://github.com/hrydgard/ppsspp/blob/408d3aa1addc1b6bc115fe7ab0a4233a0fb8b56b/assets/lang/ru_RU.ini#L731) doesn't fit the button, can its font be scaled as in settings?

One language is shown as "???" and every its string does too, don't know which one is it.


hrydgard:
A language showing as ??? means there's no font on your system to display that language.

I'll see what I can do about the overly long string.

TotalCaesar659:
Thanks for the confirmation, I supposed similar thing but decided to clarify.

PR #18772, opened 2024-01-28
Add volume slider for RetroAchievements sound effects

### LINKED ISSUE #18770: Achievement sounds need a volume control
Requested on Discord, and I think it's needed indeed. Some games are a lot quieter than the default achievement sound effects.

hrydgard:
Fixes #18770

Our volume sliders kinda suck actually, not enough steps and linear instead of exponential. That's for later though.


NABN00B:
does this not affect other UI sounds?

hrydgard:
It does not. Though, we could make it do that, or add another slider. Not sure what's best or motivated.

PR #18781, opened 2024-01-29
Disable 16-bit textures on PowerVR with Vulkan

### LINKED ISSUE #18780: Some objects render green on certain PowerVR devices
image

~~Doesn't happen on my PowerVR devices (Galaxy A21, Redmi 9A).~~

Does happen on Galaxy A21, just had to play Tekken: Dark Resurrection instead of Tekken 6.

Originally reported on [Infinix hot 30 play](https://www.gsmarena.com/infinix_hot_30_play_nfc-12289.php) .. 

Reported by CSxDM on Discord.

The weird thing is, this doesn't seem to be a straight R/B swap or similar..

Also, this doesn't happen in 1.16, so a bisection would be helpful, if we can just find someone with a device it happens on to do it...

hrydgard:
We had a bug before that prevented them from being enabled, but now that they are enabled properly, it seems there's some wacky driver bug.

So disable them again, specifically on PowerVR.

Fixes #18780


Auwal-Ac:
My account number is 0005680086 jaiz bank

PR #18786, opened 2024-01-29
Release all keys on pause.

### LINKED ISSUE #18767: When bringing up the pause menu (with run-behind), inputs may stay pressed
Reported by HeyImFranco on Discord.

https://github.com/hrydgard/ppsspp/assets/130929/1c6251e7-f2da-40ac-8550-c71864c8ee40

hrydgard:
Fixes #18767

Not sure about merging for the 1.17 series though, maybe someone really likes to bring up the pause menu briefly while holding a direction.. we could actually handle that with some extra logic.

### DIFF
--- a/Core/ControlMapper.cpp
+++ b/Core/ControlMapper.cpp
@@ -231,6 +231,42 @@ void ControlMapper::ForceReleaseVKey(int vkey) {
 	KeyMap::UnlockMappings();
 }
 
+void ControlMapper::ReleaseAll() {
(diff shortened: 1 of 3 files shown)


LunaMoo:
I recall using keys lock on pause as feature to create some cheats easier in the past(to ensure required inputs are ON while searching) or using it to lock a key that I otherwise had to press all the time, but that's just me being lazy, same things can be done with external makros or just holding the buttons.

PR #18791, opened 2024-01-29
Refresh iOS/macOS versions

Florin9doi:
- Fix CFBundleExecutable for QT version
- Fix CFBundleVersion by searching git-version in BINARY_DIR
- Fix TARGETED_DEVICE_FAMILY value
- Remove i386 version for iOS simulator
- Remove minus sign from CFBundleShortVersionString. Now the version is shown in iOS>Settings>General>Storage>PPSSPP

I also had to force [ZSTD_ENABLE_ASM_X86_64_BMI2=0](https://github.com/facebook/zstd/blob/4a74eb8cfb4a1887140c7581cbf0dfcf2fedec7a/lib/common/portability_macros.h#L136) to build the iOS version for simulator

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -221,12 +221,6 @@ jobs:
         sudo apt-get update -y -qq
         sudo apt-get install libsdl2-dev libgl1-mesa-dev libglu1-mesa-dev libsdl2-ttf-dev libfontconfig1-dev
 
-    - name: Create macOS git-version.cpp for tagged release
-      if: startsWith(github.ref, 'refs/tags/') && runner.os == 'macOS' && matrix.extra == 'test'
-      run: |
-        echo "const char *PPSSPP_GIT_VERSION = \"${GITHUB_REF##*/}\";" > git-version.cpp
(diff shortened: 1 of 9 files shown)


hrydgard:
Hm, does bumping the iOS target restrict deployment to older versions of iOS?

Florin9doi:
Yes, iOS 11 is no longer supported by the latest SDK. Should I revert it to iOS 11 to be used with older SDKs?

Florin9doi:
Won't work for https://github.com/hrydgard/ppsspp/blob/2a69c96a49da75f85270a032b3915740c013c55e/.github/workflows/build.yml#L224

anr2me:
For the git-version.cpp generation, you can try using the method i use for https://github.com/hrydgard/ppsspp/blob/master/.github/workflows/manual_generate_ipa.yml

Florin9doi:
It is already generated, but its location is different from build to build, sometimes it is placed in root directory and sometime it is placed in ./build[-version]

hrydgard:
I'm happy to merge, but not sure if I should wait until after the 17.x series is done. (Likely won't go higher than 17.2 this time, maybe 17.1 will be enough)...

Issue #18798, opened 2024-01-31
Performance problems with CHD 


master1274:
### Game or games this happens in

1. Dante's Inferno 2. God of War - Chains of Olympus 3. God of War - Ghost of Sparta 4. Colin McRae Rally 2005 5. Tekken 5 Dark Resurrection 6. The Treasures of Montezuma (minis) (PSN) 7. N.O.V.A. Near Orbit Vanguard Alliance (minis) (PSN)

### What area of the game / PPSSPP

Some performance problems after converting to chd

### What should happen

Performance shouldn't suffer

### Platform

Android

### Mobile device model or graphics card (GPU)

Various devices 

### PPSSPP version affected

Since the launch of support chd, before the current build

### Last working version

Since the launch of support chd, before the current build

### Graphics backend (3D API)

OpenGL / GLES

### Checklist

- [X] Test in the [latest git build](https://buildbot.orphis.net/ppsspp/) in case it's already fixed.
- [X] [Search for other reports](https://github.com/hrydgard/ppsspp/search?q=my+issue&type=issues) of the same issue.
- [ ] Try resetting settings or older versions and include if the issue is related.
- [ ] Try without any cheats and without loading any save states.
- [ ] Include logs or screenshots of issue.

hrydgard:
You tried the same game in both CHD and ISO, and you can measure a difference? You sure?

master1274:
This list was compiled by several contributors. Of course, the games were initially tested in iso. There were no problems. I personally checked Tekken 5: Dark Resurrection and The Treasures of Montezuma. Tekken 5: Dark Resurrection was also checked in cso. There were no problems either. Because We don’t know the reasons for the problem, we just decided to pass the list here so we can check this observation.

hrydgard:
How does the problem manifest itself? Are the framerates low? Choppy or slow loading of the next level?

I need to know what to look for in order to investigate. Got any video comparison?

The thing is, CHD should only affect the speed of UMD disc reading, and games are already tolerant of slow speed, so a bit of additional compression and hashing really shouldn't make much of a difference, other than slightly affecting load times maybe.

hrydgard:
Seems very odd to me that CHD would cause that, and if it does, it's really libchd performing poorly. We use it just as if it was the CSO backend, no difference there.

I recommend you use CSO on lower powered devices.

hrydgard:
[...]

PR #18803, opened 2024-01-31
Warn the user about bad CHDs

hrydgard:
If they're not created with "chdman createdvd", they will perform really badly.

See #18798

Sorry @TotalCaesar659 and @Saramagrean , another string for you :)

image

### DIFF
--- a/Common/System/OSD.h
+++ b/Common/System/OSD.h
@@ -15,6 +15,7 @@ enum class OSDType {
 	MESSAGE_ERROR_DUMP,  // displays lots of text (after the first line), small size
 	MESSAGE_FILE_LINK,
 	MESSAGE_CENTERED_WARNING,
+	MESSAGE_CENTERED_ERROR,
 
 	ACHIEVEMENT_UNLOCKED,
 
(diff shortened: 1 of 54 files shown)


TotalCaesar659:
It's alright, I'm already used to it ;-)

nameless1:
So, it warns about bad chds that were made with createdvd... am I missing something?

E:\Playstation Portable\PSP Backups>chdman.exe info -i "Ape Escape on the Loose (USA).chd"
chdman - MAME Compressed Hunks of Data (CHD) manager 0.262 (unknown)
Input file:   Ape Escape on the Loose (USA).chd
File Version: 5
Logical size: 749,240,320 bytes
Hunk Size:    4,096 bytes
Total Hunks:  182,920
Unit Size:    2,048 bytes
Total Units:  365,840
Compression:  lzma (LZMA), zlib (Deflate), huff (Huffman), flac (FLAC)
CHD size:     279,415,049 bytes
Ratio:        37.3%
SHA1:         aabdebdb87d0363c4ba911dc72819dc5e014ad95
Data SHA1:    373098a849dcdc573499a28c89c92b63376e666d
Metadata:     Tag='DVD '  Index=0  Length=1 bytes

Screenshot 2024-02-05 220959

hrydgard:
Yup it's the hunk size. No other hunk size than 2048 makes sense to use, since that's the sector size.

pocafon:
How can we lower the hunksize to 2048 if it's not much to ask?

hrydgard:
@pocafon using `chdman createdvd` will do it automatically.

You need a recent build of chdman though, such as the one shipped with the latest build of MAME.

Issue #18607, opened 2023-12-24
MotorStorm Artic Edge too slow 


003198:
### Game or games this happens in

UCES-01250

### What area of the game

*

### Speed seen in PPSSPP

60% (20/30)

### GE frame capture and debug statistics

[UCES01250.ppdmp.zip](https://github.com/hrydgard/ppsspp/files/13762487/UCES01250.ppdmp.zip)

### Platform

Android

### Mobile device model or graphics card (GPU)

Redmi Note 10 Pro

### PPSSPP version affected

1.16.6

### Graphics backend (3D API)

Vulkan

### Any other notes or things you've tried

GPU and CPU usage are really low. It only runs well if I use 'skip GPU readbacks,' but then the game looks so bright that I can't see anything ahead. The resolution doesn't make any difference, whether I use 1x or 4x.
![Screenshot_2023-12-24-14-22-50-681_org ppsspp ppsspp](https://github.com/hrydgard/ppsspp/assets/153548024/7997ef9e-4190-4220-92f7-3237e62f69fc)
![Screenshot_2023-12-24-14-25-45-456_org ppsspp ppsspp](https://github.com/hrydgard/ppsspp/assets/153548024/17b0f92a-0246-4532-abbd-b7beb3e299af)

https://github.com/hrydgard/ppsspp/assets/153548024/7663f443-11da-4be8-8113-c454a1fb2f22

### Checklist

- [X] Test in the [latest git build](https://buildbot.orphis.net/ppsspp/) in case it's already fixed.
- [X] [Search for other reports](https://github.com/hrydgard/ppsspp/search?q=my+issue&type=issues) of the same issue.
- [ ] Try resetting settings or older versions and include if the issue is related.
- [X] Try changing graphics settings to determine if one causes the slowdown.
- [X] Include logs or screenshots of issue.

anr2me:
Readbacks are needed for some games, but there is performance/accuracy problem according to this https://github.com/hrydgard/ppsspp/issues/16900

sum2012:
git version have a "skip" option
![3](https://github.com/hrydgard/ppsspp/assets/2177532/19d56c7c-2ceb-4210-b63e-784dbb39c4e9)

003198:
[...]

PR #18807, opened 2024-02-01
Add "Half skip" Mode

sum2012:
My random idea from https://github.com/hrydgard/ppsspp/issues/16900#issuecomment-1903906739

Added as another mode to the speedhack setting Skip GPU readbacks.

May help #18607

### DIFF
--- a/Core/Config.h
+++ b/Core/Config.h
@@ -241,7 +241,7 @@ public:
 	float fCwCheatScrollPosition;
 	float fGameListScrollPosition;
 	int iBloomHack; //0 = off, 1 = safe, 2 = balanced, 3 = aggressive
-	int iSkipGPUReadbackMode;  // 0 = off, 1 = skip, 2 = to texture
+	int iSkipGPUReadbackMode;  // 0 = off, 1 = skip, 2 = to texture, 3 = half skip
 	int iSplineBezierQuality; // 0 = low , 1 = Intermediate , 2 = High
 	bool bHardwareTessellation;
 	bool bShaderCache;  // Hidden ini-only setting, useful for debugging shader compile times.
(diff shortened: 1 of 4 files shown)


LunaMoo:
Why making it random if you want "half"(which would imply every other frame)?

 Random pretty much means it might break things sometimes and sometimes it will not, doesn't feel like an useful thing to have, even less for performance reasons as lower framerate is better than unpredictable one.

hrydgard:
One argument for making it random is that it will work to halve the work "evenly" whether the game does 1, 2 or 3 readbacks per frame for example - if you do every second readback, you run the risk of falling into stuck patterns and always skipping the same read every frame.  But, obvious drawbacks are of course uneven framerates etc.

Either way, this idea of skipping readbacks randomly highly depends on what the game is doing, if it makes sense or not...

What game do you feel this is the most beneficial in, sum2012?

sum2012:
MotorStorm Artic Edge, gpu readback the sun to adjustment the sky

hrydgard:
Oh, yeah, I guess it can be a pretty good option for that game, since the effect isn't very frame-critical.. 

Hm. I'm gonna play with it soon.

sum2012:
@hrydgard how about this ? ( although need rebase )

hrydgard:
Is it actually a good gameplay improvement, and doesn't just get weirdly choppy?

sum2012:
Yes

在 2024年9月17日週二 下午9:02，Henrik Rydgård ***@***.***> 寫道：

hrydgard:
ok. i'm gonna manually rebase this and do some minor changes (naming, initialization).

hrydgard:
Replacing this with #19469 

Not super succesful for me, something weird is going on in Motorstorm.

Issue #18806, opened 2024-02-01
UFC Undisputed 2010: Crash on device lost on some ARM GPUs


hrydgard:
After the first few punches in a fight, the game just crashes with a device loss on Mali G52 with driver v18 (model: HUAWEI:JNY-LX1).

```
#02 pc 00000000008675f8  arm64/libppsspp_jni.so (HandleAssert(char const*, char const*, int, char const*, char const*, ...)+344) (BuildId: e2a665cce99374cec395cfef5394541ea8505f76)
#03 pc 00000000008398cc  arm64/libppsspp_jni.so (VulkanRenderManager::BeginFrame(bool, bool)+280) (BuildId: e2a665cce99374cec395cfef5394541ea8505f76)
#04 pc 0000000000dd18b0  arm64/libppsspp_jni.so (Draw::VKContext::BeginFrame(Draw::DebugFlags)+32) (BuildId: e2a665cce99374cec395cfef5394541ea8505f76)
#05 pc 00000000008851b0  arm64/libppsspp_jni.so (NativeFrame(GraphicsContext*)+748) (BuildId: e2a665cce99374cec395cfef5394541ea8505f76)
```

Vulkan validation doesn't flag any problems, so this one might be real tricky to track down...

Runs fine in OpenGL.

Reported by a user on Google Play.

Confirmed affected devices:

* Galaxy S8+ (Mali G71)
* Huawei (Mali G52)
* Galaxy S21 Ultra (Mali G78)

The last one behaved slightly differently, after the bug hit it stumbles along at 1fps for a few frames, hangs, and dies after a delay. While it stumbles, our tracked GPU memory consumption does not seem to increase.
```
02-01 19:26:42.704 22312 22745 E Fence   : waitForever: Throttling EGL Production: fence 158 didn't signal in 3000 ms
02-01 19:26:42.704 22312 22745 I Fence   : waitForever: fence(mali-mali.timeline1512757-357) status(0)
02-01 19:26:42.704 22312 22745 I Fence   : waitForever: sync point: timeline(mali.timeline) drv(mali) status(0) timestamp(0.000000)
```

I'm trying to rule out causes, some notes:

* Confirmed that this happens on a Galaxy S8+ with PPSSPP 16.4, so not new. It also performs horrendously!
* Enabling the robustBufferAccess feature on the device doesn't do anything.
* Running with Android validation layers doesn't catch anything
* Skip buffer effects doesn't help
[...]

PR #18813, opened 2024-02-02
Mali: Turn off any depth writes in the shader if depth test == NEVER

hrydgard:
Works around #18806.

If this breaks graphics in any game, they already crashed on Mali. NEVER is very rare too.

So I think there's no way this is a risk, except for the very tiny change of adding DIRTY_FRAGMENTSHADER_STATE on ZTEST changes. We could gate it on Mali, though I don't think it's gonna regress anything.


hrydgard:
Hm, this was strangely reported to cause lost devices on some Intel GPUs by Flannette on Discord:

image

PR #18824, opened 2024-02-04
Update libchdr to the latest, which supports zstd blocks

hrydgard:
Apparently these perform better, and I assume people will start using these by default and expect them to work soon.

Sigh, I thought CHD was a fixed thing...

Also updated the guide on https://www.ppsspp.org/docs/getting-started/dumping-games/ with this information.

### DIFF
--- a/UWP/libchdr_UWP/libchdr_UWP.vcxproj
+++ b/UWP/libchdr_UWP/libchdr_UWP.vcxproj
@@ -80,7 +80,8 @@
   <ImportGroup Label="PropertySheets">
     <Import Project="$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props" Condition="exists('$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props')" Label="LocalAppDataPlatform" />
   </ImportGroup>
-  <PropertyGroup Label="UserMacros" /><PropertyGroup />
+  <PropertyGroup Label="UserMacros" />
+  <PropertyGroup />
   <PropertyGroup>
     <GenerateManifest>false</GenerateManifest>
   </PropertyGroup>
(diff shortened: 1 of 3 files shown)


LunaMoo:
I find if funny that chd got a zstd compression due to popular requests caused by performance problems with lzma^_^, but I doubt it's users will switch to zstd.

 I mean with zstd it's only selling point aka filesize ends up worse than cso's made by maxcso, of course it's an irrelevant difference, but so was the other way around before and they still tried to make it most important thing ever.

thingsiplay:
> I mean with zstd it's only selling point aka filesize ends up worse than cso's made by maxcso

To me the selling point of CHD is to use a single program to convert (and manage) the files, to work with a single uniform file extension. Even if CHD files would be larger in filesize against CSO, I would still prefer CHD.

What about the performance issues caused by lzma? Is this something only noticeable in weak hardware and only then zstd makes a difference? And what performance do we speak here? Does it have an effect on while gaming or is it just the loading times when starting the game?

hrydgard:
Some games stream data in-game (like GTA), causing stutters if the storage is very slow.

However, I think standard CHDs created with createdvd are fast enough anyway to not really notice. Though, CHDs created with createcd are indeed very, very slow for PPSSPP to read from due to the mismatching sector sizes.

LunaMoo:
I guess there's also a reason to think about performance that can't be detected by stutters/glitches or FPS - on battery powered devices every watt matters for battery life.

 Generally it's all for humans, having extra option is cool for most of of us, the primary reasons some of us were against the format were:
- lies and missinformation spread by people who literally wanted nothing else, but to cash-in bounties and advertise their awful software abusing other people's work further on the emulation scene,
- maintenance,
- already having 2 compressed formats which meets all needs of PSP emulation with no downsides and being issue free.

 CHD for PSP games was always like reinwenting wheel, just with lots of rought edges due to the format being designed around games and hardware from completely different era and the whole "retro" mindset of cluttering a thing which could be best in one thing into being average at everything.

 Compressing backups is basically done once, there's 0 advantage of having 1 software for all console backups and for many PSP games it's not worth doing anything as when bought of the store they come with already well compressed sony's container called PBP which also had a lot of missinformation spread coming from retro-people either just lying for their personal benefits or just being dumb and confusing PSP with PSOne.

 CHDMAN isn't anything super cool or user friendly either dunno why would anyone want to make everything using such thing, just a really basic commandline software with many options that are badly documented or lack any real documentation, software which doesn't even have standalone download and is part of a large mame project which really is on the darkest parts of the emulation grayzone with it's users in majority not even being able to own physical game copies due to nature of the arcades.

 People that think they can just keep using that software with the same settings for all their backups will hit even harder performance issues like #18798 which Henrik mentioned. And then if you really would care about maximizing performance and to be safe for all your future needs just use zstd, you end up with larger filesize than simply sticking to maxCSO closing the "wheel of chd worth" in PSP emulation and squishing it back to when it wasn't supported.

PR #18825, opened 2024-02-04
ffmpeg: Improved fix for checking if const AVCodec* is necessary

Tatsh:
The last way to do this in #18670 was a bit flawed and just added a lot of duplicated code. I think testing the code at configure time is better. Sadly we cannot use something like `-DAVCodec="const AVCodec"` because any `-D` on the command line comes before the includes.

This change also creates a new file that can be used for more macros in the future.

Releated #15308

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -116,6 +116,7 @@ if(NOT IOS)
 	list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake/sdl)
 endif()
 
+include(CheckCXXSourceCompiles)
 include(GNUInstallDirs)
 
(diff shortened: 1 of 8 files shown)


hrydgard (approved):

brad0:
@Tatsh @hrydgard I was wondering why the HAVE_LIBAVCODEC_CONST_AVCODEC check_cxx_source_compiles test was failing and I noticed it does not pull in the FFmpeg header or library paths so it can't find either the headers or the libraries when linking.

hrydgard:
Hm, that's a good point, so it never worked at all? @Tatsh ?

Tatsh:
This was always tested on the assumption that it's for those willing to link against system ffmpeg instead of bundled. Because of that it doesn't need to pull in include paths or similar because the test assumes normal locations for those things.

Did this break using bundled ffmpeg? I feel like this would have been caught at the time.

PR #18858, opened 2024-02-12
Make the i18n T function use std::string_view

hrydgard:
Just a code refactoring I've wanted to get in for a while.


hrydgard:
If I missed a case somewhere, this could lead to some new weird crash in the UI. Should notice it fairly quickly though. I did scroll through most of the UI to be sure..

PR #18867, opened 2024-02-15
Allow users to change UPnP timeout setting

anr2me:
In case their UPnP device took longer to be detected.

Will need to get feedbacks first from user who reported that their iPhone 12 pro can't detect UPnP device anymore (might be related to newer iOS version?)  https://forums.ppsspp.org/showthread.php?tid=29904

PS: If the reporter can solve the issue by increasing the timeout, i'll probably need to increase the default value too, a long with this PR.

**Edit:** Apparently it wasn't timeout issue, may be miniupnpc no longer compatible with newer iOS :( since it (said to) works on PC on the same network/router.
Or may be we just need to update the miniupnp library?


Core/Config.cpp
```diff
@@ -865,6 +865,7 @@ static const ConfigSetting networkSettings[] = {
 	ConfigSetting("ForcedFirstConnect", &g_Config.bForcedFirstConnect, false, CfgFlag::PER_GAME),
 	ConfigSetting("EnableUPnP", &g_Config.bEnableUPnP, false, CfgFlag::PER_GAME),
 	ConfigSetting("UPnPUseOriginalPort", &g_Config.bUPnPUseOriginalPort, false, CfgFlag::PER_GAME),
+	ConfigSetting("UPnPTimeout", &g_Config.iUPnPTimeout, 2000, CfgFlag::PER_GAME),
```
anr2me, line 868: Hmm.. since upnp timeout usually external issue, may be not supposed to be PER_GAME setting.

anr2me:
We probably just need to update our miniupnp library, may be.

hrydgard:
Hm, ok. we can leave this sitting around for now.

hrydgard:
Alright, do we do anything else with this?

anr2me:
This is probably not needed, i'll close it but i'll keep the branch, just in case

PR #18891, opened 2024-02-25
Fix "Savestate Slot" string

LunaMoo:
Not sure how long it was broken nor if there are any similar to fix, very minor thing that nobody notices.

### DIFF
--- a/UI/NativeApp.cpp
+++ b/UI/NativeApp.cpp
@@ -1200,7 +1200,7 @@ bool HandleGlobalMessage(UIMessage message, const std::string &value) {
 		return true;
 	} else if (message == UIMessage::SAVESTATE_DISPLAY_SLOT) {
 		auto sy = GetI18NCategory(I18NCat::SYSTEM);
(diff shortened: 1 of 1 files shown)


UI/NativeApp.cpp
```diff
@@ -1200,7 +1200,7 @@ bool HandleGlobalMessage(UIMessage message, const std::string &value) {
 		return true;
 	} else if (message == UIMessage::SAVESTATE_DISPLAY_SLOT) {
 		auto sy = GetI18NCategory(I18NCat::SYSTEM);
-		std::string msg = StringFromFormat("%s: %d", sy->T("Savestate Slot"), SaveState::GetCurrentSlot() + 1);
+		std::string msg = StringFromFormat("%s: %d", std::string(sy->T("Savestate Slot")), SaveState::GetCurrentSlot() + 1);
```
hrydgard, line 1203: Good catch, but that's not gonna work, can't pass an std::string as a format parameter.

Either use T_cstr() which remains for this use, or .c_str() after constructing the std::string.

LunaMoo: Well, it's only for ("Savestate Slot") part and works as such, but ok, I'll change it.

PR #18896, opened 2024-02-26
FS: fix crash overreading data on flash0

rofl0r:
noticed while using intrafont with some homebrew code. when using musl libc's hardened malloc, the overread in the code caused PPSSPP to segfault.


hrydgard (approved):

hrydgard:
good catch!

PR #18913, opened 2024-03-07
fix libretro compile error

hsoftxl:


libretro/jni/Application.mk
```diff
@@ -2,4 +2,4 @@ NDK_TOOLCHAIN_VERSION ?= 4.8
 APP_ABI := armeabi-v7a,arm64-v8a,x86
 APP_STL := c++_static
 APP_GNUSTL_CPP_FEATURES := exceptions
-APP_PLATFORM := android-9
+APP_PLATFORM := android-21
```
hrydgard, line 5: Hm, is this needed?

hrydgard:
(Looks good, just wondering about the platform bump. We try to specify as low an android version as possible in various places).

libretro/jni/Application.mk
hsoftxl: yes, if 9, compile error

PR #18915, opened 2024-03-07
utils: rewrite on optimized function SplitString

GermanAizek:
@hrydgard, new function affects performance on Debug and Release with optimizations.

Release: 2 times faster than before
Debug: 3.6 times faster than before

Benchmark results (not ideal input data): https://quick-bench.com/q/rJ3XmSH1EoG8Ms-CgFUYzIINb8k

### DIFF
--- a/Common/StringUtils.cpp
+++ b/Common/StringUtils.cpp
@@ -316,12 +316,15 @@ void SplitString(std::string_view str, const char delim, std::vector<std::string
 
 void SplitString(std::string_view str, const char delim, std::vector<std::string> &output) {
 	size_t next = 0;
-	for (size_t pos = 0, len = str.length(); pos < len; ++pos) {
-		if (str[pos] == delim) {
-			output.emplace_back(str.substr(next, pos - next));
-			// Skip the delimiter itself.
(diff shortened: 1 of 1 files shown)


Common/StringUtils.cpp
```diff
@@ -316,12 +316,17 @@ void SplitString(std::string_view str, const char delim, std::vector<std::string
...
 void SplitString(std::string_view str, const char delim, std::vector<std::string> &output) {
 	size_t next = 0;
-	for (size_t pos = 0, len = str.length(); pos < len; ++pos) {
-		if (str[pos] == delim) {
-			output.emplace_back(str.substr(next, pos - next));
-			// Skip the delimiter itself.
-			next = pos + 1;
+	size_t pos = 0;
+	size_t len = str.length();
+	output.reserve(len);
```
hrydgard, line 321: that's a really large reservation, as if every character would be split into their own string...

hrydgard:
Remove the reserve call (or reserve like, 4) and I'll merge this. As mentioned, reserving the same number of output strings as input characters as the length of the input string is always far too much.

Issue #18919, opened 2024-03-08
CHD files don't work when using the FileProvider API from a frontend application


leonstyhre:
### Game or games this happens in

All games

### What area of the game / PPSSPP

When launching games from the ES-DE Frontend using the FileProvider API everything works fine, except when launching CHD files. If I use SAF URIs instead then everything works fine on my devices, but some people have reported weird/random issues with this mechanism on some devices so it would be really nice to be able to switch to the FileProvider API as the default.

What happens when attempting to launch CHD files using the FileProvider API is that the message "Could not load game. Unknown file type:" is displayed inside PPSSPP, followed by the file/content URI.

Do you perhaps have some specific logic in place to handle CHD files differently than other file formats, which could explain the behavior? PPSSPP is the only emulator I have come across that behaves like this so I don't think there is anything on my side that is causing the issue.

Many thanks in advance!

### What should happen

CHD files should launch correctly when using the FileProvider API in the same way as they do when using SAF URIs.

### Platform

Android

### Mobile device model or graphics card (GPU)

All devices

### PPSSPP version affected

v1.17.1 (regular and Gold)

### Graphics backend (3D API)

Vulkan

### Checklist

- [X] Test in the [latest git build](https://buildbot.orphis.net/ppsspp/) in case it's already fixed.
- [X] [Search for other reports](https://github.com/hrydgard/ppsspp/search?q=my+issue&type=issues) of the same issue.
- [ ] Try resetting settings or older versions and include if the issue is related.
- [ ] Try without any cheats and without loading any save states.
- [ ] Include logs or screenshots of issue.

leonstyhre:
And btw, I did not test the latest Git build, it seems like the buildbot system is not responding so I couldn't get hold of the latest build.

GlazedBelmont:
Went ahead and built latest commit of PPSSPP, when using %ROMPROVIDER% in ES-DE, this is what I get

Android 13
Odin 2
Roms on external storage

![image](https://github.com/hrydgard/ppsspp/assets/48196637/5306abbf-d53f-4d20-857d-6bcfee26f977)

GlazedBelmont:
Here's the log of what happens the game is launched

```
03-08 17:47:00.667 10255 10255 I PpssppActivity: Found Shortcut Parameter in data: content://org.es_de.frontend.files/root/storage/30FC-BE41/Games/psp/Jeanne%20d'Arc.chd
[...]

PR #18924, opened 2024-03-12
Fix a bunch of cases where we forgot to check for CHD files

hrydgard:
Should fix some issues like possibly #18919 .

### DIFF
--- a/UWP/PPSSPP_UWPMain.cpp
+++ b/UWP/PPSSPP_UWPMain.cpp
@@ -502,7 +502,7 @@ bool System_MakeRequest(SystemRequestType type, int requestId, const std::string
 		std::vector<std::string> supportedExtensions = {};
 		switch ((BrowseFileType)param3) {
 		case BrowseFileType::BOOTABLE:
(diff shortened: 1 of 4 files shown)


UWP/PPSSPP_UWPMain.cpp
```diff
@@ -502,7 +502,7 @@ bool System_MakeRequest(SystemRequestType type, int requestId, const std::string
 		std::vector<std::string> supportedExtensions = {};
 		switch ((BrowseFileType)param3) {
 		case BrowseFileType::BOOTABLE:
-			supportedExtensions = { ".cso", ".bin", ".iso", ".elf", ".pbp", ".zip" };
+			supportedExtensions = { ".chd", ".cso", ".bin", ".iso", ".elf", ".pbp", ".zip" };
```
anr2me, line 505: just wondering why `.prx` extension can't be found in this list :) since prx files can be run on PPSSPP too (ie. sample files from PSPSDK)

hrydgard: was probably just forgotten when the list was created. I'll add it.

PR #18931, opened 2024-03-16
CHD: Fix unnecessary reloads of "hunks" during large reads

### LINKED ISSUE #18925: PPSSPP warns about bad performant CHD while using ZSTD

None

image

using the follow command:
`chdman createdvd -c "zstd,zlib,huff,flac" -i "game.iso" -o "game.chd"`

the emulator identifies as bad performant CHD, zstd should be more performant than lzma...

PS: using the MAME 0.263 chdman.

### What should happen

It should identify the CHD as good one.

### Logs

_No response_

### Platform

Windows

AMD 7900 XT

### PPSSPP version affected

1.17.1 (stable)

### Last working version

_No response_

### Graphics backend (3D API)

Vulkan

### Checklist

hrydgard:
This fixes the worst performance problem of using large hunk sizes with CHDs in PPSSPP.

It seems that with this, performance is good enough that we can remove the warning for bad CHDs, but I still recommend 2048-byte hunk size and will continue to do so in the [website instructions](https://www.ppsspp.org/docs/getting-started/dumping-games/).

Fixes #18925

### DIFF
--- a/Core/FileSystems/BlockDevices.cpp
+++ b/Core/FileSystems/BlockDevices.cpp
@@ -644,12 +644,6 @@ CHDFileBlockDevice::CHDFileBlockDevice(FileLoader *fileLoader)
 	impl_->chd = file;
 	impl_->header = chd_get_header(impl_->chd);
 
-	if (impl_->header->hunkbytes != 2048) {
-		badCHD_ = true;
-	} else {
-		badCHD_ = false;
-	}
-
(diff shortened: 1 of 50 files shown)


LunaMoo:
They say it's about file size, but it really differs a lot between games and in most cases 2048 produces smaller size, so the argument is just silly.

hrydgard:
Well, either way, we had a bug, so fixing it.

PR #18942, opened 2024-03-21
OpenXR - One APK for Android and standalone VR

lvonasek:
The obstacles to include VR in the Android APK are now gone:
* AndroidManifest is backward compatible
* Meta Quest has proper support for scope storage
* OpenXR libs are available on every CPU architecture

After merging this:
* If Android APK is installed on a compatible VR headset then it starts in VR mode
* `OPENXR` define indicates the VR is available on the platform but not necessary enabled
* There is no longer need for https://github.com/hrydgard/ppsspp/issues/17065


hrydgard:
Very nice!

How is the scope storage support, smooth?

I suppose people with the old app installed will just need to remove it and install this. So, I can just upload the regular APK to SideQuest, after this?

lvonasek:
The scope storage flow is smooth.

It isn't the ppssppvr package anymore. It might be that the listing on Sidequest have to be recreated. If update will install on the headset as a new app.

hrydgard:
OK. Let's get it in!

hrydgard:
Hm, this broke the arm32 build, which the buildbot doesn't catch:

```
C:/Android/sdk/ndk/21.4.7075529/toolchains/llvm/prebuilt/windows-x86_64/lib/gcc/arm-linux-androideabi/4.9.x/../../../../arm-linux-androideabi/bin\ld: error: cannot find -lGLESv3
Common/VR/VRFramebuffer.cpp:74: error: undefined reference to 'glGetError'
Common/VR/VRFramebuffer.cpp:314: error: undefined reference to 'glDeleteFramebuffers'
```

Weird...

lvonasek:
I'll fix it by removing GLES3 and multiview dependency (which isn't used).

hrydgard:
Cool - though, just ifdef it away for arm32 for now?

lvonasek:
Ok, as you wish

hrydgard:
thanks!

anr2me:
I was planning to add the VR build variants on my manual generate apk workflow today, but after i saw this i guess it's not longer needed >.< goodworks on combining it.

Btw, does iOS or any other platforms still have a separate build variant/configuration for VR? if they're existed i should probably add them to the manual generate workflows.

lvonasek:
Nope, Android is that the moment the only platform supporting VR. AFAIK iOS/VisionOS cannot use this due to missing OpenXR support. There is plan to add Windows support at some point but that will be probably also included in one binary.

lvonasek:
I was testing this more and it might be better to revert it. The data migration gets on the latest QuestOS very wonky.

hrydgard:
Hm, what goes wrong?

I'm ok with reverting if needed, though.

lvonasek:
The dialog to select directory often doesn't open. The access to folder is often refused. Opening recents from the previous version freezes the headset.

Could you do the revert? (For you it is less clicks).

anr2me:
> The dialog to select directory often doesn't open. The access to folder is often refused. Opening recents from the previous version freezes the headset.
> 
> Could you do the revert? (For you it is less clicks).

Sounds like similar issue to some Android TV O.o
Does the standalone VR use a different permission than the regular ppsspp?

lvonasek:
Thank you @hrydgard.

@anr2me before this PR it used legacy and everything worked perfectly. With this PR it uses scoped storage (just like regular Android app) and as I mentioned above it is currently pain to use.

hrydgard:
@lvonasek VR builds are now available on the buildbot, I set a cutoff to start from this very version (-106): https://www.ppsspp.org/devbuilds/

PR #18953, opened 2024-03-26
Using constexpr std::size() C++17 more faster and safer from typos

GermanAizek:
@hrydgard, I think it will be very useful for emu, and also how do you think about replacing new operator with unique_ptr for safe memory cleanup and code reduce?

### DIFF
--- a/Core/HW/SasAudio.cpp
+++ b/Core/HW/SasAudio.cpp
@@ -165,14 +165,16 @@ void VagDecoder::DoState(PointerWrap &p) {
 	if (!s)
 		return;
 
+	const int size = std::size(samples);
 	if (s >= 2) {
-		DoArray(p, samples, ARRAY_SIZE(samples));
+		DoArray(p, samples, size);
 	} else {
-		int samplesOld[ARRAY_SIZE(samples)];
-		DoArray(p, samplesOld, ARRAY_SIZE(samples));
-		for (size_t i = 0; i < ARRAY_SIZE(samples); ++i) {
+		int* samplesOld = new int[size];
+		DoArray(p, samplesOld, size);
+		for (size_t i = 0; i < size; ++i) {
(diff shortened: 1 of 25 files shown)


Nemoumbra:
`#define ARRAY_SIZE(a) std::size(a)` looks funny. The idea is great, to be honest, but the real modern C++ code doesn't even use C-arrays all that much like we do.
Why did you make the arrays `constexpr`?

GermanAizek:
> Why did you make the arrays `constexpr`?

@Nemoumbra, indeed, you are right, static does not make sense outside scope function
https://iamsorush.com/posts/cpp-constexpr/#constexpr-vs-static-constexpr

GermanAizek:
@Nemoumbra, fix it here: https://github.com/hrydgard/ppsspp/pull/18953/commits/5efc328681635c39bdf5282b65d4c798a9cb75b5

Core/HW/SasAudio.cpp
```diff
@@ -165,14 +165,16 @@ void VagDecoder::DoState(PointerWrap &p) {
...
 
+	const int size = std::size(samples);
 	if (s >= 2) {
-		DoArray(p, samples, ARRAY_SIZE(samples));
+		DoArray(p, samples, size);
 	} else {
-		int samplesOld[ARRAY_SIZE(samples)];
-		DoArray(p, samplesOld, ARRAY_SIZE(samples));
-		for (size_t i = 0; i < ARRAY_SIZE(samples); ++i) {
+		int* samplesOld = new int[size];
```
Nemoumbra, line 172: Oh no, unnecessary dynamic memory allocation!

hrydgard:
I don't think it's any safer than our existing ARRAY_SIZE macro. I don't really feel like detail-reviewing every change here to make sure it's right, since there's essentially no gain to be had.

PR #18957, opened 2024-03-29
[C++17] Conversion to same type, as well more correct use constexpr strlen and strings

GermanAizek:
More info here: https://iamsorush.com/posts/cpp-constexpr/#constexpr-values


hrydgard:
Nothing here is gonna improve anything, just risk. It's all const enough already for the compiler to do a good job. Closing.

PR #18959, opened 2024-03-29
More strict const ref in range-loops and immutable vars + reserve()

GermanAizek:
More info: https://blog.knatten.org/2013/10/13/prefer-using-references-with-range-based-for-loops/

### DIFF
--- a/UI/MainScreen.cpp
+++ b/UI/MainScreen.cpp
@@ -104,7 +104,7 @@ static bool IsTempPath(const Path &str) {
 	item = ReplaceAll(item, "/", "\\");
 #endif
 
-	std::vector<std::string> tempPaths = System_GetPropertyStringVec(SYSPROP_TEMP_DIRS);
+	const auto &tempPaths = System_GetPropertyStringVec(SYSPROP_TEMP_DIRS);
 	for (auto temp : tempPaths) {
 #ifdef _WIN32
 		temp = ReplaceAll(temp, "/", "\\");
@@ -674,7 +674,7 @@ bool GameBrowser::DisplayTopBar() {
 bool GameBrowser::HasSpecialFiles(std::vector<Path> &filenames) {
 	if (path_.GetPath().ToString() == "!RECENT") {
 		filenames.clear();
-		for (auto &str : g_Config.RecentIsos()) {
+		for (const auto &str : g_Config.RecentIsos()) {
 			filenames.push_back(Path(str));
 		}
 		return true;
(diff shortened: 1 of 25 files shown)


UI/MainScreen.cpp
```diff
@@ -1389,7 +1389,7 @@ void MainScreen::update() {
 	UpdateUIState(UISTATE_MENU);
 
 	if (searchChanged_) {
-		for (auto browser : gameBrowsers_)
+		for (auto* browser : gameBrowsers_)
```
anr2me, line 1392: just wondering, is this made a difference? since `gameBrowsers_`'s elements is already a pointer (ie. `GameBrowser*`), thus `browser` type will became `GameBrowser**` after this change isn't?

GermanAizek: fix it here https://github.com/hrydgard/ppsspp/pull/18959/commits/83cb897b1fd92965583adc608edf4b246f3c41b1

Nemoumbra:
Are you amending your commits and force-pushing them, @GermanAizek? This completely erases all the possibilities for other people to see the history on the PR. If you're afraid that 2 commits would pollute the commit history, first of all, you're wrong, and second of all... Git can *squash on merge*. If hrydgard wants to, even 100 commits can be made into 1 commit, so don't concern yourself with that. You just make it harder to review.

UI/MainScreen.cpp
Nemoumbra: > just wondering, is this made a difference? since `gameBrowsers_`'s elements are already a pointer (ie. `GameBrowser*`), thus `browser` type will became `GameBrowser**` after this change isn't?

Auto is like a template, so no.
* `auto browser` => `auto` turns into `GameBrowser*`.
* `auto* browser` => `auto` turns into `GameBrowser`.
The type of `browser` is still `GameBrowser*`.

GermanAizek:
@Nemoumbra, I'm used to doing this, probably because in other repositories I'm always asked to make squash commits. you are first one who was against it, but I will make next commits as separate new commits.

Florin9doi:
> Are you amending your commits and force-pushing them, @GermanAizek? This completely erases all the possibilities for other people to see the history on the PR. If you're afraid that 2 commits would pollute the commit history, first of all, you're wrong, and second of all... Git can _squash on merge_. If hrydgard wants to, even 100 commits can be made into 1 commit, so don't concern yourself with that. You just make it harder to review.

You can click here to see the history: 
image

hrydgard:
Not a single one of these are in code that matters at all for performance, so it barely seems worth the risk for regressions... It's a lot of changes to review and test.

GermanAizek:
> Not a single one of these are in code that matters at all for performance, so it barely seems worth the risk for regressions... It's a lot of changes to review and test.

U can only take what you think is necessary and close PR.

PR #18960, opened 2024-03-29
Using constexpr C++17 strlen implementation

GermanAizek:
More info: https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0426r1.html


Nemoumbra:
And why do you want that to be used in PPSSPP?

GermanAizek:
> And why do you want that to be used in PPSSPP?

constexpr is much better codegenerated by compiler than c function, also do not forget that there are places in PPSSPP where constexpr can be assigned to constexpr variables, which precisely guarantees optimization at compile-time.

hrydgard:
Not a fan of how the code looks, and very marginal benefit if any. closing this one.

PR #18962, opened 2024-04-01
Fixed zeroise SHA_CTX struct, not pointer size (CWE Information Disclosure)

GermanAizek:
Maybe CWE Information Disclosure if size pointer less struct size.


hrydgard (approved):
Hardly a security issue since this is inside the context of a game executing. Still, good to make the code do what's intended.

PR #18963, opened 2024-04-01
[C++17] Recheck some memset with zeros on buffer underflow

GermanAizek:
More readable-accurate calculation size arrays filled zeros for x86, x64 and C++17 guarantee compile-time constexpr calculate size for memset


hrydgard:
I think this is less readable. sizeof does what we want here (compute the complete size of the thing being zeroed).

GermanAizek:
> I think this is less readable. sizeof does what we want here (compute the complete size of the thing being zeroed).

Then do not take this PR, I will later review all memcpy, memset, memmove where there really may be errors in sizes types or buffers.

PR #18966, opened 2024-04-02
Fix memory leak on shutdown in sceHeap

### LINKED ISSUE #18952: Potential memory leak in sceHeap

Not game-specific

CLion's code inspection noticed that the function `sceHeapCreateHeap` (defined [here](https://github.com/hrydgard/ppsspp/blob/master/Core/HLE/sceHeap.cpp)) allocates the `Heap` object with `operator new` and saves its address in the `heapList`, after which the variable that holds onto the pointer goes out of scope as the function ends. CLion thought it's a leak, but I didn't believe it and went to check how the `heapList` is managed.

1) `heapList` is a static variable and it's only available in its `*.cpp` file.
2) `getHeap` is a static function.
3) The only `operator delete` that can be used to deallocate the previously saved heap instance is located in the HLE function `sceHeapDeleteHeap`.

I don't think the emu can prevent the leak if the user reloads the game, exists the game or loads a savestate.

### What should happen

I don't think it's a very important leak (considering I couldn't remember a single game that uses `sceHeap` right away, it's not too widely used), but it's just way too blatant. Gotta fix that.

Interestingly enough, we've got 3 unimplemented functions in this module. Maybe for another time...

(By the way, please consider renaming `found` to `it` or `iter` [here](https://github.com/hrydgard/ppsspp/blob/master/Core/HLE/sceHeap.cpp#L50), because `auto found` suggests a boolean value).

### Logs

_No response_

### Platform

Windows

AMD Radeon(TM) Graphics

### PPSSPP version affected

### Last working version

_No response_

### Graphics backend (3D API)

Direct3D 11

### Checklist

hrydgard:
Fixes #18952

### DIFF
--- a/Core/HLE/sceHeap.cpp
+++ b/Core/HLE/sceHeap.cpp
@@ -47,11 +47,11 @@ struct Heap {
 static std::map<u32, Heap *> heapList;
 
 static Heap *getHeap(u32 addr) {
-	auto found = heapList.find(addr);
-	if (found == heapList.end()) {
-		return NULL;
+	auto it = heapList.find(addr);
+	if (it == heapList.end()) {
+		return nullptr;
 	}
-	return found->second;
+	return it->second;
 }
 
 void __HeapDoState(PointerWrap &p) {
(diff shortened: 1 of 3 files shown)


GermanAizek:
> Fixes #18952

I have just started fixing this issue and you are fixing it, really fast, since I do not communicate with you via social discord, I would prefer to fix old issues if possible.

PR #18967, opened 2024-04-02
[Common/File/UI] Fixed minor memleaks

GermanAizek:
@hrydgard I think it's also simple PR for code review and merge in master.

### DIFF
--- a/UI/DevScreens.cpp
+++ b/UI/DevScreens.cpp
@@ -1405,7 +1405,6 @@ void FrameDumpTestScreen::CreateViews() {
 		c->SetTag(url);
 		c->OnClick.Handle<FrameDumpTestScreen>(this, &FrameDumpTestScreen::OnLoadDump);
 	}
-	delete dumpsScroll;
 }
 
 UI::EventReturn FrameDumpTestScreen::OnLoadDump(UI::EventParams &params) {
--- a/UI/DevScreens.cpp
+++ b/UI/DevScreens.cpp
@@ -1395,7 +1395,7 @@ void FrameDumpTestScreen::CreateViews() {
 	LinearLayout *dumps = new LinearLayoutList(ORIENT_VERTICAL);
 	dumps->SetSpacing(0);
 	dumpsScroll->Add(dumps);
-	tabHolder->AddTab("Dumps", dumps);
+	tabHolder->AddTab("Dumps", dumpsScroll);
 
 	dumps->Add(new ItemHeader("GE Frame Dumps"));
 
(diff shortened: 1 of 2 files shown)


Common/Thread/ParallelLoop.cpp
```diff
@@ -106,6 +106,7 @@ void ParallelRangeLoop(ThreadManager *threadMan, const std::function<void(int, i
 	// and process the remainder right here. If there's no remainer, we'll steal a whole chunk.
 	if (counter) {
 		counter->WaitAndRelease();
+		delete counter;
```
hrydgard, line 109: WaitAndRelease performs the delete already, so this is not needed (and would be a double-free).

UI/DevScreens.cpp
```diff
@@ -1405,6 +1406,7 @@ void FrameDumpTestScreen::CreateViews() {
 		c->SetTag(url);
 		c->OnClick.Handle<FrameDumpTestScreen>(this, &FrameDumpTestScreen::OnLoadDump);
 	}
+	delete dumpsScroll;
```
hrydgard, line 1409: this is wrong. this will be deleted later, it gets added to the view hiearchy so it's being kept track of.

Or, hm, turns out it doesn't - that's a bug. We don't get a scrollview at all.

So instead, above, change the last parameter of "tabHolder->addTab("Dumps", dumps)" to dumpsScroll.

UI/DevScreens.cpp
```diff
@@ -845,6 +845,7 @@ void SystemInfoScreen::CreateTabs() {
 	LinearLayout *internals = AddTab("DevSystemInfoInternals", si->T("Internals"));
 	CreateInternalsTab(internals);
 #endif
+	delete tabHolder;
```
hrydgard, line 848: Like the one below, this is a different kind of bug. I'll fix it separately, you can delete this line.

UI/DevScreens.cpp
GermanAizek: > Or, hm, turns out it doesn't - that's a bug. We don't get a scrollview at all.
> 
> So instead, above, change the last parameter of "tabHolder->addTab("Dumps", dumps)" to dumpsScroll.

you have eye eagle, you noticed other mistakes through my PR.

Common/Thread/ParallelLoop.cpp
GermanAizek: fix it https://github.com/hrydgard/ppsspp/pull/18967/commits/9c3c23daa6d3a70921f3aeb3560c81b497065eb4

UI/DevScreens.cpp
GermanAizek: fix it https://github.com/hrydgard/ppsspp/pull/18967/commits/9c3c23daa6d3a70921f3aeb3560c81b497065eb4

UI/DevScreens.cpp
GermanAizek: fix it https://github.com/hrydgard/ppsspp/pull/18967/commits/9c3c23daa6d3a70921f3aeb3560c81b497065eb4

UI/DevScreens.cpp
hrydgard: This line still needs to be removed - as now the object is properly in the view hiearchy and will be deleted later.

PR #18970, opened 2024-04-02
Cleanup excess redefines and maybe fixed result in HLE sceIo::sceIoGetDevType()

GermanAizek:
There are questions about offset and sceIoGetDevType, as far as I have fixed correctly.


hrydgard:
I restored the offset thing but made it clearer.

PR #18972, opened 2024-04-02
Fixed to free() C version, since memory block was allocated using malloc()

GermanAizek:
@hrydgard,
Memory was allocated using malloc, as far as it is correct to apply C++ new and delete to these memory blocks. I didn't think about it and never did it, but most people on Internet forums write that you can't do that, maybe it's UB, or memory just won't clear.

Example discussion: https://stackoverflow.com/questions/10854210/behaviour-of-malloc-with-delete-in-c


hrydgard (approved):
yeah, good find.

PR #18973, opened 2024-04-02
Fixed return copies from functions and added const ref params

GermanAizek:
Most likely, const qualifiers were added by accident, I think it was meant that it was necessary to move object and not copy.

More info: https://stackoverflow.com/questions/57761077/return-const-value-prevent-move-semantics


hrydgard:
It's correct that const on return values is entirely unnecessary here, and the rest looks safe and good.

PR #18974, opened 2024-04-02
Fixed possible accidental conversion memsize type to strict 32 bit

GermanAizek:
@hrydgard, may have forgotten to delete conversion when there were old parameters functions with integer type.


hrydgard (approved):
This looks fine.

BTW, I assume you're working from the output of a static analyzer? Which one?

GermanAizek:
> BTW, I assume you're working from the output of a static analyzer? Which one?

if you meant this PR, then these are usual warnings Clang compiler.

PR #18975, opened 2024-04-02
Protect against undefined behavior when bitwise shift occurs with negative value

GermanAizek:
Can such case happen if values come in negative from GetNumVectorElements() function or from encodedSize(Size)?


hrydgard:
None of these functions can ever return a negative number, so not much of an issue in practice.. However, it does make things clearer to use an unsigned type here, so I'll merge.

PR #18980, opened 2024-04-03
memoryAreas knows size, containers haven replaced with more optimized and reserve()

GermanAizek:
@hrydgard, I think fixes are useful for improving code generation by compiler, especially since it affects arm64jit and x64jit where it is more important.

With regard to container replacement, it is assumed that it is more optimized when used in this section code, but if there is search or index access or other operations, container will have to be changed to a more suitable one.

### DIFF
--- a/Common/Math/expression_parser.cpp
+++ b/Common/Math/expression_parser.cpp
@@ -407,6 +407,7 @@ bool parsePostfixExpression(PostfixExpression& exp, IExpressionFunctions* funcs,
 	size_t num = 0;
 	uint32_t opcode;
 	std::vector<uint32_t> valueStack;
+	valueStack.reserve(exp.size());
 	unsigned int arg[5]{};
 	float fArg[5]{};
 	bool useFloat = false;
--- a/Core/HLE/sceGe.cpp
+++ b/Core/HLE/sceGe.cpp
@@ -228,7 +228,7 @@ void __GeDoState(PointerWrap &p) {
 	if (s >= 2) {
 		Do(p, ge_pending_cb);
 	} else {
-		std::list<GeInterruptData_v1> old;
+		std::vector<GeInterruptData_v1> old;
 		Do(p, old);
 		ge_pending_cb.clear();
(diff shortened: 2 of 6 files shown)


Nemoumbra:
You say it like it's a well-known fact that `deque` is better than the `vector`. I find it confusing... What static code analysis program do you use for finding the places to improve in your PRs? I would really like to know why it thinks the deque would work better here.

GermanAizek:
> You say it like it's a well-known fact that `deque` is better than the `vector`. I find it confusing... What static code analysis program do you use for finding the places to improve in your PRs? I would really like to know why it thinks the deque would work better here.

More info about this situation: https://cengizhanvarli.medium.com/when-should-we-std-deque-instead-of-std-vector-97971028764e
In same place, code shows that container only has an insertion call `push_back()` and getting container size, in this context it means that this is micro-optimization for compiler. And this is not a drastic change in performance, because there are not many calls and this new list vector is used only once.

Nemoumbra:
Not the best article to mention... considering it's poorly written... Even if we forget the broken English, they make some odd assertions...
> Vector elements are placed in contiguous storage so that they can be accessed and traversed using iterators

This sounds like we need the contiguous storage to implement iterators... The point of iterators is to traverse the container with a simple universal syntax no matter how complex the container is.
They also say the deque is usually a linked list of chunks and if one wants to index it, the deque navigates through the chunks until it finds the one it needs. This is not how `std::deque` works, because that's not `O(1)` for the `operator[]`.
The article doesn't take the cache-friendliness into account at all.

Lastly, the point of the article is to show there is a container that extends the vectors's interface with efficient insertions to the front and **if** you need that, you should use the deque and not the vector. Now please tell me... What does it have to do with our use case? I think you just linked some unrelated stuff.

hrydgard:
If a workload suits a vector, which means not a lot of dynamic insertions and deletions all over, only pushing/popping at the end, a deque has no advantage, only disadvantages.

hrydgard:
Simliar to many of the others, there's just no benefit to any of these changes. Closing.

PR #18981, opened 2024-04-03
[GPU/System/Thread/Core/FS] Simplify loops

GermanAizek:
this is more code refactor than valid optimization.

### DIFF
--- a/Core/Config.cpp
+++ b/Core/Config.cpp
@@ -1514,16 +1514,12 @@ void Config::RemoveRecent(const std::string &file) {
 
 	private_->ResetRecentIsosThread();
 	std::lock_guard<std::mutex> guard(private_->recentIsosLock);
-	const std::string filename = File::ResolvePath(file);
-	for (auto iter = recentIsos.begin(); iter != recentIsos.end();) {
-		const std::string recent = File::ResolvePath(*iter);
-		if (filename == recent) {
-			// Note that the increment-erase idiom doesn't work with vectors.
-			iter = recentIsos.erase(iter);
-		} else {
-			iter++;
-		}
-	}
+	
+	const auto &filename = File::ResolvePath(file);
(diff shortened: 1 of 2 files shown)


hrydgard:
Flipping ++ / -- to the start has zero effect on performance, as far as I know it's an old myth that prefixing them would be faster. I would prefer to not do those changes.

The rest is fine.

GermanAizek:
fix it https://github.com/hrydgard/ppsspp/pull/18981/commits/f9777ceced95916e20b5a5f4dda3eb71fba8dd02

hrydgard (approved):

Core/Config.cpp
```diff
@@ -1514,16 +1514,12 @@ void Config::RemoveRecent(const std::string &file) {
...
-		if (filename == recent) {
-			// Note that the increment-erase idiom doesn't work with vectors.
-			iter = recentIsos.erase(iter);
-		} else {
-			iter++;
-		}
-	}
+	
+	std::remove_if(recentIsos.begin(), recentIsos.end(), [file](const auto &str) {
+		const auto &filename = File::ResolvePath(file);
```
hrydgard, line 1519: I missed this before, but there's no need to resolve the file path over and over, this can be done outside the loop like before.

hrydgard:
Can you fix this last thing I commented above? Then this can go in.

Core/Config.cpp
GermanAizek: > I missed this before, but there's no need to resolve the file path over and over, this can be done outside the loop like before.

Yes, you're right, for some reason I stuffed it inside

GermanAizek:
done https://github.com/hrydgard/ppsspp/pull/18981/commits/4b92848bdcb0291a0e9fc01b8891f6bfc432077a

hrydgard (approved):
The remove_if is a nice change. I should use that more.

PR #18987, opened 2024-04-03
Update rcheevos, switch back to upstream. 

hrydgard:
Avoid using rcheevos' hash.c, by computing game hashes ourselves.

### DIFF
--- a/Core/RetroAchievements.cpp
+++ b/Core/RetroAchievements.cpp
@@ -4,6 +4,14 @@
 // Derived from Duckstation's RetroAchievements implementation by stenzek as can be seen
 // above, relicensed to GPL 2.0.
 
+// Actually after the rc_client rewrite, barely anything of that remains.
+
+// The PSP hash function:
+// md5_init
+// md5_hash(PSP_GAME/PARAM.SFO)
+// md5_hash(PSP_GAME/EBOOT.BIN)
+// hash = md5_finalize()
+
 #include <algorithm>
 #include <atomic>
 #include <cstdarg>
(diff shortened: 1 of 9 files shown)


Core/RetroAchievements.cpp
```diff
@@ -51,8 +58,59 @@
...
+		return false;
+	}
+
+	uint32_t sz = fs->SeekFile(handle, 0, FILEMOVE_END);
+	fs->SeekFile(handle, 0, FILEMOVE_BEGIN);
+	if (!sz) {
+		return false;
+	}
+
+	std::unique_ptr<uint8_t[]> buffer(new uint8_t[sz]);
```
Nemoumbra, line 76: How about
`auto buffer = std::make_unique<uint8_t[]>(sz);`
?

Core/RetroAchievements.cpp
```diff
@@ -51,8 +58,59 @@
...
+}
+
+// Consumes the blockDevice.
+std::string ComputePSPHash(BlockDevice *blockDevice) {
+	md5_context md5;
+	ppsspp_md5_starts(&md5);
+
+	SequentialHandleAllocator alloc;
+	{
+		std::unique_ptr<ISOFileSystem> fs(new ISOFileSystem(&alloc, blockDevice));
```
Nemoumbra, line 93: `auto fs = std::make_unique<ISOFileSystem>(&alloc, blockDevice);`
?

Core/RetroAchievements.cpp
hrydgard: I'll sneak this into an upcoming change, thanks.

PR #18989, opened 2024-04-03
[IRJit] Excessive precision was removed by changing double to float

GermanAizek:
Floating-point numbers are executed much faster than exact ones with double type, this can greatly affect performance weak devices. If accuracy is really not important.
More info: https://stackoverflow.com/a/4584666


hrydgard:
So, on modern CPUs that is not accurate - double is just a little slower, much of it because of higher memory bandwidth requirements. Also, this is nowhere near any kind of hot path, this is just for computing statistics, which is not done during gameplay. It's preferable to have easy code with no worry about running out of precision over speed here.

PR #18990, opened 2024-04-03
[HLE] sceDisplay: maybe necessary to increase accuracy timePerVblank is defined as double

GermanAizek:
Does such precision need to be used for timePerVblank? if don't need it, can convert it to float type.


hrydgard:
I don't think this matters at all. I'll clean it up as a part of another change later.

PR #18991, opened 2024-04-03
[HLE/GPU/UI/IR] Replaced push and insert on emplace C++11 and try_emplace C++17

GermanAizek:
@hrydgard, it probably only affects on Debug configuration performance, but it takes up less space if you read.

More info about try_emplace: https://stackoverflow.com/questions/46046828/is-there-any-reason-to-use-stdmapemplace-instead-of-try-emplace-in-c17

### DIFF
--- a/GPU/Common/TextureShaderCommon.cpp
+++ b/GPU/Common/TextureShaderCommon.cpp
@@ -236,7 +236,7 @@ Draw2DPipeline *TextureShaderCache::GetDepalettizeShader(uint32_t clutMode, GETe
 std::vector<std::string> TextureShaderCache::DebugGetShaderIDs(DebugShaderType type) {
 	std::vector<std::string> ids;
 	for (auto &iter : depalCache_) {
-		ids.push_back(StringFromFormat("%08x", iter.first));
+		ids.emplace_back("%08x", iter.first);
 	}
 	return ids;
 }
--- a/Common/Data/Format/JSONWriter.cpp
+++ b/Common/Data/Format/JSONWriter.cpp
@@ -19,17 +19,17 @@ JsonWriter::~JsonWriter() {
 
 void JsonWriter::begin() {
 	str_ << "{";
-	stack_.push_back(StackEntry(DICT));
+	stack_.emplace_back(DICT);
 }
 
 void JsonWriter::beginArray() {
 	str_ << "[";
-	stack_.push_back(StackEntry(ARRAY));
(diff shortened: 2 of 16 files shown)


Nemoumbra:
I vote for the `push_back(Object(/* args */))` -> `emplace_back(/* args */)`, but... Why are you trying to introduce `try_emplace` again, after you were told not to in https://github.com/hrydgard/ppsspp/pull/18576?

GermanAizek:
> I vote for the `push_back(Object(/* args */))` -> `emplace_back(/* args */)`, but... Why are you trying to introduce `try_emplace` again, after you were told not to in #18576?

yes, u're right, I probably forgot that I already tried to offer to u.

GermanAizek:
@Nemoumbra fix it https://github.com/hrydgard/ppsspp/pull/18991/commits/b951e7b0b68b6738e2beb38fcc8565e82237ea4a

GPU/Common/TextureShaderCommon.cpp
```diff
@@ -236,7 +236,7 @@ Draw2DPipeline *TextureShaderCache::GetDepalettizeShader(uint32_t clutMode, GETe
 std::vector<std::string> TextureShaderCache::DebugGetShaderIDs(DebugShaderType type) {
 	std::vector<std::string> ids;
 	for (auto &iter : depalCache_) {
-		ids.push_back(StringFromFormat("%08x", iter.first));
```
Nemoumbra, line 239: How does this even compile?? You're pushing a compile-time string literal + the iterator into a vector of `std::strings`!

GermanAizek: > How does this even compile?? You're pushing a compile-time string literal + the iterator into a vector of `std::strings`!

ha ha ha, you're really right, it's amazing how my compiler didn't mention this, as well as all the others. It's good that all contributors can keep track where and how errors are hidden when reviewing code. It's pity that UB is so impossible to identify.

Nemoumbra: Ok, why did I say it's the iterator...? It's `iter.first`, not `iter`. That makes more sense - the compiler used some valid `std::string` constructor, probably this one:

image


And no, it's not UB, that's a logic error.

hrydgard:
`emplace_back` is nice for just a little extra performance in some cases, but not really in any of these, it's all very small and simple objects and I think `push_back` with the struct name visible is clearer to read. So I'm choosing not to merge.

PR #18993, opened 2024-04-04
Remove excess const references which will only reduce performance

GermanAizek:
More info: https://stackoverflow.com/a/24612241


hrydgard:
Too risky, not enough benefit, sorry. Closing.

PR #18995, opened 2024-04-04
[Core/GPU] Fixed incorrect bitwise shift when expanding to 64 bytes and type overflow

GermanAizek:
More info: https://stackoverflow.com/questions/14846566/how-to-shift-32-bits-in-uint64-t

### DIFF
--- a/Common/Arm64Emitter.cpp
+++ b/Common/Arm64Emitter.cpp
@@ -4295,7 +4295,7 @@ bool ARM64FloatEmitter::TryAnyMOVI(u8 size, ARM64Reg Rd, uint64_t elementValue) 
 
 	uint64_t value = elementValue;
 	if (size != 64) {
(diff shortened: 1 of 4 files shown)


hrydgard:
The stuff in [MIPSVFPUUtils.cpp](https://github.com/hrydgard/ppsspp/pull/18995/files#diff-4846e6b68dbee946c1a2c97c0c345ed3b3f708b30915fecf5dde46db753ff1c2) has been carefully checked to match hardware closely by @fp64 who wrote it. There is no point in meddling there, what's there is correct.

The one worthwhile change here seems to be the one in ARM64Emitter - it is a bug for sure, just not sure we actually hit it in practice, but worth fixing.

hrydgard:
Fixed the emitter fix to compile and merged it separately. Closing.

PR #18996, opened 2024-04-04
[GPU/HLE/HW/IR] Code condition blocks clean up

GermanAizek:
Nothing special.


hrydgard:
This type of change is very dangerous because what if something not-super-visible mutates the checked variable in the first block?  Like, a ref argument to a function call. 

I'm not seeing any immediately, but makes me hesistant. We had a similar thing a while back where you moved things into blocks, and it broke stuff.

This is also super-low-benefit, like many of your recent changes. 

Can I ask you to do less of these pure cleanups, and focus on things that seem to be actual bugs?

PR #18999, opened 2024-04-04
Reordered priority initialize class fields and removed unused int index

GermanAizek:
More info: https://stackoverflow.com/a/58212322


hrydgard:
I don't see the point, this doesn't change anything.

I'll delete that extra `int index;` separately.

PR #19002, opened 2024-04-05
Add initial RAIntegration support through rc_client

hrydgard:
[RAIntegration](https://github.com/RetroAchievements/RAIntegration) is the (unfortunately Windows-only) toolkit for developing achievements for [RetroAchievements](https://www.ppsspp.org/docs/reference/retro-achievements/).

[rc_client]() is right now getting [support](https://github.com/RetroAchievements/rcheevos/wiki/rc_client_raintegration) for direct RAIntegration with only a small amount of extra code, so that's what we're using here. The required DLL is not yet released, so this is not directly usable.

Only enabled for Win32/Win64 builds (RAIntegration only supports Windows, and doesn't yet release officially for ARM windows, as far as I can tell).

Extended the return value of System_GetPropertyInt to int64_t since it's now used to pass pointer-sized things (HWND of main window).

This requires the upcoming `RAIntegration-x64.dll` version 1.3. Just plonk it in the same directory where you have PPSSPP. Also, you must enable RAIntegration support on the Developer tab in the RetroAchievement settings in PPSSPP and restart.

Part of #17631

### DIFF
--- a/Core/RetroAchievements.cpp
+++ b/Core/RetroAchievements.cpp
@@ -25,6 +25,7 @@
 
 #include "ext/rcheevos/include/rcheevos.h"
 #include "ext/rcheevos/include/rc_client.h"
+#include "ext/rcheevos/include/rc_client_raintegration.h"
 #include "ext/rcheevos/include/rc_api_user.h"
 #include "ext/rcheevos/include/rc_api_info.h"
 #include "ext/rcheevos/include/rc_api_request.h"
@@ -38,8 +39,6 @@
 #include "Common/Log.h"
 #include "Common/File/Path.h"
 #include "Common/File/FileUtil.h"
-#include "Core/FileLoaders/LocalFileLoader.h"
-#include "Core/FileSystems/BlockDevices.h"
 #include "Common/Net/HTTPClient.h"
 #include "Common/System/OSD.h"
 #include "Common/System/System.h"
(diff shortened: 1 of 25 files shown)


Core/RetroAchievements.cpp
```diff
@@ -461,6 +469,93 @@ static void login_token_callback(int result, const char *error_message, rc_clien
...
+	case RC_MISSING_VALUE:
+		g_OSD.Show(OSDType::MESSAGE_WARNING, ac->T("RAIntegration is enabled, but RAIntegration-x64.dll was not found."));
+		break;
+	default:
+		{
+			// If not successful, just report the error and bail. Shouldn't happen.
+			if (result != RC_OK) {
+				g_OSD.Show(OSDType::MESSAGE_ERROR, StringFromFormat("RAIntegration init failed: %s", error_message));
+				return;
+			}
```
Jamiras, line 536: With a switch statement, you should just have lines 534-535 in the default clause and use RC_OK for 538+

Core/RetroAchievements.cpp
```diff
@@ -477,14 +572,27 @@ void Initialize() {
...
+	if (g_Config.bAchievementsEnableRAIntegration) {
+		wchar_t szFilePath[MAX_PATH];
+		GetModuleFileNameW(NULL, szFilePath, MAX_PATH);
+		for (int64_t i = wcslen(szFilePath) - 1; i > 0; i--) {
+			if (szFilePath[i] == '\\') {
+				szFilePath[i] = '\0';
+			}
+		}
+		HWND hWnd = (HWND)System_GetPropertyInt(SYSPROP_MAIN_WINDOW_HANDLE);
+		rc_client_begin_load_raintegration(g_rcClient, szFilePath, hWnd, "PPSSPP", "1.0", &load_integration_callback, hWnd);
```
Jamiras, line 592: "1.0" should be replaced with an actual version constant. The client name and version constant will be combined and used as the UserAgent for requests not being redirected through the HTTP handler (like editing code notes).

PR #19005, opened 2024-04-05
Make const and static class members, const pointers and using static access

GermanAizek:
@hrydgard,
If you like it so much, I've looked through most of classes for possible changes.


hrydgard:
There's a lot in here that's good but also a bunch of things that aren't really worth it, I'm gonna make a mini version of this and then close it.

hrydgard:
#19468 replaces this, closing.

PR #19006, opened 2024-04-05
[Core/HLE/Dialog] Fixed memleak after first return error and increase lifetime in current scope

GermanAizek:


Core/HLE/sceFont.cpp
```diff
@@ -1123,15 +1122,14 @@ static u32 sceFontOpenUserFile(u32 libHandle, const char *fileName, u32 mode, u3
...
 
-	Font *f = new Font(buffer);
+	auto f = std::make_unique<Font>(buffer);
 	FontOpenMode openMode = mode == 0 ? FONT_OPEN_USERFILE_HANDLERS : FONT_OPEN_USERFILE_FULL;
-	LoadedFont *font = fontLib->OpenFont(f, openMode, *errorCode);
+	LoadedFont *font = fontLib->OpenFont(f.get(), openMode, *errorCode);
 	if (font) {
 		*errorCode = 0;
 		return hleLogSuccessInfoX(SCEFONT, font->Handle());
 	}
```
hrydgard, line 1132: This and the other font parts are wrong, on success, we need the font object to live on. The fontlib takes ownership.

hrydgard (commented):
Some of the things in here are incorrect (the font ones), and I don't see a single likely actual leak being fixed by these changes. I'm closing the PR.

PR #19025, opened 2024-04-10
[Core/GPU] Rewrited loops to C++17 range-based and modernize std::bind() to lambda

GermanAizek:
@hrydgard, I think you'll like it.

### DIFF
--- a/Core/Debugger/DisassemblyManager.cpp
+++ b/Core/Debugger/DisassemblyManager.cpp
@@ -403,9 +403,9 @@ void DisassemblyManager::clear()
 {
 	auto memLock = Memory::Lock();
 	std::lock_guard<std::recursive_mutex> guard(entriesLock_);
-	for (auto it = entries.begin(); it != entries.end(); it++)
+	for (auto &ent : entries)
 	{
-		delete it->second;
+		delete ent.second;
 	}
 	entries.clear();
 }
@@ -496,11 +496,9 @@ void DisassemblyFunction::getBranchLines(u32 start, u32 size, std::vector<Branch
 	u32 end = start+size;
 
 	std::lock_guard<std::recursive_mutex> guard(lock_);
-	for (size_t i = 0; i < lines.size(); i++)
+	for (auto &line : lines)
 	{
-		BranchLine& line = lines[i];
-
-		u32 first = line.first;
+        u32 first = line.first;
 		u32 second = line.second;
 
 		// skip branches that are entirely before or entirely after the window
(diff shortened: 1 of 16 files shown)


GPU/Common/TextureScalerCommon.cpp
```diff
@@ -698,22 +698,22 @@ const int MIN_LINES_PER_THREAD = 4;
 
 void TextureScalerCommon::ScaleXBRZ(int factor, u32* source, u32* dest, int width, int height) {
 	xbrz::ScalerCfg cfg;
-	ParallelRangeLoop(&g_threadManager, std::bind(&xbrz::scale, factor, source, dest, width, height, xbrz::ColorFormat::ARGB, cfg, std::placeholders::_1, std::placeholders::_2), 0, height, MIN_LINES_PER_THREAD);
+	ParallelRangeLoop(&g_threadManager, [factor, source, dest, width, height, cfg](auto && PH1, auto && PH2) { return xbrz::scale(factor, source, dest, width, height, xbrz::ColorFormat::ARGB, cfg, std::forward<decltype(PH1)>(PH1), std::forward<decltype(PH2)>(PH2)); }, 0, height, MIN_LINES_PER_THREAD);
```
hrydgard, line 701: This doesn't look very good. Does this have any huge advantage over just using std::bind ?

Core/Debugger/DisassemblyManager.cpp
```diff
@@ -555,12 +553,12 @@ void DisassemblyFunction::generateBranchLines()
...
-	for (size_t i = 0; i < lines.size(); i++)
+	for (auto &line : lines)
 	{
-		for (int l = 0; l < NUM_LANES; l++)
+		for (auto &lain : lanes)
 		{
-			if (lines[i].first > lanes[l].end)
-				lanes[l].used = false;
+			if (line.first > lain.end)
+                lain.used = false;
```
hrydgard, line 561: indentation is wrong

hrydgard (changes requested):
It seems once again you haven't tested. Those any_of loops with returns in them will not work as expected.

GPU/Common/TextureScalerCommon.cpp
GermanAizek: > This doesn't look very good. Does this have any huge advantage over just using std::bind ?

All answers in https://stackoverflow.com/questions/24852764/stdbind-vs-lambda-performance

hrydgard:
The speedups shown in the linked example won't happen in most of these cases - you'll only get such overhead from bind if you do a really small amount of work every iteration, and in the performance critical loops here, we do a lot of work per iteration (texture scaling) so won't be noticable. Not worth it, and since you obviously did not test this and check that it did the same thing as before, I'm closing.

PR #19027, opened 2024-04-10
Minor optimize methods std::string using param as char with simpler implementation

GermanAizek:
Example compare algorithm functions as in screenshot.

optimize-char-as-parameter-for-std-string-methods


hrydgard (approved):

hrydgard:
not much of a win, but correct, clean and safe. i'm ok with it.

PR #19028, opened 2024-04-10
More readable methods std::string and fix repeat calling methods

GermanAizek:
I think it's a good refactoring that doesn't break and reduces size code functions.

### DIFF
--- a/Common/GPU/OpenGL/thin3d_gl.cpp
+++ b/Common/GPU/OpenGL/thin3d_gl.cpp
@@ -406,15 +406,16 @@ public:
 		stencilWriteMask_ = writeMask;
 		stencilCompareMask_ = compareMask;
 		// Do we need to update on the fly here?
+		const auto depthStencil = curPipeline_->depthStencil;
 		renderManager_.SetStencil(
-			curPipeline_->depthStencil->stencilEnabled,
-			curPipeline_->depthStencil->stencilCompareOp,
+			depthStencil->stencilEnabled,
+			depthStencil->stencilCompareOp,
 			refValue,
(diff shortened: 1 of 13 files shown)


UI/OnScreenDisplay.h
```diff
@@ -58,7 +58,7 @@ enum class NoticeLevel {
 class NoticeView : public UI::InertView {
 public:
 	NoticeView(NoticeLevel level, std::string_view text, std::string_view detailsText, UI::LayoutParams *layoutParams = 0)
-		: InertView(layoutParams), level_(level), text_(text), detailsText_(detailsText), iconName_("") {}
+		: InertView(layoutParams), level_(level), text_(text), detailsText_(detailsText), iconName_() {}
```
hrydgard, line 61: can even skip iconName_ here, it defaults to empty

Common/GPU/OpenGL/thin3d_gl.cpp
```diff
@@ -406,15 +406,16 @@ class OpenGLContext : public DrawContext {
 		stencilWriteMask_ = writeMask;
 		stencilCompareMask_ = compareMask;
 		// Do we need to update on the fly here?
+		const auto depthStencil = curPipeline_->depthStencil;
```
hrydgard, line 409: I think whenever it's not too hard, let's be explicit about the type for these instead of using auto. Clearer and doesn't really take up a lot of code space.

hrydgard (changes requested):

hrydgard:
Decided not to merge this one, it's not really worth it.

PR #19029, opened 2024-04-10
[Net] Reduced width line an simplify asm instruction on debug configuration

GermanAizek:
Rather, it refers to code refactor, I think modern compilers even on debug optimize such design into more suitable one.

### DIFF
--- a/Common/Net/Resolve.cpp
+++ b/Common/Net/Resolve.cpp
@@ -269,7 +269,7 @@ int inet_pton(int af, const char* src, void* dst)
 			else if (ch < '0' || ch > '9')
 				return 0;
 			else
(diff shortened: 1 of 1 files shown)


Common/Net/Resolve.cpp
```diff
@@ -269,7 +269,7 @@ int inet_pton(int af, const char* src, void* dst)
 			else if (ch < '0' || ch > '9')
 				return 0;
 			else
-				x = x * 10 + ch - '0';
+				x *= 10 + ch - '0';
```
hrydgard, line 272: Come on now. This is just wrong.

This means `x = x * (10 + ch - '0')`, not `x = x * 10 + ch - '0';`

Again, please test your changes.

PR #19031, opened 2024-04-10
[Common/Core/GPU] Set strict keyword default C++11 for default (con/de)structors

GermanAizek:
Strict keyword definition C++11 to modern compiler.

More info: https://stackoverflow.com/questions/20828907/the-new-syntax-default-in-c11


hrydgard:
Sorry, I don't like this. I think it's less clear.

PR #19032, opened 2024-04-11
Fixed conditions always false using unsigned type and logical AND 0x0 & 0x4

GermanAizek:


Common/Buffer.cpp
```diff
@@ -111,9 +111,6 @@ void Buffer::Printf(const char *fmt, ...) {
 		// Output was truncated. TODO: Do something.
 		ERROR_LOG(IO, "Buffer::Printf truncated output");
 	}
-	if (retval < 0) {
```
anr2me, line 114: This is to catch encoding error during `vsnprintf` isn't?
This should probably be: `if ((int)retval < 0) {` instead of removing it, since `vsnprintf` returns an integer.

GermanAizek: fix it here: https://github.com/hrydgard/ppsspp/pull/19032/commits/8e580045cd423b6cfff0a2c30e53e32f675cf9e8

anr2me: i'm not even sure whether the conversion from `int` to `size_t` and back to `int` will still have the correct sign on 64-bit CPU, since they have different bit size.

hrydgard (approved):

PR #19033, opened 2024-04-11
Break out the Atrac3/Atrac3+ decoders from FFMPEG to a separate library

hrydgard:
Followup to #19026

I grabbed a chunk of ffmpeg, then pared it down to only contain the decoders themselves, plus some math utils and stuff.

This I put in /ext/at3_standalone, and it's exposed by a very simple interface in /ext/at3_standalone/at3_decoders.h

Also, we now use this instead of FFMPEG for decoding music in the menu, and wherever SimpleDecoder was previously used for AT3 playback (sceAudiocodec, MediaEngine video playback).

It's not yet used for background music in-game, porting sceAtrac to use this instead of ffmpeg will be a bigger job. But worthwhile, since we'll be able to debug much easier now.

This is one more step towards getting rid of our huge pre-built FFMPEG as a dependency. Much work remains. Also this isn't yet using SIMD so might be slightly slower than the ffmpeg build.

EDIT: Profiled a little bit, it's not really substantially slower. The functions `decode_qu_spectra` and `ff_atrac3p_ipqf` are the hotspot and the former is a huffman decoder, so not very easy to optimize. The latter though is very SIMD-able.


hrydgard:
Noticed a crash in crisis core, converting to draft.

LunaMoo:
Is the same planned to encoders?(Recording feature)
I guess the plan is to unify experience and stop linux users break theirs by using system ffmpeg^_^;.

hrydgard:
Encoders no, I think we'll still gonna use ffmpeg for that, or OS-native encoders - or remove encoding and leave it to OBS, although there are some arguments that we shouldn't - no decision there

And yes, getting rid of bugs and bug reports caused by weird combinations of system ffmpegs is definitely a big goal! But the biggest goal is make debugging and working on decoder stuff easier.

PR #19034, opened 2024-04-11
Unsigned char pos always >= 0, removed excess condition

GermanAizek:
@hrydgard, also minor, rather code cleanup.


hrydgard:
I kinda like the consistency with the other components above, the compiler will remove it anyway. Closing.

PR #19035, opened 2024-04-11
[GPU/HLES] 'fs' its already NULL, removed excess delete and bufferState_ always ATRAC_STATUS_HALFWAY_BUFFER

GermanAizek:


hrydgard (approved):
These are all correct.

PR #19036, opened 2024-04-11
Safer checking pointers before use, as well check malloc result

GermanAizek:


Common/Data/Format/PNGLoad.cpp
```diff
@@ -27,6 +27,13 @@ int pngLoad(const char *file, int *pwidth, int *pheight, unsigned char **image_d
 
 	int stride = PNG_IMAGE_ROW_STRIDE(png);
 	*image_data_ptr = (unsigned char *)malloc(PNG_IMAGE_SIZE(png));
+	if (!image_data_ptr)
+	{
+		WARN_LOG(IO, "pngLoad: %s (%s)", png.message, file);
+		*image_data_ptr = nullptr;
```
hrydgard, line 33: What's this supposed to do? This seems to be a write to null.

Common/Data/Format/PNGLoad.cpp
```diff
@@ -56,6 +63,13 @@ int pngLoadPtr(const unsigned char *input_ptr, size_t input_len, int *pwidth, in
 	}
 
 	*image_data_ptr = (unsigned char *)malloc(size);
+	if (!image_data_ptr)
+	{
+		ERROR_LOG(IO, "pngLoad: malloc image_data_ptr failed");
+		*image_data_ptr = nullptr;
```
hrydgard, line 69: This too. will crash.

hrydgard (changes requested):

hrydgard:
Closing this one due to no response to feedback.

PR #19037, opened 2024-04-11
Fixed __KernelCheckResumeMsgPipeReceive and __KernelCheckResumeMsgPipeSend not return result back to calling function

GermanAizek:
Maybe I misunderstood something and then I just had to cut out the parameter?

https://github.com/hrydgard/ppsspp/blob/bab6c5997b12df9f7b65bb367255b63057598947/Core/HLE/KernelWaitHelpers.h#L231-L238


hrydgard:
The result is already passed along in waitInfo.Complete(m->GetUID(), result); 

I don't think anything else is needed.

PR #19040, opened 2024-04-11
[Core/GPU] Fixed condition blocks where result always true

GermanAizek:


Core/Util/DisArm64.cpp
```diff
@@ -477,7 +477,7 @@ static void DataProcessingRegister(uint32_t w, uint64_t addr, Instruction *instr
 		int imm3 = (w >> 10) & 0x7;
 		if (Rd == 31 && sub && S) {
 			// It's a CMP
-			snprintf(instr->text, sizeof(instr->text), "%s%s %c%d, %c%d, %s", "cmp", S ? "s" : "", r, Rn, r, Rm, extendnames[option]);
```
anr2me, line 480: just wondering, how did a conditional `S` became a constant "s"? 
Since `w` is a type of `uint32_t` that came from argument and `bool S = (w >> 29) & 1;` Which part of the code causing `S` to have a fixed/constant value of `true`?

GermanAizek:
This PR is waiting for actions.

hrydgard (approved):

Issue #19041, opened 2024-04-11
Invalid opcode or SetConst dst


GermanAizek:
@hrydgard,

Here `IROp::SetConstF` or `IROp::SetConst`,
and
`out.AddConstant(gpr.GetImm(inst.src1))` or `out.AddConstantFloat(gpr.GetImm(inst.src1))`

https://github.com/hrydgard/ppsspp/blob/bab6c5997b12df9f7b65bb367255b63057598947/Core/MIPS/IR/IRPassSimplify.cpp#L607-L614

and same here:

https://github.com/hrydgard/ppsspp/blob/bab6c5997b12df9f7b65bb367255b63057598947/Core/MIPS/IR/IRPassSimplify.cpp#L2005-L2014

GermanAizek:
Im fixed it here: https://github.com/hrydgard/ppsspp/pull/19042

hrydgard:
Did you test that this fixes anything?

I'm pretty sure this stuff is actually right, the first one is a move from a GPR (general purpose register) to an FPR (floating point registers), which looks a bit odd but it's correct.

The second one just takes a float value that happens to be stored in an integer, and sets a float register to its contents. It's also correct.

GermanAizek:
> Did you test that this fixes anything?
> 
> I'm pretty sure this stuff is actually right, the first one is a move from a GPR (general purpose register) to an FPR (floating point registers), which looks a bit odd but it's correct.
> 
> The second one just takes a float value that happens to be stored in an integer, and sets a float register to its contents. It's also correct.

so maybe it's really easier to combine Write with all enums into one common function?
so less code comes out to call Write function.

PR #19042, opened 2024-04-11
Fixed invalid setconstfloat and rewrited to constexpr template (this should count typos in future)

GermanAizek:
It PR fixes this issue: https://github.com/hrydgard/ppsspp/issues/19041


hrydgard:
No thank you, sorry. I don't see how this is an improvement, especially since #19041 isn't even buggy.

PR #19045, opened 2024-04-12
[Common/Core/Windows] Removed excess check pointer before delete or free()

GermanAizek:
This method rechecking ptr came from C, now it doesn't make sense in C++, you can make cleaner and less code.
More info in this answer: https://stackoverflow.com/a/615408


anr2me:
Probably to avoid deleting/freeing an already freed/deleted pointer, if there are another clean up code somewhere else that set the pointer to null (since delete/free doesn't set the pointer to null automatically). 

Also, if it involves a PSP syscall, sometime games could called somethingTerm syscall multiple times (either due to a bug or unexpected returned value causing the game to behave strangely), which could attempt to clean up something more than once.

So, you will need to test this changes to make sure it doesn't cause a crash due to an attempt to free/delete an already freed/deleted allocation.

anr2me:
Ah right, if there are codes that intentionally set those pointer to NULL/nullptr/0 after freeing/deleting them somewhere else, it should be safe to delete/free them again without checking.
So, this changes should be okay i guess.

GermanAizek:
@anr2me @hrydgard,
Do I understand correctly that I need to cut out entire atrac in PR?
https://github.com/hrydgard/ppsspp/pull/19045/conflicts

hrydgard:
I'll resolve it.

hrydgard:
Replaced by #19467

PR #19047, opened 2024-04-12
Reduced struct sizes using align for x64 platforms, fields carefully moved

GermanAizek:
Tested only on Debug

Before RAM util (master):

image

After RAM util (align-struct-64):

image

### DIFF
--- a/Core/HW/SimpleAudioDec.h
+++ b/Core/HW/SimpleAudioDec.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2013- PPSSPP Project.
+﻿// Copyright (c) 2013- PPSSPP Project.
 
 // This program is free software: you can redistribute it and/or modify
 // it under the terms of the GNU General Public License as published by
@@ -74,26 +74,7 @@ public:
 	u32 GetCtxPtr() const { return ctxPtr; }
 
 private:
-	void Init();
-	bool OpenCodec(int block_align);
-
-	u32 ctxPtr;
-	int audioType;
-	int sample_rate_;
-	int channels_;
-	int outSamples; // output samples per frame
-	int srcPos; // bytes consumed in source during the last decoding
-	int wanted_resample_freq; // wanted resampling rate/frequency
-
-	AVFrame *frame_;
-#if HAVE_LIBAVCODEC_CONST_AVCODEC // USE_FFMPEG is implied
-	const
-#endif
-	AVCodec *codec_;
-	AVCodecContext  *codecCtx_;
-	SwrContext      *swrCtx_;
-
-	bool codecOpen_;
+	uint32_t ctxPtr = 0xFFFFFFFF;
 };
 
 void AudioClose(SimpleAudio **ctx);
--- a/Core/MIPS/MIPSAnalyst.h
+++ b/Core/MIPS/MIPSAnalyst.h
@@ -144,10 +144,6 @@ namespace MIPSAnalyst
 		u32 opcodeAddress;
 		MIPSOpcode encodedOpcode;
 
-		// shared between branches and conditional moves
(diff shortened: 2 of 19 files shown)


Common/GPU/Shader.h
```diff
@@ -102,9 +102,9 @@ enum class SamplerFlags {
 ENUM_CLASS_BITOPS(SamplerFlags);
 
 struct SamplerDef {
-	int binding;  // Might only be used by some backends.
-	const char *name;
+	const char* name;
 	SamplerFlags flags;
+	int binding;  // Might only be used by some backends.
```
hrydgard, line 107: This particular one won't matter for performance, here I prefer the original order, looks better in the table using the struct later.

Core/MIPS/MIPSAnalyst.h
```diff
@@ -144,25 +144,25 @@ namespace MIPSAnalyst
 		u32 opcodeAddress;
 		MIPSOpcode encodedOpcode;
 
-		// shared between branches and conditional moves
-		bool isConditional;
-		bool conditionMet;
-
 		// branches
 		u32 branchTarget;
+		int branchRegisterNum;
```
hrydgard, line 149: move this one int down again, I liked having it next to isBranchToRegister. It's aligned with no waste anyway since it's after exactly four bools.

Windows/W32Util/Misc.h
```diff
@@ -47,8 +47,8 @@ struct GenericListViewColumn
 struct GenericListViewDef
 {
 	const GenericListViewColumn* columns;
-	int columnCount;
 	int* columnOrder;
+	int columnCount;
```
hrydgard, line 51: This is another one I'd prefer to revert because it makes the tables that use the struct look less nice.

hrydgard (commented):
These are good for things like the Vulkan/OpenGL structs.

However, when structs are used to define arrays like SamplerDef which are not accessed very often, I think clarity matters more than saving a few bytes.

So, I have marked a couple that I would like to be reverted.

Common/GPU/Shader.h
GermanAizek: fix it https://github.com/hrydgard/ppsspp/pull/19047/commits/87bd591cee2a17d537b75330467cc699318fd3b9

Core/MIPS/MIPSAnalyst.h
GermanAizek: fix it https://github.com/hrydgard/ppsspp/pull/19047/commits/87bd591cee2a17d537b75330467cc699318fd3b9

Windows/W32Util/Misc.h
GermanAizek: fix it https://github.com/hrydgard/ppsspp/pull/19047/commits/87bd591cee2a17d537b75330467cc699318fd3b9

Core/HW/SimpleAudioDec.h
```diff
@@ -49,7 +49,26 @@ class AudioDecoder {
...
+
+	bool codecOpen_;
+
 	uint32_t ctxPtr = 0xFFFFFFFF;
+	int audioType;
+	int sample_rate_;
+	int channels_;
+	int outSamples; // output samples per frame
+	int srcPos; // bytes consumed in source during the last decoding
+	int wanted_resample_freq; // wanted resampling rate/frequency
```
hrydgard, line 71: Remove these additions, seems a merge went wrong here.

GermanAizek: @hrydgard fix it

hrydgard:
Replacing this with #19464, that keeps the changes that make the most sense. The other changes are low-impact and not worth it.

PR #19079, opened 2024-04-19
iOS: Update VulkanLoader for MoltenVK 1.2.8-style framework finding

warmenhoven:
The MoltenVK 1.2.8 docs recommend adding "one or both" of 
> @executable_path/Frameworks
> @executable_path/Frameworks/MoltenVK.framework

This change assumes both.


hrydgard:
ok, let's try it.

PR #19089, opened 2024-04-27
Fix building on OpenBSD/riscv64.

brad0:
Unbreak on riscv64, we don't have sys/auxv.h or getauxval().

OpenBSD/riscv64 assumes RV64GC, aka RV64IMAFDC.

Our kernel provides no support for the V extension.


hrydgard (approved):

hrydgard:
I'm gonna fix/workaround the mac build failure separately (#19093 ), it's not the fault of this PR.

PR #19091, opened 2024-04-28
Use the DWM API to disable rounded corners on Windows 11.

hrydgard:
Don't think this suits the app, and it cuts off a corner pixel. So, let's turn off the auto rounded corners that Windows 11 gives our main window.

Hopefully just ignored on lower windows versions. The DWM API exists since Windows Vista so I don't think this will cause linking errors (we don't supports older versions of Windows than Vista).


hrydgard:
Error is fixed on master, merging.

PR #19093, opened 2024-04-29
Update unittests for compatibility with ARM Macs

hrydgard:
Not sure why it started crashing, CI OS upgrade?

### DIFF
--- a/Common/CPUDetect.cpp
+++ b/Common/CPUDetect.cpp
@@ -542,3 +542,31 @@ std::string CPUInfo::Summarize() {
 }
 
 #endif // PPSSPP_ARCH(X86) || PPSSPP_ARCH(AMD64)
+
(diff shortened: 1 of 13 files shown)


hrydgard:
Aha, it seems it has started to run as ARM on CI... And, maybe the unittest hasn't worked correctly on ARM. The sampler test is obviously broken (since we don't have a sampler jit for ARM but the test insists on it).

hrydgard:
The remaining failures are a minor different in the output of the VFPU instruction vh2f, (convert half to float, which the PSP's CPU does slightly different to the built-in instruction on ARM), and a slight difference in the software renderer in the test `pspautotests/tests/gpu/vertices/texcoords.prx`: 

Correct:

image

Incorrect:

image

I'm gonna ignore that test for now, though it is a good one for checking the software renderer.

PR #19096, opened 2024-04-29
RetroAchievements: Add support for hashing homebrew

hrydgard:
We finally defined a method in https://github.com/RetroAchievements/rcheevos/issues/244 .

I made a change to the proposed method to cap the hashed file size to 16MB to avoid memory allocation issues. This should be by far enough considering that normal EBOOTs (that don't contain ISOs) just can't be too large if they want to fit into RAM.


LucasKell:
Would it be worth using 64MB purely because it's the max buffer size already used by rcheevos for retail PSP hashing?

PR #19097, opened 2024-04-29
implement sysclib_sprintf

Kethen:
It gets through https://github.com/Kethen/psp-string-test/ with the following result

```
module started
test1 |1 2 3 4 5 6 7 8 9 10 11 12| test1
test2 |1 2 3 4 5 6 7 8 9 10 11 12| test2
test3 |1 2 3 4 5 6 7 8 9 10 11 12| test3
test4 |1 2 3 4 5 6 7 8 9 a b c| test4
test5 |123 45677654 321| test5
test6 |305419896 2427178479 1311768467294899695 305419896 -1867788817 1311768467294899695 12345678 90abcdef 1234567890abcdef| test6
test7 |12345678 90abcdef 1234567890abcdef 12345678 90abcdef 1234567890abcdef 12345678 90abcdef 1234567890abcdef| test7
test8 |123 456 1234567887654321 654 321| test8
```

PR #19100, opened 2024-05-01
sysclib_sprintf psp memory bound checking instead of length limits

Kethen:
Hi, thanks for reviewing https://github.com/hrydgard/ppsspp/pull/19097, wonder if this alternative way of handling memory would be appreciated


hrydgard:
Note that we also have Memory::IsValidRange. Though I guess you need to scan anyway.

Maybe add a Memory::IsValidNullTerminatedString ?

hrydgard:
That's better, yeah. The function could be more optimal by using an approach more similar to IsValidRange, but it doesn't really matter, so we'll go with it for simplicity.

PR #19103, opened 2024-05-03
optimize Memory::IsValidNullTerminatedString

Kethen:
Is this the optimization https://github.com/hrydgard/ppsspp/pull/19100#issuecomment-2088964171 described?


hrydgard:
Exactly, very nice.

PR #19108, opened 2024-05-04
[what][build] add `dwmapi` to CMakeLists.txt

HR1025:
[why] use cmake build directly in win32


HR1025:
According to `https://github.com/hrydgard/ppsspp/commit/2cb3bd1cbe4706c8b50fac9800e92ab99e473e74`

Adapt build simply ~

PR #19115, opened 2024-05-07
 Allow load sceLibFont module

### LINKED ISSUE #19111: Feature Request: Add an option switch to use font files from the game ROM.
### What should happen

If some Chinese translation games use their own fonts, text will be missing, but as long as you change the name of "PSP_GAME\USRDIR\fonts.ppgf" in the game rom and replace it with "ppsspp\assets\flash0\font\jpn0.ppgf", you will be fine. Perfect solution to this problem.

before use：

After use：

### Who would this benefit

Many people who play Chinese translation games. I have been troubled by this problem for several years. I hope a switch can be added to toggle it.

### Platform (if relevant)

None

ULJS-00371-凉宫春日的追忆 汉化版

_No response_

### Checklist

- [X] Check the [latest git build](https://www.ppsspp.org/devbuilds) in case it's already implemented.

sum2012:
It is using custom font library.
Fix #19111
Also remove zh_gb.pgf hack (Now no need)
Test on chinese patched game of Zettai Meikyuu Grimm Nanatsu no Kagi to Rakuen no Otome (ULJM05632)

### DIFF
--- a/Core/HLE/sceKernelModule.cpp
+++ b/Core/HLE/sceKernelModule.cpp
@@ -810,10 +810,12 @@ void ImportFuncSymbol(const FuncSymbolImport &func, bool reimporting, const char
 }
 
 void ExportFuncSymbol(const FuncSymbolExport &func) {
-	if (FuncImportIsSyscall(func.moduleName, func.nid)) {
-		// HLE covers this already - let's ignore the function.
-		WARN_LOG(LOADER, "Ignoring func export %s/%08x, already implemented in HLE.", func.moduleName, func.nid);
-		return;
+	if (strcmp(func.moduleName, "fontfuck") == 0) {
+		if (FuncImportIsSyscall(func.moduleName, func.nid)) {
+			// HLE covers this already - let's ignore the function.
+			WARN_LOG(LOADER, "Ignoring func export %s/%08x, already implemented in HLE.", func.moduleName, func.nid);
+			return;
+		}
 	}
 
 	u32 error;
--- a/Core/HLE/sceKernelModule.cpp
+++ b/Core/HLE/sceKernelModule.cpp
@@ -810,7 +810,7 @@ void ImportFuncSymbol(const FuncSymbolImport &func, bool reimporting, const char
 }
 
 void ExportFuncSymbol(const FuncSymbolExport &func) {
-	if (strcmp(func.moduleName, "fontfuck") == 0) {
+	if (strcmp(func.moduleName, "sceLibfont") == 0) {
 		if (FuncImportIsSyscall(func.moduleName, func.nid)) {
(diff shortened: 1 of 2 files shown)


Core/HLE/sceKernelModule.cpp
```diff
@@ -810,10 +810,12 @@ void ImportFuncSymbol(const FuncSymbolImport &func, bool reimporting, const char
 }
 
 void ExportFuncSymbol(const FuncSymbolExport &func) {
-	if (FuncImportIsSyscall(func.moduleName, func.nid)) {
-		// HLE covers this already - let's ignore the function.
-		WARN_LOG(LOADER, "Ignoring func export %s/%08x, already implemented in HLE.", func.moduleName, func.nid);
-		return;
+	if (strcmp(func.moduleName, "sceLibfont") == 0) {
+		if (FuncImportIsSyscall(func.moduleName, func.nid)) {
```
hrydgard, line 814: Can you explain this logic? Seems to me that now, FuncImportIsSyscall will only be checked if func.moduleName == "sceLibFont". Couldn't this change behavior or break things for other modules? I don't really understand.

sum2012:
@hrydgard fixed

hrydgard:
Ah okay, I understand now.

I still wonder if this could somehow break other games that might have previously not loaded sceLibFont but now will, might do some testing myself after I send the iOS release for review. Although seems more likely to fix than to break stuff, if we can successfully load the library and use it...

xeonliu:
Many Chinese Localization Games uses "fontfuck" moudule, which is basically a wrapper around the original sceFontLib. The sceLibFont is renamed to sceLibFttt, and then can be loaded by the "fontfuck" module.

A particular fontfuck module I found recently will load `fonts.pgf` located in USRDIR.

xeonliu (commented):
I think this will solve a lot of problem.
But sceFont HLE seems to be useless then.
sceFttt & sceFont will be loaded as normal modules.
I think that HLE still has a lot of problem on loading PGF fonts compared to jpcsp. Not sure if it's good to leave a buggy implementation here.

hrydgard:
Yeah I suppose ideally we should skip the HLE here entirely. I'm not willing to mess with this until I've gotten first a good stable iOS build out, and then 1.18. Then we can go experimental for a while.

hrydgard:
This will be implemented in #20196, first as an optional developer setting and later we may enable it by default for libfont. Closing. Thanks for the reminder @sum2012

sum2012:
@hrydgard  I will remind you remove zh_gb.pgf hack when you enable it by default for libfont

hrydgard:
Yup, will do. Does it do any harm if left in though? (of course, should still be removed)

sum2012:
I test before , it should be safe to remove.

PR #19119, opened 2024-05-09
Advanced logging for the ~SCE header modules

Nemoumbra:
This is how the loading of PRX files looks now:

image

If any chars are not printable, they are replaced with question marks. The odd formatting is due to the name containing spaces.


hrydgard:
Looks good to me! It's great to have this little previously-unknown piece of the PRX format documented this way, and to have this information logged out.

PR #19166, opened 2024-05-21
Vulkan on iOS using MoltenVK

### LINKED ISSUE #10654: Feature: Enable Vulkan on macOS/iOS using MoltenVK and the LunarG Apple SDK for Vulkan

hrydgard:
This sets us up to link statically with MoltenVK, which is required on the App Store (actually, seems there are ways to do it dynamically too, but let's not bother).

Fixes #10654 finally (the macOS part is done already).

I don't think Vulkan ever worked even in sideloaded iOS builds previously, since code to start it up is missing. It said it was using Vulkan but wasn't really.

We're using a separate ViewController in order to not pull in GL stuff when we don't need it.

Prep needed before merge:

- [x] #19168
- [x] #19170 Refactor iOS input message processing so it can be shared between viewcontrollers
- [x] Rendering should work
- [x] Backgrounding and foregrounding the app should work
- [ ] Screen resize (optional, might deal with this later)
- [x] Switching between OpenGL and Vulkan should work smoothly without manual restart

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -59,3 +59,6 @@
 [submodule "ext/OpenXR-SDK"]
 	path = ext/OpenXR-SDK
 	url = https://github.com/KhronosGroup/OpenXR-SDK.git
+[submodule "ios/MoltenVK"]
+	path = ios/MoltenVK
+	url = git@github.com:hrydgard/ppsspp-moltenvk.git
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1313,7 +1313,7 @@ elseif(IOS AND NOT LIBRETRO)
 		Common/Battery/AppleBatteryClient.m
 	)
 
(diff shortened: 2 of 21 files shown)


hrydgard:
All the basics work now, just need to correctly handle task switching and resize by managing the render thread correctly.

hrydgard:
That took some cleanup work, but now switching between Vulkan and OpenGL works smoothly. Let's get it in.

PR #19176, opened 2024-05-23
[libretro] Replace `std::clamp()` with local `clamp()`

RobLoach:
This replaces libretro's use of `std::clamp()` with a local implementation of `clamp()`.

### DIFF
--- a/Common/Data/Color/RGBAUtil.cpp
+++ b/Common/Data/Color/RGBAUtil.cpp
@@ -1,7 +1,7 @@
 #include "Common/Data/Color/RGBAUtil.h"
 
 template <typename T>
-static T clamp(T f, T low, T high) {
+T clamp(T f, T low, T high) {
 	if (f < low)
 		return low;
 	if (f > high)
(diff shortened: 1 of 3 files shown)


hrydgard:
Why?

RobLoach:
When building with LibreELEC, even when we had the `-std=c++17` flag set, `clamp` was still being reported as not a member of `std`. @ToKe79 would know which platform this was an issue on, but it was likely some older target like busybox or something.

While we could define a small macro or something to replicate `std::clamp`, thought it would be good to leverage an existing implementation. If you'd rather we take that direction, I'd be happy to make a follow up commit. Thanks!

hrydgard:
Make a local static one instead. Reaching into the color conv code from there is more awkward than duplicating.

RobLoach:
@hrydgard Updated to a local implementation. Thanks for the review :+1:

ToKe79:
works for me - thanks @RobLoach & @hrydgard

hrydgard (approved):

PR #19184, opened 2024-05-25
Native text drawing on macOS/iOS

hrydgard:
Will fix problems with languages rendering as ????? caused by our font atlas not supporting that set of characters.

Helvetica doesn't quite feel right though... Gonna have to pick another of the shipped-with-the-OS fonts maybe.


Merucza-25 (commented):
_#02 pc 00000000008675f8  arm64/libppsspp_jni.so (HandleAssert(char const*, char const*, int, char const*, char const*, ...)+344) (BuildId: e2a665cce99374cec395cfef5394541ea8505f76)
#03 pc 00000000008398cc  arm64/libppsspp_jni.so (VulkanRenderManager::BeginFrame(bool, bool)+280) (BuildId: e2a665cce99374cec395cfef5394541ea8505f76)
#04 pc 0000000000dd18b0  arm64/libppsspp_jni.so (Draw::VKContext::BeginFrame(Draw::DebugFlags)+32) (BuildId: e2a665cce99374cec395cfef5394541ea8505f76)
#05 pc 00000000008851b0  arm64/libppsspp_jni.so (NativeFrame(GraphicsContext*)+748) (BuildId: e2a665cce99374cec395cfef5394541ea8505f76)_

PR #19206, opened 2024-05-28
More libretro maintenance

sonninnos:
- Upload one frame worth of audio per frame
- Enable 30hz duplicate frames option by default
- Add CHD to valid extensions
- Allow debug building without editing makefile
- Remove unused StereoResampler


hrydgard (approved):
Not sure about the duplicate frames thing, but it can't hurt that much, so let's go for it if you think it's best for RetroArch.

PR #19218, opened 2024-05-31
i18n: Remove our awkward &-escape support, except in [DesktopUI]

hrydgard:
Windows supports adding "&" before a text character in a string, and using that as an accelerator key. To actually produce a "&", you need to use "&&". This is mainly used in window menus and in classic win32 dialog boxes. Elsewhere in the app, it's not used. The Win32 "DrawText" API supports discarding & for drawing, and we've stuck to that standard for all the text backends, leading to awkward workarounds.

So, to simplify the code (before improving our text rendering code in other ways) let's get rid of it almost everywhere (except in win32 menus).


UI/SavedataScreen.cpp
```diff
@@ -268,8 +268,7 @@ UI::EventReturn SavedataPopupScreen::OnDeleteButtonClick(UI::EventParams &e) {
 }
 
 static std::string CleanSaveString(const std::string &str) {
-	std::string s = ReplaceAll(str, "&", "&&");
-	s = ReplaceAll(s, "\n", " ");
+	std::string s = ReplaceAll(s, "\n", " ");
```
oltolm, line 271: This probably should look like this
```
std::string s = ReplaceAll(str, "\n", " ");
```

hrydgard: Indeed, thanks!

hrydgard: Fixed in eb317378e3

PR #19225, opened 2024-06-01
Add a 32-bit AppImage Script for Linux

ghost:
A script to create a 32-bit AppImage. Right-click on the script and open "Properties". In the "Permissions" section check if the file is allowed to be used as a program; If not, allow it to be used as a program.

Enter the following code in a command-line window with ~/ppsspp/build/ as the directory after a build has been made:

~/ppsspp/scripts/makeappimage_32-bit.sh


sandboxgamedev123:
A different pull request was added.

PR #19230, opened 2024-06-01
Libretro: Add video renderer core option

### LINKED ISSUE #14866: (Libretro) Expose "Backend" in core options
### What should happen

Users should be able to override retroarch's global video driver with a per core option. 

### Who would this benefit

People who want to set a global video driver in retroarch but prefer to use something specific in the ppsspp core.

### Platform (if relevant)

libretro / Retroarch

Any

Some cores already give us this option. Like Beetle PSX, Mupen, and PCSX2. Hopefully this hasn't been requested already. If it has I apologize.

### Checklist

sonninnos:
- Add core option for video renderer
  - Defaults to current auto mode
  - Was not possible to add both gl and glcore, since frontend will return glcore regardless if it exists, unless gl is already the current frontend driver
  Closes #14866
- Add core option for cropping to 16x9
- Add separate core option category for speed hacks
- Add missing speed hacks as core options
- Set the only available emulated controller type with `SET_CONTROLLER_INFO`
- Reorganized core options to match better with the original regarding labels and order

Also hooked up software renderer, and please tell me if there is a better way of getting the image than using `TranslateDebugBufferToCompare()`. It has a minor issue though with the image getting full of garbage in the background if frontend pauses the core in the menu, but otherwise seems fine. Also wasn't able to figure out how to reset context cleanly with Vulkan after toggling the 16x9 crop mode while the core is running, so left that for later times.

### DIFF
--- a/libretro/LibretroGraphicsContext.cpp
+++ b/libretro/LibretroGraphicsContext.cpp
@@ -90,8 +90,8 @@ LibretroGraphicsContext *LibretroGraphicsContext::CreateGraphicsContext() {
 	if (!Libretro::environ_cb(RETRO_ENVIRONMENT_GET_PREFERRED_HW_RENDER, &preferred))
 		preferred = RETRO_HW_CONTEXT_DUMMY;
 
-	if (Libretro::renderer != RETRO_HW_CONTEXT_DUMMY)
-		preferred = Libretro::renderer;
+	if (Libretro::backend != RETRO_HW_CONTEXT_DUMMY)
+		preferred = Libretro::backend;
 
 #ifndef USING_GLES2
 	if (preferred == RETRO_HW_CONTEXT_DUMMY || preferred == RETRO_HW_CONTEXT_OPENGL_CORE) {
--- a/libretro/LibretroGraphicsContext.cpp
+++ b/libretro/LibretroGraphicsContext.cpp
@@ -90,6 +90,9 @@ LibretroGraphicsContext *LibretroGraphicsContext::CreateGraphicsContext() {
 	if (!Libretro::environ_cb(RETRO_ENVIRONMENT_GET_PREFERRED_HW_RENDER, &preferred))
 		preferred = RETRO_HW_CONTEXT_DUMMY;
 
+	if (Libretro::renderer != RETRO_HW_CONTEXT_DUMMY)
+		preferred = Libretro::renderer;
+
 #ifndef USING_GLES2
 	if (preferred == RETRO_HW_CONTEXT_DUMMY || preferred == RETRO_HW_CONTEXT_OPENGL_CORE) {
 		ctx = new LibretroGLCoreContext();
@@ -131,7 +134,6 @@ LibretroGraphicsContext *LibretroGraphicsContext::CreateGraphicsContext() {
 		delete ctx;
 
 		ctx = new LibretroD3D9Context();
-
 		if (ctx->Init()) {
 			return ctx;
 		}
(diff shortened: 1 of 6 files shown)


hrydgard:
For the software renderer, you should let at least part of the hardware presentation pipeline do its job, that is, `SoftGPU::CopyToCurrentFboFromDisplayRam` ? Otherwise, special stuff like our workaround for Darkstalkers won't work properly. https://github.com/hrydgard/ppsspp/pull/12443

sonninnos:
Thanks, but I can't figure out how to go ahead in letting it do that.. That function is triggered, but won't do anything because draw context does not exist. And looks like getting the draw context implies getting the HW render interface, which kind of defeats the whole purpose of software render, if I understand correctly.

Indeed that same particular game behavior happens also when using the screenshot feature in standalone, since it is using the same method.

anr2me:
HW render interface will always be needed isn't? because PPSSPP's UI is using it, right? (regardless whether the game is using hardware/software GPU emulation)

hrydgard:
In PPSSPP, the software GPU option performs all actual rendering operations on the host CPU, writing properly to the PSP's video memory just like the real PSP (doesn't virtualize framebuffers etc like the hardware backends do in order to support higher resolutions). However, presentation is still always handled by the currently selected hardware backend, so really it's a submode.

I don't know how that maps to RetroArch, i guess it has its own presentation pipeline for raw framebuffer data. I suggest not to use it and instead treat it as an option of the current hardware backend - it shouldn't be an option along with Vulkan and OpenGL for example, it should be a sub-option.

sonninnos:
Okeydoke, I separated "software rendering" to its own core option, and kept the current none-context for science. Also fixed rendering resolution option to work without restarting, and reordered and relabeled more core options.

hrydgard:
What fonts are garbled? Fonts shouldn't be garbled if you have the /assets files from PPSSPP. And if you don't, your distribution method is broken since PPSSPP requires them.

libretro/LibretroGraphicsContext.cpp
```diff
@@ -143,3 +145,61 @@ LibretroGraphicsContext *LibretroGraphicsContext::CreateGraphicsContext() {
 	ctx->Init();
 	return ctx;
 }
+
+std::vector<u32> TranslateDebugBufferToCompare(const GPUDebugBuffer *buffer, u32 stride, u32 h) {
```
hrydgard, line 149: Hm, this shouldn't be needed anymore right?

bslenul:
Just messed a bit with the PR out of curiosity, noticed 2 things:

1. RetroArch crashes when using the "None" backend: [crash-240603-105345.log](https://github.com/user-attachments/files/15531357/crash-240603-105345.log), if it's there "for science" maybe it should be behind a `#ifdef DEBUG` or whatever? Otherwise it feels like we're giving a new tool for the user to shoot itself in the foot 😅 
2. OGL backend + SW gives me a black screen (Windows 10 if it matters). However it seems to work fine if I turn ON SW renderer AFTER booting a game then restart it.

libretro/LibretroGraphicsContext.cpp
sonninnos: The previous "software software" is still there as none-context, and it would not work at all otherwise.

sonninnos:
None is not crashing here, and I'm not getting those garbled UI fonts with any hardware context software rendering, but only with none-context.

bslenul:
Ah yeah with default core options I don't get the crash, looks like it's because of the internal res option, if it's set to anything above 1x it crashes for me with the "None" backend.

sonninnos:
Ok, sorted that resolution oopsie, and also completely disabled resolution switching with none-context. None is also now forcing software on.

Oops, forgot that gl+software startup blackness..

hrydgard:
Can you explain or take a screenshot of the garbled fonts? That just shouldn't happne.

sonninnos:
None-context + software render enabled:
retroarch_2024_06_03_20_51_24_223

None-context and software rendering disabled:
retroarch_2024_06_03_20_52_08_451

I forced software rendering on always with the none-context now, but the same font effect can be done with any hardware context after disabling software render after it was enabled at startup. I guess I could just force core restart to the option like in standalone, since it can't be changed while running.

But that opengl + software blackness at startup is getting a bit annoying. It might have something to do with shaders, since the software image starts showing immediately after log has shader precompiling rows due to being started with non-software. I was able to hack it so that with opengl it disables software before init, sets it back and resets, but rather not. Might have to resort to it sooner or later..

sonninnos:
Ok, opengl software init hack added, and prevented realtime switching of software rendering. I'll leave the opengl Darkstalkers issue for later, since it is like that before these changes, and thus seems out of scope.

hrydgard:
OK, let's get it in.

PR #19239, opened 2024-06-03
OpenXR - Cleanup

lvonasek:
OpenXR SDK update and removal for stereo projection matrix.

### DIFF
--- a/Common/VR/PPSSPPVR.cpp
+++ b/Common/VR/PPSSPPVR.cpp
@@ -808,8 +808,7 @@ void UpdateVRParams(float* projMatrix) {
 	}
 }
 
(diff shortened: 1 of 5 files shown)


hrydgard:
Cool, but can you explain the reason for the projection matrix change?

lvonasek:
The projection matrices are almost the same for left and right eye. Already in 2022 I averaged the values because separated projection matrices caused ghosting on texts rendered in 3D. This is just a cleanup to stop sending the same projection matrix into the shaders twice.

I will fix the compile issue(s) tomorrow.

lvonasek:
I removed from this PR OpenXRLoader removal commit. We need to keep it because the SDK's loader doesn't compile on all platforms.

PR #19240, opened 2024-06-04
Store IR instructions in a bump-allocated vector instead of loose allocations

hrydgard:
This allows us to store the offset directly inside the pseudo instructions, instead of block numbers which require an extra lookup in the dispatcher, which is expensive enough to matter.

On PC, this is a 5-10% performance improvement in the IRInterpreter, which is very decent. Not benchmarked on Apple yet.

The IR JITs are still working.

TODO before merge: 

- [x] We need to stop and reset once the offsets go beyond 24 bits. However, that is a lot of instructions so in practice this will be fine in 99% of games - still want to fix it though.
- [x] This crashed on iOS when switching games from GTA to Outrun. (have not seen this again)
- [x] Savestates don't seem to be working :(

### DIFF
--- a/Core/MIPS/ARM64/Arm64IRJit.cpp
+++ b/Core/MIPS/ARM64/Arm64IRJit.cpp
@@ -65,10 +65,11 @@ static void NoBlockExits() {
 	_assert_msg_(false, "Never exited block, invalid IR?");
 }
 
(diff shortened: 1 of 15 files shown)


hrydgard:
Fixed a bug, it seems pretty solid now. Let's get it in.

PR #19250, opened 2024-06-05
Libretro: More maintenance

sonninnos:
- Fixed logging to start early and continue to show items after `g_Config.Load`
- Fixed crash when opengl + vsync swap interval detection is switching rates
- Fixed crash when autostate load happens too early (currently not from command line, but from menu playlist)
- Remove `RETRO_ENVIRONMENT_GET_FASTFORWARDING` trickery since FF works fine without it, and it makes opengl to show the output rate wrong


sonninnos:
Might need to rethink or bring back that fastforwarding hackery, since at least ios appears to have some issues with it, as in no FF at all without it, yet Windows flies without problem with every driver.

hrydgard:
OK, so I should hold off on merging?

Issue #19256, opened 2024-06-05
30fps games running at 60fps.


foxsevent:
### Game or games this happens in

ULUS10200 Bltiz Overtime, ULES00841 Brain Lara 2007 Preasure

### What area of the game

The entire game.

### What happens

Both are 30fps games but run at 60fps doubling the game logic speed.

### What should happen

Both games should run at 30fps.
https://youtu.be/kywy1fe8XOQ?si=XrC0fnPcUZDpPb4j
https://youtu.be/QrLOGQAURT8

### Platform

Windows

### Mobile device model or graphics card (GPU)

AMD RX7900GRE

### PPSSPP version affected

1.17.1

### Last working version

v1.17.1-727

### Graphics backend (3D API)

Direct3D 11

### Checklist

- [X] Test in the [latest git build](https://www.ppsspp.org/devbuilds) in case it's already fixed.
- [X] [Search for other reports](https://github.com/hrydgard/ppsspp/search?q=my+issue&type=issues) of the same issue.
- [X] Try resetting settings or older versions and include if the issue is related.
- [X] Try changing graphics settings to determine if one causes the glitch (especially speed hacks or enhancements/replacements.)
- [ ] Include logs or screenshots of issue.

LunaMoo:
Since it looks as a timing issue that's not fixable in an easy way I provided cwcheat workarounds for both of those in my repo long time ago: https://github.com/LunaMoo/PPSSPP_workarounds

Double-0-seven7:
> Since it looks as a timing issue that's not fixable in an easy way I provided cwcheat workarounds for both of those in my repo long time ago: https://github.com/LunaMoo/PPSSPP_workarounds

Also happens in Crazy Taxi but there is no cwcheat in your repo.
Also same thing for a few other games but only in adhoc mode.

LunaMoo:
> Also happens in Crazy Taxi but there is no cwcheat in your repo.

 Never said it is.
 Merely mentioned those two games you initially wrote about as I remembered making workarounds for them.

 This is a valid issue, it's just that as I said, those are timing issues, it's not about some code working incorrectly that could be found and fixed, but about PPSSPP being faster than the real console and some very poorly coded games, depend on it and that means it will not be fixed easily if at all.

PR #19264, opened 2024-06-08
Change GOWFramerateHack30 to FramerateHack30 and workaround timing issues

LunaMoo:
Implements workarounds for the two games from #19256:
- Blitz: Overtime,
- Brian Lara 2007: Pressure Play.

 Which both are running incorrectly at 60 fps with double speed.

 Makes use of GOWFramerateHack30, now renamed to just FramerateHack30, but it's not universal, compat is only used for safety and to have an option to easily disable it, actual hacks are in function replacements.


hrydgard (approved):
Not ideal, but it'll have to do.

Issue #19267, opened 2024-06-11
iOS App Store version’s category should be changed to Games, to enable iOS 18 Game Mode support


schm1dtxbox:
### What should happen

As the title suggests, iOS 18/iPadOS 18 brings over the Game Mode feature seen in macOS Sonoma to iPhones and iPads, and enabling it is likely the same as on macOS too, i.e. the application category type (LSApplicationCategoryType) should be set to something ending in ‘games’ for the feature to activate (e.g. http://public.app-category.games/). Currently PPSSPP appears to set this to ‘Developer Tools’ for the App Store build, hence Game Mode doesn’t currently kick in.

### Who would this benefit

This would particularly benefit those using wireless headphones (e.g. AirPods) or game controllers with PPSSPP, as Game Mode doubles the Bluetooth sampling rate when active, which should reduce latency for the above devices. Game Mode in theory may also provide sustained performance and frame time consistency improvements by prioritising the current game over background processes and by applying more aggressive scheduling (as was observed with macOS’s flavour of Game Mode at least).

### Platform (if relevant)

iOS

### Games this would be useful in

All titles (as this is an app-wide request)

### Other emulators or software with a similar feature

Any application on macOS/iOS with the app category set to something ending in games would be a good example of this, can be tested on the iOS 18 developer preview by launching said app and checking for a ‘Game Mode’ notification at the top of the screen.

### Checklist

- [X] Check the [latest git build](https://www.ppsspp.org/devbuilds) in case it's already implemented.
- [X] [Search for other requests](https://github.com/hrydgard/ppsspp/search?q=my+issue&type=issues) of the same feature.

schm1dtxbox:
1.17.3 is out on the App Store now with all relevant changes (and more it turns out, very nice that Apple finally supports explicitly declaring Game Mode support with GCSupportsGameMode!) so should be safe to close this issue now.

schm1dtxbox:
1.17.3 is out on the App Store now with all relevant changes (and more it turns out, very nice that Apple finally supports explicitly declaring Game Mode support with GCSupportsGameMode!) so should be safe to close this issue now.

hrydgard:
Thanks for closing :)

PR #19270, opened 2024-06-11
Set GCSupportsGameMode in info.plist files for iOS and Mac

hrydgard:
See https://youtu.be/SgshkDg1QN8?t=519

This seems to be the right way to do it.

See #19267

### DIFF
--- a/macOS/InfoGold.plist
+++ b/macOS/InfoGold.plist
@@ -94,6 +94,10 @@
 	<true/>
 	<key>NSHumanReadableCopyright</key>
 	<string></string>
+	<key>LSApplicationCategory</key>
+	<string>public.app-category.games</string>
+	<key>GCSupportsGameMode</key>
+	<true/>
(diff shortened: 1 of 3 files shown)


macOS/InfoGold.plist
```diff
@@ -94,6 +94,10 @@
 	<true/>
 	<key>NSHumanReadableCopyright</key>
 	<string></string>
+	<key>LSApplicationCategory</key>
+	<string>public.app-category.games</string>
```
anr2me, line 98: This value should be used for `LSApplicationCategoryType` instead of `LSApplicationCategory` isn't? https://developer.apple.com/documentation/bundleresources/information_property_list/lsapplicationcategorytype

PS: `LSApplicationCategory` key might not even exist

hrydgard: Oops, that's a typo by me indeed. I'll fix it.

PR #19272, opened 2024-06-15
Fix potential data race

HR1025:


hrydgard:
Hm, care to explain?

HR1025:
- Add lock protection in `GetSystem` and `Unmount` (I think it miss)
- Move `GetHandleOwner` from `public` to `private` to reduce the lock condition
- Use `MetaDummyFileSystem` to remove confuse `null` check
> At first I want use `EmptyFileSystem` directly, but in order to keep `Ioctl ` result as orign try `MetaDummyFileSystem`.

hrydgard:
Well, what I meant is, what problem does this solve?

Do we actually have a serious race condition here that causes problems?

HR1025:
Actually not, not problem ...

hrydgard:
I'm going to add the two missing lock-guards separately, I don't think getting rid of the null checks is worth adding a dummy system, though.

PR #19274, opened 2024-06-16
Updated Brazilian Portuguese Translation

Felipefpl:
Added the new strings and fixed a couple ones.

I'd like to use this PR to avoid creating an issue, the GUI should auto-resize itself in this area (perhaps others?) because the string isnt displayed inside the dialog box.
HOME00000_00001

Also, in the menu bar on top the strings found in the tools of debug section arent translated (beware of the lack of space for them like the one i reported above ;) ).


hrydgard:
Good find, I'll fix the dialog sizing later!

hrydgard:
I don't think we ever had that functionality? I'm confused.

PR #19281, opened 2024-06-19
iOS: Disable the swipe-back gesture in-game, to maximize touch responsiveness

hrydgard:
Got a lot of complaints about lost responsiveness on the left edge of the screen after adding the gesture.

The gesture remains in menus, but it's now disabled in-game.


2xxxhp (approved):

2xxxhp (commented):
disable-swipe-ingame

PR #19283, opened 2024-06-19
Use fuse3 compatible appimagetool

Samueru-sama:
appimagetool: https://github.com/probonopd/go-appimage

Fixes a common issue with ubuntu and some forks where `libfuse2` no longer comes installed.

This is a problem because ubuntu doesn't let you have both `libfuse2` and `libfuse3` installed together, even though it is fully possible and distros like arch let you do it.

This appimagetool also uses a zstd compression, which results in a slightly smaller appimage. It also works if the distro only has `libfuse2` installed.

Edit: If more info is needed, Cemu uses this appimagetool to make their appimage as well.

### DIFF
--- a/scripts/makeappimage_32-bit.sh
+++ b/scripts/makeappimage_32-bit.sh
@@ -1,7 +1,8 @@
 #!/usr/bin/env bash
 
 if [ ! -f appimagetool-i686.AppImage ]; then
-    wget https://github.com/AppImage/AppImageKit/releases/download/continuous/appimagetool-i686.AppImage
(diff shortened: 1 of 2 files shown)


sandboxgamedev123:
I support this pull request. A 32-bit AppImage and a 64-bit AppImage was made (based on SDL) using the changed "appimagetool" and they seem to work well.

sandboxgamedev123:
I support this pull request. I made a 32-bit AppImage and a 64-bit AppImage (based on SDL) with the changed "appimagetool" and they seem to work well.

hrydgard (approved):

PR #19287, opened 2024-06-20
iOS: Enable "double-swipe" to switch apps

hrydgard:
To prevent accidental app switching.

Unfortunately, the way to do this is to disable auto-hiding of the swipe indicator. Another strange Apple quirk.


HiracleGinan:
Hi let play

PR #19295, opened 2024-06-22
Update build.yml for Linux

ghost:
Commented code was added to the "build.yml" file. The commented code might be used as a reference.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -158,6 +158,18 @@ jobs:
           args: ./b.sh --libretro_android ppsspp_libretro
           id: android-libretro
 
+#        - os: linux-32-bit-appimage
+#          extra: sdl
+#          cc: gcc
+#          cxx: g++
+#          args: ./b.sh && cd build && ~/ppsspp/scripts/makeappimage_32-bit.sh
+#          id: sdl
+#        - os: linux-64-bit-appimage
+#          extra: sdl
+#          cc: gcc
+#          cxx: g++
+#          args: ./b.sh && cd build && ~/ppsspp/scripts/makeappimage_64-bit.sh
+#          id: sdl
         - os: ubuntu-latest
           extra: qt
           cc: gcc


anr2me:
Doing `./b.sh && cd build && ~/ppsspp/scripts/makeappimage_32-bit.sh` will still built 64-bit version of PPSSPP because PPSSPP detects the CPU arch, and doesn't have cross-compiling option i think.

It might be better to create a separate yml file to test (since you're not creating the artifact) AppImage building instead of using build.yml, which reduce the complexity.

PS: I'm currently working on manual workflow to generate linux build on various arch at https://github.com/ANR2MERefork/ppsspp/blob/test_workflow/.github/workflows/manual_generate_ipa.yml
(borrowing `manual_generate_ipa.yml` file for testing as i can only run jobs on existing yml file)
Was planning to add AppImage format too later if all arch supported by PPSSPP can be built successfully and passed the Headless and Unit tests (currently failed on some of the headless test on non-intel arch).
PPS: Dealing with `musl` toolchain on Alpine Linux can be a pain :( i might consider using a different distro that support riscv64 too later.

sandboxgamedev123:
The commented code was added as it might be used as a reference. The commented code was not necessarily made to be directly used.

The 32-bit script was previously used on Linux 32-bit.

The 64-bit script was previously used on Linux 64-bit.

I am not planning on making additional contributions. This pull request will be closed.

Samueru-sama:
Uh oh

@anr2me I was helping sandboxgamedev123 with setting up the appimage, but I see there has been a small problem lol. 

I just checked your commits, I don't think you will be able to use linuxdeploy on alpine unless you use their compatibility layer iirc, it has a dependency to glibc (all appimages do unless you bother to use the static runtime with a statically linked program, which is basically none, and linuxdeploy itself doesn't use the static runtime anyway).

anr2me:
Yeah, i just realized that and stopped testing it after using `gcompat` and still didn't work LOL

On the bright side, i noticed something strange about action runner where doing `sudo update-binfmts --disable appimage-type2` (which is supposedly to be one of the workaround i found [here] (https://github.com/AppImage/AppImageKit/issues/1027#issuecomment-1219537909)) seems to have permanent effects (can't be enabled again without reinstalling the package)

Samueru-sama:
> Yeah, i just realized that and stopped testing it after using `gcompat` and still didn't work LOL
> 
> On the bright side, i noticed something strange about action runner where doing `sudo update-binfmts --disable appimage-type2` (which is supposedly to be one of the workaround i found [here](https://github.com/AppImage/AppImageKit/issues/1027#issuecomment-1219537909)) seems to have permanent effects (can't be enabled again without reinstalling the package). I thought each run will have a fresh start, but apparently not, after facing this issue.
> 
> I'll probably replace the `jirutka/setup-alpine` action with `uraimo/run-on-arch-action` later for a more versatile selection of distro.

Worth mentioning that `go-appimage`, which is the other tool that gets downloaded alongside linuxdeploy uses the static runtime and probably can run in alpine with the `--appimage-extract-and-run` flag, `go-appimage` can also do the job of linuxdeployqt, however with qt applications I haven't had much luck with it. Maybe it works for the sdl build though.

Samueru-sama:
@anr2me Can you try this on alpine to see if it can deploy the appimage with go-appimagetool? Here is the modified version of the appimage script:

```
#!/usr/bin/env bash

if [ ! -f appimagetool-x86_64.AppImage ]; then
	APPIMAGETOOL=$(wget -q https://api.github.com/repos/probonopd/go-appimage/releases -O - | sed 's/"/ /g; s/ /\n/g' | grep -o 'https.*continuous.*tool.*86_64.*mage$')
	wget -q "$APPIMAGETOOL" -O ./appimagetool-x86_64.AppImage
    chmod +x appimagetool-x86_64.AppImage
fi

mkdir ./AppDir/
mkdir ./AppDir/usr/
mkdir ./AppDir/usr/bin/
mkdir ./AppDir/usr/share/
mkdir ./AppDir/usr/share/applications/
mkdir ./AppDir/usr/share/icons/
mkdir ./AppDir/usr/share/icons/hicolor/
mkdir ./AppDir/usr/share/icons/hicolor/256x256/
mkdir ./AppDir/usr/share/icons/hicolor/256x256/apps/

cp ~/ppsspp/SDL/PPSSPPSDL.desktop ./AppDir/
cp ~/ppsspp/SDL/PPSSPPSDL.desktop ./AppDir/usr/share/applications/
cp ~/ppsspp/build/PPSSPPSDL ./AppDir/usr/bin/
cp -R ~/ppsspp/build/assets ./AppDir/usr/bin/
cp ~/ppsspp/icons/hicolor/256x256/apps/ppsspp.png ./AppDir/usr/share/icons/hicolor/256x256/apps/

DESTDIR=AppDir make install
chmod +x AppDir/AppRun
./appimagetool-x86_64.AppImage --appimage-extract-and-run -s deploy AppDir/usr/share/applications/*.desktop
rm AppDir/ppsspp.png
pushd AppDir
ln -s usr/share/icons/hicolor/256x256/apps/ppsspp.png
popd
ARCH=x86_64 VERSION=$(./AppDir/AppRun --version) ./appimagetool-x86_64.AppImage --appimage-extract-and-run AppDir

```

Note that I'm using `-s deploy` with go-appimagetool, this makes it deploy everything and makes the appimage compatible with any distro (including musl distros) whereas linuxdeploy has an internal list of libs to skip as those are always expected to be in any linux distro, however it might fail as it might try to deploy something that wasn't available originally when the script was made.

It also doesn't support exclude libs which sandboxdev added a few extra libs to exclude which I don't know if there was a good reason behind those being excluded.

anr2me:
For flexibility, it's better to use relative path to the current directory (`./../`) instead of relative to $HOME (`~`) with fixed repository name, like the changes i made on my PR https://github.com/ANR2MERefork/ppsspp/pull/13/files

Samueru-sama:
> For flexibility, it's better to use relative path to the current directory (`./../`) instead of relative to $HOME (`~`) with fixed repository name (which doesn't works well on CI), like the changes i made on my PR https://github.com/ANR2MERefork/ppsspp/pull/13/files

Oh I know, I didn't make the script and just did the necessary changes, it also has like no error checking and all of that. I'm also not sure why there is a `ln -s usr/share/icons/hicolor/256x256/apps/ppsspp.png` in there, same with the `pushd` and `popd`, I don't think they are needed but anyway lol.

> Btw, which line numbers are being changed from existing script?
>Edit: nevermind, i use https://text-compare.com/ to compare them.

Here is an explanation anyway.

I removed everything related to linuxdeploy (the check and download for linuxdeploy) and the line where linuxdeploy gets called to deploy the appimage was replaced for this: 

`./appimagetool-x86_64.AppImage --appimage-extract-and-run -s deploy AppDir/usr/share/applications/*.desktop`

All the `--exclude-library` lines were also removed since go-appimagetool doesn't support that.

And the `chmod +x AppRun` was replaced for `chmod +x AppDir/AppRun` and it is done sooner now instead of later which in the original script was being done after deploy was done, this is just in case go-appimagetool complains about that because it has more integrity checks than linuxdeploy.

anr2me:
Btw, since i don't have a linux box (don't even have WSL2 due to low free space) to test this locally, and can only test it at github action.

What is the final output file?

Is it a single file named AppRun?
If it is, can i simply rename it to PPSSPPSDL.AppImage or do i need another step to create AppImage file?

Samueru-sama:
image

The appimage was made! but looks like zip failed to find it? 

> Btw, since i don't have a linux box (don't even have WSL2 due to low free space) to test this locally, and can only test it at github action.
> 
> What is the final output file?
>
> Is it a single file named AppRun? If it is, can i simply rename it to PPSSPPSDL.AppImage or do i need another step to create AppImage file?

The final output file is a function of the name of the application in the `ppsspp.desktop` and the `version` from the `$VERSION` env variable. This time it was `PPSSPP-v1.17.1-925-gef6d974e69-x86_64.AppImage` but that will change in the future as the VERSION changes. 

However you can modify the script so that it does a `mv *.AppImage PPSSPP.AppImage` for example after it is done and that way you will always have a fixed name. Make sure there isn't any other file that ends in `.AppImage` before doing the mv btw.

anr2me:
Great!
Yeah, it failed to zip the artifact because the location of the file is in the build folder instead of the AppDir

So now, i need to modify the script to accept `--arch=x86_64` argument to make it more flexible, or may be better to create a new script without the 32/64 bit in the name.

Samueru-sama:
> Great! Yeah, it failed to zip the artifact because the location of the file is in the build folder instead of the AppDir
> 
> So now, i need to modify the script to accept `--arch=x86_64` argument to make it more flexible, or may be better to create a new script without the 32/64 bit in the name.
> 
> Btw, what is the `pubkey.asc` issue? is it something needed for AppImage? never seen this error before

I'm not sure how it works, but that is related to signature verification function that go-appimage does to the AppDir.

It can also add a zsync release, this way it can be quickly and safely updated without having to download the entire appimage again. But I haven't bothered to figure out how to get that working with go-appimage lol.

anr2me:
LOL okay, it;s probably only needed if we want to publish the AppImage file to AppImageHub which probably need the file to be signed.

Anyway, thanks a lot for the help!
Well, we still need to test the output AppImage file to see whether it can run on any linux box LOL.

Btw, is it possible for you to release the `appimagetool` for the riscv64 arch? since PPSSPP support riscv too.

Samueru-sama:
> Anyway, thanks a lot for the help! Well, we still need to test the output AppImage file to see whether it can run on any linux box LOL.

I can test it but I need the artifact link. I have distrobox setup so I can try it in different distros as well. 

The original appimage that sandboxgamedev made worked fine, the Qt version had issues with vulkan but the SDL version had no problem running vulkan.
 
> Btw, is it possible for you to release the `appimagetool` for the riscv64 arch? since PPSSPP support riscv too.

I'm just a linux user. I only know shell and a little bit of awk lol.

Someone has to make the PR adding RISC-V support to [go-appimage](https://github.com/probonopd/go-appimage) and the [runtime](https://github.com/AppImage/type2-runtime) which there is actually an open PR adding `loong64` support to the runtime but nobody has done one for RISC-V.

anr2me:
Btw, i tried to create the x86 version, but it seems to get error https://github.com/ANR2MERefork/ppsspp/actions/runs/9901616542/job/27354278342
```
2024/07/12 03:00:26 Copying to libTargetPath: AppDir//usr/lib/libstdc++.so.6
2024/07/12 03:00:26 Working on /usr/lib/libgcc_s.so.1
2024/07/12 03:00:26 Copying to libTargetPath: AppDir//usr/lib/libgcc_s.so.1
2024/07/12 03:00:26 Copying in copyright files...
2024/07/12 03:00:26 Done
LD_LIBRARY_PATH='' find AppDir -type f -exec ldd {} 2>&1 \; | grep '=>' | grep -v AppDir
2024/07/12 03:00:26 To check whether it is really self-contained, run:
rm: can't remove 'AppDir/ppsspp.png': No such file or directory
~/work/ppsspp/ppsspp/build/AppDir ~/work/ppsspp/ppsspp/build
~/work/ppsspp/ppsspp/build
PPSSPP
Architecture from $ARCH: i386
2024/07/12 03:00:26 git root: /home/runner/work/ppsspp/ppsspp
2024/07/12 03:00:26 Target AppImage filename: PPSSPP-v1.17.1-926-g8114701d25-i386.AppImage
2024/07/12 03:00:26 Icon file: AppDir/ppsspp.png
2024/07/12 03:00:26 WARNING: AppStream upstream metadata is missing, please consider creating it in
         AppDir/usr/share/metainfo/PPSSPPSDL.appdata.xml
2024/07/12 03:00:26 Cannot find /tmp/appimage_extracted_6fbe33b1f5ab1254679ea1449a64fc1e/usr/bin/runtime-i386, exiting
2024/07/12 03:00:26 It should have been bundled, but you can get it from https://github.com/AppImage/AppImageKit/releases/continuous
         Please see https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html#sect-Quickstart-DesktopApps
         for more information or use the generator at
         https://appimagecommunity.github.io/simple-appstream-generator/
Error: Process completed with exit code 1.
```

Samueru-sama:
i686 support was something that gameboxdev wanted. The error indicates that the i686 runtime is missing, which is a big oof and someone that likely needs to be raised to go-appimage.

anr2me:
Well there is no `i386` at https://github.com/probonopd/go-appimage/releases only a better version of it (i686).
May be i should use `ARCH=i686` instead of `ARCH=i386` ?

Btw, you can download the 64 bit PPSSPP AppImage build artifact at https://github.com/ANR2MERefork/ppsspp/actions/runs/9901841457 if you want to test it.

The zip issue before was because i mistakenly thought that AppImage is an archive containing AppDir and it's contents, which later i found out that AppImage is an ELF format but forgot to remove the zip part from the script LOL

Samueru-sama:
> Well there is no `i386` at https://github.com/probonopd/go-appimage/releases only a better version of it (i686). May be i should use `ARCH=i686` instead of `ARCH=i386` ?

I thought those two were the same lol. but maybe try it anyway?

> Btw, you can download the 64 bit PPSSPP AppImage build artifact at https://github.com/ANR2MERefork/ppsspp/actions/runs/9901841457 if you want to test it.
> 
> The zip issue before was because i mistakenly thought that AppImage is an archive containing AppDir and it's contents, which later i found out that AppImage is an ELF format but forgot to remove the zip part from the script LOL

Alright the appimage doesn't work

```
./PPSSPP-v1.17.1-927-g9a2509c2c4-x86_64.AppImage                   
VulkanMayBeAvailable: Device allowed ('SDL:Linux')
Vulkan loader: Library not available: 
DEBUG: Vulkan is not available, not using Vulkan.
Failed to initialize SDL with joystick support. Retrying without.
Unable to initialize SDL: dsp: No such audio device

```

However this may have been caused because I used the deploy everything mode with go-appimage and it failed to add everything (yes that happens because it only uses ldd to know what the app needs lol)

Try changing the line that contains `./appimagetool-x86_64.AppImage --appimage-extract-and-run -s deploy AppDir/usr/share/applications/*.desktop` 

for 

`./appimagetool-x86_64.AppImage --appimage-extract-and-run deploy AppDir/usr/share/applications/*.desktop` and hopefully that was the issue.

anr2me:
Here is the new 64 bit build https://github.com/ANR2MERefork/ppsspp/actions/runs/9902215003
And here is the 32 bit build (using ARCH=i686 works) https://github.com/ANR2MERefork/ppsspp/actions/runs/9902218252

Samueru-sama:
> Here is the new 64 bit build https://github.com/ANR2MERefork/ppsspp/actions/runs/9902215003 And here is the 32 bit build (using ARCH=i686 works) https://github.com/ANR2MERefork/ppsspp/actions/runs/9902218252 But the 32 bit version is still using the `-s deploy` LOL

Yeah it doesn't work: 

```
./PPSSPP-v1.17.1-929-g9b1e2b6442-x86_64.AppImage                                                                  
/tmp/samuel/.mount_i3LCFFND/AppRun: 104: exec: /tmp/samuel/.mount_i3LCFFND/usr/bin/PPSSPPSDL: not found
```

What's odd is that the `usr/bin/PPSSPPSDL` is there, however if I try to run it it fails with a not found error wut: 

image

There are two possibilities here: 

* go-appimagetool has a bug that breaks PPSSPPSDL binary (this can happen as it does some binary patching hacks)

* There is something wrong with how the original binary was compiled 👀

I will have to setup alpine in my end and make the PPSSPP appimage with both linuxdeploy and go-appimage and see what happens.

anr2me:
If i'm not mistaken, if it couldn't file the file during exec while it existed, either it's considered as non-executable or it linked to a library of the wrong arch (ie. tried to use 32 bit version library on 32 bit executable)

Samueru-sama:
>Btw, what is the AppDir/ppsspp.png image used for? is it as default icon?

That becomes the `.DirIcon` file which is the icon that gets displayed for the AppImage thumbnail. 

I don't think that part of the script is needed either, because go-appimage/linuxdeploy automatically look in `usr/share/icons` for the icon defined in the `*.desktop`

Usually when there is multiple icons in `usr/share/icons` one has to make a symlink to the top level of the AppDir to prevent the wrong icon from being picked for `.DirIcon` but there is only one icon in `usr/share/icons` so it should be ok to remove that part of the script.

anr2me:
> > Here is the new 64 bit build https://github.com/ANR2MERefork/ppsspp/actions/runs/9902215003 And here is the 32 bit build (using ARCH=i686 works) https://github.com/ANR2MERefork/ppsspp/actions/runs/9902218252 But the 32 bit version is still using the `-s deploy` LOL
> 
> Yeah it doesn't work:
> 
> ```
> ./PPSSPP-v1.17.1-929-g9b1e2b6442-x86_64.AppImage                                                                  
> /tmp/samuel/.mount_i3LCFFND/AppRun: 104: exec: /tmp/samuel/.mount_i3LCFFND/usr/bin/PPSSPPSDL: not found
> ```
> 
> What's odd is that the `usr/bin/PPSSPPSDL` is there, however if I try to run it after extracting the appimage it fails with a not found error wut:
> 
> image
> 
> There are two possibilities here:
> 
>     * go-appimagetool has a bug that breaks PPSSPPSDL binary (this can happen as it does some binary patching hacks)
> 
>     * There is something wrong with how the original binary was compiled 👀
> 
> 
> I will have to setup alpine in my end and make the PPSSPP appimage with both linuxdeploy and go-appimage and see what happens.
> 
> EDIT:
> 
> image
> 
> ldd works on the binary but it has that weird error when trying to run.

Oh regarding that `ldd` i think it's related to the `/lib/ld-musl-x86_64.so.1` if it's similar issue to this https://stackoverflow.com/questions/57104635/why-wont-my-compiled-c-binary-execute

Samueru-sama:
> > > Here is the new 64 bit build https://github.com/ANR2MERefork/ppsspp/actions/runs/9902215003 And here is the 32 bit build (using ARCH=i686 works) https://github.com/ANR2MERefork/ppsspp/actions/runs/9902218252 But the 32 bit version is still using the `-s deploy` LOL
> > 
> > 
> > Yeah it doesn't work:
> > ```
> > ./PPSSPP-v1.17.1-929-g9b1e2b6442-x86_64.AppImage                                                                  
> > /tmp/samuel/.mount_i3LCFFND/AppRun: 104: exec: /tmp/samuel/.mount_i3LCFFND/usr/bin/PPSSPPSDL: not found
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > What's odd is that the `usr/bin/PPSSPPSDL` is there, however if I try to run it after extracting the appimage it fails with a not found error wut:
> > image
> > There are two possibilities here:
> > ```
> > * go-appimagetool has a bug that breaks PPSSPPSDL binary (this can happen as it does some binary patching hacks)
> > 
> > * There is something wrong with how the original binary was compiled 👀
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I will have to setup alpine in my end and make the PPSSPP appimage with both linuxdeploy and go-appimage and see what happens.
> > EDIT:
> > image
> > ldd works on the binary but it has that weird error when trying to run.
> 
> Oh regarding that `ldd` i think it's related to the `/lib/ld-musl-x86_64.so.1` if it's similar issue to this https://stackoverflow.com/questions/57104635/why-wont-my-compiled-c-binary-execute

Yeah it is that, I just booted into alpine with distrobox and it **almost** works now.

```
./PPSSPP-v1.17.1-929-g9b1e2b6442-x86_64.AppImage                                                                                                                                                                                     5ms 
Error loading shared library libGL.so.1: No such file or directory (needed by /tmp/samuel/.mount_PPSSPPmjaENE/usr/bin/PPSSPPSDL)
Error relocating /tmp/samuel/.mount_PPSSPPmjaENE/usr/bin/PPSSPPSDL: glDrawArrays: symbol not found
Error relocating /tmp/samuel/.mount_PPSSPPmjaENE/usr/bin/PPSSPPSDL: glFrontFace: symbol not found
Error relocating /tmp/samuel/.mount_PPSSPPmjaENE/usr/bin/PPSSPPSDL: glColorMask: symbol not found
Error relocating /tmp/samuel/.mount_PPSSPPmjaENE/usr/bin/PPSSPPSDL: glClear: symbol not found
Error relocating /tmp/samuel/.mount_PPSSPPmjaENE/usr/bin/PPSSPPSDL: glDisable: symbol not found
Error relocating /tmp/samuel/.mount_PPSSPPmjaENE/usr/bin/PPSSPPSDL: glViewport: symbol not found
Error relocating /tmp/samuel/.mount_PPSSPPmjaENE/usr/bin/PPSSPPSDL: glTexImage2D: symbol not found
Error relocating /tmp/samuel/.mount_PPSSPPmjaENE/usr/bin/PPSSPPSDL: glReadPixels: symbol not found
Error relocating /tmp/samuel/.mount_PPSSPPmjaENE/usr/bin/PPSSPPSDL: glGetTexImage: symbol not found
Error relocating /tmp/samuel/.mount_PPSSPPmjaENE/usr/bin/PPSSPPSDL: glPixelStorei: symbol not found
Error relocating /tmp/samuel/.mount_PPSSPPmjaENE/usr/bin/PPSSPPSDL: glGetError: symbol not found
Error relocating /tmp/samuel/.mount_PPSSPPmjaENE/usr/bin/PPSSPPSDL: glXGetClientString: symbol not found
Error relocating /tmp/samuel/.mount_PPSSPPmjaENE/usr/bin/PPSSPPSDL: glDepthMask: symbol not found
Error relocating /tmp/samuel/.mount_PPSSPPmjaENE/usr/bin/PPSSPPSDL: glTexParameterf: symbol not found
Error relocating /tmp/samuel/.mount_PPSSPPmjaENE/usr/bin/PPSSPPSDL: glGetString: symbol not found
Error relocating /tmp/samuel/.mount_PPSSPPmjaENE/usr/bin/PPSSPPSDL: glDepthFunc: symbol not found
Error relocating /tmp/samuel/.mount_PPSSPPmjaENE/usr/bin/PPSSPPSDL: glGetTexLevelParameteriv: symbol not found
Error relocating /tmp/samuel/.mount_PPSSPPmjaENE/usr/bin/PPSSPPSDL: glReadBuffer: symbol not found
Error relocating /tmp/samuel/.mount_PPSSPPmjaENE/usr/bin/PPSSPPSDL: glScissor: symbol not found
Error relocating /tmp/samuel/.mount_PPSSPPmjaENE/usr/bin/PPSSPPSDL: glEnable: symbol not found
Error relocating /tmp/samuel/.mount_PPSSPPmjaENE/usr/bin/PPSSPPSDL: glStencilMask: symbol not found
Error relocating /tmp/samuel/.mount_PPSSPPmjaENE/usr/bin/PPSSPPSDL: glGetIntegerv: symbol not found
Error relocating /tmp/samuel/.mount_PPSSPPmjaENE/usr/bin/PPSSPPSDL: glDrawElements: symbol not found
Error relocating /tmp/samuel/.mount_PPSSPPmjaENE/usr/bin/PPSSPPSDL: glStencilFunc: symbol not found
Error relocating /tmp/samuel/.mount_PPSSPPmjaENE/usr/bin/PPSSPPSDL: glXGetProcAddressARB: symbol not found
Error relocating /tmp/samuel/.mount_PPSSPPmjaENE/usr/bin/PPSSPPSDL: glClearDepth: symbol not found
Error relocating /tmp/samuel/.mount_PPSSPPmjaENE/usr/bin/PPSSPPSDL: glStencilOp: symbol not found
Error relocating /tmp/samuel/.mount_PPSSPPmjaENE/usr/bin/PPSSPPSDL: glBindTexture: symbol not found
Error relocating /tmp/samuel/.mount_PPSSPPmjaENE/usr/bin/PPSSPPSDL: glCullFace: symbol not found
Error relocating /tmp/samuel/.mount_PPSSPPmjaENE/usr/bin/PPSSPPSDL: glDeleteTextures: symbol not found
Error relocating /tmp/samuel/.mount_PPSSPPmjaENE/usr/bin/PPSSPPSDL: glTexParameteri: symbol not found
Error relocating /tmp/samuel/.mount_PPSSPPmjaENE/usr/bin/PPSSPPSDL: glGenTextures: symbol not found
Error relocating /tmp/samuel/.mount_PPSSPPmjaENE/usr/bin/PPSSPPSDL: glXQueryVersion: symbol not found
Error relocating /tmp/samuel/.mount_PPSSPPmjaENE/usr/bin/PPSSPPSDL: glClearColor: symbol not found
Error relocating /tmp/samuel/.mount_PPSSPPmjaENE/usr/bin/PPSSPPSDL: glTexSubImage2D: symbol not found
Error relocating /tmp/samuel/.mount_PPSSPPmjaENE/usr/bin/PPSSPPSDL: glDepthRange: symbol not found
Error relocating /tmp/samuel/.mount_PPSSPPmjaENE/usr/bin/PPSSPPSDL: glGetFloatv: symbol not found
Error relocating /tmp/samuel/.mount_PPSSPPmjaENE/usr/bin/PPSSPPSDL: glClearStencil: symbol not found
Error relocating /tmp/samuel/.mount_PPSSPPmjaENE/usr/bin/PPSSPPSDL: glLogicOp: symbol not found
```

Those symbols error might be fixed by setting the `NO_STRIP=true` env variable before deploying. 

How difficult is it to statically link musl?

anr2me:
> How difficult is it to statically link musl?

Not sure, i'm pretty new to linux, and even less with Alpine/musl

Samueru-sama:
> > How difficult is it to statically link musl?
> 
> Not sure, i'm pretty new to linux (i only use Ubuntu on WSL2 to test PPSSPP before), and even less with Alpine/musl

May I know why alpine is being used to build ppsspp instead of ubuntu?

anr2me:
> May I know why alpine is being used to build ppsspp instead of ubuntu?

Because it's the image commonly used to build things on container like Docker LOL

Well i was looking for a github action that allows me to build on various different arch using QEMU without the need to setup my own (or anyone who forked my workflow just to generate the artifacts) Docker account/secret token at github, and the first action i found was `jirutka/setup-alpine` action which looks easy to use, and later found out about 'uraimo/run-on-arch-action' which can use many distros (can even submit a PR of Dockerfile to add more distro)

anr2me:
> Those symbols error might be fixed by setting the NO_STRIP=true env variable before deploying.

where should i add this env var?

Btw, if i use Ubuntu/Debian to build (Debian seems to have riscv64 support at `uraimo/run-on-arch-action`), should i go back using `-s deploy` or removing the `-s` will be better for compatibility with other distros?

Samueru-sama:
>where should i add this env var?

anywhere in the script before appimagetool is ran.

> > May I know why alpine is being used to build ppsspp instead of ubuntu?
> 
> Because it's the image commonly used to build things on container like Docker LOL
> 
> Well i was looking for a github action that allows me to build on various different arch using QEMU without the need to setup my own (or anyone who forked my workflow just to generate the artifacts) Docker account/secret token at github, and the first action i found was `jirutka/setup-alpine` action which looks easy to use, and later found out about `uraimo/run-on-arch-action` which can use many distros (can even submit a PR of Dockerfile to add more distro)

That's ok, I asked because normally when musl is used to build binaries they are statically linked. It surprised me that wasn't the case here.

Alpine is fine if musl is statically linked, if it is not we have this problem that musl is only used in alpine and the musl version of voidlinux and nothing else lol. 

And also I don't know if musl has the versioning compatibility issues that glibc has when it is linked dynamically. which brings me to this other thing: 

>Btw, if i use Ubuntu/Debian to build (Debian seems to have riscv64 support at uraimo/run-on-arch-action), should i go back using -s deploy or removing the -s will be better for compatibility with other distros?

If we go back to ubuntu it has to be ubuntu 20:04 or debian stable ideally. because now ppsspp will be linked dynamically with glibc and that means that it has to be a old version of glibc, because the application will only work with the current version of glibc it was built on and newer but not older.

This sometimes can be a pain because those distros are very old meaning that they might be missing certain dependency versions needed to build ppsspp. some people instead use ubuntu 22:04 and pray it never causes issues.

The deploy everything mode can prevent this issue (and it might increase the size of the appimage by about 30 MiB since it will bundle all of glibc lol) however worth mentioning that when I tried alpine with the older artifact that had the `-s` flag it was still broken so I don't know if it will work. worth the try I guess. 

another good thing of using the deploy everything mode is that the appimage will work on alpine as well. I've tested this with other appimages, I am surprised that go-appimage doesn't bundle musl when deploy everything is used in alpine.

qurious-pixel:
i'm a little late to this one, but i have had success building with glibc using an older build 20.04 as mentioned as the base. Using the musl build also works if building for a musl distro. The biggest issue is building a cross-distro appimage that uses a lib like QT. Dynamically linking would be the best way, but there are some hidden surprises. Statically building will require all bundled libs to be statically built against musl. @ivan-hc builds a small arch environment, like a chroot, that i am sure could be built against musl, and then it may have everything needed to be  a truly cross-distro, glibc independent appimage. 

[ArchImage](https://github.com/ivan-hc/ArchImage)

ivan-hc:
Hi @qurious-pixel , the Archimage method needs improvements.

There is a point I'm stuck on, hardware accelleration support, see https://github.com/ivan-hc/ArchImage/issues/20

hrydgard:
Wait, how did this get closed? And why can't I re-open it?

Samueru-sama:
@qurious-pixel  The idea to bundle everything and be glibc independent came because @anr2me tried to get the original appimage script that used `linuxdeploy` to work in alpine, which it can't work lol.

`go-appimage` can deploy in alpine, and I hoped that using the deploy everything mode it would make the appimage work outside alpine but that didn't work. It might still be possible if musl is statically linked but I don't have high hopes that it will work though.

anr2me:
> Wait, how did this get closed? And why can't I re-open it?
> 
> Oh I guess because the creator deleted his account. I see.

This PR also modify the build.yml in the wrong way LOL

anr2me:
> How difficult is it to statically link musl?

I think musl is already statically linked by default, not sure why it's still linking to libc.so.6, etc, may be i should use `musl-gcc/musl-g++` instead of `gcc/g++`.
Also, there might be issue when the `ld-musl` interpreter doesn't exist as it will be substituted with other interpreter by the kernel, which might not work, based on [the test here](https://stackoverflow.com/questions/65429331/how-can-i-verify-what-dynamic-linker-is-used-when-a-program-is-run) to make the dynamic-linker non-exist by renaming it (even though it's substituted with the original interpreter)

> Error loading shared library libGL.so.1: No such file or directory

Hmm.. you've the mesa-gl apk installed, right?

Samueru-sama:
@anr2me As far as I know graphics libraries cannot be or are not recommended to be statically linked.

When I see programs that release statically linked musl binaries that work everywhere, it is stuff like CLI utilities and similar, not something to the scale of an emulator.

>Hmm.. you've the mesa-gl apk installed, right?

Oh my bad I didn't have it installed. Since I created the distrobox container to test if the "not found" issue was musl related, I actually didn't notice that error until now.

Just installed mesa-gl and tested the [recent appimage](https://github.com/ANR2MERefork/ppsspp/actions/runs/10020679913) that was made:

```
VulkanMayBeAvailable: Device allowed ('SDL:Linux')
Vulkan library loaded ('libvulkan.so.1')
VulkanMayBeAvailable: Enumerating instance extensions
VulkanMayBeAvailable: Instance extension count: 23
VulkanMaybeAvailable: Instance extension found: VK_KHR_device_group_creation (00000001)
VulkanMaybeAvailable: Instance extension found: VK_KHR_display (00000017)
VulkanMaybeAvailable: Instance extension found: VK_KHR_external_fence_capabilities (00000001)
VulkanMaybeAvailable: Instance extension found: VK_KHR_external_memory_capabilities (00000001)
VulkanMaybeAvailable: Instance extension found: VK_KHR_external_semaphore_capabilities (00000001)
VulkanMaybeAvailable: Instance extension found: VK_KHR_get_display_properties2 (00000001)
VulkanMaybeAvailable: Instance extension found: VK_KHR_get_physical_device_properties2 (00000002)
VulkanMaybeAvailable: Instance extension found: VK_KHR_get_surface_capabilities2 (00000001)
VulkanMaybeAvailable: Instance extension found: VK_KHR_surface (00000019)
VulkanMaybeAvailable: Instance extension found: VK_KHR_surface_protected_capabilities (00000001)
VulkanMaybeAvailable: Instance extension found: VK_KHR_wayland_surface (00000006)
VulkanMaybeAvailable: Instance extension found: VK_KHR_xcb_surface (00000006)
VulkanMaybeAvailable: Instance extension found: VK_KHR_xlib_surface (00000006)
VulkanMaybeAvailable: Instance extension found: VK_EXT_acquire_drm_display (00000001)
VulkanMaybeAvailable: Instance extension found: VK_EXT_acquire_xlib_display (00000001)
VulkanMaybeAvailable: Instance extension found: VK_EXT_debug_report (0000000a)
VulkanMaybeAvailable: Instance extension found: VK_EXT_debug_utils (00000002)
VulkanMaybeAvailable: Instance extension found: VK_EXT_direct_mode_display (00000001)
VulkanMaybeAvailable: Instance extension found: VK_EXT_display_surface_counter (00000001)
VulkanMaybeAvailable: Instance extension found: VK_EXT_surface_maintenance1 (00000001)
VulkanMaybeAvailable: Instance extension found: VK_EXT_swapchain_colorspace (00000004)
VulkanMaybeAvailable: Instance extension found: VK_KHR_portability_enumeration (00000001)
VulkanMaybeAvailable: Instance extension found: VK_LUNARG_direct_driver_loading (00000001)
VulkanMayBeAvailable: Calling vkCreateInstance
VulkanMayBeAvailable: Vulkan test instance created successfully.
VulkanMayBeAvailable: Eligible device found: 'AMD Radeon RX 580 2048SP (RADV POLARIS10)'
VulkanMayBeAvailable: Ineligible device found and ignored: 'llvmpipe (LLVM 17.0.6, 256 bits)'
VulkanMayBeAvailable: Found working Vulkan API!
VulkanMayBeAvailable: Destroying instance
DEBUG: Vulkan might be available.
Failed to initialize SDL with joystick support. Retrying without.
Unable to initialize SDL: dsp: No such audio device
```

Doesn't work though.

anr2me:
Hmm.. i guess it doesn't have audio device too, may be installing sof-firmware  and/or pipewire/pipewire-pulse, and install pavucontrol to check whether audio is working or not.

Samueru-sama:
> Hmm.. i guess it doesn't have audio device too, may be installing sof-firmware and/or pipewire/pipewire-pulse, and install pavucontrol to check whether audio is working or not.

Didn't work, same error. And according to pavucontrol I have sound

anr2me:
I have no idea why PPSSPP couldn't detect audio device on Alpine, may be it relies on alsa only.

Btw, does it run on Ubuntu if you have musl-tools package installed?

Samueru-sama:
I'm on artix linux, the `/lib/x86_64-linux` is something from debian.

I installed the musl package in artix, which added a `/lib/ld-musl-x86_64.so.1`, I no longer get the not found errors but I get this instead.

```
Error relocating /lib/libstdc++.so.6: __fprintf_chk: symbol not found
Error relocating /lib/libstdc++.so.6: __openat_2: symbol not found
Error relocating /lib/libstdc++.so.6: __strftime_l: symbol not found
Error relocating /lib/libstdc++.so.6: __libc_single_threaded: symbol not found
Error relocating /lib/libstdc++.so.6: __memcpy_chk: symbol not found
Error relocating /lib/libstdc++.so.6: __cxa_thread_atexit_impl: symbol not found
Error relocating /lib/libstdc++.so.6: __mbsnrtowcs_chk: symbol not found
Error relocating /lib/libstdc++.so.6: __read_chk: symbol not found
Error relocating /lib/libstdc++.so.6: strfromf128: symbol not found
Error relocating /lib/libstdc++.so.6: __wmemset_chk: symbol not found
Error relocating /lib/libstdc++.so.6: __mbsrtowcs_chk: symbol not found
Error relocating /lib/libstdc++.so.6: __memset_chk: symbol not found
Error relocating /lib/libstdc++.so.6: strtof128: symbol not found
Error relocating /lib/libstdc++.so.6: __wmemcpy_chk: symbol not found
Error relocating /lib/libstdc++.so.6: __sprintf_chk: symbol not found
Error relocating /lib/libstdc++.so.6: __memmove_chk: symbol not found
Error relocating /lib/libstdc++.so.6: __isoc23_strtoul: symbol not found
Error relocating /lib/libgcc_s.so.1: __cpu_indicator_init: symbol not found
Error relocating /lib/libgcc_s.so.1: __cpu_model: symbol not found
Error relocating /lib/libgcc_s.so.1: _dl_find_object: symbol not found
Error relocating /lib/libgcc_s.so.1: __memset_chk: symbol not found
Error relocating /lib/libGLdispatch.so.0: __vfprintf_chk: symbol not found
Error relocating /lib/libGLdispatch.so.0: __isoc23_strtol: symbol not found
Error relocating /lib/libGLX.so.0: __vsnprintf_chk: symbol not found
Error relocating /lib/libGLX.so.0: __isoc23_strtol: symbol not found
Error relocating /lib/libGLX.so.0: __snprintf_chk: symbol not found
Error relocating /lib/ld-linux-x86-64.so.2: unsupported relocation type 37
Error relocating /lib/libX11.so.6: __strcat_chk: symbol not found
Error relocating /lib/libX11.so.6: __isoc23_strtoul: symbol not found
Error relocating /lib/libX11.so.6: __snprintf_chk: symbol not found
Error relocating /lib/libX11.so.6: __vfprintf_chk: symbol not found
Error relocating /lib/libX11.so.6: __read_chk: symbol not found
Error relocating /lib/libX11.so.6: __isoc23_sscanf: symbol not found
Error relocating /lib/libX11.so.6: __fdelt_chk: symbol not found
Error relocating /lib/libX11.so.6: __memmove_chk: symbol not found
Error relocating /lib/libX11.so.6: __memcpy_chk: symbol not found
Error relocating /lib/libX11.so.6: __isoc23_strtol: symbol not found
Error relocating /lib/libX11.so.6: __strncpy_chk: symbol not found
Error relocating /lib/libX11.so.6: __strcpy_chk: symbol not found
Error relocating /lib/libX11.so.6: __memset_chk: symbol not found
Error relocating /lib/libX11.so.6: __fprintf_chk: symbol not found
Error relocating /lib/libX11.so.6: __sprintf_chk: symbol not found
Error relocating /lib/libxcb.so.1: __isoc23_strtoul: symbol not found
Error relocating /lib/libxcb.so.1: __snprintf_chk: symbol not found
Error relocating /lib/libxcb.so.1: __memcpy_chk: symbol not found
Error relocating /lib/libxcb.so.1: __strcpy_chk: symbol not found
Error relocating /lib/libxcb.so.1: __memset_chk: symbol not found
```

>Also, may be we can include the ld-musl-x86_64.so.1 in the AppImage

It is in the appimage, you can check it by extracting the appimage and checking `squashfs-root/lib/libc.musl-x86_64.so.1`

The rest of the libraries that were bundled are in `squashfs-root/usr/lib` instead btw.

>We won't be able to statically link libc too, because we use gethostbyname which won't works when statically linked.

The deploy everything mode of go-appimage bundles glibc in the appimage, which for all purpose has the same reason as having glibc statically linked.

I have this [mpv appimage](https://github.com/Samueru-sama/mpv-AppImage/releases/tag/continuous) that I release with and without the deploy everything mode, the bigger one is the one that deployed everything. 

However I don't have high hopes that it will work with ppsspp, with the mpv appimage I have it titled WIP because I have an issue that I cannot play youtube videos with yt-dlp for some reason, which might be related with the gethostbyname issue you mention there lol

Not to mention that recently the duckstation dev shared some concerns [about the deploy everything mode](https://github.com/stenzek/duckstation/pull/3251#issuecomment-2226820698) might cause issues with graphics libraries.

I think at this point it would be better to just switch back to ubuntu to build the appimage, ideally 20:04 but I've seen people use 22:04 and they don't seem to get too many issues opened about glibc version issues lol. 

Also two appimages can be made at once, one with the all trusty linuxdeploy and another with go-appimage deploying everything, it is basically 5 extra lines of shell doing it this way.

Samueru-sama:
@anr2me The appimage now works in alpine, I'm surprised.

image

Games with vulkan launch as well.

anr2me:
Any additional step needed from the last time?

Samueru-sama:
> Any additional step needed from the last time?

None, that's what surprised me, before it wouldn't even launch, and now the gui works and games open. although I just noticed that sound isn't working, might be an issue with alpine though.

------------------------------------------------------

Outside of alpine it doesn't work though, if I don't have musl installed it gives the not found error, and if I install musl I get the long list of `symbol not found` errors.

anr2me:
i tried to use system ffmpeg but the executable is still linked to stdc++ :( i'm confused, unless the alpine's ffmpeg was built with stdc++
```
Dynamic section at offset 0xe95b60 contains 34 entries:
  Tag        Type                         Name/Value
 0x0000000000000001 (NEEDED)             Shared library: [libGL.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libavformat.so.60]
 0x0000000000000001 (NEEDED)             Shared library: [libavcodec.so.60]
 0x0000000000000001 (NEEDED)             Shared library: [libswresample.so.4]
 0x0000000000000001 (NEEDED)             Shared library: [libswscale.so.7]
 0x0000000000000001 (NEEDED)             Shared library: [libavutil.so.58]
 0x0000000000000001 (NEEDED)             Shared library: [libstdc++.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libgcc_s.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libc.musl-x86_64.so.1]
```

anr2me:
Btw @Samueru-sama i tried installing Alpine WSL on Win11, which seems to be Alpine 3.17

After installing fuse3 mesa-gl and gcompat i'm still seeing missing libstc++ library :(
```
Error loading shared library libstdc++.so.6: No such file or directory (needed by /tmp/.mount_PPSSPPOOMPNP/usr/bin/PPSSPPSDL)
Error loading shared library libgcc_s.so.1: No such file or directory (needed by /tmp/.mount_PPSSPPOOMPNP/usr/bin/PPSSPPSDL)
```
Anything else i should install? (was hoping to not needing any development package just like most end users)

Samueru-sama:
I just tested the recent appimage from 5 hours ago, **now sound works**.

image

but it only works in alpine though. And it seems that it has to be a recent version of alpine (3.20.1 in my case). 

------------------------------------------

>After installing fuse3 mesa-gl and gcompat i'm still seeing missing libstc++ library :(

This looks like a bug in go-appimage, because indeed ldd shows that the binary links to those libraries, but they didn't get bundled.

```
/tmp/samuel/Volatile/squashfs-root/usr/bin/ ldd ./PPSSPPSDL | grep -E "(libst|libgcc)"                                                                                                                                                  
	libstdc++.so.6 => /usr/lib/libstdc++.so.6 (0x0000746f28400000)
	libgcc_s.so.1 => /usr/lib/libgcc_s.so.1 (0x0000746f2882b000)
/tmp/samuel/Volatile/squashfs-root/usr/bin/ ldd ./PPSSPPSDL | head                                                                                                                                                                     
	linux-vdso.so.1 (0x000075e773215000)
	libSDL2-2.0.so.0 => /tmp/samuel/Volatile/squashfs-root/usr/bin/./../lib/libSDL2-2.0.so.0 (0x000075e7730a6000)
	libGL.so.1 => /usr/lib/libGL.so.1 (0x000075e773002000)
	libavformat.so.60 => /tmp/samuel/Volatile/squashfs-root/usr/bin/./../lib/libavformat.so.60 (0x000075e772d6f000)
	libavcodec.so.60 => /tmp/samuel/Volatile/squashfs-root/usr/bin/./../lib/libavcodec.so.60 (0x000075e7718ab000)
	libswresample.so.4 => /tmp/samuel/Volatile/squashfs-root/usr/bin/./../lib/libswresample.so.4 (0x000075e77188b000)
	libswscale.so.7 => /tmp/samuel/Volatile/squashfs-root/usr/bin/./../lib/libswscale.so.7 (0x000075e7717f5000)
	libavutil.so.58 => /tmp/samuel/Volatile/squashfs-root/usr/bin/./../lib/libavutil.so.58 (0x000075e7706dc000)
	libstdc++.so.6 => /usr/lib/libstdc++.so.6 (0x000075e770400000)
	libgcc_s.so.1 => /usr/lib/libgcc_s.so.1 (0x000075e7706ad000)
```

Notice that `libstdc++` and `libgcc` point to `/usr/lib` instead of a location in the `squashfs-root` directory.

anr2me:
btw how come it shows full path on your output, meanwhile it only shows relative path on mine:
```
/mnt/c/TMP/shared/squashfs-root/usr/bin # ldd ./PPSSPPSDL | head
        /lib/ld-musl-x86_64.so.1 (0x7f7f04644000)
        libSDL2-2.0.so.0 => ./../lib/libSDL2-2.0.so.0 (0x7f7f044db000)
        libGL.so.1 => /usr/lib/libGL.so.1 (0x7f7f0443d000)
        libavformat.so.60 => ./../lib/libavformat.so.60 (0x7f7f041aa000)
        libavcodec.so.60 => ./../lib/libavcodec.so.60 (0x7f7f02ce8000)
        libswresample.so.4 => ./../lib/libswresample.so.4 (0x7f7f02cc8000)
        libswscale.so.7 => ./../lib/libswscale.so.7 (0x7f7f02c32000)
        libavutil.so.58 => ./../lib/libavutil.so.58 (0x7f7f01b19000)
        libstdc++.so.6 => /usr/lib/libstdc++.so.6 (0x7f7f018cb000)
        libgcc_s.so.1 => /usr/lib/libgcc_s.so.1 (0x7f7f018ad000)
```

Samueru-sama:
> btw how come it shows full path on your output, meanwhile it only shows relative path on mine:
> 
> ```
> /mnt/c/TMP/shared/squashfs-root/usr/bin # ldd ./PPSSPPSDL | head
>         /lib/ld-musl-x86_64.so.1 (0x7f7f04644000)
>         libSDL2-2.0.so.0 => ./../lib/libSDL2-2.0.so.0 (0x7f7f044db000)
>         libGL.so.1 => /usr/lib/libGL.so.1 (0x7f7f0443d000)
>         libavformat.so.60 => ./../lib/libavformat.so.60 (0x7f7f041aa000)
>         libavcodec.so.60 => ./../lib/libavcodec.so.60 (0x7f7f02ce8000)
>         libswresample.so.4 => ./../lib/libswresample.so.4 (0x7f7f02cc8000)
>         libswscale.so.7 => ./../lib/libswscale.so.7 (0x7f7f02c32000)
>         libavutil.so.58 => ./../lib/libavutil.so.58 (0x7f7f01b19000)
>         libstdc++.so.6 => /usr/lib/libstdc++.so.6 (0x7f7f018cb000)
>         libgcc_s.so.1 => /usr/lib/libgcc_s.so.1 (0x7f7f018ad000)
> ```

That's a difference in `ldd`. The `./../` is still shown in both cases though.

If I do the ldd in alpine it does look like in your case. 

image

anr2me:
Btw, i tried to build it using Alpine 3.17 could you test it whether it still works on 3.20 or not?

And apparently it's still not working on my Alpine WSL :( and the file size became much larger than before (with so many libraries bundled)

Samueru-sama:
> Btw, i tried to build it using Alpine 3.17 could you test it whether it still works on 3.20 or not?
> 
> And apparently it's still not working on my Alpine WSL :( and the file size became much larger than before (with so many libraries bundled)

Now it segfaults in alpine: 

```
VulkanMayBeAvailable: Device allowed ('SDL:Linux')
Vulkan library loaded ('libvulkan.so.1')
VulkanMayBeAvailable: Enumerating instance extensions
zsh: segmentation fault  ./PPSSPP-v1.17.1-1020-g0ba124188b-x86_64.AppImage
```

I think we have to switch to he old trusty method of using ubuntu and linuxdpeloy to make the appimage.

anr2me:
Yeah, i agreed, especially when stdc++ is mandatory, and will probably need to build it using old version of glibc for compatibility with both old and new version.

anr2me:
Btw @Samueru-sama could you do another test, i got curious when you said you got audio working the last time.
I wondered whether it was because you upgrade to v3.20.x or because i switched to system sdl2 & ffmpeg instead of statically link them.

So here are the one with statically linked sdl2 & ffmpeg:
This one build on Alpine v3.17: https://github.com/ANR2MERefork/ppsspp/actions/runs/10120046689/artifacts/1746050937
This one build on Alpine v3.20: https://github.com/ANR2MERefork/ppsspp/actions/runs/10119877862/artifacts/1746025901
Could you test both of them, thanks.

anr2me:
> I went back and checked several workflows to find the ones with working sound.
> 
> [239](https://github.com/ANR2MERefork/ppsspp/actions/runs/10103628197) [240](https://github.com/ANR2MERefork/ppsspp/actions/runs/10104071293)
> 
> The one that follows 240 has a segfault, and the one that comes before 239 doesn't have working sound.

Thanks, both 239 and 240 are using system (dynamically linked) sdl2 and ffmpeg :( i guess something is missing when i made sdl2 statically, causing audio not working properly.

anr2me:
@Samueru-sama could you do one more test :)

Out of those "not found" libraries/frameworks used by SDL2 i can only found `pulseaudio-dev sndio-dev jack-dev` at https://pkgs.alpinelinux.org/packages

Here is the build with those packages found (built on Alpine edge) https://github.com/ANR2MERefork/ppsspp/actions/runs/10123458214/artifacts/1746693459

Samueru-sama:
> @Samueru-sama could you do one more test :)
> 
> Out of those "not found" libraries/frameworks used by SDL2 i can only found `pulseaudio-dev sndio-dev jack-dev` at https://pkgs.alpinelinux.org/packages
> 
> Here is the build with those packages found (built on Alpine edge) https://github.com/ANR2MERefork/ppsspp/actions/runs/10123458214/artifacts/1746693459
> 
> ```
> -- Performing Test HAVE_OSS_SYS_SOUNDCARD_H
> -- Performing Test HAVE_OSS_SYS_SOUNDCARD_H - Success
> -- Found ALSA: /usr/lib/libasound.so (found version "1.2.12")
> -- dynamic libasound -> libasound.so.2
> -- Checking for module 'jack'
> --   Found jack, version 1.9.22
> -- dynamic libjack -> libjack.so.0
> -- Checking for module 'libpipewire-0.3>=0.3.20'
> --   Found libpipewire-0.3, version 1.2.1
> -- dynamic libpipewire-0.3 -> libpipewire-0.3.so.0
> -- Checking for module 'libpulse>=0.9.15'
> --   Found libpulse, version 17.0
> -- dynamic libpulse -> libpulse.so.0
> -- Checking for module 'esound'
> --   Package 'esound' not found
> -- Looking for audio/audiolib.h
> -- Looking for audio/audiolib.h - not found
> -- Checking for module 'sndio'
> --   Found sndio, version 1.9.0
> -- dynamic libsndio -> libsndio.so.7
> ```
> 
> PS: This should be the last test for Alpine Linux before moving on to a new branch to test Debian-based container.

Sound works but for some reason it is distorted lol.

anr2me:
Thanks for testing! So it really was due to missing some frameworks that supported by SDL2.

> Sound works but for some reason it is distorted lol.

May be out of those many supported frameworks/libraries, the one being used by the statically linked SDL2 (v2.30.5) is different than the one used by the dynamically linked SDL2(v2.28.5-r1) that came from alpine package, or may be the statically linked SDL2 have sound effect/enhancement (ie. reverb or something) being enabled.

PR #19304, opened 2024-06-25
Allow "Alt" to act like a normal keyboard input, if it's been mapped to something

### LINKED ISSUE #19298: When I press alt, the buttons is automatically pressed.

All

When I press alt, the button is automatically pressed byttons.
I don't know if this is a bug? If this is a function, then please change the button in the settings. I need to name the action on alt, but I can't do it because of that.

### Logs

_No response_

### Platform

Windows

GTX 1050ti

### PPSSPP version affected

Latest build

### Last working version

_No response_

### Graphics backend (3D API)

Direct3D 11

### Checklist

- [X] Test in the [latest git build](https://www.ppsspp.org/devbuilds) in case it's already fixed.

hrydgard:
Or is it worth a specific setting?

Fixes #19298

### DIFF
--- a/Windows/MainWindow.cpp
+++ b/Windows/MainWindow.cpp
@@ -1093,6 +1093,10 @@ namespace MainWindow
 		// doesn't work on Vista or higher.
 		case WM_SYSCOMMAND:
 			{
+				// Disable Alt key for menu if "Ignore Windows Key" is on (likely related)
+				if (g_Config.bIgnoreWindowsKey && wParam == SC_KEYMENU && (lParam >> 16) <= 0) {
+					return 0;
+				}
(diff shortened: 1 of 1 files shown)


Gamerppsspp:
well, everything suits me, but it would be better if it was only when assigned to these keys. So that we can minimize the game and press the windows button

hrydgard:
Hm, yeah, only doing it when the key is mapped to something maybe would be ideal - and only in-game.

hrydgard:
I changed to the mentioned approach. Now, if you map the Alt key, it will no longer open the window menu and you can just use it as an action button, like back when you played Wolfenstein in MS-DOS :)

PR #19316, opened 2024-07-06
Updated Polish translation & fixed strings not being translated properly

eveninglatte:
Added some strings and fixed some old ones in the Polish translation.

While translating, I noticed three strings that are not getting translated properly on the Windows version:
1. "Restart Graphics" in the "Debug" window menu (`MainWindowMenu.cpp`):
image

2. "ISO OK according to the Redump project" (along with the commented out "Game ID unknown - not in the Redump database") after triggering a CRC calculation for a game (`GameScreen.cpp`):
image

3. "Smart 2D texture filtering" in the Game settings -> Texture Filtering window menu:
image

I was able to fix these for the Windows version (**I'm currently unable to test and possibly fix it for other platforms**) by:
1. Adding a `TranslateMenuItem()` method for `ID_DEBUG_RESTARTGRAPHICS`,
2. Changing strings passed to the `T()` function of the `I18NCategory` class,
3. Changing the "Smart 2D Texture Filtering" token to be capitalised on every letter in each location.

For the sake of consistency, I also capitalised every letter of "Smart 2D Texture Filtering" in the `[Graphics]` section and updated the `T()` call accordingly (though it does get translated correctly without these changes, as you can see in the background of the last picture).


hrydgard:
Thank you for the fixes!

Would you mind changing the capitalization the other way though, to always lowercase except the first word? I'm (slowly) moving towards less excessive capitalization. Capitalizing every word used to be the most common style for settings and menus in Windows, but it's falling out of favor.

eveninglatte:
Sure thing, just pushed the changes.

hrydgard (approved):

Issue #19321, opened 2024-07-11
Reconfigure Android Manifest to avoid "Switch to Controllers" nag on oculus quest


jack4455667788:
### What should happen

PPSSPP VR should run without displaying the "Switch to Controllers" nag screen on the oculus quest.

### Who would this benefit

When using a gamepad controller, this nag means you have to get/find your oculus controller in order to continue beyond it and actually run PPSSPP VR. 

Some users may also have broken controllers (or none at all), so this would be particularly useful for them.

Below is a hacked version of the latest PPSSPP VR I edited that accomplishes this, but with the side effect of having an annoying white dot shown in the center of the menu/config screen. The dot does not interfere with normal use, and does not display while playing the games.

https://mega.nz/file/XRRkhC5J#Bi0QmugBNkc5sjLcDxeDe21bejOEZjOo9MiPvdo5Dmo

### Platform (if relevant)

Other

### Games this would be useful in

All

### Other emulators or software with a similar feature

CitraVR and most all android apps.

### Checklist

- [X] Check the [latest git build](https://www.ppsspp.org/devbuilds) in case it's already implemented.
- [X] [Search for other requests](https://github.com/hrydgard/ppsspp/search?q=my+issue&type=issues) of the same feature.

lvonasek:
The gaze dot disappears if you enable hand tracking. Hopefully Meta makes the gaze pointer optional at some point.

lvonasek:
https://github.com/hrydgard/ppsspp/pull/19322 PR merged. You can close this.

PR #19322, opened 2024-07-12
OpenXR - Do not require controllers connected

lvonasek:
When using VR version with gamepad, users are bothered with "Controllers not connected" dialog once the VR controller lost tracking. This PR fixes this annoying behaviour by indicating the app supports hand tracking (although it doesn't).

https://github.com/hrydgard/ppsspp/issues/19321 fixed


hrydgard:
Heh, sneaky.

